PR #70, opened 2020-04-26
fix bad error msg

XK4MiLX:


ZelBack/src/services/explorerService.js
```diff
@@ -424,14 +424,16 @@ async function initiateBlockProcessor(restoreDatabase) {
       if (zelbenchInfo.status === 'success') {
         const { version } = zelbenchInfo.data;
         zelbenchVersion = serviceHelper.ensureNumber(version.split('.').join(''));
+        if (zelbenchVersion < 110) {
+          log.warn('Your ZelBench version is outdated! ZelNode may not behave correctly');
+          log.warn('Block processing will try to reinitiate in 10 minutes');
+          throw new Error('ZelBench update required. Block processing halted.');
```
TheTrunk, line 430: Similar error needs to be thrown in case we are unable to communicate with zelbench too. So the block processing wont be initiated.

XK4MiLX: so so simmilar not info that version is incorrect that is not treu in this case

TheTrunk: Yes instead of` log.error('Unable to communicate with zelbench!');` it will be` throw new Error('Unable to communicate with zelbench!);` instead of the log.error

XK4MiLX: ok

TheTrunk:
#87

PR #73, opened 2020-04-27
running zelcash as service

XK4MiLX:
after first reboot service will be activate no reason to not using it from runing zelcash in this way is possibe better track errors and prevent from crash and chceck when update is tigerred


dk808:
If you want to do this let's change it to if service is detected than start daemon if not use zelcashd cmd. Not everyone uses scripts to install and this would cause issues for them. So something in the lines of `if sudo systemctl status zelcash >/dev/null; then sudo systemctl start zelcash; else $COIN_DAEMON`

XK4MiLX:
yes i think it is good solution

PR #187, opened 2020-10-29
v1.4.1

TheTrunk:
App Removal Fixes
Background: When Flux decides that application shall be removed, this removal could have been interrupted by another Flux process deciding that application shall be running. Tracing like this:
1) Removal process stops ZelApp
2) Check Running Apps starts the ZelApp which was supposed to be removed
3) Removal process fails to remove ZelApp properly leaving some data of ZelApp on the node.
- Correct Removal of application with force. Using force parameter will now further check global specifications, locally available specifications and permanent messages ensuring that some specifications are found.
- Wait with starting a stopped application if removal is in progress

Other Tweaks
- Increase timeout from when Flux begins application installation from 10 mins to 16 mins
-> this allows more zelapprunning messages to be present to better determine if an application shall be spawned
- Remove logging of failing to broadcast a message to peer

### DIFF
--- a/ZelBack/src/services/serviceManager.js
+++ b/ZelBack/src/services/serviceManager.js
@@ -55,10 +55,10 @@ async function startFluxFunctions() {
     }, 4 * 60 * 1000);
     log.info('Flux Block Processing Service started');
     setTimeout(() => {
-      // after 10 minutes of running ok.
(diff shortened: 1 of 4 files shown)


Cabecinha84 (approved):
Ack

PR #192, opened 2020-11-04
v1.4.2

TheTrunk:
Fixes typo issues of paddress instead of ipaddress on adjusting cruxid/kadena. This had no effect as it was correctly overwritten by auto check later on
Fixes an issue where a new version of element UI was installed but Flux was using a deprecated event
-> updates element-ui package and event
Furthermore updates axios package


Cabecinha84 (approved):
Ack

PR #202, opened 2020-12-24
v1.5.0

TheTrunk:
- General bug preventions
- Add deep rescan when duplication of key is found in database
- slow down block processing of transaction
- correct zelid logo
- KDA application description to explain heathy status
- Minor dev corrections
- Remove CruxID from UI


Cabecinha84 (approved):
Ack

PR #205, opened 2020-12-30
v1.6.0

TheTrunk:
- Handling of Application volumes
-> Volumes are now properly mounted after system reboot
-> Fixes an issue with disappearing volumes, files from application after a system restart
- Adjustment of Visit button for KDA application
- Improved KDA application image

Update can take up to 5 minutes as of additional dependency


Cabecinha84 (approved):
Ack

PR #208, opened 2021-01-11
v1.7.0

TheTrunk:
Release targets zelcash daemon stability and updates KDA applicaiton

- Flux will now only start some services once zelcash is synced.
- Flux will now always use unfiltered list of determinstic nodes and mainly its cached value instead of casking for a filtered one, this results in less load for daemon
- Kadena Chainweb application v2.4


Cabecinha84 (approved):
Ack

PR #210, opened 2021-01-13
v1.8.0

TheTrunk:
- Introduce periodical check for zelcash sync status with expectation of minimum header
- Use inner flux function of zelcash sync to check zelcash sync status
- correct crontab adjustment
- additional check for cron job validity


Cabecinha84 (approved):
Ack

PR #216, opened 2021-01-20
v1.9.0

TheTrunk:
- Kadena application hard redeploy fork ( + temporary lower probability of application update)
- Better caching on ZelNode list
- stop block processor if zel daemon not synced


Cabecinha84 (approved):
Ack

PR #222, opened 2021-02-06
v1.10.0

TheTrunk:
- Adds domain, https support for official load balanced solution available at https://home.runonflux.io with api at https://api.runonflux.io
- Improves performance and system usage 
- adds invalid PubKeys to temporarily blocked list
- node list cache optimization

This version will be enforced on all nodes on 13 February and also latest zelcash version.


Cabecinha84 (approved):
Ack

PR #241, opened 2021-03-20
v1.13.0

TheTrunk:
- Adjusts to Flux network fork
- Local calls for getting current Flux directory
- Flux can now be installed in any directory
- Prohibit global application installation on Cumulus nodes
- Coinbase Fusion index - tracks coinbase transactions for Fusion application that determines additional rewards on parallel chains
- Rewrites Flux communication, use native ping
- Updates KDA application

This version will be enforced on ALL nodes by 25th of March


Cabecinha84 (approved):
Before merging, we should revert the delete of file start.sh, or we might have many node operators asking for support as install scripts uses that file to start flux when machine reboots, and inside pm2.
Ack!

TheTrunk:
readded, tests passed

PR #242, opened 2021-03-21
First steps for no static ip needed

Cabecinha84:
Changed flux coms.
This should be released when we release new bins.


Cabecinha84:
If people don't have new bins installed, it will act like before.

Cabecinha84:
So this can be reviewed and merged.

ZelBack/src/services/benchmarkService.js
```diff
@@ -139,6 +139,14 @@ async function getInfo(req, res) {
   return res ? res.json(response) : response;
 }
 
+async function getBenchPublicIp(req, res) {
```
TheTrunk, line 142: please rename to getPublicIp so it is the same name as the rpc method and different function name to the function in fluxCommunicaation

ZelBack/src/services/fluxCommunication.js
```diff
@@ -93,6 +93,28 @@ async function getMyFluxIP() {
   return myIP;
 }
 
+async function getBenchPublicIp() {
```
TheTrunk, line 96: this function getBenchPublicIp and restartNodeBenchMarks are actually not needed, can be removed and simplified.
Please paste similar code directly to checkMyFluxAvailability where you are calling the 
await daemonService.restartNodeBenchmarks(); directly

ZelBack/src/services/fluxCommunication.js
```diff
@@ -1081,7 +1103,14 @@ async function checkMyFluxAvailability(nodelist) {
     }
     if (resMyAvailability.data.status === 'error' || resMyAvailability.data.data.message.includes('not')) {
       log.error(`My Flux unavailability detected from ${askingIP}`);
-      // Asked Flux cannot reach me
+      // Asked Flux cannot reach me lets check if ip changed
+      const publicIp = await getBenchPublicIp();
+      if(publicIp && publicIp !== myIP){
+        myIP = publicIp;
+        await restartNodeBenchMarks();
+        await serviceHelper.delay(2*60*1000); //lets wait two minutes
```
TheTrunk, line 1111: spaces between * for readability

TheTrunk (changes requested):
Concept ack, cosmetic and simplification changes requested. 
Shall be merged/released AFTER fork when we are sure such a method exists (bench 2.0.0 is mandatory)

ZelBack/src/services/benchmarkService.js
Cabecinha84: Adjusted.

ZelBack/src/services/fluxCommunication.js
Cabecinha84: Done

ZelBack/src/services/fluxCommunication.js
Cabecinha84: Sorry, this part didn't understood

Cabecinha84:
Tadeas I would like to have this on the version 1.14.0 that you have made the PR.
It's not mandatory for having new bench version installed. Has you can see if there is no success calling the getpublicip it will mark the DOS status like it was doing before these new calls.

TheTrunk (approved):

PR #247, opened 2021-03-24
Helper update

XK4MiLX:


TheTrunk (changes requested):
helpers shall for a period of time contain both zelcash/zelbench and flux/fluxbench

PR #252, opened 2021-04-03
fix download/show debug logs for daemon/benchmark

XK4MiLX:


ZelBack/src/services/fluxService.js
```diff
@@ -3,7 +3,7 @@ const path = require('path');
 const config = require('config');
 const fullnode = require('fullnode');
 const fs = require('fs').promises;
-
+const fsPromises = fs.promises;
```
TheTrunk, line 6: This is not correct as above fs is fs promise already

TheTrunk (changes requested):

PR #260, opened 2021-04-06
v1.14.0

TheTrunk:
- First steps for no need for static ip
- Fix debug logs for daemon/benchmark
- Bump benchmark version
- Minor service helper improvements
- Dynamic IP support


Cabecinha84 (approved):
Ack, should be released after apt is updated with bench v2.1.1

PR #268, opened 2021-04-22
added  fallocate

XK4MiLX:
- removed verification by dd
- replaced dd by fallocate


Cabecinha84 (approved):
Ack. Good Job. Install process is so fast now.

TheTrunk (approved):

PR #271, opened 2021-05-01
v1.15.0

TheTrunk:
- Improvements on application volume creation using fallocate instead of dd resulting in very fast creation
- Creation of Confirmation Transaction in case of IP address changes. A restart of a node is needed for correct functionality.
- UI improvements
- Naming corrections to FluxNodes
- Kadena application version bump

This release is mandatory for Kadena operators to be applied before 6th of May.


Cabecinha84 (approved):
Ack

Cabecinha84:
Please adjust our chainweb docker before release new kda docker image.
The healthy check should be adjusted the same way you did on the api to check if a node is synced.

Cabecinha84:
And please also update kda bootstrap on flux servers. Thank you.

TheTrunk:
both comments addressed

PR #294, opened 2021-05-27
Add userVolumeOwnership to appSpecs

Cabecinha84:
Polka Docker Creates a user inside docker file and that user needs to have ownership on the volume that was created for the docker app or the docker doesn't start with a permission error.

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -1312,8 +1312,8 @@ async function createAppVolume(appSpecifications, res) {
       res.write(serviceHelper.ensureString(mountingStatus2));
     }
 
-    if (appSpecifications.userVolumeyOwnership) {
-      const execUserOwner = `sudo chown ${appSpecifications.userVolumeyOwnership}.${appSpecifications.userVolumeyOwnership} ${appsFolder + appId} -R`;
+    if (appSpecifications.userVolumeOwnership) {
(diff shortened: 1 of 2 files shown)


Cabecinha84:
To test it has a globalapp you can use this:
{ // app specifications
      version: 2,
      name: 'KusamaNode', // corresponds to docker name and this name is stored in apps mongo database
      description: 'Kusama is a "canary network" for Polkadot, an early unaudited release of the code that is available first and holds real economic value. For developers, Kusama is a proving ground for runtime upgrades, on-chain governance, and parachains.',
      repotag: 'parity/polkadot:latest',
      owner: '1hjy4bCYBJr4mny4zCE85J94RXa8W6q37',
      ports: [31113, 31112, 31111],
      containerPorts: [30333, 9933, 9944],
      domains: ['', '', ''],
      tiered: false,
      cpu: 0.8, // true resource registered for app. If not tiered only this is available
      ram: 1800, // true resource registered for app
      hdd: 20, // true resource registered for app
      enviromentParameters: [],
      commands: ['--base-path', '/chaindata', '--chain', 'kusama', '--name', 'RunOnFlux', '--rpc-external',
        '--ws-external', '--rpc-methods', 'Safe'],
      containerData: '/chaindata', // cannot be root todo in verification
      hash: 'localSpecificationsVersion1', // hash of app message
      height: 0, // height of tx on which it was
      userVolumeOwnership: '1000',
    },

Cabecinha84:
More info about it here: https://github.com/paritytech/polkadot/blob/master/doc/docker.md

TheTrunk:
- option to be evaluated for future fork, specification update.

PR #301, opened 2021-06-03
v1.17.0

TheTrunk:
- Movement to runonflux/flux repository
- Folding@Home for ARM is now available as local application
- Adjustments of requirements for KDA and Folding@Home applications
- Disabling web ui on Folding@Home
- Helper scripts updated for runonflux organization and new apt repository
- Adjustments for fluxd, fluxbenchd
- Dashboard now reflects complete available resources of Flux network, not just minimum requirements
- Improvements on temporary application installation (local apps)
- Prohibit application installation if node is not confirmed
- Smoother curve on application installation depending on number of apps on the network
- Minor bug fixes and improvements


ZelBack/src/services/appsService.js
```diff
@@ -3598,19 +3648,32 @@ async function updateAppGlobalyApi(req, res) {
...
-async function installTemporaryLocalApplication(req, res, applicationName) {
+async function installTemporaryLocalApplication(req, res) {
   try {
+    let { appname } = req.params;
+    appname = appname || req.query.appname;
+
+    if (!appname) {
+      throw new Error('No Flux App specified');
+    }
+    const isNodeConfirmed = await isNodeStatusConfirmed();
```
TheTrunk, line 3659: we shall allow application installation of local apps if user is offline

ZelBack/src/services/appsService.js
```diff
@@ -3598,19 +3648,32 @@ async function updateAppGlobalyApi(req, res) {
...
       const allApps = await availableApps();
-      const appSpecifications = allApps.find((app) => app.name === applicationName);
+      const appSpecifications = allApps.find((app) => app.name === appname);
       if (!appSpecifications) {
         throw new Error('Application Specifications not found');
       }
 
+      const tier = await nodeTier();
+      if (appname === 'KadenaChainWebNode' && tier === 'basic') {
+        throw new Error('KadenaChainWebNode is not allowed to install on Cumulus Node.');
```
TheTrunk, line 3673: should be prevented by default as of not enough available space (?)

ZelBack/src/services/appsService.js
```diff
@@ -1936,6 +1958,10 @@ async function removeAppLocallyApi(req, res) {
 
 async function checkAppRequirements(appSpecs) {
   // appSpecs has hdd, cpu and ram assigned to correct tier
+  const isNodeConfirmed = await isNodeStatusConfirmed();
```
TheTrunk, line 1961: I would add the check just to tries on spawning an app

ZelBack/src/services/appsService.js
```diff
@@ -389,6 +389,20 @@ async function nodeTier() {
   throw new Error('Unrecognised Flux Node tier');
 }
 
+async function isNodeStatusConfirmed() {
+  const response = await daemonService.getZelNodeStatus();
```
TheTrunk, line 393: do we want to throw an error or just simple true false function?

ZelBack/src/services/appsService.js
Cabecinha84: For now yes (forgot about disk space), but If we start using machine specs instead of pre defined specs for tier, this will solve the situation. :)

ZelBack/src/services/appsService.js
Cabecinha84: saw that you already resolved all the situations :D

ZelBack/src/services/appsService.js
```diff
@@ -2759,10 +2753,9 @@ async function availableApps(req, res) {
       hddsuper: 5,
       hddbamf: 5,
       enviromentParameters: [`FOLD_USER=${userconfig.initial.zelid}`, 'FOLD_TEAM=262156', 'FOLD_ANON=false'],
-      commands: [
-      ],
+      commands: [],
       containerData: '/config',
-      hash: 'localSpecificationsVersion1', // hash of app message
+      hash: 'localSpecificationsFoldingVersion1', // hash of app message
```
Cabecinha84, line 2758: Don't we also have to adjust Folding@Home and Kadena hash, for the soft redeploy to happen with the new specs?

Cabecinha84 (approved):
Ack, please just check my comment about hashes. if hashes of current folding@home and kadena remain, will the app still be soft redeploy to the new specs?

ZelBack/src/services/appsService.js
TheTrunk: I updated this hash because somewhere we are refferencing applications by hash. And I think we already used localSpecificationsVersion1 for KDA before. The hash should be strictly unique. It shall not cause any issue since that version of app does not exist on the network but just to be safe I changed it on this app.

TheTrunk: And yes. If we dont change the hash of kadena app and folding app, they will not be redeployed and will continue run according to old specifications

TheTrunk: KDA has a network update in 2 weeks so I was thinking we just leave it for now without changing the hash so nodes dont redeploy the app ( as there is ton of space on the network we do not really need it)

Cabecinha84:
So let's update also the hash of ***@***.*** for amd64 devices, so that
one generates a soft redeploy for those running, it will lower the ram used
and disable the web client.

A quinta, 3/06/2021, 13:30, TheTrunk ***@***.***> escreveu:

PR #324, opened 2021-09-06
v1.21.0

TheTrunk:
- Updates update scripts for smooth transition to FluxOS v2
- This version is mandatory to be enforced prior Flux v2 update otherwise update will fail.


Cabecinha84 (approved):
Ack

PR #329, opened 2021-09-08
v2.0.2

TheTrunk:
Fix HW check


Cabecinha84 (approved):
Ack. Code ok, but I can't test it now.

PR #335, opened 2021-09-11
v2.1.0

TheTrunk:
- Update economics dashboard
- Fixes several conditions on applications
- Fixes message processing
- Resolves missing application messages, specifications caused by wrong message processing

Previous message processing was purely chronological and did not count cases of application messages coming in random order. Previous processing was always looking just for its local knowledge of application that is in global specifications - application that is up to date right now running on the network based on local node knowledge. The update was based on local knowledge of existing specifications not on exactly previous message of specifications. This was wrong. In case node joins the network, there was no way for the node to update the apps specifications based on previous messages. Application that already expired based on local node knowledge would never registered as application available on the network as a whole.


Cabecinha84 (approved):
Ack. Please review my last two commits. I also suggest to reduce two half the flux costs on apps spawning reflecting the rise of $Flux price on the last weeks.

Cabecinha84:
Lets delete old front end?

PR #343, opened 2021-09-30
v2.1.5

Cabecinha84:
- Update FluxBench min version allowed 2.3.1;
- New Anti-Cheating Measures;
- Show on bottom right FluxOs instead of Flux version.

This version should be released after enforcement of FluxBench 2.3.1;

### DIFF
--- a/ZelBack/src/services/fluxCommunication.js
+++ b/ZelBack/src/services/fluxCommunication.js
@@ -23,7 +23,7 @@ const incomingPeers = []; // array of objects containing ip
 let dosState = 0; // we can start at bigger number later
 let dosMessage = null;
 
-const minimumFluxBenchAllowedVersion = 223;
+const minimumFluxBenchAllowedVersion = 231;
 let storedFluxBenchAllowed = null;
 
 // my external Flux IP from benchmark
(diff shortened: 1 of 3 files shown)


ZelBack/src/services/fluxCommunication.js
```diff
@@ -52,7 +52,20 @@ async function isFluxAvailable(ip) {
   try {
     const fluxResponse = await serviceHelper.axiosGet(`http://${ip}:${config.server.apiport}/flux/version`, axiosConfig);
     if (fluxResponse.data.status === 'success') {
-      return true;
+      try {
+        const fluxBenchInfoResponse = await serviceHelper.axiosGet(`http://${ip}:${config.server.apiport}/benchmark/getinfo`, axiosConfig);
```
TheTrunk, line 56: Note: Need to be careful if the request of the other node won't time out as we now have 2 calls to the node which may exceed timeout limit of the node asking

ZelBack/src/services/fluxCommunication.js
```diff
@@ -1103,7 +1116,7 @@ async function checkMyFluxAvailability(retryNumber = 0) {
   }
   let availabilityError = null;
   const axiosConfig = {
-    timeout: 7000,
+    timeout: 11000,
```
TheTrunk, line 1119: Good I see it reflected here, shall be sufficient

TheTrunk: Maybe we shall do it in 2 step release, as if this check is not applied it may be an issue. However not getting a response of the 2 requests within 7 seconds is highly unlikely anyway

TheTrunk (approved):
todo rebuild UI so change is reflected

Cabecinha84:
Maybe we should change to make both calls at same time and put back the 7
seconds to get the answer? What do you think?

A domingo, 3/10/2021, 12:36, TheTrunk ***@***.***> escreveu:

TheTrunk:
@Cabecinha84 I just realised that we are actually doing 2 of the same calls there
1. We are asking for verison of Flux - and so determining that the Flux is accessible from outside if the call is successful
2. We are asking for benchmark version
-> In 2nd call we are asking for benchmark through Flux, this ensures that 1st call is not needed as with this call we already verified that Flux is working since we got a response
-> in 1st call we are getting version of Flux. Version of Flux ensures some enforcement on the FluxBench so 2nd call is essentially not needed.

=> Asking version of Flux is sufficient and it is the lightest possible call. Lets just use this ask for version and ensure its higher than 2.1.4 (this would not enforce the newest benchmark but we can bump it in the next release OR enforce this v2.1.5 through timestamp)
=> This would be more light and faster verification while doing the same.

Cabecinha84:
Just reading on mobile but seems fine. We release first a version just with
bump min fluxbench. And a next release to on that validity check check
fluxOS min version that needs to be the one we release first with bumped
fluxbench. Only one call, and we could put timeout again of 7 seconds.
Seems a good call!

A domingo, 3/10/2021, 15:08, TheTrunk ***@***.***> escreveu:

Cabecinha84:
Looks good.

PR #353, opened 2021-11-07
v2.2.2

Cabecinha84:
Update Economics to Use api.runonflux.io to get nodecount


Cabecinha84:
this should be merged also to development and deleted after.

TheTrunk (approved):

TheTrunk:
Can be changed to local daemon call in next update

PR #365, opened 2021-11-11
Add my zelID to whitelist

AlenasAlcauskas:
Hi,
I'd like my zelID whitelisted, if possible. I'm creating an app, which is not yet available, and one of it's features is a payment gateway for flux, and it appears I cant create or manage transactions without a whitelisted zelID.
If you need more info or I'm doing something incorrectly, please let me know.
Thanks


TheTrunk:
Hi @AlenasAlcauskas
We would like to see you developing on Flux! Would you still like us to whitelist your zelid?

AlenasAlcauskas:
Hi,
Thank you. It appears to be unnecessary at this point, as it will not provide a way to create transactions.
I will submit it once again when I have my project ready on docker.
Thanks again

PR #370, opened 2021-11-24
v3.0.0

TheTrunk:
v3.0.0 Compose
- v4 application specifications define Composed applications. Application can now be composed from up to 5 components.
Parameters:
```
    const appSpecs = {
      version: 4, // int
      name: 'FoldingAtHomeB', // string
      description: 'Folding @ Home is cool :)', // string
      owner: '1CbErtneaX2QVyUfwU7JGB7VzvPgrgc3uC', // string
      compose: [ // array of max 5 objects of following specs
        {
          name: 'Daemon', // string
          description: 'Main ddaemon for foldingAtHome', // string
          repotag: 'yurinnick/folding-at-home:latest',
          ports: '[30001]', // array of ints
          containerPorts: '[7396]', // array of ints
          domains: '[""]', // array of strings
          enviromentParameters: '["USER=foldingUser", "TEAM=262156", "ENABLE_GPU=false", "ENABLE_SMP=true"]', // array of strings
          commands: '["--allow","0/0","--web-allow","0/0"]', // array of strings
          containerData: '/config', // string
          cpu: 0.5, // float
          ram: 500, // int
          hdd: 5, // int
          tiered: true, // bool
          cpubasic: 0.5, // float
          rambasic: 500, // int
          hddbasic: 5, // int
          cpusuper: 1, // float
          ramsuper: 1000, // int
          hddsuper: 5, // int
          cpubamf: 2, // float
          rambamf: 2000, // int
          hddbamf: 5, // int
        },
      ],
      instances: 10, // int
    };
```
Additional specs improvements:
- Application updates of v4 can now specify different repotags and ports
- ports and containerPorts are now treated as integers (previously were recorded as strings, no impact change)
- application name and composition name cannot be changed with an update
- Installation of application follow an order in array. Uninstallation is reversed order. This can serve as depends on functionality

Additional changes
- Improvements on Flux communication
- Fixes bug where app was able to be deployed from node that was not properly communicating with network
- 5x App price reduction 
- Code reorganisation and improvements
- Separation of docker, general and fluxshare services
- Nodes will now use all their available hardware to accompany applications
- Cumulus with SSD disks will now be spawning applications as well
- Adjustments in permissions to API calls
- Additional API calls
  - /apps/verifyappregistrationspecifications
  - /apps/verifyappupdatespecifications
  - /apps/deploymentinformation
- Code simplification, removal of old unused code
- Additional minor improvements and fixes
- UI adjustments
- UI uses more Flux calls rather than relying on its own functions
- Possibility to to connect to peer on UI
- Adjustments of inputs on UI
- **Blockheight fork 1004000**

_Scheduled release at blockheight 1000000._


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "flux",
-  "version": "2.2.4",
+  "version": "2.2.53",
```
Cabecinha84, line 3: this needs to be adjusted to 3.0.0

Cabecinha84:
Can't we update Mongodb version? Probably would solve most of the problems of mongodb crashing when it's syncing the explorer from scratch.

TheTrunk:
Yes upgrading mongodb is planned.
However as of this PR complexity and lots of breaking changes in v4 of mongodb I would prefer for it to be done later. The upgrade should be more carefully evaluated, flux code adjusted and tested.
for future reference mongodb changelog update path
https://github.com/mongodb/node-mongodb-native/blob/HEAD/docs/CHANGES_4.0.0.md
We aren't using any deprecated methods but some code adjustments are certainly needed

Cabecinha84 (approved):
Ack

PR #372, opened 2021-11-29
v3.0.2

TheTrunk:
- Fixes v4 specifications typo of enviromentParameters to environmentParameters
- removes deprecated api calls
- removes obsolete functions
- adds rate limiting on Flux communication
- adds origin checks for blocking public keys
- fixes ports uniqueness check


Cabecinha84 (approved):
Ack

PR #378, opened 2021-12-01
Update zelids.json

Esmeralddd:
Added my zelID so I can deploy apps on Flux


Cabecinha84:
ZelId no longer is required to be on whitelist. If you check discord ann you will see it. Anyone can spawn a app that is whitelisted. Cheers.

PR #382, opened 2021-12-04
Update repositories.json

Esmeralddd:


TheTrunk:
That repository is already whitelisted

PR #392, opened 2021-12-09
Marketplace

sygem:


package.json
```diff
@@ -93,6 +93,7 @@
     "qs": "~6.10.1",
     "store": "~2.0.12",
     "stream": "~0.0.2",
+    "vue-form-wizard": "^0.8.4",
```
TheTrunk, line 96: please move to devDependencies

rorysan333 (approved):

package.json
sygem: Done

TheTrunk:
tack. We can add additional confirmation button that will prevent clicking on Finish button if the app was not yet paid for as it is impossible to pay for it if the pop-up window closes

PR #399, opened 2021-12-11
Update repositories.json

Archibaldwuwu:
Hi,this is docker images support sshd.
This is the source code,https://github.com/Archibaldwuwu/centos-sshd


TheTrunk:
Hi, sorry but this is not an application we can allow. It basically creates a dockerized OS to which you can log into and in the end run anything. That is not a purpose of Flux at this time. Highly appreciate the transparency and source.
We have had a team discussion internally how to handle similar applications, images and are looking into

PR #408, opened 2021-12-15
Update repositories.json

miningtheblock:
Add my test repository to discover flux


TheTrunk (approved):
tACK
POC: http://173.212.252.18:3000/

PR #428, opened 2021-12-29
added fluxwatchtower for docker image autoupdate

XK4MiLX:
- fluxwatchtower will be installed on first run on FluxOS
- check interval random from 7500-35000 sec. 
- --cleanup flag enable ( removes old images after updating )
- --label-enable flag enable for updating only image with  label  (  com.centurylinklabs.watchtower.enable="true" )


Cabecinha84:
@XK4MiLX can you make the PR to the development branch instead of master? Thanks.

TheTrunk:
base changed

Cabecinha84:
Cool, didn't know I had permissions to do it :)

Cabecinha84:
@XK4MiLX is this still valid? Will update all docker images? No need to create new version of PreSearch Node app because of that flag. Like we discussed, update all docker images to latest if a new version is released? Thanks

PR #435, opened 2022-01-07
Fixed 'My Apps' name error under Global Apps

sygem:


Cabecinha84:
No need for PR. We will merge the dev branch with more things that needs to be added.

PR #436, opened 2022-01-07
Add zelid

arkitoure:
add zelid to whitelist for larger project deployment


TheTrunk:
Hi, zelid whitelist is no longer required

PR #441, opened 2022-01-11
Development fluxwatchtower

XK4MiLX:
- fluxwatchtower will be install on first run on FluxOS
- check interval random from 7500-35000 sec.
- cleanup flag enable ( remove old image after update )
- autoupdate every docker image ( in future possible add label to continer that will prevent autoupdate )

### DIFF
--- a/ZelBack/src/services/fluxService.js
+++ b/ZelBack/src/services/fluxService.js
@@ -2,6 +2,7 @@ const cmd = require('node-cmd');
 const path = require('path');
 const config = require('config');
 const fullnode = require('fullnode');
+const util = require('util');
 const fs = require('fs');
 
 const fsPromises = fs.promises;
@@ -744,6 +745,15 @@ async function getNodeTier(req, res) {
   }
 }
 
+async function InstallFluxWatchTower() {
+    const nodedpath = path.join(__dirname, '../../../helpers');
+    const exec = `cd ${nodedpath} && bash fluxwatchtower.sh`;
+    const cmdAsync = util.promisify(cmd.get);
+    const cmdres = await cmdAsync(exec);
+    console.log(cmdres);
+}
+
+
 module.exports = {
   startDaemon,
   updateFlux,
@@ -780,4 +790,5 @@ module.exports = {
   adjustKadenaAccount,
   fluxBackendFolder,
   getNodeTier,
+  InstallFluxWatchTower,
 };
--- a/ZelBack/src/services/fluxService.js
+++ b/ZelBack/src/services/fluxService.js
@@ -746,13 +746,16 @@ async function getNodeTier(req, res) {
 }
 
 async function InstallFluxWatchTower() {
+  try {
     const nodedpath = path.join(__dirname, '../../../helpers');
-    const exec = `cd ${nodedpath} && bash fluxwatchtower.sh`;
+    const exec = `cd ${nodedpath} && sh fluxwatchtower.sh`;
     const cmdAsync = util.promisify(cmd.get);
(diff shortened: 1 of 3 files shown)


Cabecinha84 (approved):
Ack

helpers/fluxwatchtower.sh
```diff
@@ -0,0 +1,33 @@
...
+
+if [[ "$apps_check" == "" ]]; then
+
+  echo "Stopping fluxwatchtower..."
+  docker stop fluxwatchtower > /dev/null 2>&1
+  sleep 2
+  echo "Removing fluxwatchtower..."
+  docker rm fluxwatchtower > /dev/null 2>&1
+  echo "Downloading containrrr/watchtower image..."
+  docker pull containrrr/watchtower:latest > /dev/null 2>&1
```
TheTrunk, line 14: I think we shall fork it and host it on our docker hub to have control over what exactly is being installed since its part of FluxOS now.
Since it has a prefix of flux in the name it would fall under exceptions of flux applications and same rules would apply. We shall reflect it on FluxOS UI (that this app has exception) and forbid application message broadcasting.
Furthermore we shall rename it from fluxwatchtower to flux_watchtower. This would resolve all potential application name conflicts as apps on flux can only use a-zA-Z0-9. This also simplifies/unifies other exceptions we can introduce to some special apps with the _ symbol in front.

ZelBack/src/services/serviceManager.js
```diff
@@ -6,6 +6,8 @@ const explorerService = require('./explorerService');
 const fluxCommunication = require('./fluxCommunication');
 const appsService = require('./appsService');
 const daemonService = require('./daemonService');
+const fluxService = require('./fluxService');
+
```
TheTrunk, line 10: double line

ZelBack/src/services/fluxService.js
```diff
@@ -744,6 +745,15 @@ async function getNodeTier(req, res) {
...
 }
 
+async function InstallFluxWatchTower() {
+    const nodedpath = path.join(__dirname, '../../../helpers');
+    const exec = `cd ${nodedpath} && bash fluxwatchtower.sh`;
+    const cmdAsync = util.promisify(cmd.get);
+    const cmdres = await cmdAsync(exec);
+    console.log(cmdres);
+}
+
```
TheTrunk, line 755: double line

helpers/fluxwatchtower.sh
```diff
@@ -0,0 +1,33 @@
...
+  echo "Interval: $random sec."
+  apps_id=$(docker run -d \
+  --restart unless-stopped \
+  --name fluxwatchtower \
+  -v /var/run/docker.sock:/var/run/docker.sock \
+  containrrr/watchtower \
+  --cleanup --interval $random 2> /dev/null)
+
+  if [[ $apps_id =~ ^[[:alnum:]]+$ ]]; then
+    echo "FluxWatchtower installed successful, id: $apps_id"
```
TheTrunk, line 26: Reflection of name flux_watchtower. Unification of name

TheTrunk (changes requested):
Just very slight adjustments needed mainly name change is needed and later on transition to our docker hub organisation
Later on we can build more in depth control of watchtower inside of fluxos

helpers/fluxwatchtower.sh
TheTrunk: watchtower will not broadcast status as its not located in installed apps database

TheTrunk (approved):

PR #442, opened 2022-01-14
Update repositories.json

champ8644:
add itzg/minecraft-server:java8-multiarch for playing modded minecraft on forged <1.17
according to this
https://github.com/itzg/docker-minecraft-server/blob/f29f098b2628666c50a2cb3448eecb61614fef6d/README.md?plain=1#L302-L304


TheTrunk:
Hi, please add your repository at the end of the array :)
Thank you for looking into Flux!

champ8644:
I've added it to the position after "itzg/minecraft-server:latest" already, but if you'd like this to be append to the end of the array then fine. done

TheTrunk (approved):

PR #444, opened 2022-01-17
v3.5.0

TheTrunk:
- Flux Watch Tower
  - Flux Apps are now automatically updated to latest version of available image every couple hours
- Automatically stop all non Flux Apps on the node
  - Docker containers that are not tied to FluxOS are being automatically stopped


Cabecinha84 (commented):
Ack.

Cabecinha84 (approved):

PR #445, opened 2022-01-19
v3.6.0

TheTrunk:
- Fix flux communication issues where multiple nodes can share the same node key

Previous system did not expected light node addition that share the same node key. Current expansion of zelcore light nodes has a bigger impact where many nodes have the same pubKey resulting in finding wrong node, refusing their messages and adding them to blocked list.


Cabecinha84 (approved):
Ack. Great work!

PR #446, opened 2022-01-23
v3.7.0

TheTrunk:
- Fix installation of application on additional instances
- Update Kadena Chainweb Node application

TBD: Next week we will be releasing another update of Kadena Chainweb Node - we will use latest tag, change our monitoring system and mainly bump the space requirements to recommended 150GB of SSD. 
After that preparation for global launch will happen in 2 ways - 2 x 100 instances for composed app consisting of chainweb node and chainweb data. 3 x 100 instances of chainweb node only. Exact specs TBD.


TheTrunk:
I believe we can just launch it globally on the network next week with couple of instances, skipping the middle step.

PR #448, opened 2022-01-29
Update repositories.json

ibettokens:
Added docker image for dapp to run on flux.

https://github.com/ibettokens/ibettoken

Source code for docker image is public in guthub.


TheTrunk:
Hi,
Thank you for your interest in Flux.
Can you please write some summary of the application and what the image is doing? Is it a gambling application? Are some real funds being used? // as that may possess legal issues in some countries, testnet coins are fine.
POC: http://62.171.163.150:24555

ibettokens:
These run off of matic tokens currently. we are planning to base it off of Kadena smart contacts and use KDA tokens. It does not pose any legal issues.  We will gradually enhance the dapp to encompass more compliance.

The basic Idea is to have shared, dispersed, and distributed tokens used for peer-to-peer betting.

Shared
Users of IBET tokens will receive bonus token for sharing it with others. It is designed to proliferate sharing. over all goal is to reach 22 million users.

group
Distributed
Contract owner only owns 1 coin at the beginning. only way to mint new coins is by sharing coins with others. this will continue until max supply token is shared.

settings
Dispersed
IBET coins are designed to be dispered among many users rather than just few people. the are rules in places as "sharing limit within 7 days" and others just to maintain a disperse set of users.

TheTrunk (approved):

TheTrunk:
Cool, sorry for waiting little longer for thus review.
Thank you for contributing 👍

PR #458, opened 2022-02-14
Update repositories.json

HexxaVerse:
Added whitelisting entry for HexxaVerse/hexxaverse-portal


TheTrunk:
Hi, the repository is not on docker hub and also is missing a tag

PR #473, opened 2022-02-28
JSDoc documentation - generalService.js file

johnhanlon86:
Created generalService.js.html file which shows function descriptions, parameter types and return types.


johnhanlon86:
New pull request opened to replace this one.

PR #474, opened 2022-03-01
Message helper + more tests

JacekAdamczyk:


ZelBack/src/services/messageHelper.js
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @param {object} message
+ * @param {string} address
+ * @param {string} signature
+ * @param {string} strMessageMagic
+ * @param {string} checkSegwitAlways
+ *
+ * @returns {bool} isValid
+ */
+function verifyMessage(message, address, signature, strMessageMagic, checkSegwitAlways) {
```
TheTrunk, line 114: The verifyMessage and signMessage are unrelated to the creation of error/data messages.
They are more related to signing, verifying, id services. I think they shall be located in different file

TheTrunk (changes requested):

ZelBack/src/services/messageHelper.js
JacekAdamczyk: @TheTrunk I can move them to verification helper if that works?

TheTrunk: Yes that works 👍

TheTrunk (approved):

PR #479, opened 2022-03-06
v3.10.0

TheTrunk:
- Multiport support
- UPnP
- Additional tests
- JSdoc source documentation
- code reorganisation


ZelBack/src/services/fluxCommunication.js
```diff
@@ -1258,7 +1263,7 @@ async function checkMyFluxAvailability(retryNumber = 0) {
   }
   let askingIpPort = config.server.apiport;
   if (askingIP.includes(':')) { // has port specification
-    // it is ipv6
```
w2vy, line 1261: How does this work for IPv6 and :: in the IP portion

ZelBack/src/services/idService.js
```diff
@@ -23,52 +29,52 @@ async function confirmNodeTierHardware() {
...
     } else if (tier === 'basic' && collateral === 1000) {
       if (nodeRam < 3) {
         throw new Error(`Node Total Ram (${nodeRam}) below new Cumulus requirements`);
       }
-      if (nodeCpuCores < 4) {
-        throw new Error(`Node Cpu Cores (${nodeCpuCores}) below new Cumulus requirements`);
+      if (nodeCpuThreads < 4) {
+        throw new Error(`Node Cpu Threads (${nodeCpuThreads}) below new Cumulus requirements`);
       }
     }
```
w2vy, line 79: else throw new Error undefined tier or incorrect collateral

Stylistically I might have made this a table to search for tier and collateral

docs/idService.js.html
```diff
@@ -0,0 +1,874 @@
...
+        throw new Error(`Node Cpu Threads (${nodeCpuThreads}) below new Nimbus requirements`);
+      }
+    } else if (tier === 'basic' &amp;&amp; collateral === 1000) {
+      if (nodeRam &lt; 3) {
+        throw new Error(`Node Total Ram (${nodeRam}) below new Cumulus requirements`);
+      }
+      if (nodeCpuThreads &lt; 4) {
+        throw new Error(`Node Cpu Threads (${nodeCpuThreads}) below new Cumulus requirements`);
+      }
+    }
```
w2vy, line 107: else throw new error undefined tier

w2vy (commented):
Just had a few comments/questions

HomeUI/src/App.vue
```diff
@@ -103,11 +103,25 @@ export default {
...
     document.documentElement.setAttribute('dir', isRTL ? 'rtl' : 'ltr');
+    this.adjustHostnameAndPort();
   },
   created() {
     this.getZelIdLoginPhrase();
   },
   methods: {
+    adjustHostnameAndPort() {
+      const { hostname, port } = window.location;
+      const regex = /[A-Za-z]/g;
```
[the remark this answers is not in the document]
TheTrunk: This method is to determine if node is accessed over IP or via some domain.  If it is an IP, then we store the IP and port to know how to communicate with the flux backend from this UI

TheTrunk:
Flux is not compatible with ipv6. It is not possible to have IPv6 node on the network. When that comes, many more places will need to be rewritten

docs/idService.js.html
TheTrunk: Adjusted 👍

ZelBack/src/services/idService.js
TheTrunk: Adjusted 👍

ZelBack/src/services/fluxCommunication.js
TheTrunk: Flux is not compatible with ipv6, obsolete comment removal

ZelBack/src/services/fluxCommunication.js
```diff
@@ -873,7 +875,7 @@ async function initiateAndHandleConnection(connection) {
   websocket.onopen = () => {
     outgoingConnections.push(websocket);
     const peer = {
-      ip: connection,
+      ip, // can represent just one ip address, multiport
```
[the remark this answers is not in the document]
TheTrunk: `ip,` is a shorthand for `ip: ip,`

Cabecinha84:
Please check my commit -> https://github.com/RunOnFlux/flux/pull/479/commits/8951caa40580844ecc2ab77ce3277e995e95b4f2

Hardcoded port 16126, I think we can release it this way now.

Cabecinha84 (approved):
Ack

TheTrunk:
> Won't the app name start with Zel_ in this specific case?

This function `getAppName` removes the zel or flux prefix from the real app name (if present). Then in Flux backend those prefixes are added/removed per the function needs.

PR #495, opened 2022-03-14
Update repositories.json

SmarticoApps:


TheTrunk:
Hi, please do not modify existing whitelisted images. Add a new image at the end of the array.

PR #501, opened 2022-03-18
Flux communication tests

JacekAdamczyk:


JacekAdamczyk:
This pr ready for review.

TheTrunk (changes requested):
Can 
```
/utils/outgoingConnections
/utils/incomingConnections
/utils/outgoingPeers
/utils/incomingPeers
```
be just one file named /utils/establishedConnections for example?
Does it work correctly as in 2 files I see eg.
`const { outgoingPeers } = require('./utils/outgoingPeers');
`
and then we are pushing to the array etc. So I am curious if both of the files have the same list in the end or they have separate list. Then fluxCommunication is also exporting the outgoingPeers.

JacekAdamczyk:
@TheTrunk 
Yes, I can move that into single file, I will test it, but I'm pretty sure it'll work the same
And yes, this is why I extracted it into a separate module, so the data is shared between other modules and array keeps its contents throughout other modules.
Exports in flux communication of outgoing peers etc. - no need for them, forgot to remove it, my bad.

TheTrunk (approved):

PR #509, opened 2022-03-21
Update repositories.json

Kyuubi2709:
Added oasis node for test deployment


TheTrunk:
This seems to be whitelisting more repositories including mining which is prohibited

PR #517, opened 2022-03-24
v3.14.0

TheTrunk:
- fix app spawning on nodes without any app running

$and/$or/$nor must be a nonempty array
- If no running app was present on a node, our query to mongo was following
```
const query = {
    $or: appsQuery,
  };
  ```


Cabecinha84 (approved):
Ack

PR #530, opened 2022-03-31
Static ports for marketplace apps

sygem:
The ports are now defined in the marketplaceApps.json file in fluxapi. Eventually the 'portSpecs' can be removed from there as well. Port-based domains will no longer be created.


TheTrunk (approved):
Note:
FDM is not creating automatic domains for marketplace apps as of enourmous amount of the same application. Those apps are excluded from fdm.

Cabecinha84 (approved):
This willl fix, installing the same app on multiportNodes.

PR #534, opened 2022-04-04
v3.15.0

TheTrunk:
- Adjust Dashboard for halving fork
- More test suite
- Additional documentation
- General improvements and fixes
- Fix UI on updating application instances


Cabecinha84 (approved):
Code review ack. I am not sure if we should be allowing the deploy of apps with all node tier resources, I think the way it was before was fine for now.

TheTrunk:
> Code review ack. I am not sure if we should be allowing the deploy of apps with all node tier resources, I think the way it was before was fine for now.

I think we should, it may be interesting to see some apps using that. I know of 1 we want to also deploy which is Cardano graphql app. That needs 30GB+ of RAM, 12+ cores and some 500GB ssd.

PR #544, opened 2022-04-14
Add newfluxdev/play2earnflux:latest

newfluxdev:
Add newfluxdev/play2earnflux:latest to repositories whitelist


TheTrunk:
Hi, can you please write some information about the application? What does it do?

newfluxdev:
> Hi, can you please write some information about the application? What does it do?

We have actually a running BO game and are having a planning to migrate our system to run on Flux. First we'll try to test with frontend (Angular) and if ok, we'll migrate our backend to run on flux (nodejs)

PR #551, opened 2022-04-18
Added ZellID for ggoldie/noderank

GGoldie:
repo already added to whitelist, adding ZellID


TheTrunk:
Zelid whitelist is no longer needed, file will be removed soon.

PR #556, opened 2022-04-20
Update repositories.json

Lazzeruz:
To add nodes for Kaspa so people don't have to boot up their own one, anytime they want to start mining Kaspa, to keep the network healthy.

So to keep it all secure and decentralized it would be nice to have more nodes to validate all the blocks and to mine to while we wait for mining pools.
Sincerely, Lazzeruz (sent by a bunch of excited Kaspa miners)


TheTrunk:
Hi, Thank you for the submission.

It will be for a bit longer discussion as we can not have mining applications on the network however just the node/validator piece is completely welcomed. 

Does the app always mine? Can the image be restricted in a way so it can never do mining?

blondfrogs:
@Lazzeruz does this node mine blocks or act like a mining pool in which miners connect with to obtain mining information required to mine?

Lazzeruz:
Hi, thank you for the consideration.
I'll check with the original person that made it into a docker image.

So, it should just act as an IP to mine towards regarding the mining. With seperate mining on the node itself, but that can be turned off. (Should be off by default)
And I'll see if I can ask for or make a docker that has the mining turned off permanently.

I'll get back to you once I have a new docker image or confirmation that this one can't mine.

Lazzeruz:
Once mining is turned off it should just act like a pool, yes.
And validating isn't intensive as far as I've noticed. As it just accepts or rejects the blocks that have already been mined.

Lazzeruz:
Ok, one of the devs checked the image and says it seems to only be running their node.
Which I think is true as the image running on my machine is only taking at most 0.1% of my CPU with no GPU load.

Lazzeruz:
Capture
This is the discord convo I had with them.

TheTrunk (approved):

PR #562, opened 2022-04-30
Add ZeroClassic daemon

zzzpotato:
Add ZeroClassic daemon

https://hub.docker.com/r/zzzpotato/zeroclassic/tags


Cabecinha84:
Hi. Thanks for the PR. I have checked and your app image exceeds the max. allowed by FluxOs, that is 500MB.
If you are able to fix it, also please resolve conflicts or create a new PR.
Also please give information of what is the app and what is the use case of it. Thank you.

zzzpotato:
Okay thanks, I will look into reducing it's size and update the PR.

This is simply a container to stand up a full node for ZeroClassic. See https://github.com/zeroclassic/ZeroClassic and https://www.zeroclassic.org/ In an effort to help decentralize nodes during sync, I'd like to use the help of flux nodes to support the network.

PR #563, opened 2022-05-03
Geolocation

Cabecinha84:
Update new app specification v5.
- Add email contacts to get fluxOs notifications (ex. app about to expire);
   -> App contacts are not visible on UI, only on app management to avoid people using it to send spam;
- Possibility to set Geolocation to spawn your dapp, both possible to choose continent and specific country;
   -> Front end allows only one continent/country but app specs is a array for both properties and backend is ready for multiselect in case we want to offer it in future;
- Added geolocation to fluxinfo so it can be removed from fluxstats call.
   
TODO: Set block height fork!

### DIFF
--- a/HomeUI/src/views/apps/GlobalApps.vue
+++ b/HomeUI/src/views/apps/GlobalApps.vue
@@ -52,6 +52,17 @@
                         title="Hash"
                         :data="row.item.hash"
                       />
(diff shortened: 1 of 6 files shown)


Cabecinha84:
New PR will be created with Geolocation V2.0 Array meeting proposal changes.

PR #571, opened 2022-05-10
Titan Shared Nodes on FluxOS

sygem:
This PR encompases the changes required to support the UI for the Titan Shared Nodes feature.

In addition to the UI, this moves the Marketplace to the top level on the side menu (because Shared Nodes and other future enhancements don't fit under 'Apps' any more). There is also a mechanism in place to restrict access to a set of ZelIDs. This will be removed for the final release.


HomeUI/src/views/apps/marketplace/SharedNodesView.vue
```diff
@@ -0,0 +1,1837 @@
...
+        },
+      });
+    };
+
+    const tier = ref('');
+    tier.value = props.tier;
+    const userZelid = ref('');
+    userZelid.value = props.zelid;
+
+    const apiURL = 'http://titantest.runonflux.io:54978';
```
Cabecinha84, line 1205: Here is the duplicated variable.

sygem: Sorry, not sure how that made it's way in there. Keep the titantest apiURL.

Cabecinha84:
Stake period between 1 and 3 weeks? This is just for alpha internal testing correct?
Max stake 1000 flux? Also just for alpha? I am thinking on some with much more flux but don't know how or don't wanna run a flux node.

Cabecinha84:
Active Nodes show me 0, is Titan active Nodes? Maybe changing the label.

Cabecinha84 (approved):
Ack despite not "able" to test staking (but I know you have been doing it and will continue. Code looks clean and loved the simplicity UI. Great Work.

Cabecinha84:
> > 
> 
> Yes, these are just alpha values. The lockup periods are defined in the backend database. I think for beta / production we plan to go with a range of 50-40,000 for the staking amount.

I would change the label to Titan Active Nodes. Great job!

TheTrunk (approved):
tack

PR #574, opened 2022-05-12
Submitting request to whitelist dapp

kadcarnfts:
Entering dapp dockerhub repo name to request whitelisting on flux io


TheTrunk:
Looks like the app is already whitelisted

TheTrunk:
Looks like the app is already whitelisted

PR #586, opened 2022-05-19
Update repositories.json | Add ngoduykhanh/powerdns-admin:latest

FliKites:


TheTrunk:
Looking forward for this deployement. We are running PDNS as well but currently on centralised servers preparing migration to flux. There are some issues to overcome - secrets and mainly syncing of database between instances

FliKites:
> Looking forward for this deployment. We are running PDNS as well but currently on centralised servers preparing migration to flux. There are some issues to overcome - secrets and mainly syncing of database between instances

I just realized that Flux won't allow me to push port 9191 to 80 which was the default config. I will need to recreate the image using a higher point. I will likely do that tonight, and then do a pull request for a new image. 

When I do that, I will also be adding the Docker image for a handshake full node. 

Is their anyway to access the shell and/or logs for a flux deployment?

TheTrunk:
Flux now whitelists entire images. So all tags of ngoduykhanh/powerdns-admin are allowed to be deployed and tags are not longer needed for a whitelist.

Flux gives app owner access to docker logs. So logging into the node where the app is running and going into the app management will show you an option for docker logs for example

FliKites:
Got it. The original repo isn't mine, so not 100% sure if I'll be able to get the updated port number ported in the original repo. I'll try though.

FliKites:
I was able to get it deployed using the original image that was whitelisted: https://pdnsadmin.app.runonflux.io/login

PR #589, opened 2022-05-20
Update repositories.json | Add Handshake Full Node Image

FliKites:
Deploy handshake full nodes on flux


FliKites:
Is their a variable that I can use when executing command that would allow me to call in the IP of the flux node? 

If so I could automate the proper configuration when deploying to multiple instances. 

["hsd", "--bip37=true", "--listen =true", "--public-host={flux_node_varibale_for_ip_address?}", "--public-port=39001", "--port=12038", "--max-inbound=3000"]

TheTrunk:
> Is their a variable that I can use when executing command that would allow me to call in the IP of the local flux node running the container?
> 
> If so I could automate the proper configuration when deploying to multiple instances.
> 
> ["hsd", "--bip37=true", "--listen=true", "--public-host={flux_node_varibale_for_ip_address?}", "--public-port=39001", "--port=12038", "--max-inbound=3000"]

It is not possible to know on which IP or IP:port an application will be spawned before it is actually spawned. So an app should use some internal service to detect the ip of a server it is running on. We cannot use it as a static parameter in applications specifications. 
FluxOS currently does not expose any global variables - that is a great suggestions 👍 

Can you use something like 0.0.0.0?

FliKites:
Unfortunately 0.0.0.0 didn't work. Loopback also didn't work. 

I am confirming with handshake core devs but i may be able to get away with removing the --public-host argument and it still working the way it needs to. 

If that is not the case we will create a script that uses wget to call an API and save the IP address then run the command.

One thing that I am kinda stumped on is getting the proper array command syntax for this command 

``` hsd-cli rpc getpeerinfo | jq -r '.[]|select(.inbound==true)|.addr|sub(":.*$";"")' | curl -s -XPOST --data-binary @- "ipinfo.io/tools/summarize-ips?cli=1"|jq -r .reportUrl ```

Hoping you can point me in the right direction. This is how I can determine if the node is advertising itself to the network properly.

TheTrunk:
That is quite tricky, would need to perform testing/searching as I only tried simple commands.

Applications migrate from node to node, so this command should be build directly into a node. Today your app can run on IPs X and one week later on IPs Y. 

I think best might be to adjust the docker image

FliKites:
Yes thinking that might be the best route. Thanks for your help.

PR #590, opened 2022-05-21
Fix arbitrary file upload vulnerability

gorgarp:


TheTrunk:
Note: Needs some testing as of major version bump and potential breaking changes

gorgarp:
Yeah, I imagine it should roll through test branch first.

TheTrunk:
Looks like v3 is still marked as not latest but as next upcoming release. Recommending staying on current v2 and observing formidable update path

PR #618, opened 2022-06-14
v3.21.0

TheTrunk:
- Geolocation service
- Fixes spawning of applications
- Fix circular import
- Adjusts checks for application geolocation spawn


Cabecinha84 (approved):
Ack

TheTrunk:
> Code review ack. Despite with your change is possible to set a country without setting continent what I not 100% fan. I don't have time to test it now but I think the changes are minimal and safe to release.

Correct that is an intentional change.
Later on the function will get rewritten to further support more in the backend - eg. X continents, X countries, negations etc

PR #622, opened 2022-06-16
I need to Deploy our Tomi website on flux

adrehman11:
we need to whitelist our dockerhub repos so that i can deploy it on flux please guide me through this process that how can i whitelist my docker hub repos.


alihm:
Please submit a pr on https://github.com/RunOnFlux/flux/blob/master/helpers/repositories.json and describe what each of your images will be doing.

PR #627, opened 2022-06-18
Update repositories.json

coolerspeex:


Cabecinha84:
Hi @coolerspeex ! Thank you for your PR and willing to use Flux Network.
Can you provide some details about what is your app and how it is going to be used?

PR #633, opened 2022-06-23
Update repositories.json

divstrong:
Added docker image for whitelist/deployment

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -275,5 +275,6 @@
    "spacepumpkin/flux-api-playground:latest",
    "megachips/flux_looking_glass:latest",
    "alekslitvinenk/openvpn",
-   "fluxfolio/fluxfolio:latest"
+   "fluxfolio/fluxfolio:latest",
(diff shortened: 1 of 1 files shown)


TheTrunk:
Hi, can you provide information about the image? What does it do and how can we test it?
Mining is currently not allowed.

divstrong:
It is just a ReactJS content build that is currently live on a Digital Ocean droplet @ https://minestrong.co/

We want to move hosting from DO to FLUX.

I don't understand the comment about 'mining not being allowed' since FLUX is 50% PoW???

TheTrunk:
> It is just a ReactJS content build that is currently live on a Digital Ocean droplet @ https://minestrong.co/
> 
> We want to move hosting from DO to FLUX.
> 
> I don't understand the comment about 'mining not being allowed' since FLUX is 50% PoW???

Perfect thank you.

People are launching for example monero miners or bandwidth miners on the Flux network or other even worse content. So we have to verify new images to stay legally compliant.

PR #635, opened 2022-06-26
Ragnar Finance protocol Add whitelist

l-dussoulier:
Protocol DEFI to Yeti on Avalanche
@RagnarFinance on twitter
App develop in react js


Cabecinha84 (approved):
Ack

PR #639, opened 2022-06-28
Update repositories.json

Kyuubi2709:
Docushield IPFS @TheTrunk @Cabecinha84


Cabecinha84 (changes requested):
The array is not valid. Should add a "," at the previous ending repo.

PR #642, opened 2022-06-29
Update repositories.json

xpiREC:
@TheTrunk @Cabecinha84


TheTrunk:
Sorry, Flux is meant to run apps not OSs as that way you can then launch practically anything

PR #647, opened 2022-07-06
Update repositories.json

xpiREC:


TheTrunk:
Hi, can you please provide description of the application?
What does it do? Is source code available? How can we tested? Is there som poc running?

PR #653, opened 2022-07-09
+ packetcrypt mining

figurestudios:
Here's a mining use-case with ballpark numbers ranging from $60 to $90 right now on. These estimates are based of the configuration 700 vCPU (100x7), 1 GB memory, & 1 GB disk and [this calculator](https://calculator.pkt.watch/).

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -1,4 +1,5 @@
 [
+   "thomasjp0x42/packetcrypt:latest",
    "yurinnick/folding-at-home:latest",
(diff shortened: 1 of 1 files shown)


Cabecinha84:
Thank you for your interest in using Flux network, unfortunately will have to decline your PR, mining apps are not allowed.

figurestudios:
Where can I see the rules of Flux? Just seems distributed at this point - not decentralized - but such a network can be useful too for cheap deployments of different kinds.

PR #654, opened 2022-07-12
Update repositories.json

wabato5634:
Add node rank for test


wabato5634:
It's a simple webserver with node rank information from tutorial to create dapp is for trying how to work.

PR #655, opened 2022-07-12
update repositories.json

Sikbik:
added ergoauctionhouse
Source code:
https://github.com/anon-real/ErgoAuctionHouse


Cabecinha84 (approved):
Ack

PR #657, opened 2022-07-14
Final changes to support Titan Beta

sygem:
- Removed the ZelID checks so Titan Beta is accessible to everyone
- Added a "maintenance mode" overlay
- API now resides on 'titan.runonflux.io'
- Added an additional warning about locking up funds when creating a stake


Cabecinha84 (approved):
Ack

PR #659, opened 2022-07-19
Update repositories.json

samirma:
This app to work as a vpn node to the peer2profit where you earn by sharing the internet.

How to test:

Register into https://peer2profit.io/

set the registered e-mail into the docker env var: email_p2p


TheTrunk:
Mining on Flux Nodes of any kind is not allowed yet.

samirma:
@TheTrunk I am sorry, since the presearch docker was approved and it is a mining project I thought that would be ok to add this project as well, since it requires even less resources than presearch.

Am I wrong?  Do you have a site with some oriendation about what is accepable or not for he project.

I don;t want to be argumentative here, I just want to understand better what this project supports use it in the future without violete any guideline.

Thanks,
Samir

PR #666, opened 2022-07-28
Titan - latest fixes

sygem:
Mostly cosmetic changes. Should also reduce the number of incorrect collateral txs we are seeing, as it is no longer possible to change the collateral amount during staking.


Cabecinha84 (approved):
Ack

blondfrogs (approved):
utACK

PR #693, opened 2022-09-02
fix functions

JacekAdamczyk:
For whatever reasons these functions were not using existing functions;
Also fixed a test that checked number of connections.


TheTrunk:
Oo nice catch 👍 
That was most likely introduced by me in the last commit. Should only be in dev branch. I still live in the era where all was in serviceHelper 😅

PR #696, opened 2022-09-05
Added WisconsinWaters Repo to whitelist

DanZirnhelt:


Cabecinha84:
Can you provide more information about the application? What does it do? How can we test it?

DanZirnhelt:
> Can you provide more information about the application? What does it do? How can we test it?

This is a website for a woodworking business. It is used to display furniture that is built by me and for potential new clients to reach out to me to commission projects. Testing should be pretty simple. I have a self hosted database so the app that is deploying on Flux can be tested by checking that the website loads.  I am very new to flux and thought this was a much better option for hosting then something like AWS or BlueHost.

Cabecinha84 (approved):
Ack. Thanks for using/test flux network.

PR #704, opened 2022-09-09
Changed owncloud core to server.

speed400m:


TheTrunk:
Hi, please do not remove whitelists. You can add it as another whitelisted image instead.

PR #712, opened 2022-09-12
Geolocation 2.0

Cabecinha84:
Geolocation 2.0:
- Created new AppSpecification to admit the new geolocation information (TODO set fork height)
- Geolocation options live from all nodes on the network;
- Continents, Countries and Regions are available to choose;
- Choose up to 5 geolocations where your app can be spawn;
- Choose up to 5 geolocations where your app cannot be spawn;
- Geolocation is an array of strings separated by '_';

Positive geo example: continentCode_continentText_countryCode_countryText_regionCode_regionText - Negative geo starts with !

The Text were added so it's not needed to be accessing FluxStats to get all geolocations on pages where data is only readable.

Management.vue was the last thing changed in difficult circunstances so might still need some love. Ex. would be nice to only get the Stats GeoInformation when and if user press update specifications tab.


TheTrunk:
Adjusted to style
- new geolocation has prefix ac for allowed and a!c for not allowed location
- fully compatible with previous geolocation scheme
- one allowed or forbidden geolocation is now in one string of the array
- Continents are represented as codes, countries are represented as codes but regions are full regionName. This is to truly know the geolocation and be nicely visible to user especially when stats api is not available or region is missing
- auto adjust max allowed instances
- works for some continents and countries even if stats api fails
- geolocations file created
Example of style: [acEU], [acEU_CZ], [acEU_CZ_Hlavni mesto praha], [a!cNA], [a!cNA_US], [a!cNA_US_Washington]

Sadly it requires a little fork of Flux currently set to happen in 2 weeks on 1230000 as current app specifications disallow submission of geolocation string bigger than 5 characters. That will change to 50 character. Limit of 10 geolocation specifications stays.
Before 1230000 only continent can be specified in app registration. Application updates that do not change its geolocation specifications will work well. In case of geolocation specification change, only continent will be taken in account.
-> That can be also caught on FDM where we specify that nodes running old version are in load balancer prior 1230000, after that height, node running new version are in load balancer

All nodes should update by the height 1230000 but it is not necessary as they can sync later on too.

One test that was not conducted is to lower the blockheight in test enviornment and try to truly register it and launch it - that part of code is quite easy so not problems expected.
After 1230000, the 1230000 mentions can be safely removed.

Cabecinha84:
Love the static data with all geolocation continents and countries. Basically this fix my worries to don't have to be always calling stats to get the get the text of codes to show on Ui. Amazing "little" rework! 👍

PR #737, opened 2022-10-03
v3.25.0

TheTrunk:
- Flux Apps Geolocation specifications update
  - Scheme changes from aXX, bYY to acXX_YY_RR and a!cXX_YY_RR, where XX is continent code, YY is country code and RR is region name. Old scheme is now considered deprecated although supported.
   - An allowed or forbidden geolocation is now in one string of the array
  - ac prefix means allowed location while a!c prefix means location is forbidden
  - It is now also possible to specify with Region accuracy
  - Geolocation options are now live from stats api with fallback to default dataset of continents, countries if unreachable - see geolocations file
  - Maximum of 10 total geolocation specification can be chosen with UI restricted to 5 allowed and 5 banned.
  - new geolocation has prefix ac for allowed and a!c for not allowed location
  - fully compatible with previous geolocation scheme
  - Continents are represented as codes, countries are represented as codes but regions are full regionName. This is to truly know the geolocation and be nicely visible to user especially when stats api is not available or region is missing
  - auto adjust max allowed instances based on selected geolocation (UI)
  - Example of style: [acEU], [acEU_CZ], [acEU_CZ_Hlavni mesto praha], [a!cNA], [a!cNA_US], [a!cNA_US_Washington]
  - **Geolocation is a geo app specifications fork set to go live on 1230000** All nodes should update to the release otherwise can be rejecting some application submissions and so falling off the network.
 - Apps Monitoring
   - Provides statistics of resources usage of app such as CPU, RAM, Disk, Net and more
   - Apps stats are now being monitored with option to:
     - stream real time app statistics
     - get per minute information of app stats for the last hour
     - get per 15 minute information of app stats for the last day
- API changes:
  - /apps/appmonitor/:appname?
    - Get monitoring statistics information about specific app or app component
  - /apps/appmonitorstream/:appname?
   - Get stream monitoring statistics information about specific app or app component
  - /apps/startmonitoring/:appname?
    - starts monitoring of an application or application component
  - /apps/stopmonitoring/:appname?/:deletedata?
    - stops monitoring of an application or application component with a possibility to delete gathered statistics data
- Minor scripts improvements
- UI improvements
- Dependencies update
   
_All nodes are highly recommended to update prior block 1230000 - within 2 weeks._


Cabecinha84:
(Geolocations) Registering app, only continent dropdown is populated with values. Most probably updating specs will suffer from the same issue.

TheTrunk:
> (Geolocations) Registering app, only continent dropdown is populated with values. Most probably updating specs will suffer from the same issue.

That is correct behaviour. Till blockheight 1230000 only continents will be displayed as fluxos would reject mote precise locations anyway.
Updating of an app is also affected but only if user wants to change the geolocation. So if location of app remains (continent, country), UI would submit old specifications of geolocation.

TheTrunk:
https://github.com/RunOnFlux/flux/pull/737/files#diff-5e6baaa67c548fc842c757ce4c7131eb28d31c40f5199dbd8fb53451400ae6d8L3750

This is the line of reason why this is a geo app specs fork. As our geolocation string was restricted too much. Which is a good thing as expanding is easier than shrinking

Cabecinha84:
> > (Geolocations) Registering app, only continent dropdown is populated with values. Most probably updating specs will suffer from the same issue.
> 
> That is correct behaviour. Till blockheight 1230000 only continents will be displayed as fluxos would reject more precise locations anyway. Updating of an app is also affected but only if user wants to change the geolocation. So if location of app remains (continent, country), UI would submit old specifications of geolocation.

sorry lack of sleeping. I saw it on geo branch.
Still I believe we should remove forbidden until blockheight 1230000 as on latest ui there wasn't that option. What do you think?

Cabecinha84:
Also, shouldn't we have a i for information saying when selecting multiple geolocations will not mean that will have instances installed on all geolocations?

TheTrunk:
> Also, shouldn't we have a i for information saying when selecting multiple geolocations will not mean that will have instances installed on all geolocations?

Yes can add this

TheTrunk:
> > > (Geolocations) Registering app, only continent dropdown is populated with values. Most probably updating specs will suffer from the same issue.
> > 
> > 
> > That is correct behaviour. Till blockheight 1230000 only continents will be displayed as fluxos would reject more precise locations anyway. Updating of an app is also affected but only if user wants to change the geolocation. So if location of app remains (continent, country), UI would submit old specifications of geolocation.
> 
> sorry lack of sleeping. I saw it on geo branch. Still I believe we should remove forbidden until blockheight 1230000 as on latest ui there wasn't that option. What do you think?

I think that would not be good as fluxos would reject it saying 'Invalid geolocation' when user tries to submit such a geolocation. So its locked on UI till that height is reached as well so errors wont occur

Cabecinha84:
Nice! Was going to ack :D

TheTrunk:
> Nice! Was going to ack :D

Yeah we will maybe have some minor point releases soon but I don't want to push the forking height, giving time to node ops to update 👍

PR #740, opened 2022-10-10
DNS Over HTTPS Server - Update repositories.json

FliKites:
This docker image will be used as a second component in a multi component deployment that should allow me to deploy a Handshake (HNS) capable DN over HTTPS Server. 

Thanks for the consideration.


Cabecinha84 (approved):
Ack

PR #745, opened 2022-10-14
Update repositories.json

fegauthier:
I want to add my docker container to the verified list.

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -361,5 +361,6 @@
    "fusenet/spark-netstat:1.0.0",
    "siomiz/softethervpn:latest",
    "w2vy/vault_demo:latest",
-   "satishweb/doh-server:latest"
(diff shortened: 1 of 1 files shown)


TheTrunk:
Hi, can you please describe what your application does?

fegauthier:
@TheTrunk It's a NodeJS app that fetch and parse some IPTV schedule data. The app provide an endpoint to retrieve these data.

TheTrunk:
@fegauthier Hi, can you be more specific? As we cannot allow streaming of copyrighted content for example.
Is the app running somewhere? Do you have source available?

fegauthier:
@TheTrunk The app is currently running at https://epg.globomine.ca/diablo . I wanted to test FluxOS, but if it's not allowed, let me know.

TheTrunk:
I believe it is fine as I do not see any prohibited content on the above link

PR #763, opened 2022-10-24
Add OE Fuse

honsontran:


Cabecinha84 (changes requested):
Please remove "fusenet/node:2.0.2" as fusenet/node is already on approved list, no need to add specific versions,all are whitelisted.

PR #765, opened 2022-10-25
Update repositories.json

Kyuubi2709:
This image provides a complete node setup for ETHO protocol (https://ethoprotocol.com/) The container relies on 3 env variables:

NODE can be gn for gateway node, mn for master node, sn for service node.

UID The user id from the node dashboard (https://nodes.ethoprotocol.com/)

WALLET An ETHO wallet with the node collateral

After the launch of the container start the terminal in the container and start the post processing script: bash ./launch.sh


Cabecinha84 (approved):
Ack

PR #767, opened 2022-10-26
Remove old repo and add new repo

fegauthier:
Removing old repo and adding new repo for Front-End admin panel and public website Vue3 App


TheTrunk:
Please do not modify/remove existing repositories (would cause difficulties on Flux) but rather whitelist a new

fegauthier:
> Please do not modify/remove existing repositories (would cause difficulties on Flux) but rather whitelist a new

Ok good! I readd the old repositories.

PR #775, opened 2022-10-28
Update repositories.json

marcoleder:
By combining Kahoot with Spotify, RaveWave creates a competitive and fun music guessing game to play among friends.


Cabecinha84 (changes requested):
repositories.json contains a array of strings of whitelisted repos, the strings should be separated by a comma. Your PR like it is it's not a valid array. Please fix it.

Cabecinha84 (approved):
Ack

PR #779, opened 2022-10-30
v3.26.2

TheTrunk:
- UI: updated Titan guide link
- allow ALL to be a valid location (as some services submit acALL for example)
- isDecimalLimit function
- additional decimal checks on parameters and message formatting
- minor UI package version updates


Cabecinha84 (approved):
Ack

PR #789, opened 2022-11-04
Add Monero Node - Update repositories.json

FliKites:


TheTrunk:
Hi, we do not allow monero and other cryptocurrency miners

PR #802, opened 2022-11-09
Update repositories.json

alelipona:


TheTrunk:
peer2profit is not allowed

PR #808, opened 2022-11-13
add new repo

fegauthier:
This is a container that can received files via HTTP server and it will serve the Vue3 App received and a Persistent test storage container test


helpers/repositories.json
```diff
@@ -395,6 +395,7 @@
    "bigmandave/doge-node",
    "treyyoder/quakejs:latest",
    "biosmarcel/scribble.rs:latest",
+   "fegauthier/persistent-storage-test:latest"
```
TheTrunk, line 398: Hi, can you please add comma here and ideally move it to the end of the array?

TheTrunk (changes requested):

helpers/repositories.json
fegauthier: fegauthier/persistent-storage-test:latest was already there. I just removed the duplicate that I commited.

TheTrunk (approved):

PR #815, opened 2022-11-18
added rest apis for config

alihm:
added these endpoint functions:
  getConfigRestartRequired
  getConfigFolders
  getConfigDevices
  putConfigFolders
  putConfigDevices
  patchConfigFolders
  patchConfigDevices
  deleteConfigFolders
  deleteConfigDevices
  getConfigOptions
  putConfigOptions
  patchConfigOptions
  getConfigGui
  putConfigGui
  patchConfigGui

### DIFF
--- a/ZelBack/src/routes.js
+++ b/ZelBack/src/routes.js
@@ -1044,43 +1044,43 @@ module.exports = (app, expressWs) => {
     syncthingService.postConfigDevices(req, res);
   });
 
-  app.put('/syncthing/config/folders', (req, res) => {
+  app.post('/syncthing/config/folders', (req, res) => {
     syncthingService.putConfigFolders(req, res);
   });
 
-  app.put('/syncthing/config/devices', (req, res) => {
+  app.post('/syncthing/config/devices', (req, res) => {
     syncthingService.putConfigDevices(req, res);
   });
 
-  app.patch('/syncthing/config/folders', (req, res) => {
+  app.post('/syncthing/config/folders', (req, res) => {
     syncthingService.patchConfigFolders(req, res);
   });
 
-  app.patch('/syncthing/config/devices', (req, res) => {
+  app.post('/syncthing/config/devices', (req, res) => {
     syncthingService.patchConfigDevices(req, res);
   });
 
-  app.delete('/syncthing/config/folders', (req, res) => {
+  app.post('/syncthing/config/folders', (req, res) => {
     syncthingService.deleteConfigFolders(req, res);
   });
 
-  app.delete('/syncthing/config/devices', (req, res) => {
+  app.post('/syncthing/config/devices', (req, res) => {
     syncthingService.deleteConfigDevices(req, res);
   });
 
-  app.put('/syncthing/config/options', (req, res) => {
+  app.post('/syncthing/config/options', (req, res) => {
     syncthingService.putConfigOptions(req, res);
   });
 
-  app.patch('/syncthing/config/options', (req, res) => {
+  app.post('/syncthing/config/options', (req, res) => {
     syncthingService.patchConfigOptions(req, res);
   });
 
-  app.put('/syncthing/config/gui', (req, res) => {
+  app.post('/syncthing/config/gui', (req, res) => {
     syncthingService.putConfigGui(req, res);
   });
 
-  app.patch('/syncthing/config/gui', (req, res) => {
+  app.post('/syncthing/config/gui', (req, res) => {
     syncthingService.patchConfigGui(req, res);
   });
 
--- a/ZelBack/src/routes.js
+++ b/ZelBack/src/routes.js
@@ -434,6 +434,15 @@ module.exports = (app, expressWs) => {
   app.get('/syncthing/config/folders', cache('30 seconds'), (req, res) => {
     syncthingService.getConfigFolders(req, res);
   });
+  app.get('/syncthing/config/restart-required', cache('30 seconds'), (req, res) => {
+    syncthingService.getConfigRestartRequired(req, res);
+  });
+  app.get('/syncthing/config/options', cache('30 seconds'), (req, res) => {
+    syncthingService.getConfigOptions(req, res);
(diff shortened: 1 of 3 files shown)


ZelBack/src/services/syncthingService.js
```diff
@@ -101,6 +101,42 @@ async function getConfigFolders(req, res) {
   return res ? res.json(response) : response;
 }
 
+async function getConfigDevices(req, res) {
+  const response = await performRequest('get', '/rest/config/devices');
+  return res ? res.json(response) : response;
+}
+
+async function putConfigFolders(req, res, id, data) {
```
TheTrunk, line 109: The id and data shall be part of req params/query

TheTrunk (changes requested):

ZelBack/src/services/syncthingService.js
TheTrunk: Eg from other fluxos code
```
  let { id } = req.params;
  id = id || req.query.id;
  ```
  then 
  if (!id) {
  // shall be handled if errorMessage of no id specified etc
  }

Comment (alihm) [reply ZelBack/src/services/syncthingService.js:109]: oh, ok, fixing those

File: ZelBack/src/routes.js [hunk 2]
```diff
@@ -1027,6 +1036,54 @@ module.exports = (app, expressWs) => {
...
 
+  app.post('/syncthing/config/folders', (req, res) => {
+    syncthingService.postConfigFolders(req, res);
+  });
+
+  app.post('/syncthing/config/devices', (req, res) => {
+    syncthingService.postConfigDevices(req, res);
+  });
+
+  app.put('/syncthing/config/folders', (req, res) => {
```
TheTrunk, line 1047: please use only get and post methods in routes

ZelBack/src/services/syncthingService.js
```diff
@@ -96,11 +99,258 @@ async function systemPing(req, res) {
...
 }
 
+async function getConfigRestartRequired(req, res) {
+  const response = await performRequest('get', '/rest/config/restart-required');
+  return res ? res.json(response) : response;
+}
+
 async function getConfigFolders(req, res) {
-  const response = await performRequest('get', '/rest/config/folders');
+  let { id } = req.params;
```
TheTrunk, line 108: this expects params or query but not specified in path

ZelBack/src/services/syncthingService.js
```diff
@@ -96,11 +99,258 @@ async function systemPing(req, res) {
...
+  id = id || req.query.id;
```
TheTrunk, line 109: perhaps using ?? what if id is 0?

ZelBack/src/services/syncthingService.js
```diff
@@ -96,11 +99,258 @@ async function systemPing(req, res) {
...
+  let path = '/rest/config/folders';
+  if(id){
```
TheTrunk, line 111: linting

ZelBack/src/services/syncthingService.js
```diff
@@ -19,6 +20,8 @@ const syncthingURL = `http://${config.syncthing.ip}:${config.syncthing.port}`;
 
 let syncthingApiKey = '';
 
+const disableAuth = true;
```
TheTrunk, line 23: for now we can just comment it out instead of introducing this variable

ZelBack/src/services/syncthingService.js
```diff
@@ -96,11 +99,258 @@ async function systemPing(req, res) {
...
+  return res ? res.json(response) : response;
+}
+
+async function postConfigFolders(req, res) {
+  let { data } = req.params;
+  data = data || req.query.data;
+  if (!data) {
+    throw new Error('data is not provided');
+  }
+  const authorized = await verificationHelper.verifyPrivilege('admin', req);
```
TheTrunk, line 135: authorized has to happen if res is present. if not its always true as fluxos shall be having access to all

ZelBack/src/services/syncthingService.js
```diff
@@ -96,11 +99,258 @@ async function systemPing(req, res) {
...
+async function getConfigOptions(req, res) {
+  const response = await performRequest('get', `/rest/config/options`);
+  return res ? res.json(response) : response;
+}
+
+async function putConfigOptions(req, res) {
+  let { data } = req.params;
+  data = data || req.query.data;
+  if (!data) {
+    throw new Error('No options data is provided');
```
TheTrunk, line 288: throw without catching

ZelBack/src/services/syncthingService.js
```diff
@@ -96,11 +99,258 @@ async function systemPing(req, res) {
...
+  return res ? res.json(response) : response;
+}
+
+async function postConfigDevices(req, res) {
+  let { data } = req.params;
+  data = data || req.query.data;
+  if (!data) {
+    throw new Error('data is not provided');
+  }
+  const authorized = await verificationHelper.verifyPrivilege('admin', req);
```
TheTrunk, line 151: verificationHelper.verifyPrivilege CAN throw, needs to be caught. But I think we shall make it to not throw and default to false

TheTrunk (changes requested):
All applies to all the functions
Further usage of post methods is not done well as data is not obtained

ZelBack/src/services/syncthingService.js
alihm: if id is 0 or not provided, '/rest/config/folders' will be called. syncthing rest api accepts both forms. if id is not passed it will return an array of folders.

ZelBack/src/services/syncthingService.js
alihm: is there any implementation of this elsewhere? will add it after the tests are done.

TheTrunk: You can do
`const authorized = res ? await verificationHelper.verifyPrivilege('admin', req) : true;`

PR #818, opened 2022-11-19
v2ray vmess and vless vpn protocol server

sohrabsameny:
v2ray is an official repo supported by the Docker and is used by several developers as a tool for initiating a vmess protocol service provider. It is really needed to be accessible via flux.

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -447,4 +447,5 @@
    "runonflux/ipfs:latest",
    "honsontran/uniswap-frontend:latest",
    "runonflux/blockbook-docker:latest"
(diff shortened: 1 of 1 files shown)


helpers/repositories.json
```diff
@@ -447,4 +447,5 @@
    "runonflux/ipfs:latest",
    "honsontran/uniswap-frontend:latest",
    "runonflux/blockbook-docker:latest"
```
TheTrunk, line 449: please add comma here to make it a valid array

TheTrunk (changes requested):

PR #833, opened 2022-11-24
Add FluxRun Apps to repository.json

fluxrun-dev:
These images are similar to `bunnyanalyst/flux-function-python` (See https://github.com/RunOnFlux/flux/pull/804) but with a different name. These images are required to prepare for Beta testing for FluxRun.


TheTrunk:
This looks highly suspicious, please provide way more information about the applications.
We do not approve app placeholders.

PR #837, opened 2022-11-27
v3.28.0

TheTrunk:
- Synced container data across instances
  - Uses Syncthing
  - Complex Syncthing API, at /syncthing
  - Custom configuration to fully fit Flux
  - Port scheme is +2 (eg 16129, 16139), upnp support. Port is mandatory to be opened
- containerData flags support
  - Currently only supported flag is 's'
  - Use as s:/path/to/data
- Node owner can now remove any app from their node via UI
- Whitelist, Blacklist now applies to also entire organisations
- Fixes ability to adjust instances if ALL geolocation is selected
- Fixes allowance of repositories containing space character
- Stop sending fluxapprequests after some retries of app not found on the network, reducing a lot the bandwidth in use on nodes
- Various minor improvements, fixes

This release is heavy and requires successful syncthing installation. 
Node operators should be monitoring proper FluxOS update.


Cabecinha84:
Sorry made some bad commits and forgot to git pull before but I think it should be ok now. Lack of sleeeeeeeeeeeeeeeeeeep!
Will continue tomorrow so I don't brake anything else :D!

Cabecinha84 (approved):
Ack. Ui needs to be rebuild before the release.

PR #846, opened 2022-12-01
Adding Multipaper repo

Lazzeruz:
Hey, I'd like to add Multipaper Minecraft servers.

It's a special kind of Minecraft server that distributes the workload over multiple servers.

The master is the one distributing the workload and saving the data.

The regular ones act like normal minecraft servers but on a 16x16 meter basis. Getting assigned those areas by the master.

I don't know if the Master one can work on flux, but the regular ones should. I can always test it.


Lazzeruz:
I meant image, not repo.

PR #848, opened 2022-12-02
Adding Stable diffusion dapp

Lazzeruz:
Hi, I would like to add Stable Diffusion, with a web ui, to Flux.

It's a text2image generator, and I got the cpu version working with Docker.

People need to use the "CLI_ARGS --use-cpu" for it to work, I might change this in a future update to the docker. So it's standard. Also "CLI_ARGS --no-half" and "CLI_ARGS --precision full" are also recommended to use, afaik.

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -562,5 +562,6 @@
    "byblaaz/celestianodelight:latest",
    "byblaaz/celestianodefull:latest",
    "byblaaz/celestianodebridge:latest",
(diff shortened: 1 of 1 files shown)


Lazzeruz:
I am currently redoing the container to work out of the box within Flux.
I'll update this pull request once I am done with it.

Lazzeruz:
Ok, should be done now. It's basically plug and play for CPUs now, only needing the ports.

PR #859, opened 2022-12-08
Adding Multipaper image. fixed environment

Lazzeruz:
Hey, the environment parameter was borked on the normal Multipaper image when ran in flux. for now I added a test IP to see if it can even connect and I will look into a better way of setting the master IP.

After I've found a better way to set Master IPs I will make a new pull request to take out this image and the old MultiPaper ones. And put the image in that can work out of the box.

Sorry for any inconveniences.


Cabecinha84 (approved):
Ack

PR #862, opened 2022-12-09
Update repositories.json

Kyuubi2709:
@Cabecinha84 @TheTrunk


Cabecinha84:
What is the "kyuubi2709/test:latest,"
"kyuubi2709/test2:latest," for?

PR #874, opened 2022-12-20
v3.30.0

TheTrunk:
- Resolves Syncthing permission issues
- Allows installation of app of a specific version - hash (even of temporary apps or expired apps)
- App price increase to following specifications
cpu: 0.15 per 0.1 cpu core
ram: 0.05  per 100mb
ssd: 0.02 per 1gb
- This is a forking release with activation at blockheight 1288000
**This version is enforced on the 2nd of January 2023**


alihm (approved):

Cabecinha84 (approved):
Ack. We should reactivate nodes to go DOS if syncthing is not working properly on this version.

PR #895, opened 2023-01-04
Update repositories.json

ngcheewee:


TheTrunk:
Hi, what application do you plan to run? Mining of any sort including bandwidth mining is strictly prohibited.
Please correct your PR as there is a missing comma which would cause the array to be invalid.

ngcheewee:
> Hi, what application do you plan to run? Mining of any sort including bandwidth mining is strictly prohibited. Please correct your PR as there is a missing comma which would cause the array to be invalid.

It is a html/css application. I've added comma to the PR, please review

PR #900, opened 2023-01-07
Update repositories.json

geeksquadd:
Preconfigured Dante Server https://github.com/geeksquadd/dante-preauth


TheTrunk:
Hi sorry for the late response.
Please provide way more information and state how do you plan to use it.
As we had similar app on the network that was doing malicious behaviour

geeksquadd:
I wanted to use socks5 with scrapbox seo tool

PR #901, opened 2023-01-07
Update repositories.json

adarshaketh:
Added Mysterium Network
More Info: https://www.mysterium.network/nodes


TheTrunk:
Hi,
Mining including bandwidth mining is prohibited on Flux.

PR #904, opened 2023-01-15
Update repositories.json

vmvelev:
I am building an app that I want to test.

The app consists of the following:
index page - where the customer would see their zel id if logged
login page - where the customer will be asked to login (through zelcore)


TheTrunk:
Perfect, looking forward to it 👍

PR #906, opened 2023-01-18
v3.32.0

TheTrunk:
- Visit button now accounts for more components and ports, should handle more apps for proper opening
- Minimum FluxBench allowed version is 3.7.0
- Restoration of Ports and UPNP will be happening every 6 blocks (~12 muntes) instead of every 25 blocks. This should lead to more reliable upnp mapping and ports management
- UPNP mapping is now assigning time for 2 hours, instead of indefinite. This should lead to more reliable mapping as routers reflect ttl of 0 differently
- Syncthing is now properly running on assigned port of fluxos on upnp nodes. NAT management of syncthing got disabled and is controlled by fluxos.
- Applications Docker Network is now deterministic based on the LAST character of application name:
Docker network is created in following scheme
```
172.XXX.0.0
where XXX is character code of last character of the application name.
```
- Home UI and Syncthing properly running and be accessible is now being checked by other fluxos on the network. This should cause nodes that do not have properly set firewall rules to go offline
- Syncthing running properly is now enforced. This will cause nodes that have issue running syncthing properly to go offline
- Corrections of start and end options on some daemon rpc calls
- Reduced number of application running messages and broadcast of those messages. From every 20 minutes to 60 minutes. This should decrease node load while still maintain high accuracy.
- Applications are now being spawned on a node after 90 minutes of running fine (previously 20 minutes)
- Better removal logging of applications
- Corrects Home UI to give apps deployment multisig address
- Apps Service Test Suite
- /flux/checkappavailability post api calls is used internally by the nodes to test if app ports are properly responding verifying that application is running smoothly.
- Periodical checks of availability of applications installed on the node. This asks other flux nodes to check if application installed is truly available from outside. This check is in testing phase and will be enforced in the next version. Most likely many nodes will fall off the network if they have misconfigured firewall
- Automation of hash whitelist - this ensure hash list is always up to date, support for 'development' nodes and potential human factor mistakes on hash whitelist
- /flux/entermaster and /flux/enterdevelopment api calls are allowing to switch between master and development branches of fluxos. // for development purposes
- development flag added to userconfig file // for development purposes
- development nodes are running edge source code and receive updates first. Further periodically pull latest code to run. Development nodes are participating in main net testing latest features of Flux before entering mass production. The testing phase is so adding a crucial step to ensure mass production is unaffected from potential critical bugs.
- decryptionkey flag added to userconfig. Flux is preparing for introducing 'Trusted Nodes' to solve secrets and privacy of applications. // experimental
- Other minor improvements and bug fixes


Cabecinha84 (approved):
Ack. Great work!

PR #911, opened 2023-01-21
Update repositories.json

post-opulence:


TheTrunk:
Hi,
Can you please write a little information on what do you plan to run?

post-opulence:
Hey, thanks for the quick reply!
I plan on running a small node.js script for a project I have for an assignment. The script fetches data from an external API every 2 seconds then store it in an external database. You can find more details regarding the script here https://github.com/post-opulence/birdnest-backend

PR #913, opened 2023-01-21
Update repositories.json

kikundipixel:
Adding https://hub.docker.com/_/ubuntu version 20.04 docker  image

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -258,5 +258,6 @@
    "jktuned",
    "vmvelev",
    "timpiltd",
-   "postopulence/birdnestbackend"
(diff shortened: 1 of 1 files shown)


TheTrunk:
Sorry, OS is not an app. That would allow to run practically anything which is not the intention for Apps.

kikundipixel:
> Sorry, OS is not an app. That would allow to run practically anything which is not the intention for Apps.

Thanks Trunk for the clarification this was my first, merge request ever but now that you mention it is clear lol. proper noob when it is said out loud it's more precise.
At least I now understand the process of getting onto Flux as we plan to launch a healthcare referral app later this month 
Cheers tony

PR #914, opened 2023-01-21
Update repositories.json

kikundipixel:
Please add Ubuntu Docker account to the repo


TheTrunk:
Sorry, OS is not an app. That would allow to run practically anything which is not the intention for Apps.

PR #918, opened 2023-01-24
Update repositories.json

adarshaketh:
Added https://hub.docker.com/r/adarshak103/redis-api-cache
Code: https://github.com/adarshaketh/redis-api-cache
This image will cache the Mysterium Network's Discovery API
You can get a list of API's available
https://discovery.mysterium.network/swagger/index.html

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -259,5 +259,6 @@
    "vmvelev",
    "timpiltd",
    "postopulence/birdnestbackend",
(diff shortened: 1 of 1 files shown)


TheTrunk:
Alright, but please be noted that ANY sort of mining including bandwidth mining is prohibited. If we get reports from node owners that applications are doing some mining activity, we have to blacklist it. Merging.

PR #919, opened 2023-01-25
v3.33.0

TheTrunk:
- Adjustment of Firewall and UPNP is now done every 10 minutes instead of relying on Flux block times.
- Fixes bug in pricing on application updates - will come to effect on blockheight 1315000
- Fixes typo bugs on flux node flux apps availability that was causing some checks to be skipped
- Improvements int reindexing of global applications specifications

This is a forking release. Nodes MUST update before blockheight 1315000


alihm (approved):

Cabecinha84 (changes requested):
Why are we setting ttl on upnp call to 0. Like this if Node is reinstalled on host the ports will be remain in use on the router upnp table and none app using that ports will ever install on all nodes running under that router.

TheTrunk:
> Why are we setting ttl on upnp call to 0. Like this if Node is reinstalled on host the ports will be remain in use on the router upnp table and none app using that ports will ever install on all nodes running under that router.

- This is changing a test suite. In production we are using 0 currently.
- Sadly we tested a lot with the different ttls for upnp. But we get a lots of complains on custom setups of users. We tried 30 mins, 1 hour - none were working for them. The routers they have support it but I think their routers implement ttl as minutes and not as seconds (guess) so it just throws an error and flux does not work for them. 
- ttl of 0 n most cases defaults to default value - usually 30 mins or 1 hour on most routers. I think having 0 is not ideal as we do not really control this but it 'works' - also a reason of move to setinterval for adjustment of it rather than relying flux block times. Most routers do not allow borrow time more than 24 hour, and 0 should always default to default ttl of usually 1 hour.
- On upstream branch of nat-upnp that we are using, their is a few issues with ttl being set different than 0 causing the error that our users expeirienced. So perhaps it is a bug in our library. -> we shall probably examine our library more what the issue is.

Cabecinha84:
> > Why are we setting ttl on upnp call to 0. Like this if Node is reinstalled on host the ports will be remain in use on the router upnp table and none app using that ports will ever install on all nodes running under that router.
> 
> * This is changing a test suite. In production we are using 0 currently.
> * Sadly we tested a lot with the different ttls for upnp. But we get a lots of complains on custom setups of users. We tried 30 mins, 1 hour - none were working for them. The routers they have support it but I think their routers implement ttl as minutes and not as seconds (guess) so it just throws an error and flux does not work for them.
> * ttl of 0 n most cases defaults to default value - usually 30 mins or 1 hour on most routers. I think having 0 is not ideal as we do not really control this but it 'works' - also a reason of move to setinterval for adjustment of it rather than relying flux block times. Most routers do not allow borrow time more than 24 hour, and 0 should always default to default ttl of usually 1 hour.
> * On upstream branch of nat-upnp that we are using, their is a few issues with ttl being set different than 0 causing the error that our users expeirienced. So perhaps it is a bug in our library. -> we shall probably examine our library more what the issue is.

I think on my router sending ttl 0 makes the setting remain forever on upnp table.
Ok, it's something we can see in further releases.

TheTrunk:
> > > Why are we setting ttl on upnp call to 0. Like this if Node is reinstalled on host the ports will be remain in use on the router upnp table and none app using that ports will ever install on all nodes running under that router.
> > 
> > 
> > 
> > * This is changing a test suite. In production we are using 0 currently.
> > * Sadly we tested a lot with the different ttls for upnp. But we get a lots of complains on custom setups of users. We tried 30 mins, 1 hour - none were working for them. The routers they have support it but I think their routers implement ttl as minutes and not as seconds (guess) so it just throws an error and flux does not work for them.
> > * ttl of 0 n most cases defaults to default value - usually 30 mins or 1 hour on most routers. I think having 0 is not ideal as we do not really control this but it 'works' - also a reason of move to setinterval for adjustment of it rather than relying flux block times. Most routers do not allow borrow time more than 24 hour, and 0 should always default to default ttl of usually 1 hour.
> > * On upstream branch of nat-upnp that we are using, their is a few issues with ttl being set different than 0 causing the error that our users expeirienced. So perhaps it is a bug in our library. -> we shall probably examine our library more what the issue is.
> 
> I think on my router sending ttl 0 makes the setting remain forever on upnp table. Ok, it's something we can see in further releases.

Can you check your routers manual? There should be something about upnp ttl. From the community routers they usually had 1 hour. We should definitely examine the upnp more, its ttl values, various routers

johnhanlon86 (approved):

Cabecinha84 (approved):
Ack

PR #930, opened 2023-02-01
v3.33.1

TheTrunk:
- Fixes missing opening of port before asking if receiver applications port is available
- Fixes 'Flux does not have sufficient peers' error happening on nodes that are not yet confirmed on the network. FluxOS can't have peers if it is not confirmed on Flux network first.
- Better logging


Cabecinha84 (approved):
Ack

PR #932, opened 2023-02-04
v3.34.0

TheTrunk:
- Corrections in app port checks, allowance and deletions
- Enforce application port checks
- Minor UI and other fixes


ZelBack/src/services/fluxNetworkHelper.js
```diff
@@ -254,12 +255,40 @@ async function checkAppAvailability(req, res) {
...
+      const apiport = processedBody.apiport || config.server.apiport;
+
+      if (ip !== remoteIP4 && authorized === false) {
+        throw new Error(`Request ip ${remoteIP4} of ${remoteIP} doesn't match the ip: ${ip} to check App port Availability.`);
+      }
+
+      if (authorized === false) { // can only be called from a valid fluxnode if not authorized call
+        const nodeList = await fluxCommunicationUtils.deterministicFluxList();
+        let ipPortOnNodeList = ip;
+        if (apiport !== '16127' && apiport !== 16127) {
```
Cabecinha84, line 278: This is wrong, the api port is from the node calling, not from the node where this is being executed. Because it was added as a new property on the post, it cannot be used here until all nodes are updated. That is why I left this part commented.

Cabecinha84 (changes requested):
There are things that needs to be changed before this is released.

ZelBack/src/services/fluxNetworkHelper.js
Cabecinha84: the apiport should come from:
const {
        ip, apiport, ports, appname,
      }
but only updated nodes will send that information.

TheTrunk: I think it is fine to leave it enabled - majority of the nodes upgrade very fast to the latest release making it a non issue for those nodes.
If apiport is not supplied we fall back to the default one basically searching just an IP in the deterministic list. Which usually is present if people are using upnp anyway. This makes it fine as well.
Ideally we should be submitting public key and signature if we really want to make this very strict but I think above ip check will serve the needs very well.

But I am actually seeing a different issue now. As unconfirmed nodes will basically get dossed immediately as they are not on the confirmed list. 
I think we have to add the signature to it actually to be 100% sure.

Cabecinha84 (approved):
Ack

Cabecinha84:
The last change requires bomebuild.

ZelBack/src/services/appsService.js
```diff
@@ -8926,18 +8940,22 @@ async function checkMyAppsAvailability() {
         ip: myIP,
         appname: app.name,
         ports: appPorts,
+        pubKey,
```
johnhanlon86, line 8943: Should we assign `pubKey` to an object property?
```suggestion
        pubKey: pubKey,
```

ZelBack/src/services/fluxNetworkHelper.js
```diff
@@ -254,16 +255,36 @@ async function checkAppAvailability(req, res) {
...
+      // pubkey of the message has to be on the list
+      const zl = await fluxCommunicationUtils.deterministicFluxList(pubKey); // this itself is sufficient.
+      const node = zl.find((key) => key.pubkey === pubKey); // another check in case sufficient check failed on daemon level
+      const dataToVerify = processedBody;
+      delete dataToVerify.signature;
+      const messageToVerify = JSON.stringify(dataToVerify);
+      const verified = verificationHelper.verifyMessage(messageToVerify, pubKey, signature);
+      if ((verified !== true || !node) && authorized !== true) {
+        log.error('Unable to verify request authenticity');
+        // throw new Error('Unable to verify request authenticity');
```
johnhanlon86, line 275: Do we want to do a `return` here if authenticity is not verified or let it keep going?

TheTrunk: Currently keep it going. Once nodes update we will most likely uncomment this throw

ZelBack/src/services/appsService.js
TheTrunk: pubKey, is a shorthand of pubKey: pubKey,

blondfrogs (approved):
utACK - Checks are much better

PR #943, opened 2023-02-13
Adds makdi project to whitelisted repositories

teammakdi:
https://github.com/teammakdi


TheTrunk:
crawler looks interesting :+1:

PR #950, opened 2023-02-23
Update repositories.json

pvilalta:


TheTrunk:
Hi,
- What applications do you want to Run On Flux?
- Are your desired applications running somewhere already? (optional)
- Is your application open source? (optional)

Be noted that illegal content, mining applications including bandwidth mining are prohibited.

pvilalta:
> Hi,
> 
> * What applications do you want to Run On Flux?
> * Are your desired applications running somewhere already? (optional)
> * Is your application open source? (optional)
> 
> Be noted that illegal content, mining applications including bandwidth mining are prohibited.

Hi there,

I would like to run the staging version of my App in order to test Flux.
It’s not open source and some datas (i.e: env variables) are very critical and shouldn’t be viewable by the node operator 

Could you tell me if Flux can assume that ?

Also, what happen if my subscription end ? Does Flux keep the data some time in case I want to renew my sub ?

thanks in advance,
Paul

TheTrunk:
Hi,

Entire Flux is open source public blockchain. So anyone can go and currently see any variables. We do have Flux Vault, that can inject env variables at runtime. And we also provide Flux Storage (UI is not yet released, will be on next version), that allows for huge env vars and slight obfuscation - that they are not in plain text but replaced by a link to flux storage where only fluxos can download it.

But if you really mean some privacy at runtime, that is not possible on Flux. Node owners have root access to the machine, so they can see pretty much anything that is running on their machine.

For expiration, you can subscribe an app for up to 1 year, there is notification by email in place that can notify about an app going. to expire. However when it expires, application is removed from the network entirely, without any trace, data left. Only the application specifications are still stored on the fluxos permanent messages, and hash is stored on flux blockchain.

pvilalta:
Hi @TheTrunk 

Thanks you for your answer, it couldn't be more precise
Is the privacy at runtime a direction for Flux team ? If it's the case, is there a date to expect this feature ?

Paul

PR #952, opened 2023-02-24
Update repositories.json

teeverc:


TheTrunk:
Hi,
- What applications do you want to Run On Flux?
- Are your desired applications running somewhere already? (optional)
- Is your application open source? (optional)

Be noted that illegal content, mining applications including bandwidth mining are prohibited.

PR #956, opened 2023-02-25
Updated "cybergate" to "pgerringer"

pgerringer:
Originally put the wrong repo name to whitelist.


TheTrunk:
Hi, whitelists can be removed, adjusted. Please add a new whitelist to the array without adjusting/removing previous whitelist

PR #958, opened 2023-02-26
v3.36.0

TheTrunk:
- Allow installation of multiple applications using the same image on the same instance (exclude presearch)
- Check for applications availability is adjusted
  - previous scheme was testing availability of installed applications. That had some limitations as ports may have been in use by the application and so respond in a different than expected way. Furthermore the test was only done on nodes on which some application was running
  - FluxOS will now choose a random unused port within application range and test availability of that port by running a simple listening server. This test is way more deterministic, predictable and properly tests even nodes without any app running.
- FluxOS Home Dashbord adjusted for halving with taking data about rewards directly from daemon.
- Introducing Flux Storage on Home UI
  - Storage for big variables and other parameters: https://github.com/RunOnFlux/fluxstorage 
  - Upload Envionment parameters, Commands or Contacts to central Flux Storage managed by Flux team.
  - This allows for long list of Environment parameters or commands going beyond FluxOS character limitation. 
  - Furthermore Environment parameters and Commands provide slight obfuscation as they are not directly stored on FluxOS but only as a link to Flux Storage that only FluxOS itself can access. This should definitely not be treated as private storage but just as a slight obfuscation mechanism
  - Contacts on the other hand if used in Flux Storage are fully private. Only Flux team managing Flux Storage can access it. This provides a set of anonymity where you email won't be exposed publicly. Contacts field is used primarily for notifications about applications - expiration, launches and more. 
  - Flux Storage can be directly used on Home UI in app registration or update.
  - The Flux backend scheme for Flux Storage is F_S_[CMD, ENV, CONTACTS]=URL
- Introducing Apps Monitor https://github.com/RunOnFlux/appsmonitor 
  - Very basic tool on which we will expand to add functionality of application monitoring, notification system via email and other provided contacts
- As Flux grows we are forced to protect development team, node operators and users of Flux by applying Terms of Service. This forces more strict rules on what can Run on Flux. https://cdn.runonflux.io/Flux_Terms_of_Service.pdf
- Don't allow incoming connections from non-routable ips


HomeUI/src/views/apps/RegisterFluxApp.vue
```diff
@@ -1287,6 +1336,15 @@
         </b-col>
       </b-row>
     </div>
+    <div>
+      <br>
```
Cabecinha84, line 1340: same here. clickable to accept conditions.

ZelBack/src/services/appsService.js
```diff
@@ -8933,92 +8940,89 @@ async function checkMyAppsAvailability() {
...
+    }
+    // now open this port properly and launch listening on it
+    await fluxNetworkHelper.allowPort(testingPort);
+    if ((userconfig.initial.apiport && userconfig.initial.apiport !== config.server.apiport) || isUPNP) {
+      await upnpService.mapUpnpPort(testingPort, 'Flux_Test_App');
+    }
+    await serviceHelper.delay(5 * 60 * 1000);
+    testingAppserver.listen(testingPort);
+    await serviceHelper.delay(10 * 60 * 1000);
+    // eslint-disable-next-line no-await-in-loop
```
Cabecinha84, line 8978: 5 + 10m of service delay here. Isn't that too much time?

ZelBack/src/services/fluxNetworkHelper.js
```diff
@@ -304,7 +313,7 @@ async function checkAppAvailability(req, res) {
         res.json(dataResponse);
         return;
       }
-      const successResponse = messageHelper.createSuccessMessage(`Flux App ${appname} is available.`);
+      const successResponse = messageHelper.createSuccessMessage(`Flux Application on ${ip}:${ipPort} are available.`);
```
johnhanlon86, line 316: @TheTrunk I think that this message refers to multiple apps?
```suggestion
      const successResponse = messageHelper.createSuccessMessage(`Flux Applications on ${ip}:${ipPort} are available.`);
```

ZelBack/src/services/appsService.js
```diff
@@ -8933,92 +8940,89 @@ async function checkMyAppsAvailability() {
...
+    const signature = await signCheckAppData(stringData);
+    data.signature = signature;
+    // first check against our IP address
+    // eslint-disable-next-line no-await-in-loop
+    const resMyAppAvailability = await axios.post(`http://${askingIP}:${askingIpPort}/flux/checkappavailability`, JSON.stringify(data), axiosConfig).catch((error) => {
+      log.error(`${askingIP} for app availability is not reachable`);
+      log.error(error);
+    });
+    if (resMyAppAvailability && resMyAppAvailability.data.status === 'error') {
+      log.warn(`Applications port range unavailability detected from ${askingIP}:${askingIpPort}`);
```
Cabecinha84, line 9009: We should have here a while in error, or next time we execute the method it will be on a different random port that can be ok and we clear the DOS when in fact there are ports that aren't ok. And also because the method itself takes 15m delay on service listening creation.

Cabecinha84 (changes requested):
Some changes requested. Cheers

Cabecinha84:
Added to the PR comment: "Not allowing incoming connections from internal ip"

Cabecinha84:
HomeUI needs to be build before release.

ZelBack/src/services/appsService.js
TheTrunk: Oh good catch. Was meant to be 5 seconds, 10 seconds

ZelBack/src/services/appsService.js
TheTrunk: The chance of hitting the same testing port is 1 to 10000 basically. So it highly improbable of testing the same port. We are also not testing ports that are in use per apps. I think this is sufficient and if lets say 1000 ports are not working well in the range (aka node op is artificially blocking some port range), it is not ideal yes - majority of apps will run well but some not.
Now its a question if node op is blocking one port - that is 'fine' but if node op is blocking 50% of ports, that is not fine.
So some adjustment to this test would be in fact needed. As if 50% of ports are blocked, our test will still allow it and said that this flux is ok

ZelBack/src/services/fluxCommunication.js
```diff
@@ -96,15 +100,26 @@ function handleIncomingConnection(ws, req, expressWS) {
     }, 1000);
     return;
   }
-  incomingConnections.push(ws);
   const peer = {
     ip: ws._socket.remoteAddress,
   };
+  const ipv4Peer = peer.ip.replace('::ffff:', '');
+  for (let i = 0; i < 17; i += 1) {
```
TheTrunk, line 107: this should get some rewrite so its more programatic

ZelBack/src/services/fluxCommunication.js
```diff
@@ -25,6 +25,10 @@ let numberOfFluxNodes = 0;
 
 const blockedPubKeysCache = new LRU(LRUoptions);
 
+const privateIpsList = ['192.168.', '10.', '172.16.', '172.17.', '172.18.', '172.19.',
```
TheTrunk, line 28: Are we sure of this list? Like where does it come from?

HomeUI/src/views/apps/Management.vue
```diff
@@ -3132,6 +3181,15 @@
         class="mt-1"
       />
     </div>
+    <div>
+      <br>
+      By managing your application you agree with
```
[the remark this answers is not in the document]
TheTrunk: I was thinking about that but I think we are good with just stating it like this without clicking. Lets recheck with biz

ZelBack/src/services/appsService.js
TheTrunk: resolved. If port fails, we will be testing this port again if unused.

ZelBack/src/services/fluxCommunication.js
Cabecinha84: https://www.techtarget.com/whatis/definition/private-IP-address

"Private IPv4 addresses have the following class configurations:

Class A IP addresses. Configurations range from 10.0.0.0 to 10.255.255.255. This class is for large networks and has 8 bits for the network and 24 bits for hosts.
Class B IP addresses. Configurations range from 172.16.0.0 to 172.31.255.255. This class is for medium networks and has 16 bits for the network and 16 bits for hosts.
Class C IP addresses. Configurations range from 192.168.0.0 to 192.168.255.255. This class is for smaller networks and has 24 bits for the network and 8 bits for hosts."

Cabecinha84: Lots of sites refering the same information.

Cabecinha84:
"2023-02-27T07:53:45: Flux App appPortsTest is available. Detected from 65.108.132.80:16187 on 34555"
Goooooooooood

Cabecinha84 (approved):
Ack

PR #961, opened 2023-02-28
V3.63.3

Cabecinha84:
Fix globalapps array not matching on some nodes


TheTrunk:
It may have some impact to rescan for the apps again :+1:

TheTrunk (approved):

PR #974, opened 2023-03-15
v3.37.0

TheTrunk:
- Fixes Flux Storage Environment and command parameters
- Execution of UPNP benchmark every ~9 days
- Improvements to Flux App messages reconstruction
- Improvements to Flux App messages searching
- Fixes opening an App from running list in management
- Calculates Titan lockup APR including Nimbus nodes
- FluxBench v3.8.0 enforcement


ZelBack/config/default.js
```diff
@@ -145,6 +145,8 @@ module.exports = {
     expireFluxAppsPeriod: 100, // every 100 blocks we run a check that deletes apps specifications and stops/removes the application from existence if it has been lastly updated more than 22k blocks ago
     updateFluxAppsPeriod: 9, // every 9 blocks we check for reinstalling of old application versions
     removeFluxAppsPeriod: 11, // every 11 blocks we check for more than maximum number of instances of an application
+    reconstructAppMessagesHashPeriod: 25, // every 5 days we ask for old messages
```
Cabecinha84, line 148: this will be executed every 25 blocks, not every 5 days.

ZelBack/config/default.js
```diff
@@ -145,6 +145,8 @@ module.exports = {
...
+    benchUpnpPeriod: 27, // every 9 days execute upnp bench
```
Cabecinha84, line 149: this will be executed every 27 blocks, not every 9 days.

Cabecinha84 (changes requested):
Two things that needs to be adjusted before released. Cheers

TheTrunk:
Correct, check my last commit. It was reduced from the correct values to these fast values for testing purposes.

PR #978, opened 2023-03-26
feat: add owner filter for permanent apps

FluxST:
Adding a filter for owner for permanent messages API. 

This will enable clients to query for all the apps they own/owned whether currently live or not.


ZelBack/src/routes.js
```diff
@@ -304,9 +304,12 @@ module.exports = (app, expressWs) => {
   app.get('/apps/temporarymessages/:hash?', cache('5 seconds'), (req, res) => {
     appsService.getAppsTemporaryMessages(req, res);
   });
-  app.get('/apps/permanentmessages/:hash?', cache('30 seconds'), (req, res) => {
+  app.get('/apps/permanentmessages', cache('30 seconds'), (req, res) => {
     appsService.getAppsPermanentMessages(req, res);
   });
+  app.get('/apps/permanentmessages/:hash', cache('30 seconds'), (req, res) => {
```
TheTrunk, line 310: This should still remain as a one call
/apps/permanentmessages/:hash?/:owner?
-> now the call can take 2 parameters/queries
This should also remain just as a one function and so our query will be
```
  let query = {};
  let { hash } = req.params;
  hash = hash || req.query.hash;
  let { owner } = req.params;
  owner = owner || req.query.owner;
  if (hash) query.hash = hash
  if (owner) query.owner = owner

TheTrunk (changes requested):
Too complex, can be simplified. Unification to one function supporting multiple params/query

PR #989, opened 2023-04-12
Update repositories.json

pvilalta:
Hi there,

My code is open source, but my env variable aren’t. Adding them through the jetpack interface make them confidential ? Or viewable somehow ?

Thanks in advance,
Paul

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -293,5 +293,6 @@
    "sandmanshiri",
    "parzival2599",
    "wickedsensation",
-   "sylvain71"
+   "sylvain71",
+   "velo-drome"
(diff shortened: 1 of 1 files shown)


TheTrunk:
Hi, 
There are certain ways to accomplish it. The easiest is in the home.runonflux.io while registering the app, you can upload the env variables to Flux Storage. This way it will not be visible on the blockchain and only available to nodes running the application.
Another way might be to use Flux Vault - injecting the env parameters at runtime.
Be noted that a malicious actor on a network - malicious node on which your app spawns can still get access to it as the env variables are plain text at docker runtime.

PR #992, opened 2023-04-20
v3.39.0

TheTrunk:


Cabecinha84 (changes requested):
@TheTrunk can we change where we have = this.originalInstances; to  = this.originalInstances();
It's a method and I think it's more easy to read.

Cabecinha84 (approved):
Commited a fix for locked instances. All tested now and working good.
HomeUi needs to be rebuild before release.
Lets schedule to release this on 21st 10am UTC if it's ok for you @TheTrunk 
When Ann lets mark as a mandatory version because of the upnpbench fix.

ZelBack/src/services/benchmarkService.js
```diff
@@ -249,7 +249,7 @@ async function executeUpnpBench() {
   const isUPNP = upnpService.isUPNP();
   if ((userconfig.initial.apiport && userconfig.initial.apiport !== config.server.apiport) || isUPNP) {
     log.info('Calling FluxBench startMultiPortBench');
-    startMultiPortBench();
+    log.info(await startMultiPortBench());
```
TheTrunk, line 252: That is pretty strange @Cabecinha84

Cabecinha84: what you get:
2023-04-20T16:04:05.740Z          Calling FluxBench startMultiPortBench
2023-04-20T16:04:05.742Z          {"status":"success","data":"Multiport Benchmarks will start in the next couple minutes"}

Cabecinha84:
> @TheTrunk can we change where we have = this.originalInstances; to = this.originalInstances();
> It's a method and I think it's more easy to read.

@TheTrunk don't forget this one.

PR #1000, opened 2023-05-06
Ufw rules

TheTrunk:
UFW rules


JKTUNING:
- Check for UFW Deny Rules on FluxOS launch and purge any deny rules that are in the app port range
- Delete UFW app port allow rule, instead of deny ufw rule
- Add helper shell script to remove deny rules

Will need testing before production.

w2vy:
@Cabecinha84 in restoreFluxPortsSupport the allowPort is called 3 times without checking to see if the firewall is active same for restoreAppsPortsSupport and in checkMyAppsAvailability for testingPort

Also checkMyAppsAvailability also has unbounded recursion if all the app ports appear busy (once every 120 sec - so not horrible)

TheTrunk:
> @Cabecinha84 in restoreFluxPortsSupport the allowPort is called 3 times without checking to see if the firewall is active same for restoreAppsPortsSupport and in checkMyAppsAvailability for testingPort
> 
> Also checkMyAppsAvailability also has unbounded recursion if all the app ports appear busy (once every 120 sec - so not horrible)

I will add some more isFirewallActive checks

PR #1002, opened 2023-05-08
v3.40.0

TheTrunk:
- Adds support for any docker registry
  - Example: Github repositories, Amazon ECR and more
- To use different registry than docker hub, specify the provider in the repotag as well.
  - Examples:
    - public.ecr.aws/docker/library/hello-world:linux
    - ghcr.io/handshake-org/hsd:latest
 - It is also possible to use your own docker registry that is compliant with docker v2 API
   - Example:
     - myregistry.com:35000/mynamespace/image:latest
 - Instead of denying port, Flux is now deleting previously allowed ports. This will result in less spammed firewall rules list
 - Old Deny records will be automatically purged
 - Added mounted test for applications
   - Flux will now test to mount application volumes upon start determining if this node can truly run apps
   - If node fails this test, it will get a DOS state
   - This test will begin dossing entire node on the 1st of June. Until then a DOS message on API call of Flux Info will appear
 - Improves uninstallation of corrupted applications
 - Adjusts handling of locked instances on Home UI
 - Various minor improvements


Cabecinha84 (approved):
Ack. HomeUi needs to be rebuild and only leave needed files.

Cabecinha84:
@TheTrunk Lets schedule to release this saturday at 10am Lisbon time?

TheTrunk:
> Update HomeUI (I need to learn how to know what files are needed and …

Yeah we need to come with some better approach for building. Some deterministic build, locking dependencies

Cabecinha84:
HomeUI needs to be rebuild again and I believe all is good/ready for tomorrow release.

alihm (approved):

PR #1010, opened 2023-05-24
v4.0.0

TheTrunk:
FluxOS v4 brings enterprise capabilities to Flux!

- Application v7 specifications
  - Enterprise Nodes
    - 'nodes' specifications, array of IP addreses (or collateral txid:outidx)
    - Each node on Flux network now has an enterprise score
    - The score is determined by following
      - Collateralization of node. Stratus nodes score higher than Nimbus, Nimbus scores higher than Cumulus
      - Maturity of node. The older the node is active on the network the more score it gets
      - Public Key collateralization. The same user hosting multiple nodes scores higher
      - Known Node Public Key. Nodes operated by known people, KYC, Flux team get extra score
    - The bigger the overall score, the more reliable the node and operator is
    - User can specify up to 110 nodes to run their application
    - It is possible to target specific nodes for application deployment
    - Applications that target nodes are called enteprise and get access to private images, secrets which is achieved by encrypting app specifications.
  - Private images
    - 'repoauth' specification, string
    - it is now possible to run private images on FluxOS.
    - supported for docker hub as well as any other docker container registrars (ghcr, gcr, aws, custom) 
    - To use private images, repoauth must be specified. Repoauth consists of username:password or username:token (recommended)
    - To use private images, a set of enterprise nodes must be specified
    - repoauth is encrypted by PGP ECC key of selected enterprise nodes
  - Secrets
    - 'secrets' specifications, string
    - Environment parameters initially specified on UI as an array of strings such as standard env parameters that gets encrypted by enterprise nodes
    - To use secrets, a set of enterprise nodes must be specified
  - Static IP
    - 'staticip' specification, boolean
    - This flag will force FluxOS to spawn your applications on nodes that are known to have static IP, usually hosted on known VPS providers
- PGP identity
  - Each node now generates PGP ECC keypair to served as their encryption key.
- App Price updates
  - scope
    - Application targetting specific nodes - enterprise application have additional fee
  - staticip
    - Application requiring static IP address have additional price increment
  - enteprise port
    - application requiring an outside port in enterprise range have additional price increment
- Port range changes
  - FluxOS now requires ports range 1-65535. Previous scheme required only Flux reserved ports and port range of 30000-39999. FluxNode operators with home hosting and manual port forwarding set for the FluxNode will have to adjust router.
  - Some ports such as 22, 23 and couple more (defined in configuration file) are banned and not possible to be assigned to applications. More information about banned ports at bottom.
  - Some ports primarily in range 1-1023 and couple more (defined in configuration file) are enterprise ports. Using those ports comes with app price increment. More information about enterprise ports at bottom.
- Application internal networking scheme change
  - Previous scheme was assigning private networking of `172.${number}.0.0/16` which was wrong and sometimes resulted in assigning public IP address. This releases fixed the assigned networking to following scheme
  - `172.23.${number}.0/24`
  - Applications requiring internal communication of their components shall adjust this appropriately
- Adjust Syncthing life cycle. Syncthing was previously controlled and associated to Flux process. Which caused lots of restarts and reliability issues. In this release we have decoupled Syncthing process from FluxOS. Syncthing will now stay running eventhough FluxOS is stopped.
- User defined blocked ports list. In userconfig it is possible to defined blockedPorts variable. It is an array of ports (integers) which forces Flux to not use defined port for applications and excludes it from testing. This is useful for home hosting, port forwarding when another device on the network already uses that port
- Port checking changes and DOS state. Flux now tests random ports periodically, in terms of multiple consequent fails of some port test, DOS state will occur
- Prior application installation a proper port check test is present
- Other minor features
- Various Bug Fixes 
      
API changes
- /flux/staticip returns information if specific FluxOS has static IP
- /flux/pgp returns information about PGP key of FluxOS
- /apps/enterprisenodes returns enterprise nodes list

Activation blockheight: **1420000**
All nodes must update prior that blockheight.

bannedPorts: ['16100-16299', '26100-26299', '30000-30099', 8384, 27017, 22, 23, 25, 3389, 5900, 5800, 161, 512, 513, 5901, 3388, 4444, 123]

enterprisePorts: ['0-1023', 8080, 8081, 8443, 25565, 6667]

example of user/node operator defined blocked ports on ports 55, 70, 80
config file is userconfig.js under zelflux/config folder.
blockedPorts: [55, 70, 80]
image

_This release is heavy and comes with additional packages. FluxOS might take a while to update to this release. Node operator should monitor their node while updating_


ZelBack/src/services/appsService.js
```diff
@@ -9219,10 +9548,26 @@ async function checkMyAppsAvailability() {
...
       }
     }
+    const minPort = currentHeight.generalScannedHeight >= config.fluxapps.portBlockheightChange ? config.fluxapps.portMinNew : config.fluxapps.portMin - 1000;
+    const maxPort = currentHeight.generalScannedHeight >= config.fluxapps.portBlockheightChange ? config.fluxapps.portMaxNew : config.fluxapps.portMax;
     // choose random port
-    const min = config.fluxapps.portMin - 1000;
-    const max = config.fluxapps.portMax;
+    const min = minPort;
+    const max = maxPort;
+    if (failedPort && previouslyFailedPort === failedPort) {
```
Cabecinha84, line 9556: This needs to be removed. It was correct like it was before. This way, I can block a range of ports, the test fails, and before it's DOS, the probability of hiting a diferent port is huge and the node will never get DOS.
If there is a port that fails the test, the tests needs to continue to be on same port until it passes.

Cabecinha84 (changes requested):
Asked for a minor change. All the rest is good to be released.

ZelBack/src/services/appsService.js
TheTrunk: user defined blocked port list added
strict dos implemented

alihm (approved):

Cabecinha84 (changes requested):
The code is good, everything seems to be working but I am not 100% sure about this "enterprise nodes" feature like it is. Lets have a meeting to discuss it as business side I don't think it's ok and I think I need to explain you guys why and what can we do.

johnhanlon86 (commented):
@TheTrunk Looks great!

For max enterprise nodes, are we going with 110 or 120? Because both seem to be mentioned in different places.

It may also be worth doing a search/replace for all instances of:
- Enteprise
- Eterprise
- fetchEntepriseKey
- pgpSrevice
- userBlockedPorst

ZelBack/src/services/enterpriseNodesService.js
```diff
@@ -0,0 +1,155 @@
...
+    });
+    return enterpriseList;
+  } catch (error) {
+    log.error(error);
+    return [];
+  }
+}
+
+/**
+ * To get FluxOS tail error logs. Only accessible by admins and Flux team members.
```
johnhanlon86, line 132:
```suggestion
 * To get list of enterprise nodes via API request.
```

Cabecinha84 (approved):
Ack. I have adjusted the release notes.

PR #1035, opened 2023-06-14
Syncthing Install Fix

JKTUNING:
Running with & seems to return a non-error state from node-cmd, which results in syncthing not being installed.

Using just nohup should be sufficient to decouple the process from FluxOS


w2vy:
This will not fix it. As discussed we should specifically check to make sure syncthing is installed and install it instead of depending on a start failure to trigger the install

akhtarg:
Why not just install syncthing as a service? Becomes completely independant of fluxos then...

w2vy:
That might be a good idea! We'll kick it around

Thank You

Cabecinha84:
Thank you guys. We already have fixed it in dev branch, soon we will release the patch on mainnet.

PR #1042, opened 2023-06-24
v4.1.0

TheTrunk:
- Implements Ethereum address verification
  - Flux Node can now be owned by an Ethereum address as well
  - Possible to sign in to Flux node with an Ethereum address
  - Application can now be owned by an Ethereum address after timestamp 1688947200000 (July 10th)
  - This extends our current scheme of bitcoin p2pkh address verification
- Improvements on Syncthing implementation
- Add restrictions to Syncthing log file
- Fix bug on Syncthing unable to delete non existing folder
- Resolve dockerd high usage by disabling monitoring of logs file

This is a forking mandatory release. All nodes must update prior July 10th.


Cabecinha84 (approved):
Ack

PR #1044, opened 2023-06-24
Update repositories.json

theraw:
infos are exactly the same https://github.com/RunOnFlux/flux/pull/1033 (taken from the person that will run peertube)

the reason why i'm transfering this is because the person that was interested in running peertube on flux liked the project and wants to continue it, unfortunately i used my personal dockerhub account and now we have to swap to their account because i cannot give my dockerhub to them due to some personal/private projects..

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -314,7 +314,7 @@
    "hughparry",
    "woodpeckerci",
    "gitea",
-   "theraw/peertube:latest",
+   "privacyfreedom/peertube:latest",
    "zar1f",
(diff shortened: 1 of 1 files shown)


TheTrunk:
Hi, please do not remove or adjust current whitelists and follow the pull request form. Make it as a new whitelist submission

theraw:
looks like my branch was out of date with master, and removed helloskyy by mistake, i corrected that to my second commit or we can close/cancel this and i make a new pull request however you like

PR #1057, opened 2023-07-10
v4.3.0

TheTrunk:
- Allow WalletConnect and Metamask
  - Users can now use WalletConnect and Metamask to log into FluxOS directly using their Ethereum address as well as own, register or update applications ownership.
- Fixes v7 UI of nodes section and staticip
- Fixes ethereum regex
- Various other minor fixes


HomeUI/src/views/apps/Management.vue
```diff
@@ -116,7 +116,7 @@
             />
             <list-entry
               title="Static IP"
-              :data="callResponse.datastaticip ? 'Yes, Running only on Static IP nodes' : 'No, Running on all nodes'"
+              :data="callResponse.data.staticip ? 'Yes, Running only on Static IP nodes' : 'No, Running on all nodes'"
```
Cabecinha84, line 119: Nice catch!

Cabecinha84 (approved):
I don't have walletconnect or metamask, but it's only UI and the code seems good. When I am back to Lisbon I will start using those.
I think those are still missing on Xdao page (in case you want to add it for this release).
For now, ack, great job, love the logos!

TheTrunk:
> I don't have walletconnect or metamask, but it's only UI and the code seems good. When I am back to Lisbon I will start using those. I think those are still missing on Xdao page (in case you want to add it for this release). For now, ack, great job, love the logos!

xdao is skipped as it expects zelid on the backend, can't use eth yet. Will add to board for rechecking and allowing.

FliKites:
I can test the metamask and wallet connect once it's pushed. I have both.

PR #1063, opened 2023-07-18
Introducing apprunning message version 2

Cabecinha84:
- New fluxapprunning message version 2;
- Version 2 will be used when a fluxNode is running more than one app, instead of sending one message for each app to the network, it agregates all information in just one message (version2).

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -8347,13 +8347,25 @@ async function checkAndNotifyPeersOfRunningApps() {
         installedAndRunning.push(app);
       }
     });
-    // eslint-disable-next-line no-restricted-syntax
-    for (const application of installedAndRunning) {
-      log.info(`${application.name} is running properly. Broadcasting status.`);
-      try {
+
+    const syncStatus = daemonServiceMiscRpcs.isDaemonSynced();
+    const daemonHeight = syncStatus.data.height || 0;
+
+    if (daemonHeight >= config.fluxapps.apprunningv2 && installedAndRunning.length > 1) {
+      // eslint-disable-next-line no-restricted-syntax
+      const broadcastedAt = new Date().getTime();
+      const newAppRunningMessageV2 = {
+        type: 'fluxapprunning',
+        version: 2,
+        ip: myIP,
+        broadcastedAt,
+      };
+      const apps = [];
+      // eslint-disable-next-line no-restricted-syntax
+      for (const application of installedAndRunning) {
+        log.info(`${application.name} is running properly. Broadcasting status.`);
         // eslint-disable-next-line no-await-in-loop
         // we can distinguish pure local apps from global with hash and height
-        const broadcastedAt = new Date().getTime();
         const newAppRunningMessage = {
           type: 'fluxapprunning',
           version: 1,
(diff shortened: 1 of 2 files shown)


ZelBack/src/services/appsService.js
```diff
@@ -6034,7 +6034,20 @@ async function storeAppRunningMessage(message) {
...
+
+  if (message.version !== 1 && message.version !== 2) {
+    return new Error(`Invalid Flux App Running message for storing version ${message.version} not supported`);
+  }
+
+  if (message.version === 1 && (typeof message.hash !== 'string' || typeof message.name !== 'string')) {
+    return new Error('Invalid Flux App Running message for storing');
+  }
+
+  if (message.version === 2 && (!message.apps || !Array.isArray(message.apps))) {
```
TheTrunk, line 6050: Here we should also check the typo of every item in the array

ZelBack/src/services/appsService.js
```diff
@@ -6056,30 +6069,71 @@ async function storeAppRunningMessage(message) {
...
+    const options = {
+      upsert: true,
+    };
+    await dbHelper.updateOneInDatabase(database, globalAppsLocations, queryUpdate, update, options);
+  } else {
+    let messageNotOk = false;
+    for (let i = 0; i < message.apps.length; i += 1) {
+      const app = message.apps[i];
+      if (typeof app.hash !== 'string' || typeof app.name !== 'string') {
+        return new Error('Invalid Flux App Running v2 message for storing');
```
TheTrunk, line 6102: This type check is what should happen earlier (just so we have type validation at one place)

ZelBack/src/services/appsService.js
```diff
@@ -6056,30 +6069,71 @@ async function storeAppRunningMessage(message) {
...
+      };
+
+      // indexes over name, hash, ip. Then name + ip and name + ip + broadcastedAt.
+      const queryFind = { name: newAppRunningMessage.name, ip: newAppRunningMessage.ip, broadcastedAt: { $gte: newAppRunningMessage.broadcastedAt } };
+      const projection = { _id: 0 };
+      // we already have the exact same data
+      // eslint-disable-next-line no-await-in-loop
+      const result = await dbHelper.findOneInDatabase(database, globalAppsLocations, queryFind, projection);
+      if (result) {
+        // found a message that was already stored/bad message
```
TheTrunk, line 6119: this was only stored, can't be a bad message I think

ZelBack/src/services/appsService.js
```diff
@@ -6056,30 +6069,71 @@ async function storeAppRunningMessage(message) {
...
+        messageNotOk = true;
+        break;
```
TheTrunk, line 6121: Here we should do continue to process the other parts of the message, apps that can be running

ZelBack/src/services/appsService.js
```diff
@@ -6056,30 +6069,71 @@ async function storeAppRunningMessage(message) {
...
+      const queryUpdate = { name: newAppRunningMessage.name, ip: newAppRunningMessage.ip };
+      const update = { $set: newAppRunningMessage };
+      const options = {
+        upsert: true,
+      };
+      // eslint-disable-next-line no-await-in-loop
+      await dbHelper.updateOneInDatabase(database, globalAppsLocations, queryUpdate, update, options);
+    }
+    if (messageNotOk) {
+      return false;
```
TheTrunk, line 6132: Correct, I agree that rebroadcast should happen only in the case when we did not find any of the running messages in our database. If we find a single one, we should not rebroadcast

ZelBack/src/services/appsService.js
TheTrunk: Actually OK. Lets just break it and not process further.

ZelBack/src/services/appsService.js
TheTrunk: can leave it

ZelBack/src/services/appsService.js
```diff
@@ -8347,13 +8401,25 @@ async function checkAndNotifyPeersOfRunningApps() {
...
     });
-    // eslint-disable-next-line no-restricted-syntax
-    for (const application of installedAndRunning) {
-      log.info(`${application.name} is running properly. Broadcasting status.`);
-      try {
+
+    const syncStatus = daemonServiceMiscRpcs.isDaemonSynced();
+    const daemonHeight = syncStatus.data.height || 0;
+
+    if (daemonHeight >= config.fluxapps.apprunningv2 && installedAndRunning.length > 1) {
```
TheTrunk, line 8408: to be simplified to not have repetitive code

TheTrunk (changes requested):
1. Processing logic of running messages OK. 
- Suggestion is to move the validation check to front so its all grouped together.
(I was thinking of disallowing other keys but its not needed as its not getting stored) 
2. Logic of the function checkAndNotifyPeersOfRunningApps is all OK. 
- The suggestion is to simplify the function as there is some repetitive code included in both if and else block.
3. Test suite - some more tests are broken with this and it would be great to cover the additional logic added

ZelBack/src/services/appsService.js
Cabecinha84: ohh, I have missed this one... let me see what can I do

Cabecinha84:
@TheTrunk 
All things adjusted/changed. I thinks it's more cleaner now.
TestSuite don't have tests for the fluxapprunning.
Is one of the things I have on the list to improve is the TestSuite, for sure lots of tests can be added. Once I have a computer that can run tests locally.

TheTrunk (approved):
Nicely done :+1:

PR #1070, opened 2023-07-19
fix response type

alihm:
Adjustments on response type


ZelBack/src/services/syncthingService.js
```diff
@@ -1723,11 +1723,20 @@ async function debugCpuprof(req, res) {
...
   let response = null;
   if (authorized === true) {
-    response = await performRequest('get', '/rest/debug/cpuprof', undefined, 60000);
+    try {
+      response = await axios.get('/rest/debug/cpuprof', {
+        responseType: 'stream', // Specify response type as stream
+        timeout: 60000,
+      });
+      res.setHeader('Content-Type', 'application/octet-stream');
+      return response.data.pipe(res);
```
TheTrunk, line 1732: was this tested?

ZelBack/src/services/syncthingService.js
```diff
@@ -1757,11 +1775,15 @@ async function debugSupport(req, res) {
   const authorized = res ? await verificationHelper.verifyPrivilege('adminandfluxteam', req) : true;
   let response = null;
   if (authorized === true) {
-    response = await performRequest('get', '/rest/debug/support', undefined, 60000);
+    try {
+      response = await performRequest('get', '/rest/debug/support', undefined, 60000);
```
TheTrunk, line 1779: performRequest is has already try catch and practically can't throw

TheTrunk (changes requested):

ZelBack/src/services/syncthingService.js
alihm: yes, it pipes the response from Syncthing.

ZelBack/src/services/syncthingService.js
alihm: Oh, fixed it.

PR #1071, opened 2023-07-20
Add LogLevels and Optimize logging

valamidev:
Improve Flux ZelBack logging:

- Add configuration to adjust logging.
- Optimize logging with storing file size in memory and save an unnecessary sync file operation.


TheTrunk (approved):
This is really helpful in reducing the calls on fs stats 👍 
The config flag is NOT user editable (that would need to be specified in userconfig.js file). It is good to have it but currently user can't edit it and we probably do not want to allow editing of it either so we can ask for full logs of the node to diagnose some issues.

valamidev:
> This is really helpful in reducing the calls on fs stats 👍 The config flag is NOT user editable (that would need to be specified in userconfig.js file). It is good to have it but currently user can't edit it and we probably do not want to allow editing of it either so we can ask for full logs of the node to diagnose some issues.

Yes, I would keep this as a powerUser/Dev settings for now. There are some unknown source of logs, numbers without any further explain what they are where they came from e.g. `36900` in the logs. 
These should be addressed first to make sure that every log entry is traceable back where it came from, then we can optimize for log level an only keep those logs in warning/error which effect the Flux node stability.

Cabecinha84 (approved):
Ack!
Thank you valamidev

TheTrunk:
> > This is really helpful in reducing the calls on fs stats 👍 The config flag is NOT user editable (that would need to be specified in userconfig.js file). It is good to have it but currently user can't edit it and we probably do not want to allow editing of it either so we can ask for full logs of the node to diagnose some issues.
> 
> Yes, I would keep this as a powerUser/Dev settings for now. There are some unknown source of logs, numbers without any further explain what they are where they came from e.g. `36900` in the logs. These should be addressed first to make sure that every log entry is traceable back where it came from, then we can optimize for log level an only keep those logs in warning/error which effect the Flux node stability.

Agreed would be great to know the exact file and line that caused a log write.

PR #1072, opened 2023-07-24
v4.5.0

TheTrunk:
- App Running messages specification version 2
  - Version 2 of application running messages optimises messages on FluxOS. The message now contains all applications that run on a node effectively broadcasting just one message to do the network instead of X messages (1 per app).
  - Activation blockheight: 1444444
- Improvements on logging mechanism
- Increase on maximum amount of incoming, outgoing connections
- Increases in rate limit allowance
- Corrections on Syncthing protected calls
- Option to turn on debugging of Syncthing (automatically turned on for development nodes)
- Improvements in Syncthing integration
- Home UI fix for upgrading of non private v7 application toggle
- General minor corrections and improvements

All nodes should upgrade prior height 1444444.


alihm (approved):

Cabecinha84 (approved):
Ack

PR #1075, opened 2023-07-26
[Improve] Improve passing initial benchmark with halting Block Explorer process

valamidev:
**Reason:**
- Populating Block Explorer database is both CPU and Disk heavy operation, which can cause that Benchmark stuck in a failing/restarting loop until the Block Explorer process finish.

**Solution:**
- Only process block once benchmark in `success` status, and halt while running, failed, restarting.

### DIFF
--- a/ZelBack/src/services/explorerService.js
+++ b/ZelBack/src/services/explorerService.js
@@ -28,6 +28,8 @@ let operationBlocked = false;
 let initBPfromNoBlockTimeout;
 let initBPfromErrorTimeout;
 
+const MINUTE_IN_MS = 60 * 1000;
+
 // cache for nodes
 const LRUoptions = {
   max: 20000, // store 20k of nodes value forever, no ttl
@@ -644,7 +646,15 @@ async function processBlock(blockHeight, isInsightExplorer) {
     if (!syncStatus.data.synced) {
       setTimeout(() => {
         processBlock(blockHeight, isInsightExplorer);
-      }, 2 * 60 * 1000);
+      }, 2 * MINUTE_IN_MS);
+      return;
+    }
+    const benchmarkBenchRes = await benchmarkService.getStatus();
+    if(benchmarkBenchRes.status !== 'success') {
+      log.info(`Processing Explorer halted at Height: ${blockDataVerbose.height}, reason Benchmark running or failed`);
+      setTimeout(() => {
+        processBlock(blockHeight, isInsightExplorer);
+      }, 2 * MINUTE_IN_MS);
       return;
     }
     someBlockIsProcessing = true;
--- a/ZelBack/src/services/explorerService.js
+++ b/ZelBack/src/services/explorerService.js
@@ -29,6 +29,7 @@ let initBPfromNoBlockTimeout;
 let initBPfromErrorTimeout;
 
 const MINUTE_IN_MS = 60 * 1000;
+const BENCHMARK_STATUS_CACHE = {validUntil: 0, status: 'none'};
 
 // cache for nodes
 const LRUoptions = {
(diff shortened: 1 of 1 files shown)


ZelBack/src/services/explorerService.js
```diff
@@ -644,7 +646,15 @@ async function processBlock(blockHeight, isInsightExplorer) {
     if (!syncStatus.data.synced) {
       setTimeout(() => {
         processBlock(blockHeight, isInsightExplorer);
-      }, 2 * 60 * 1000);
+      }, 2 * MINUTE_IN_MS);
+      return;
+    }
+    const benchmarkBenchRes = await benchmarkService.getStatus();
```
TheTrunk, line 652: this call does not have any caching. My worry is that it will be executed very oftenly and can be heavy for the benchmark app side. Slowing everything down with high usage. Can we recheck this worry and test the usage or/and adding some 30 sec cache for the call?

TheTrunk (changes requested):

Cabecinha84 (commented):
Fluxbench checks both cpu usage before executing eps test and disk write speed in use before disk test. Despite they might not be 100% accurate it helps a lot in this scenarios.
I don't think we should be halting anything. We should analyze the reason benchmark is failing and if needed adjust fluxbench. Cheers

ZelBack/src/services/explorerService.js
valamidev: I add a Cache for 1 minute.

valamidev:
> Fluxbench checks both cpu usage before executing eps test and disk write speed in use before disk test. Despite they might not be 100% accurate it helps a lot in this scenarios. I don't think we should be halting anything. We should analyze the reason benchmark is failing and if needed adjust fluxbench. Cheers

I found while setting up multiple Nodes, that this scenario happen, that:

Benchmark run + MongoDB/NodeJS service under heavy load 
Benchmark fail/restart
Benchmark run + MongoDB/NodeJS service under heavy load 
Benchmark fail/restart
so on...

Cabecinha84:
> > Fluxbench checks both cpu usage before executing eps test and disk write speed in use before disk test. Despite they might not be 100% accurate it helps a lot in this scenarios. I don't think we should be halting anything. We should analyze the reason benchmark is failing and if needed adjust fluxbench. Cheers
> 
> I found while setting up multiple Nodes, that this scenario happen, that:
> 
> Benchmark run + MongoDB/NodeJS service under heavy load Benchmark fail/restart Benchmark run + MongoDB/NodeJS service under heavy load Benchmark fail/restart so on...

What is the reason of failure? Cpu? Disk?
I need the output of fluxbench-cli getbenchmarks
on that scenario, what is the cpu, what is the disk, and if it's under virtualization how many nodes running on the host.

The scenario you are pointing is the normal scenario for all new nodes on the network. And we have lots of new nodes joining the network all the time.

valamidev:
> > > Fluxbench checks both cpu usage before executing eps test and disk write speed in use before disk test. Despite they might not be 100% accurate it helps a lot in this scenarios. I don't think we should be halting anything. We should analyze the reason benchmark is failing and if needed adjust fluxbench. Cheers
> > 
> > 
> > I found while setting up multiple Nodes, that this scenario happen, that:
> > Benchmark run + MongoDB/NodeJS service under heavy load Benchmark fail/restart Benchmark run + MongoDB/NodeJS service under heavy load Benchmark fail/restart so on...
> 
> What is the reason of failure? Cpu? Disk? I need the output of fluxbench-cli getbenchmarks on that scenario, what is the cpu, what is the disk, and if it's under virtualization how many nodes running on the host.
> 
> The scenario you are pointing is the normal scenario for all new nodes on the network. And we have lots of new nodes joining the network all the time.

Mostly Disk error, the Disk I/O requirement is quite high compared to others + MongoDB is infmaus with Disk I/O usage.

I also spoke with other users, it is a very common scenario that a new Node is stuck in this condition, and just make it worst that every N minute it waste resources for benchmarking.

TheTrunk:
All load of node including block processing is accounted to the benchmark execution. We are open to different suggestions rather than halting benchmarks but network is stable and we do not observe such behaviour nor see more community inputs on this matter for months

PR #1086, opened 2023-08-14
v4.8.0

TheTrunk:
- routerIP property can now be added to configuration file (this should be filled with local router IP in case using UPNP for better support, multitoolbox option 14 will be changed to start writting this property on FluxOs);
- Fix bug that was preventing new nodes on the network asking for apps messages missing;
- Other small improvements.


Cabecinha84 (approved):
Ack

PR #1105, opened 2023-09-12
NodeJS

TheTrunk:
Support:
Ubuntu 18.04, 20.04, 22.04
NodeJS 16, 18, 20
MongoDB 4.4, 5.0, 6.0, 7.0
with 22.04, 20, 7.0 being recommended

Needs more testing


Cabecinha84:
tested ok both with node v18 and node v20.
Do you want to use this branch also to update packages?
image

Cabecinha84:
build homeui is not working on node 20 (haven't tried on 18)
image

Cabecinha84 (changes requested):
HomeUI is not building on node 20

TheTrunk:
> HomeUI is not building on node 20

Ahh I think its because you are building on windows. It now depends on grep to check for what nodejs version you run. I think non of the home commands will work on windows now, I will have to figure out better way it seems

PR #1115, opened 2023-10-08
v4.13.0

Cabecinha84:
- Gives FluxOS builds support to:
 - Ubuntu 18.04, 20.04, 22.04;
 - NodeJS 16, 18, 20;
 - MongoDB 4.4, 5.0, 6.0, 7.0;
- Small changes on development marketplace;
- Adds port 53 to the banned port list by FluxOS as it is used by Ubuntu/Debian DNS and it caused FluxOS to crash when it tried to use it;
- Adds a range of ports not available on UPNP nodes to prevent the FluxOs error "Failed: Error on FluxOs response - Applications port range is not reachable from outside!";
- Other code improvements.


alihm (approved):

package.json
```diff
@@ -89,6 +89,7 @@
     "fast-xml-parser": "~4.2.4",
     "formidable": "~2.1.2",
     "fullnode": "file:lib/fullnode",
+    "http-shutdown":"~1.2.2",
```
TheTrunk, line 92: I do not understand this move. Can you provide more insight on what is wrong on our current approach and why is this needed?

ZelBack/src/services/appsService.js
```diff
@@ -10055,28 +10064,42 @@ async function checkMyAppsAvailability() {
...
       checkMyAppsAvailability();
       return;
     }
     if (appPorts.includes(testingPort)) {
-      log.warn(`Skipped checking ${testingPort} - in use.`);
+      log.info(`checkMyAppsAvailability - Skipped checking ${testingPort} - in use.`);
       failedPort = null;
       // skip this check
-      await serviceHelper.delay(2 * 60 * 1000);
+      await serviceHelper.delay(15 * 1000);
```
TheTrunk, line 10102: intentional?

TheTrunk (changes requested):
Nice, Thank you for commenting on the shutdown package

TheTrunk (approved):
Cool I read on the shutdown package - seems cool. Did not know it exists. 
Still would appreciate your comment and if it got replaced in every scenario of our test servers.

ZelBack/src/services/appsService.js
Cabecinha84: yes, the other if's clauses had a 15 seconds await for the same kind of behaviour, the random por selected can't be used for testing. So await 15 seconds and call the method again is good, 2 minutes is a lot.

Cabecinha84:
> Cool I read on the shutdown package - seems cool. Did not know it exists. Still would appreciate your comment and if it got replaced in every scenario of our test servers.

I have explained on the other comment the use of http-shutdown. It is already replacing the two places on appsService where we are creating a server to listen and were calling the .close() event.
testingAppserver
beforeAppInstallTestingServer

Both are now changed to use http-shutdown ->
testingAppserver.shutdown((err) => {
      if (err) {
        log.error(`testingAppserver Shutdown failed: ${err.message}`);
      }
    });
    
beforeAppInstallTestingServer.shutdown((err) => {
        if (err) {
          log.error(`beforeAppInstallTestingServer Shutdown failed: ${err.message}`);
        }
      });

PR #1121, opened 2023-10-17
v4.14.0

Cabecinha84:
- Only considers error 'Applications port range is not reachable from outside' after more then 100 ports have failed;
- FluxOs now accepts a new property on configuration file blockedRepositories. In this property node operator can define up to 10 repositories that don't want to run on the fluxNode. Repos from flux Marketplace will not be take in consideration. More information bellow;
- Several Updates on HomeUI.

blockedRepositories is defined as a string array. Changing configuration file auto restarts fluxOS. It can take close to 1h for an app that you add to the blockedRepositories and is running on your node to be uninstalled.
Example of a configuration file:

module.exports = {
  initial: {
    ipaddress: '185.x.x.x',
    zelid: 'hansolo',
    kadena: 'kadena:k:blablabla?chainid=0',
    testnet: false,
    development: false,
    apiport: 16127,
    routerIP: '',
    pgpPrivateKey: `-----BEGIN PGP PRIVATE KEY BLOCK-----

blablabla
-----END PGP PRIVATE KEY BLOCK-----
`,
    pgpPublicKey: `-----BEGIN PGP PUBLIC KEY BLOCK-----

blablabla
-----END PGP PUBLIC KEY BLOCK-----
`,
    blockedPorts: [500, 501, 502, 55645],
    blockedRepositories: ['test/banthis', 'block/dontrunonme'],
  }
}


Cabecinha84:
I have added routerIP, apiport, blockedPorts and the new blockedRepositories to flux admin page on homeUI, but the last two are not working. If you can help me on that...
Or we can release like it is now.
If it is fixed we should change release notes and say that these configurations are available to edit on flux admin page.
@TheTrunk

TheTrunk (approved):

alihm (approved):

PR #1122, opened 2023-10-17
Fix Flux Storage for Commands

TheTrunk:
- Fixes using Flux Storage for Commands
- do not add Flux Storage Cmd or Env as a docker creation parameter


Cabecinha84 (approved):
Ack

PR #1131, opened 2023-10-24
v4.15.0

Cabecinha84:
- Nodes under UPNP where router is always failing to map ports (not possible to run any app) will get DOS state on fluxOS and start failing benchmarks
- All applications are now updated to version 7 on update specifications/renew
- Improvements on Application Management Page (Tabs: Information, Resources, File Changes, Processes, Logs, Execute Commands Output - commands are now a string not an array)
- Added to Flux Admin Page the ability to adjust apiport, routerIp, blockedPorts and blockedRepositories
- Prevents nodes from installing application that is IP scoped
- Other small improvements

This version will be enforced on November 7th.


package.json
```diff
@@ -102,6 +102,8 @@
     "openpgp": "~5.9.0",
     "path": "~0.12.7",
     "qs": "~6.11.2",
+    "splitargs": "~0.0.7",
```
TheTrunk, line 105: Lets move both of these to devDependencies as they are only required for UI (devDep are for UI that we are building)

Cabecinha84: moved to devDependencies

TheTrunk (approved):

PR #1132, opened 2023-10-24
Rework repoTag parsing, add dynamic registry auth provider discovery

MorningLightMountain713:
I've put this as WIP as I want to do a bit more testing on a couple of areas.  Also - feel free to reject this if you'd rather it on a seperate branch than develop.

I've run the relevant tests that I've modified / added - they're all passing. However, I'm not sure how to setup a full test environment so haven't run the full catalogue. Happy for some guidance here.

I probably need to add a few tests around the auth token fetching.

I also added a fake repositories.json respose and added that to the test stubs. This stops axios going out and fetching it on every test. 

### My Salespitch:

First off - I've added a semantic change to the way the whitelist is used. This has irked me for quite some time. Currently, the whitelist has a combination of namespace/image:tag, namespace:tag, namespace etc. Which looks like it should whitelist based on the "depth" of the image. I.e if namespace/image:tag is in the whitelist, then only namespace/image:tag should be allowed, and not any of it's ancestors. Same with images etc. However, it actually only checks the namespace.

I've changed the whitelist so it is enforced based on whatever is in the whitelist.

namespace -> any image:tag for this namespace okay
namespace:image -> any tag for this image okay
namespace:image:tag -> specific tag

Obviously, this is a breaking change. However, I've done some analysis with the following :

```bash
curl -s https://api.runonflux.io/apps/permanentmessages | jq "[[.data[] | select(.appSpecifications.expire + .height > $(curl -s https://api.runonflux.io/daemon/getblockcount | jq .data))] | .[].appSpecifications.compose[].repotag] | unique"
```
If you run the above, you will see a list of all unique tags that are running on Flux.

I don't believe any of the running images are affected, however I'm not 100% certain at this point so would like to explore this a little more. (Couple of images are close, like `theraw/s3` and `ngcheewee/gpt4summarizer:latest` and `ngcheewee/gpt3byjsm:latest` however, `nhcheewee` is in the list too so this would be fine. Again, would like to validate this. 

This makes more sense as it gives the team fine grain control on what is allowed to be whitelisted without having to whitelist a whole namespace, then blacklist some of it. It also means what people think the whitelist does, is reflected in reality.

---

### The changes: 

```
This allows for some complex registry tags to be parsed properly,
  that were causing issues prior. Eg:
    europe-west2-docker.pkg.dev/chode-400710/mugawump/cv:latest

  Removes the old splitRepoTag thing in favor of a regex (and updates the
  functions that used this). After testing, the new parseRepoTag function
  is about 40% faster; Readability has also been improved.

  Enforces the whitelist properly, if a namespace is in the whitelist,
  any child is valid. Same with an image or a tag. Note, any ancestors
  are NOT VALID. I.e. if a tag is in the whitelist, a sibling tag is not
  allowed.

  The discovery of registry realm, service and scope were previously
  hardcoded. This has since been removed. Registry auth provider discovery
  is now dynamic based on www-authenticate header. The reason for this fix is
  that not all services use the same auth endpoint as the primary, a good example
  is dockerhub or aws ecr - this was only working as axios was redirecting by default.
  This was is more robust. As part of this change, the mapping of Registry names to
  addresses has been removed. This was only used in error messages and required a
  hardcoded map of names to urls to be maintained. This offered very little value, as
  can just present the url that was tried - the user entered this, so they will know what it is.

  Have modified / added / fixed the tests for repository tag parsing.
```


MorningLightMountain713:
Can confirm no apps are broken by this change. All currently running apps are convered by the whitelist - I checked with Python.

MorningLightMountain713:
I'm pretty much finished here. If you want, let me know how to run the full test suite and if you'd like more tests.

I tested the changes to the `dockerPullStream` function manually - working.

Would be good to get this merged at some point - then I can go back to testing my enterprise app that I want to deploy - it's being hosted by Artifact registry and Flux doesn't work with this at the moment. 

Have tested `verifyRepository` against the following (all working):

// AWS ECR
"public.ecr.aws/docker/library/hello-world:linux" // public - no auth
// Google artifact registry (replacement for GCR)
"europe-west2-docker.pkg.dev/chode-400710/mugawump/cv:latest" // private - user: _json_key_base64
// Dockerhub
"megachips/sneaky:latest" // private - user: my dockerhub
"megachips/lucidwallet:latest" // public - no auth
// Github
"ghcr.io/handshake-org/hsd:latest" // public, but needs auth, my gh user

Will remove the WIP

MorningLightMountain713:
Hello? Be great to get this looked at... I can't deploy my enterprise app until the tag parsing gets fixed.

Since my image has namespacea/namespaceb/image:tag it doesn't parse the org properly. This pull fixes it.

MorningLightMountain713:
Okay, so this PR was logged over a month ago. As of now, there has been zero acknowledgement or response from the Flux development team.

This really shows the lack of a robust developer community around the Flux project, and makes it hard, if not impossible, for community members to contribute. This highlights a real risk around developer resource.

If someone logs a PR, at minimum, the PR should be responded to, acknowledged that it has been received, and will be triaged / reviewed by a team member within x timeframe; even if this means that it is ultimately rejected - that's okay.

There are areas of the Flux codebase that could certainly use some love, refactoring / bugfixing from community members should be looked upon as a good thing. The lack of developers on Flux is a risk if they were to leave the project. There should be a much simplier and robust process around onboarding and ingestion of code.

From your lack of response, I can only assume this code is not what you are looking for, outright broken, or not to your taste. Again, that's fine. However, if that is the case - I would like to learn from and it see what I can do / need to change to make the code suitable, as it does solve several defincies in the current code. 

I would also appreciate some documentation, or guidance, around setting up a proper test environment so I can run / write better tests.

Finally, being busy is not an excuse. I'm sure you have at least read the outline of this PR. It takes 30 seconds to write a comment.

Thanks,

David

TheTrunk:
We are truly sorry for this situation and will do our best to get that looked over asap and fixed for next weeks release and also to ensure that this big f* up on our end won't ever happen again.

TheTrunk:
> namespace -> any image:tag for this namespace okay
namespace:image -> any tag for this image okay
namespace:image:tag -> specific tag

Nice this is how it was intended to be rewritten but we boiled it down to just the namespace. This is appreciated addition and correction.
I believe there there are no conflicts with existing whitelist and apps running. 

I thought your repository of was already working well with the existing codebase as we have a test for 'example.repository.com:50000/complex/namespace/split/image:latest' which should cover your case as well no?

Nice to see a better more clean solution and great changes proposed 👍

ZelBack/src/services/generalService.js
```diff
@@ -313,65 +298,30 @@ function splitRepoTag(repotag) {
...
-  let isWhitelisted = false;
-  splitRepoWhitelistAllowance.forEach((allowedPart) => {
-    if (pureOrganisations.includes(allowedPart)) {
-      isWhitelisted = true;
-    }
-  });
 
-  if (!isWhitelisted) {
-    throw new Error('Repository is not whitelisted. Please contact Flux Team.');
+  const filteredImageTags = res.data.filter((imageTag) => targetDockerTag.slice(0, imageTag.length) === imageTag);
```
TheTrunk, line 319: This does not work well.
Imagine simple scenario where 'runonflux' namespace is whitelisted.
Now if you submit an app with 'runonfluxB/image:tag' it would work as runonflux is whitelisted. But it should not work as runonfluxB is not whitelisted.

ZelBack/src/services/appsService.js
```diff
@@ -4271,29 +4271,27 @@ async function verifyAppMessageUpdateSignature(type, version, appSpec, timestamp
...
 
 /**
- * To verfiy a Docker Hub repository.
- * @param {string} repotag Docker Hub repository tag.
- * @returns {boolean} True if no errors are thrown.
+ * To fetch an auth token from registry auth provider.
+ * @param {object} authDetails Parsed www-authenticate header.
+ * @param {object} AxiosConfig axios Auth object.
  */
+async function getAuthToken(authDetails, axiosConfig) {
```
TheTrunk, line 4278: can we write some tests for the new functions?

TheTrunk: tests added

ZelBack/src/services/generalService.js
TheTrunk: corrected

PR #1134, opened 2023-10-27
added config reload [POC]

XK4MiLX:
- implemented config reload base on fs.watch
- improved ui api call
-  fix typo in dockerContainerExec
- fix bad look button


HomeUI/src/views/fluxadmin/ManageFlux.vue
```diff
@@ -308,7 +308,7 @@
             </b-button>
             <confirm-dialog
               target="update-blockedRepositories"
-              confirm-button="Update Blocked Repositories"
```
Cabecinha84, line 311: It was correct like it was before, its blocked repositories. Please change it.

ZelBack/config/default.js
```diff
@@ -1,6 +1,6 @@
-const userconfig = require('../../config/userconfig');
+let userconfig = require('../../config/userconfig'); // eslint-disable-line
```
Cabecinha84, line 1: what is the  // eslint-disable-line for? if not needed please remove it.

ZelBack/config/default.js
```diff
@@ -1,6 +1,6 @@
...
 
-const isDevelopment = userconfig.initial.development || false;
+const isDevelopment = userconfig.initial.development || false; // eslint-disable-line
```
Cabecinha84, line 4: what is the  // eslint-disable-line for? if not needed please remove it.

apiServer.js
```diff
@@ -10,13 +11,47 @@ const app = require('./ZelBack/src/lib/server');
...
+        if (md5Current === md5Previous) {
+          return;
+        }
+        md5Previous = md5Current
+        if( initialMd5 === null) {
+          initialMd5 = md5Current
+        }
+        if ( initialMd5 !== md5Current ) {
+          initialMd5 = null
+          console.log(`Config file changed, reloading ${event.filename}...`);
```
Cabecinha84, line 40: no console.log, change to log.info()

apiServer.js
```diff
@@ -10,13 +11,47 @@ const app = require('./ZelBack/src/lib/server');
...
+      }
+    }
+  } catch (error) {
+      console.error(`Error watching files: ${error}`);
+  }
+}
+
+setInterval(async () => {
+  configReload();
+}, 2 * 1000);
```
Cabecinha84, line 53: lets make it slower, 60 * 1000
once per minute is more than enough.

app.js
```diff
@@ -12,15 +13,49 @@ const app = require('./ZelBack/src/lib/server');
...
+        if (md5Current === md5Previous) {
+          return;
+        }
+        md5Previous = md5Current
+        if( initialMd5 === null) {
+          initialMd5 = md5Current
+        }
+        if ( initialMd5 !== md5Current ) {
+          initialMd5 = null
+          console.log(`Config file changed, reloading ${event.filename}...`);
```
Cabecinha84, line 44: log.info instead

app.js
```diff
@@ -12,15 +13,49 @@ const app = require('./ZelBack/src/lib/server');
...
+    }
+  } catch (error) {
+      console.error(`Error watching files: ${error}`);
+  }
+}
+
+setInterval(async () => {
+  configReload();
+}, 2 * 1000);
+
```
Cabecinha84, line 58: once every 60 seconds instead

apiServer.js
```diff
@@ -10,13 +11,47 @@ const app = require('./ZelBack/src/lib/server');
...
+        if ( initialMd5 !== md5Current ) {
+          initialMd5 = null
+          console.log(`Config file changed, reloading ${event.filename}...`);
+        }
+        delete require.cache[require.resolve('./config/userconfig')];
+        userconfig = require('./config/userconfig');
+      }
+    }
+  } catch (error) {
+      console.error(`Error watching files: ${error}`);
```
Cabecinha84, line 47: use log.error() instead.

app.js
```diff
@@ -12,15 +13,49 @@ const app = require('./ZelBack/src/lib/server');
...
+        if ( initialMd5 !== md5Current ) {
+          initialMd5 = null
+          console.log(`Config file changed, reloading ${event.filename}...`);
+        }
+        delete require.cache[require.resolve('./config/userconfig')];
+        userconfig = require('./config/userconfig');
+      }
+    }
+  } catch (error) {
+      console.error(`Error watching files: ${error}`);
```
Cabecinha84, line 51: use log.error() instead.

package.json
```diff
@@ -92,6 +92,7 @@
     "http-shutdown":"~1.2.2",
     "inquirer": "~8.2.5",
     "lru-cache": "~10.0.1",
+    "md5": "~2.3.0",
```
Cabecinha84, line 95: I don't think it's needed to add the md5 lib. Check if hash lib works as we already have it included.
example to search for on the code: "const messageHash = hash(msg);"

Cabecinha84 (changes requested):
Good job. Requesting some changes.

HomeUI/src/views/fluxadmin/ManageFlux.vue
XK4MiLX: check other buttons it is too long.. look bad ( big and without space between buttons )
image

ZelBack/config/default.js
XK4MiLX: without it you won't compile home ui and it will throw an error

ZelBack/config/default.js
XK4MiLX: without it you won't compile home ui and it will throw an error

apiServer.js
XK4MiLX: i dont agree it should work fast ( hot reload ) same way as working nodemon

package.json
XK4MiLX: good idea :) tested working correct

apiServer.js
XK4MiLX: replaced :)

app.js
XK4MiLX: replaced :)

apiServer.js
XK4MiLX: replaced :)

apiServer.js
XK4MiLX: or set it global as 30s  and import function and use it when any cmd with config is use

HomeUI/src/views/fluxadmin/ManageFlux.vue
Cabecinha84: agree makes sense

apiServer.js
Cabecinha84: ok let it be every 2 seconds

Cabecinha84 (approved):
Ack

PR #1141, opened 2023-10-31
v4.16.0

Cabecinha84:
- It is now possible to define specific folders from your container to be synced by syncthing. ex: '/root|s:/root/test' this will create the root container but only the test folder will be synced accross the nodes running the app on the network;
- Marketplace now works with Environment parameters that are stored Flux Storage;
- config reload, adjusting fluxOs configuration file no longer restart fluxOS;
- Added restart FluxOS option to Flux Admin Page;
- Other improvments on Home UI;
- Improvments detecting duplicated messages, no longer processed (reduce cpu and disk usage);
- Better detection of external ip changed to broadcast the information to the network;
- Other smal improvments.


.npmrc
TheTrunk, line 1: Do we want to commit this empty file if its in gitignore?

ZelBack/config/default.js
```diff
@@ -1,6 +1,6 @@
-const userconfig = require('../../config/userconfig');
+let userconfig = require('../../config/userconfig'); // eslint-disable-line
```
TheTrunk, line 1: why is this let?

ZelBack/src/services/fluxService.js
```diff
@@ -1402,6 +1402,33 @@ async function installFluxWatchTower() {
...
+ * Restart FluxOS via nodemon (executes the command `touch ` on package.json).
+ * @param {object} req Request.
+ * @param {object} res Response.
+ */
+async function restartFluxOS(req, res) {
+  try {
+    const authorized = await verificationHelper.verifyPrivilege('admin', req);
+    if (authorized === true) {
+      const nodedpath = path.join(__dirname, '../../../package.json');
+      const exec = `touch ${nodedpath}`;
```
TheTrunk, line 1415: Can you provide more information why it is done this way? How does it work?
If we modify package.json in some way it will cause conflicts on git pulling, updating flux

TheTrunk (changes requested):
There are more tests failing than on master.

Intereting global userconfig lint comment definition

ZelBack/src/services/fluxService.js
XK4MiLX: it not modyficate file ;) it touch it and tiggering nodemon to reset .. u can normal git pull

ZelBack/config/default.js
XK4MiLX: or it need to be const but we dont want it

.npmrc
Cabecinha84: deleted

ZelBack/src/services/fluxService.js
XK4MiLX: The touch command changes certain dates for each file argument. By default, touch sets both the date of last file modification and the date of last file access to the current time.

Cabecinha84: changed to 'pm2 restart flux' command.

ZelBack/config/default.js
TheTrunk: I seem to be able to operate, build ui without it with following:

const userconfig = require('../../config/userconfig');

const isDevelopment = userconfig.initial.development || false;

TheTrunk: I pushed the correction to this. Can't figure out why let or eslint disable is needed

PR #1144, opened 2023-11-02
FluxOs Sentinel

Cabecinha84:
https://docs.google.com/document/d/1LHCyvZTkuGTe5CSUSEAdY77dKJWlxZc-Y16KCXTyGpM/edit


blondfrogs:
We should make sure that the new code has test coverage as much as possible

Cabecinha84:
Yes, I agree, I will write a xls with the tests that we should do before even merging to development.
For now is more like for code review...

TheTrunk:
Great job! We also need to write proper test suite for these changes as well

PR #1149, opened 2023-11-09
Update repositories.json - add ubuntu

Luca8991:
I'd like to have the possibility to run a more general purpose image on it, so that I can have like a "VPS".

Docker image: https://hub.docker.com/_/ubuntu


TheTrunk:
As long as content running on it is within ToS :)

PR #1161, opened 2023-11-24
WIP: Tidy entrypoints, make UPNP easier

MorningLightMountain713:
I need to validate logic and run tests but thought I'd WIP so I can get feedback if you'd like me to continue or not.

5a85cfd - The current `app.js` and `apiServer.js` are duplicates, this leads to replication errors (see +apiPort). Have removed the `apiServer.js` and imported by `app.js`. Have added the argument `--apiServer` to `app.js` and updated `package.json`. 

2f17700 - the typing in the upnp service was incorrect. While this doesn't affect app running, function calls were showing up in IDE as "await has no effect on the type of this expression"

3081c7a - This cleans up some technical debt around checking for upnp (this is what still needs testing) now validate upnp on every config reload and only use the upnp module `isUPNP` variable. Have modified the semantics of determining if upnp is activated, this is a more direct way of determining what was already used.


MorningLightMountain713:
Closing - will reopen when I source off correct branch

PR #1171, opened 2023-12-04
Marketplace Auto Select Enterprise Nodes Feature - Development

flikites:
This allows the marketplace to auto-select enterprise nodes if the ```enterpriseApplication: true```` flag is set in the apps marketplace specification in Flux Stats

### DIFF
--- a/HomeUI/src/views/apps/marketplace/AppView.vue
+++ b/HomeUI/src/views/apps/marketplace/AppView.vue
@@ -577,6 +577,16 @@ export default {
       return 'primary';
     };
 
+    const showToast = (variant, title, icon = 'InfoIcon') => {
+      toast({
+        component: ToastificationContent,
+        props: {
+          title,
+          icon,
+          variant,
+        },
+      });
+    };
     const tier = ref('');
     tier.value = props.tier;
     const userZelid = ref('');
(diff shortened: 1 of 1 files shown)


HomeUI/src/views/apps/marketplace/AppView.vue
```diff
@@ -735,8 +831,12 @@ export default {
           }
         });
       });
-
       currentComponent.value = props.appData.compose[0];
+      autoSelectNodes().then((v) => {
```
TheTrunk, line 835: this way it would be trigerred on all marketplace apps?
I think there should be first the check enterpriseApp = true before triggering this

HomeUI/src/views/apps/marketplace/AppView.vue
```diff
@@ -796,7 +885,11 @@ export default {
         }
         if (props.appData.version >= 7) {
           appSpecification.staticip = props.appData.staticip;
-          appSpecification.nodes = [];
+          if (props.appData.compose?.some((c) => (c?.enterpriseApp === true))) {
```
TheTrunk, line 888: I think this should not be in compose. 
enterpriseApp is more global affecting entire application, so it should be just props.appData.enterpriseApp === true.
Then we need to be sure that enterpriseApp flag does not get it to specifications but only acts for ent. nodes selection

TheTrunk (changes requested):

HomeUI/src/views/apps/marketplace/AppView.vue
flikites: Added the changes based on what I wrote above. Seems like the more plausible route. Please review when you get a chance @TheTrunk

PR #1173, opened 2023-12-06
Tidy entrypoints, make UPNP easier

MorningLightMountain713:
Have resubmitted based off development instead of master.

I've tested this on a live upnp node - works fine. (Haven't tested in breaking situations though)

https://github.com/RunOnFlux/flux/commit/6736bb6e9a9ffa447c042404c8b383dcf3fc8e19 - The current app.js and apiServer.js are duplicates, this leads to replication errors (see +apiPort). Have removed the direct call to apiServer.js and imported by app.js. Have added the argument --apiServer to app.js and updated package.json.

https://github.com/RunOnFlux/flux/commit/ddc43748d641e9b931de4db3b35a52fe08a25a93 - the typing in the upnp service was incorrect. While this doesn't affect app running, function calls were showing up in IDE as "await has no effect on the type of this expression"

https://github.com/RunOnFlux/flux/commit/043d9eefb8fac5a6019c7c3b3fa776db463710ad - This cleans up some technical debt around checking for upnp (this is what still needs testing) now validate upnp on every config reload and only use the upnp module isUPNP variable. Have modified the semantics of determining if upnp is activated, this is a more direct way of determining what was already used.


apiServer.js
```diff
@@ -19,9 +19,29 @@ const { watch } = require('fs/promises');
...
 const cmdAsync = util.promisify(nodecmd.get);
 const apiPort = userconfig.initial.apiport || config.server.apiport;
-const apiPortHttps = apiPort + 1;
+const apiPortHttps = +apiPort + 1;
 let initialHash = hash(fs.readFileSync(path.join(__dirname, '/config/userconfig.js')));
 
+async function loadUpnpIfRequired() {
+  // User configured UPnP node
+  if (userconfig.initial.apiport || userconfig.initial.routerIP) {
+    if (!userconfig.initial.apiport || !userconfig.initial.routerIP) {
```
TheTrunk, line 28: I think the routerIP actually does not have to be set.

Cabecinha84: Thank you very much for your work/improvements, I am unable to check/test the changes until next week or so but they will not be forgotten.
One of the things that we need to be careful is that having api port doesn't mean it's a UPnP node, as it can be set as 16127 and not be a UPnP node, and from your comments I believe you might have used that as assumption. Correct?
Cheers

MorningLightMountain713: > You're right - the routerIP is only needed for the firewall adjustment if UPnP is configured, have adjusted.
> 
> In a future PR, I would like to rework some of the natupnp module. This is the biggest area of problems users have IMO.
> 
> 1/ Add better logging around errors. 2/ Add checks to ensure multicast is working (Join 239 group and issue "self ping") for warning messages, there will be no response if snooping is enabled and filtering. 3/ Listen for igmp queries 4/ Cache SSDP response in case of IMGP snooping. This will fix most problems, as users who have misconfigured networks (igmp snooping with no querier) the bridge / switch will allow SSDP to work for the first 5 minutes (depending on switch) before the igmp join times out. After that, for the actual upnp part - multicast isn't required. 5/ Get router IP from SSDP response for firewall - then user doesn't have to specify this in the config

I checked this again - routerIP needs to be set or the firewall won't be opened to allow UPnP

MorningLightMountain713: Yes - had assumed that apiPort set in userConfig.js meant UPnP, my bad.

So from what I can see, currently it works like this:

If apiPort is set, and it's not 16127, then node is UPnP
If apiPort is set and it's 16127 AND routerIP is set it's UPnP.
if apiPort is set and it's 16127 and no routerIP - then it's a normal node.

If you can confirm the above is correct - I'll fix the code to follow the above.

Might be a good idea to add a specific isupnpnode: true/false to the config for the future? but I could see that taking years to migrate to.

I've got some ideas around auto config of upnp nodes using multicast - nodes can join a group on the lan to discover what nodes already exist, then they can allocate there own api ports amongst themselves.

Cheers!

Cabecinha84: If router IP it's UPnP, if apiport different from 16127 it's UPnP, if apiport equal to 16127 we check if it's UPnP but if it's not fluxos doesn't shutdown.

Cabecinha84 (approved):
Ack

PR #1175, opened 2023-12-07
Update marketplaceCategories.js

flikites:
Added minecraft category to marketplace category list


Cabecinha84:
Tom, don't you think just using current game category is enough? I don't know if create a specific Minecraft category is needed. Cheers

flikites:
Their are multiple server types that would be offered eventually (Vanilla, Cat, Spigot, Paper,etc.) 

As a marketer I think a separate one is needed, but its ultimately up to you and Nolan to determine this.

sygem:
Subcategories could potentially be an option in the new marketplace UI
image

PR #1178, opened 2023-12-11
v4.22.0

TheTrunk:
- Fixes Local App visit button
- XFS backing filesystem with quota mount (usrquota,prjquota,grpquota) is now recommended. XFS brings advantages to Flux Node operators for better space management of their node.
- Swap per application component now defaults to 2GB
- FileSystem storage for an application component is capped to 10GB (cumulation is possible - if first copmonent use 1GB, another component can use up to 19GB) 
- Applications violating storage allocations are now being softly redeployed. Multiple violations results in removal process
- Cache optimizations on message processing, broadcasting
- Improvements in messages broadcasts, processing of duplicate messages and processing of syncthing adjustment messages


alihm (approved):

ZelBack/src/services/dockerService.js
```diff
@@ -583,6 +585,16 @@ async function appDockerCreate(appSpecifications, appName, isComponent, fullAppS
     },
   };
 
+  // get docker info about Backing Filesystem
+  // eslint-disable-next-line no-use-before-define
+  const dockerInfoResp = await dockerInfo();
+  log.info(dockerInfoResp);
+  const driverStatus = dockerInfoResp.DriverStatus;
+  const backingFs = driverStatus.find((status) => status[0] === 'Backing Filesystem'); // d_type must be true for overlay, docker would not work if not
+  if (backingFs && backingFs[1] === 'xfs') {
```
XK4MiLX, line 594: missing mount quota check

TheTrunk: Can you explain more? It must have the 'pquota' mount option right?
Does the dockerInfo output that?

XK4MiLX: dockerInfo dont have such option i created custom function base on df and mount .
Example belown:
```
const cmd = require('node-cmd');

async function getDfDevice(directory) {
  return new Promise((resolve, reject) => {
    cmd.get(`df --output=source,target ${directory}`, (error, data, stderr) => {
      if (error) {
        resolve(false);
        return;
      }
      const lines = data.trim().split('\n');
      for (let i = 1; i < lines.length; i++) {
        const columns = lines[i].split(/\s+/);
        if (columns[0] !== "") {
          resolve(columns[0]);
          return;
        }
      }
      resolve(false);
    });
  });
}

async function hasQuotaOptionForDevice(device) {
  return new Promise((resolve, reject) => {
    cmd.get(`mount | grep "${device}"`, (error, data, stderr) => {
      if (error) {
        resolve(false);
        return;
      }
      const mountInfo = data.trim();
      if (!mountInfo) {
        resolve(false);
        return;
      }
      if (mountInfo.includes('quota') && mountInfo.includes('xfs')) {
        resolve(true);
      } else {
        resolve(false);
      }
    });
  });
}

async function main() {
  const directoryToCheck = '/var/lib/docker';
  try {
    const device = await getDfDevice(directoryToCheck);
    if (device) {
      if (await hasQuotaOptionForDevice(device)) {
        console.log('OK');
      }
    }
  } catch (error) {
    console.error('Error:', error);
  }
}

main();
```

XK4MiLX: i updated commented for pr with mount info

TheTrunk: Cool, I will include that there then. 
What are like the defaults for when we install OS on XFS? Is quota a standard?

TheTrunk: Cool your code fix commited :)

ZelBack/src/services/dockerService.js
```diff
@@ -592,7 +593,13 @@ async function appDockerCreate(appSpecifications, appName, isComponent, fullAppS
   const driverStatus = dockerInfoResp.DriverStatus;
   const backingFs = driverStatus.find((status) => status[0] === 'Backing Filesystem'); // d_type must be true for overlay, docker would not work if not
   if (backingFs && backingFs[1] === 'xfs') {
-    options.HostConfig.StorageOpt = { size: `${config.fluxapps.hddFileSystemMinimum}G` }; // must also have 'pquota' mount option
+    // check that we have quota
+    const hasQuotaPossibility = await deviceHelper.hasQuotaOptionForDevice('/var/lib/docker').catch((error) => {
```
XK4MiLX, line 597: hasQuotaOptionForDevicethat function argument is device not path -> pls check code again u need use both functions

TheTrunk: Nice catch and Thank you for the fix

PR #1179, opened 2023-12-18
v4.23.0

TheTrunk:
- Simplifies UPNP check
- Simplifies apiServer of FluxOS
- HomeUI now shows expired applications
- HomeUI now shows all marketplace deployed applications
- HomeUI - It is now possible to redeploy expired application as well as any active application
- HomeUI - Copying and import of applications specifications json (ASJ)
- Docker prune improvments


Cabecinha84 (approved):
Ack

Cabecinha84:
@TheTrunk haven't commit rebuild HomeUI because it shows 132 files changed to commit on my laptop...

Also something about the change, I don't like 100% the My Expired Apps tab because it shows up all rows...
If I had one app, that I had running for one year, paying the renew every 1 month, it will show it 12 times...

I think we should only show one row with the latest version that got expired.

But some nice improvments.

Cheers!

TheTrunk:
> @TheTrunk haven't commit rebuild HomeUI because it shows 132 files changed to commit on my laptop...
> 
> Also something about the change, I don't like 100% the My Expired Apps tab because it shows up all rows... If I had one app, that I had running for one year, paying the renew every 1 month, it will show it 12 times...
> 
> I think we should only show one row with the latest version that got expired.
> 
> But some nice improvments.
> 
> Cheers!

Hi that is exactly how it should be working. It should always show the latest specifications - the app should be shown just one time with latest specifications. Is that not the case?

Cabecinha84:
> > @TheTrunk haven't commit rebuild HomeUI because it shows 132 files changed to commit on my laptop...
> > Also something about the change, I don't like 100% the My Expired Apps tab because it shows up all rows... If I had one app, that I had running for one year, paying the renew every 1 month, it will show it 12 times...
> > I think we should only show one row with the latest version that got expired.
> > But some nice improvments.
> > Cheers!
> 
> Hi that is exactly how it should be working. It should always show the latest specifications - the app should be shown just one time with latest specifications. Is that not the case?

you are correct, I just saw so many presearch on my end that tought it was the case but the app name is different. All good! Nice work!
image

PR #1182, opened 2024-01-01
v4.25.0

TheTrunk:
- Improvements in syncthing;
- Packages updates;
- Corrects netcat installation;
- Improvements on marketplace;
- Only run checkMyAppsAvailability checks when daemon is synced;
- Improvements in docker prune;
- Improvements on application install checks so an App is not installed on new instances when it is already running on the defined number of instances.
- New Master/Slave Solution for decentralized persistent data container. With this new solution, the app will be only running on one instance (the master) and the others will be stopped (slaves), syncing the container data, and checking if master is down, when it is down, the slave start so one will be designated the new master. To use the new feature prepend g: to your container data on application specifications. There were also changes done on Flux Domain Manager (FDM) for this solution. You can read more about this new feature here: https://docs.google.com/document/d/10qGstnJnCqthTWvTsTnB--LKwC1e_HJTQv6l70l0T3Q/edit?usp=sharing

This version will be enforced on January 23th.


Cabecinha84 (approved):
Ack

PR #1196, opened 2024-01-21
v4.27.0

TheTrunk:
- Make apps to not be spawned on new nodes when they are already running on the defined number of instances;
- Uninstall apps from nodes when they are spawned on more nodes then required based on since when are the app running on the nodes, uninstall the latest one that was installed;
- Uninstall apps from nodes that are not confirmed on the network;
- Network considers now normal instances running = to app specifications. Before it was considering also ok instances specified +1;
- Add ssp, walletconnect and metamask as signing options on marketplace;
- Improved subnet/gateway ip assignment on docker network creation;
- Other small improvments.


ZelBack/src/services/appsService.js
```diff
@@ -10362,10 +10417,13 @@ async function syncthingApps() {
                 if (cache.numberOfExecutions === cache.numberOfExecutionsRequired) {
                   syncthingFolder.type = 'sendreceive';
                 } else if (cache.numberOfExecutions === cache.numberOfExecutionsRequired + 1) {
-                  log.info(`SyncthingApps starting appIdentifier ${appId}`);
+                  log.info(`SyncthingApps changing syncthing type to sendreceive for appIdentifier ${appId}`);
                   syncthingFolder.type = 'sendreceive';
-                  // eslint-disable-next-line no-await-in-loop
-                  await appDockerRestart(id);
+                  if (containerDataFlags.includes('r')) {
```
TheTrunk, line 10422: only for r?

ZelBack/src/services/appsService.js
```diff
@@ -10538,8 +10608,11 @@ async function syncthingApps() {
                   } else if (cache.numberOfExecutions === cache.numberOfExecutionsRequired + 1) {
                     log.info(`SyncthingApps starting appIdentifier ${appId}`);
                     syncthingFolder.type = 'sendreceive';
-                    // eslint-disable-next-line no-await-in-loop
-                    await appDockerRestart(id);
+                    if (containerDataFlags.includes('r')) {
```
TheTrunk, line 10611: only for r?

ZelBack/src/services/fluxNetworkHelper.js
```diff
@@ -1328,6 +1331,22 @@ async function purgeUFW() {
   }
 }
 
+/**
+ * This fix a docker security issue where docker containers can access host network, for example to create port forwarding on hosts
+ */
+async function removeDockerContainerAccessToHost() {
```
TheTrunk, line 1337: Can you explain it again why this is needed?

ZelBack/src/services/appsService.js
```diff
@@ -8986,8 +8986,32 @@ async function trySpawningGlobalApplication() {
 /**
  * To check and notify peers of running apps. Checks if apps are installed, stopped or running.
  */
+let nodeConfirmedOnLastCheck = true;
 async function checkAndNotifyPeersOfRunningApps() {
   try {
+    const isNodeConfirmed = await generalService.isNodeStatusConfirmed();
```
TheTrunk, line 8992: I disagree with removal of applications from nodes when nodes are no longer on the network.
Those applications should continue to be running. If we have some flux network issues, the apps will still be running, not affected by flux network issues and fdm remembers these old ips.

ZelBack/src/services/appsService.js
```diff
@@ -3738,11 +3739,10 @@ async function softRegisterAppLocally(appSpecs, componentSpecs, res) {
     }
 
     if (!isComponent) {
-      // last character of appName determines gateway
-      const lastCharCode = appName.charCodeAt(appName.length - 1);
-
-      const dockerNetworkAddrValue = lastCharCode;
-
+      let dockerNetworkAddrValue = Math.floor(Math.random() * (254 - 0 + 1) + 0);
```
TheTrunk, line 3742: if we are using random, we can still run into the issue that docker network is already in use. This should ideally go one by one in increments. Random is not ideal.

ZelBack/src/services/appsService.js
```diff
@@ -99,6 +99,8 @@ let masterSlaveAppsRunning = false;
 
 const hashesNumberOfSearchs = new Map();
 
+const appsThatMightBeUsingOldGatewayIpAssignment = ['HNSDoH', 'dane', 'fdm', 'Jetpack2', 'fdmdedicated', 'isokosse', 'ChainBraryDApp', 'health', 'ethercalc'];
```
TheTrunk, line 102: TODO recheck these apps so we dont have to hardcode it at all.

HomeUI/src/views/apps/Management.vue
```diff
@@ -4991,6 +4991,9 @@ export default {
           // time to encrypt
           // eslint-disable-next-line no-restricted-syntax
           for (const component of this.appUpdateSpecification.compose) {
+            component.environmentParameters = component.environmentParameters.replace('\\“', '\\"');
```
TheTrunk, line 4994: Can you explain what it is for?

tests/unit/globalconfig/default.js
```diff
@@ -138,7 +138,6 @@ module.exports = {
     maxImageSize: 2000000000, // 2000mb
     minimumInstances: 3,
     maximumInstances: 100,
-    maximumAdditionalInstances: 1, // max instances above subscribed amount. In case of min instances, this is minimumInstances + maximumAdditionalInstances
```
TheTrunk, line 141: This was not removed in default.js in zelback

ZelBack/src/services/appsService.js
```diff
@@ -9308,21 +9329,43 @@ async function checkAndRemoveApplicationInstance() {
       // eslint-disable-next-line no-await-in-loop
       const runningAppList = await getRunningAppList(installedApp.name);
       const minInstances = installedApp.instances || config.fluxapps.minimumInstances; // introduced in v3 of apps specs
-      if (runningAppList.length > (minInstances + config.fluxapps.maximumAdditionalInstances)) {
```
TheTrunk, line 9311: I think its fine to remove BUT
- By not having this flag, any small deviances in the network can cause more installations and uninstallations.
If my node does not receive some app running message just from one node, it will install this application. And then it will uninstall it again once it received new message. I think having this +1 we allow for one missing app runnning message which is effectively resulting in less random installs/uninstalls. 
Just a note to observe the network after this removal.

ZelBack/src/services/appsService.js
```diff
@@ -3364,11 +3366,10 @@ async function registerAppLocally(appSpecs, componentSpecs, res) {
     }
 
     if (!isComponent) {
-      // last character of appName determines gateway
-      const lastCharCode = appName.charCodeAt(appName.length - 1);
-
-      const dockerNetworkAddrValue = lastCharCode;
-
+      let dockerNetworkAddrValue = Math.floor(Math.random() * (254 - 0 + 1) + 0);
```
TheTrunk, line 3369: why -0 +1 + 0?

TheTrunk: 0-255?
Math. floor(Math. random() * 256)

TheTrunk (commented):
On UI we have
- Buffer is not defined

-> I will recheck more the signing with EC, MM, SSP

ZelBack/src/services/appsService.js
Cabecinha84: yeah the restart for g: is done on masterSlaveApps
but thank you for pointing it out as that leads to a bug bug, all those includes should had like r: and not only r, same for s: and g:

Cabecinha84: false alarm, all is good

ZelBack/src/services/fluxNetworkHelper.js
Cabecinha84: this fix a vulnerability where containers can access ufw/port forwading/upnp by having access to the host subnet.
vulnarability was raised by a community member.
example, go to a upnp node, manage any app, and on execute commands execute: curl http://192.168.1.1:2189/rootDesk.xml

you might need to install curl first. 

we might comment this to do more testing before we release it.

ZelBack/src/services/appsService.js
Cabecinha84: Random should be fine. If for any reason fails on a node and the app is not installed on the amount of nodes required, in 2h it will be installed again on the same node and for sure it will pick another random one.

HomeUI/src/views/apps/Management.vue
Cabecinha84: We have lots of support tickets, where apps are specified on the network but not running in any node.
The reason it's because they have copied the specifications from medium or guides, where the char for a string is not valid, the UI allows it, but when the app is selected to install on a node fails saying the JSON is not valid. And because if we gave an error most probably people would not understand the reason, I think this should be just fine.

ZelBack/src/services/appsService.js
Cabecinha84: Math.floor(Math.random() * 255);

Cabecinha84: fixed

TheTrunk: so 254 is maximum then? not 255? 255 is still valid no?

Cabecinha84: fixed

ZelBack/src/services/fluxNetworkHelper.js
Cabecinha84: it's commented for more testing

ZelBack/src/services/appsService.js
Cabecinha84: Kamil checked and almost are using componentName for networking, but I left that there just in case...

PR #1217, opened 2024-02-13
Remove duplicate repos and add specs for repositories.json

johnhanlon86:
### What does this PR do?
Removes duplicate repos and adds specs for repositories.json. The test will fail if duplicates exist.

### Ran test with existing repositories.json file.
Failing Test

### Ran test with updated repositories.json file as per changed in this PR.
Passing Test


TheTrunk (approved):
Can target master directly.

PR #1220, opened 2024-02-13
App Storage Monitoring

JKTUNING:
- Check monitored container for multiple mounts/volumes
- Loop through mounts to determine size of used space on host
- Add to container size for final storage used

image

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -1102,12 +1102,12 @@ function startAppMonitoring(appName) {
             const source = mount?.Source;
             if (source) {
               const mountInfo = nodecmd.runSync(`sudo du -sb ${source}`);
-              const mountSize = +mountInfo?.data.split(source)[0] ?? 0;
-              if (typeof mountSize === 'number' && !isNaN(mountSize)) {
+              const mountSize = mountInfo?.data.split(source)[0] ?? 0;
+              if (typeof mountSize === 'number' && !Number.isNaN(mountSize)) {
                 containerTotalSize += mountSize;
               }
             }
-          })
+          });
         }
 
         // const appFolderName = dockerService.getAppDockerNameIdentifier(appName).substring(1);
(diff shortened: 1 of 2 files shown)


XK4MiLX:
and now compare size with df  :)

ZelBack/src/services/appsService.js
```diff
@@ -1062,6 +1062,37 @@ async function getAppFolderSize(appName) {
...
+    const containerInfo = await dockerService.dockerContainerInspect(appName, { size: true });
+    let containerTotalSize = +containerInfo.SizeRootFs ?? 0;
+    // traverse mounts/volumes to find size used on host
+    if (containerInfo?.Mounts?.length) {
+      containerInfo.Mounts.forEach((mount) => {
+        let source = mount?.Source;
+        if (source) {
+          // remove /appdata to get true size of the app folder
+          source = source.replace('/appdata', '');
+          const mountInfo = nodecmd.runSync(`sudo du -sb ${source}`);
```
XK4MiLX, line 1081: u should use async function for send shell cmd and other type of loop forEach not support async

XK4MiLX: if u want get only mount volume size better use df ( it is already used df js lib in fluxos ), sudo du take long time and resources

JKTUNING: I can convert it to use cmdAsync and use a different iterator with async support.

I think du is the more accurate way to get the size of the directory, which is what is currently being used in the app monitoring solution but we can certainly look into doing this another way.

XK4MiLX: but u dont need check any folder size as bind mount is volume u need check only volume size

XK4MiLX: there is already size, used, and available info from df

Cabecinha84:
Is this ready to be merged? @JKTUNING @XK4MiLX ?

PR #1222, opened 2024-02-14
Appbackuprestore

XK4MiLX:
- removed body parser,  added express.json() and express.urlencoded({ extended: true }) for correct post parse
- adjustment all post call
- added custom user for terminal 
- added backup/restore functionality in app managment
- other UI tweaks


Cabecinha84:
Great work @XK4MiLX . lets merge it to development and test it in development 👍

PR #1223, opened 2024-02-16
Whitelist pmietlicki/dero-miner repository

AvAcalho:
# Application whitelist

dero-miner

# Checklist:

- [x] Whitelist of application is only modifying repositories.json file
- [x] repositories.json is still a valid JSON file
- [x] Only whitelists single docker image organisation (one whitelist at a time, more whitelists, more PRs)
- [x] No other whitelist has been deleted
- [x] I agree with ToS https://cdn.runonflux.io/Flux_Terms_of_Service.pdf
- [x] Application follows ToS - Application is not malicious. Application is not a scam. Application does what is meant to do and does not mislead in any way. Application does not do anything illegal. Application is not a mining application (not even bandwidth mining).
- [x] In case application receives multiple reports, behaves maliciously, it will be blacklisted and removed from the network.

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -418,5 +418,6 @@
    "tmart1010/streamlit_music-server",
    "golang",
    "prom",
-   "ghcr.io/lloesche/valheim-server"
+   "ghcr.io/lloesche/valheim-server",
+   "pmietlicki/dero-miner"
 ]


Cabecinha84:
Hello, thank you for your interest in using Flux network. Unfortunately mining applications are against our TOS and not allowed to run on our network.

AvAcalho:
> Hello, thank you for your interest in using Flux network. Unfortunately mining applications are against our TOS and not allowed to run on our network.

Hi @Cabecinha84 that's sad to hear! Why are they prohibited ? Can you elaborate more on that? 
I don't see anything mentioning that mining applications are not allowed on the network, is this something that might change in the future ?
Really very sad to hear from a web 3.0 provider and at the same time contradictory as you offer some mining nodes from your marketplace.

PR #1224, opened 2024-02-17
v4.31.0

Cabecinha84:
- Backup/Restore Container Data -> It is now possible to backup your application container data and restore it. The new feature is now available under local apps management;
- Improvements on Syncthing;
- FluxOs UI colors/icons updates;
- Other small improvements.


ZelBack/src/services/fluxshareService.js
```diff
@@ -546,8 +547,8 @@ async function fluxShareRemoveFolder(req, res) {
 
       const dirpath = path.join(__dirname, '../../../');
       const filepath = `${dirpath}ZelApps/ZelShare/${folder}`;
-      await fs.promises.rmdir(filepath);
-
+      // await fs.promises.rmdir(filepath);
```
TheTrunk, line 550: why are we not using fs for it? instead of it it executes sudo rm -rf.

ZelBack/src/services/IOUtils.js
```diff
@@ -0,0 +1,497 @@
...
+}
+
+/**
+ * Removes the specified directory and its contents or only the contents.
+ *
+ * @param {string} rpath - The path of the directory to be removed.
+ * @param {boolean} directory - Flag indicating whether to remove only the directory contents (true) or the entire directory (false).
+ * @returns {boolean} - True if the directory or its contents are removed successfully, false on failure.
+ */
+async function removeDirectory(rpath, directory = false) {
```
TheTrunk, line 348: why not using fs module?

package.json
```diff
@@ -79,6 +79,7 @@
     "bitcoin": "~3.0.3",
     "bitcoinjs-message": "~2.2.0",
     "body-parser": "~1.20.2",
+    "child_process": "~1.0.2",
```
TheTrunk, line 82: Remove. This is part of nodejs, not an npm module.

ZelBack/src/lib/socket.js
```diff
@@ -40,6 +40,7 @@ function initIO(httpServer) {
         Tty: true,
         Cmd: serviceHelper.commandStringToArray(dockerCmd),
         Env: serviceHelper.commandStringToArray(dockerEnv),
+        User: dockerUser,
```
TheTrunk, line 43: what if this is undefined? as of different fluxos versions

ZelBack/src/lib/server.js
```diff
@@ -1,14 +1,15 @@
...
+// const bodyParser = require('body-parser');
 const cors = require('cors');
 const morgan = require('morgan');
 
 const expressWs = eWS(express());
 const { app } = expressWs;
 
 app.use(morgan('combined'));
-app.use(bodyParser.json());
+app.use(express.json());
```
TheTrunk, line 11: ok our package ensure that express is on version 4.18 so this will work. There should not be any fluxos version with lower express than 4.16.

ZelBack/src/lib/server.js
```diff
@@ -1,14 +1,15 @@
 const express = require('express');
 const eWS = require('express-ws');
-const bodyParser = require('body-parser');
+// const bodyParser = require('body-parser');
```
TheTrunk, line 3: just remove it and remove from package.json

TheTrunk (changes requested):
4 tests are now failing
Missing entire test suite for IOutils
Missing tests for new functions, adjusted functions.
Version bump is missing

ZelBack/src/services/fluxshareService.js
XK4MiLX: fs.promises.rmdir will not remove not empty dir i want give ability to remove directory with files

ZelBack/src/services/IOUtils.js
XK4MiLX: fs.promises.rmdir will not remove not empty dir i want give ability to remove directory with files and others directory

ZelBack/src/services/fluxshareService.js
TheTrunk: You can actually achieve that using FS module as its described in the docs
https://nodejs.org/api/fs.html#fspromisesrmdirpath-options

ZelBack/src/lib/socket.js
XK4MiLX: if it is not set is used def username defined in dockerfile or root

ZelBack/src/services/fluxshareService.js
XK4MiLX: but it will not solve situation when i want remove all file and folders but i want keep directory ( or i miss something)

PR #1240, opened 2024-03-01
Volumebrowser

XK4MiLX:
**Volume Browser Features:**
upload multiple file
download multiple file
download zipped directory
rename file/folder
craeate directory
remove file/folder
volume storage info
Other UI tweaks


HomeUI/src/views/apps/Management.vue
```diff
@@ -7146,11 +7754,11 @@ export default {
       }
     },
 
-    cancelDownload() {
-      this.abortToken.cancel('User download cancelled');
-      this.downloaded = '';
-      this.total = '';
-    },
+    // cancelDownload() {
```
Cabecinha84, line 7757: If not needed should be removed from the code

Cabecinha84 (changes requested):
looking good. Just a small thing that can be deleted.

ZelBack/src/routes.js
```diff
@@ -636,7 +636,7 @@ module.exports = (app, expressWs) => {
     appsService.appendRestoreTask(req, res);
   });
 
-  app.post('/ioutils/fileupload/:fullpath?/:filename?/:appname?', (req, res) => {
```
TheTrunk, line 639: is this compatible with older versions?

TheTrunk (changes requested):

ZelBack/src/routes.js
XK4MiLX: no api was changed

TheTrunk: But that means its not compatible with older versions of home ui.

XK4MiLX: older version using full path as arg now all is done on backend so it is not compatible

PR #1259, opened 2024-03-23
Add checkout loading spinner

JKTUNING:
Add checkout loading spinner to give the user an indication that the request is processing when using fiat payments.


Cabecinha84 (approved):
Ack

PR #1267, opened 2024-04-05
Move syncthing to new execution model

MorningLightMountain713:
**What this pull does**

* Adds ability to stop syncthingService (doesn't get used, it's just the ability).
* Abortable axios requests
* Implements axiosCache for syncthingService, caches apikey and axios instance for 15 minutes.
* Updates fluxController to be able to loop a passed in "runner" function.
* Updates the syncthingService to use the new serviceHelper command runner.
* Updates the syncthingService to use the fluxController loop runner. 
* Adds 30 syncthingService tests (non comprehensive).
* Sets executable bit on helper scripts so they can be run directly.
* Adds userconfig global variable to eslintrc so can remove all the userconfig exceptions from each file.

**standalone**

Have added the ability to run the syncthingService as a standalone. Just added this stanza at the end of the module:

```javascript
// similar to python's if __name__ == "__main__", allows module
// to be run as standalone, useful for testing
if (require.main === module) {
  startSyncthingSentinel();

  process.stdin.on('data', async (data) => {
    const cmd = data.toString().trim();
    if (cmd === 'start') startSyncthingSentinel();
    if (cmd === 'stop') await stopSyncthingSentinel();
  });
}
```

This is helpful for testing. (It won't run under normal circumstances or during tests) You can start it like this (from the repo root):

```
NODE_CONFIG_DIR=./ZelBack/config FLUX_APIPORT=16187 node ZelBack/src/services/syncthingService.js
```

You can then type "stop" and hit enter and it will stop the service. You can then type "start" and it will start it back up again.

**Notes**

To simplify some of the functions, and maintain seperation of concerns, have started splitting the actual business logic from the express middlewares. For example:

```bash
   app.get('/syncthing/deviceid', cache('30 seconds'), (req, res) => {
-    syncthingService.getDeviceID(req, res);
+    syncthingService.getDeviceIdApi(req, res);
   });
```

Then the `getDeviceIdApi` function calls `getDevice` - this means getDevice doesn't have to be aware / create all the api responses. It can focus on returning the deviceId. This simplifies tests etc.

You may have noticed I've started refactoring some of the functions to better adhere to camel case. From a previous role, we learnt the hard way not to capitalize acronyms. This leads to confusion around where one word starts and another word ends. It's more explicit and predictable to only capitalize the first letter of an acronym. Eg instead of `deviceID` we have `deviceId` (I appreciate syncthing is using deviceID in their config and have left this)

A contrived example:

startServerHTMLTLS - this isn't obvious at first glance what it's supposed to be. Whereas, startServerHtmlTls - there is no ambiguity around what this function does.

I've been running this branch on a node for a week and all is well, however, due to the syncthing changes, would be good to get this on some other development nodes and try to break it just to make sure I've covered everything.

**Node specifier**

Something else to note, have started using the full `node:` specifier for nodeJS inbuilt modules. E.g. instead of ```const fs = require('fs);``` would use ```const fs = require('node:fs');``` This does a couple of things, makes it obvious that we are indeed using a nodeJS built in module, it also skips the require cache - so if there was a file named fs.js, it would still use the node built in module.


MorningLightMountain713:
Looking at the node specifier, I already added it in the prior version when doing the serviceHelper. (This probably shouldn't have been added)

So if there is anyone on an older version... it will be broken already

Here is what happens on node 17:

```
Welcome to Node.js v14.17.0.
Type ".help" for more information.
> const serviceHelper = require('./ZelBack/src/services/serviceHelper')
Uncaught Error: Cannot find module 'node:util'
Require stack:
- /Users/davew/code/flux/flux/ZelBack/src/services/serviceHelper.js
- <repl>
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:885:15)
    at Function.Module._load (internal/modules/cjs/loader.js:730:27)
    at Module.require (internal/modules/cjs/loader.js:957:19)
    at require (internal/modules/cjs/helpers.js:88:18) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/davew/code/flux/flux/ZelBack/src/services/serviceHelper.js',
    '<repl>'
  ]
}
> serviceHelper
Uncaught ReferenceError: serviceHelper is not defined
```

Here is node 18:

```
Welcome to Node.js v14.18.0.
Type ".help" for more information.
> const serviceHelper = require('./ZelBack/src/services/serviceHelper')
undefined
>
```

MorningLightMountain713:
Have removed all instances of `node:` from requires for tests and in serviceHelper / syncthingService

MorningLightMountain713:
All tests passing 

Screenshot 2024-04-05 at 1 53 48 PM

MorningLightMountain713:
I'm just testing this on a live node now. Will let it run for a week or so and see how it goes

MorningLightMountain713:
This has been running for the last 48hours on one of my nodes. Looks good.

Cabecinha84:
Will merge to dev after tuesday release! Nice work!

Cabecinha84 (approved):
Ack, nice job!

PR #1271, opened 2024-04-08
fluxService - update shell commands and refactor

MorningLightMountain713:
**What this pull does**

* Update fluxService to use serviceHelper command runner.
* Remove listeners for req body - the request body is fully finalized before it gets to the post handler, so don't need to use event listeners anymore. This simplifies things and makes the tests a lot easier too.
* Unify the fs calls, these are all async now.
* Fix typing.
* Add a couple of git methods for branches and stuff (unused at the moment).
* Alpha sort exports for easy lookup.
* Update all the tests (passing).

**To Do**

I'd like to revisit this again in the future and tidy up some of the path stuff, i.e. can just use `process.cwd()` for a lot of the directory stuff. Look at refactoring some of the methods, as a lot of them are quite similar and could probably remove a lot of lines of code here. Shame commonJS doesn't have decorators, as this would be a perfect place to use them.


MorningLightMountain713:
Screenshot 2024-04-08 at 9 47 48 AM

TheTrunk (approved):

MorningLightMountain713:
I forgot to make the helpers executable on this branch... have done so now.

Have also tested this on nodeJS 14.21.3 on a live node... working.

Cabecinha84 (approved):
Ack

PR #1281, opened 2024-04-21
Docker command migration

MorningLightMountain713:
**What this pull does**

* Updates deviceHelper child_process commands.
*. Uses the more modern `findmnt` instead of mount
*. Added a bunch of notes so in 6 months time I remember how things work.
*. Removes the df command - not needed anymore with findmnt. 
*. Log output if the module is run via node directly.
* Updates dockerService, with new command.
* Updates dockerService to use new command runner for FixDockerLogs.sh
* Add new deviceHelper tests
* Fix fluxNetworkService tests (not part of this pull but needed fixing)
* Remove node-cmd requirement from dockerService and deviceHelper modules

@XK4MiLX - Can you please check the deviceHelper stuff, I'm pretty sure this is correct but want to make certain. 

I haven't run this on any nodes yet - will do that this afternoon.


MorningLightMountain713:
All tests passing now

Screenshot 2024-04-21 at 11 45 47 AM

ZelBack/src/services/dockerService.js
```diff
@@ -987,60 +980,62 @@ async function dockerGetUsage() {
...
+  const cwd = path.join(__dirname, '../../../helpers');
+  const scriptPath = path.join(cwd, 'dockerLogsFix.sh');
+  const { stdout } = await serviceHelper.runCommand(scriptPath, { cwd });
+
+  // we do this so we don't log empty lines if there is no output
+  const lines = stdout.split('\n');
+  // this always has length
+  if (lines.slice(-1)[0] === '') lines.pop();
+
+  lines.forEach((line) => log.info(line));
```
Cabecinha84, line 995: Shouldn't this all continue inside a try catch?

MorningLightMountain713: Yeah nice spot, the stdout.split() could throw, I'll fix.

MorningLightMountain713: This highlighted an error I had in runCommand, stdout and stderr as allways returned as strings from child_process execFile. I've fixed this up (and tests) now that above code will never throw. Howver, for safety, have wrapped in a try / catch

PR #1284, opened 2024-04-25
Make interaction with apt more robust

MorningLightMountain713:
**Version matrix**

| OS version     | apt           | netcat-openbsd | ca-certificates |
| --------------|----------|-----------------|----------------
| Ubuntu 18.04  | 1.6.17      | 1.187                  | 20230311          |
| Ubuntu 20.04 | 2.0.10      | 1.206                 | 20230311          |
| Ubuntu 22.04 | 2.4.11      | 1.218                  | 20230311         |
| Ubuntu 24.04 | 2.7.14      | 1.226                 | 20240203         |
| Debian 12       | 2.6.1        | 1.219                  | 20230311         |

***Apt***

Any version of apt before 1.9.11 does not have access to the `-o DPkg::Lock::Timeout=x` feature. This just means that instead of using apt to wait for the lock, we have to do it manually. We have a 1 minute retry cooldown between retries if an apt command fails, and we retry 5 times.

***Netcat***

We just target a minimum version of `1.187` which will satisify the above OSes.

**What this pull does:**

* Add a 15 minute timeout to any child_process command (This is able to be overridden per command)
* Update semver check to allow just major and minor, or just major. (So not semver) As some dpkg versions don’t have minor or patch (ca-certificates, netcat-openbsd)
* Move netcat install to part of system service.
* Stop using syncthing install shell script, and move to systemService.
* Install netcat from systemService
* Checks installed packages meet requirements. (uses dpkg-query) Also makes sure package is installed.
* Move systemService start to top of serviceManager.
* Add fifo queue for apt commands. This will do the following:
  * run them sequentially, i.e. fifo.
  * If you add more than what can run, the commands will queue up.
  * If a command errors, the queue is halted and error event emitted.
  * If a command errors, it is re-added at the start of the queue (configurable).
  * On error, apt-get monitoring starts. If the error clears, the queue resumes.
  * If the error doesn’t clear and it’s not a lock, dpkg configure is run as a last resort.
  * At no point do we try kill any other apt processes, or delete any locks.
  * Sets a max queue size of 10. If more commands get added, old commands drop off.
  * You can add tasks to the queue, and optionally await them, or, just await the queue being empty.

Tested on 18.04, 20.04, 22.04 Debian 12. The minimum version for netcat-openbsd is 1.187. That is the latest package available on 18.04.

We probably need to create an object that contains the supported systems, with minimum package versions. This shouldn’t be hard to do.

**ToDo**:

***Tests:***
- [x] semver
- [x] fifoqueue
- [x] systemservice

**Risks:**
We aren’t really able to run the apt commands as detached children, as we need to get the error info back.

So until nodemon is removed, there will always be a risk that we get a messed up dpkg database from an apt command being killed. However, this should be resolved by the monitorApt method, so pretty edge case I think.

### DIFF
--- a/ZelBack/src/services/serviceHelper.js
+++ b/ZelBack/src/services/serviceHelper.js
@@ -11,6 +11,11 @@ const dbHelper = require('./dbHelper');
 const log = require('../lib/log');
 
 /**
+ * The max time a child process can run for (15 minutes)
+ */
+const MAX_CHILD_PROCESS_TIME = 15 * 60 * 1000;
+
+/**
  * Allows for exclusive locks when running child processes
  */
 const locks = new Map();
@@ -261,6 +266,9 @@ async function runCommand(userCmd, options = {}) {
   const params = options.params || [];
   delete execOptions.params;
 
+  // Default max of 15 minutes
+  if (!execOptions.hasOwnProperty('timeout')) execOptions['timeout'] = MAX_CHILD_PROCESS_TIME;
+
   if (!userCmd) {
     res.error = new Error('Command must be present');
     return res;
@@ -324,7 +332,14 @@ async function runCommand(userCmd, options = {}) {
  * @returns {{version, major, minor, patch} | null} The parsed version
  */
 function parseVersion(rawVersion) {
-  const semver = /^[^\d]?(?<version>(?<major>0|[1-9][0-9]*)\.(?<minor>0|[1-9][0-9]*)\.(?<patch>0|[1-9][0-9]*))(-(0|[1-9A-Za-z-][0-9A-Za-z-]*)(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/;
+  // Todo: tests
+
+  // modified this to allow for just major and minor. (and also ~ instead of - after version)
+  // I.e. dpkg-query --showformat='${Version}' --show netcat-openbsd
+  // 1.218-4ubuntu1
+
+  //const semver = /^[^\d]?(?<version>(?<major>0|[1-9][0-9]*)\.(?<minor>0|[1-9][0-9]*)\.(?<patch>0|[1-9][0-9]*))(-(0|[1-9A-Za-z-][0-9A-Za-z-]*)(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/;
+  const semver = /^[^\d]?(?<version>(?<major>0|[1-9][0-9]*)\.(?<minor>0|[1-9][0-9]*)(?:\.(?<patch>0|[1-9][0-9]*))?)([-~](0|[1-9A-Za-z-][0-9A-Za-z-]*)(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/
 
   const match = semver.exec(rawVersion);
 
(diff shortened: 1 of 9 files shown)


MorningLightMountain713:
Screenshot 2024-04-25 at 5 47 30 PM

MorningLightMountain713:
This is finished. All tests passing.

Screenshot 2024-04-26 at 1 43 49 PM

Cabecinha84:
We have nodes running ubuntu 18, so it needs to be compatible.
Nvidia Jetson only has ubuntu 18.

MorningLightMountain713:
> We have nodes running ubuntu 18, so it needs to be compatible. Nvidia Jetson only has ubuntu 18.

I will spin up an ubuntu 18 node, so I can test

MorningLightMountain713:
Moved this back to draft...

I've moved the syncthing install script from being a script to managed by systemService.

5.5.0 doesn't play well when installing new nodes... the syncthing install hangs and it's not great.

This should be pretty good... just needs some more tests.

Cabecinha84:
Thanks for the feedback David. No rush and don't forget to rest on weekend :)

ZelBack/src/services/syncthingService.js
```diff
@@ -2108,7 +2109,7 @@ async function installSyncthing() { // can throw
     const exec = `cd ${nodedpath} && bash installSyncthing.sh`;
     await cmdAsync(exec);
   }
-  syncthingInstalled = true;
+  syncthingExecutable = true;
   log.info('Syncthing installed');
```
akhtarg, line 2113: Doesnt look like syncthing old version enforcement is working... plenty of nodes on the old versions
{'v1.27.6': 11192, 'v1.27.7-rc.1': 2, 'v1.27.5': 54, 'v1.27.4': 51, 'v1.25.0': 8, 'v1.27.3': 5, 'v1.23.1': 2, 'v1.27.2': 20, 'v1.27.1': 8, 'v1.27.0': 2, 'v1.23.4': 4, 'v1.24.0': 2, 'v1.23.6': 1, 'v1.23.2': 1, 'v1.22.1': 1, 'v1.26.1': 5, 'v1.22.2': 1, 'v1.23.7': 1}

MorningLightMountain713: We're not enforcing it. (yet)

MorningLightMountain713:
Finsihed here (again). Have added extra notes / detail to the description. I've got this running on one of my (22.04) nodes. Can remove syncthing package and it gets reinstalled etc. Fixes issues with netcat also. 

All tests passing.

ZelBack/src/services/serviceHelper.js
```diff
@@ -378,6 +393,10 @@ function minVersionSatisfy(targetVersion, minimumVersion) {
   return true;
 }
 
+if (require.main === module) {
```
Cabecinha84, line 396: Is this if clause needed?

MorningLightMountain713: Not needed... have removed.

ZelBack/src/services/systemService.js
```diff
@@ -122,57 +362,182 @@ async function ensurePackageVersion(systemPackage, version) {
...
 
-    syncthingTimer = setInterval(async () => {
-      syncthingVersion = config.minimumSyncthingAllowedVersion;
-      response = await axios.get('https://stats.runonflux.io/getmodulesminimumversions', axiosConfig).catch((error) => log.error(error));
-      if (response && response.data && response.data.status === 'success') {
-        syncthingVersion = response.data.data.syncthing || config.minimumSyncthingAllowedVersion;
-      }
-      ensurePackageVersion('syncthing', syncthingVersion);
-    }, 1000 * 60 * 60 * 24); // 24 hours
+    syncthingTimer = setInterval(versionChecker, 1000 * 60 * 60 * 24); // 24 hours
```
Cabecinha84, line 396: nice clean up.

Cabecinha84 (approved):
Ack. Great work!

PR #1287, opened 2024-04-30
Feature: FluxDrive Backup/Restore

alihm:
This pull request adds UI communication to the FluxDrive middleware, enabling backup storage and retrieval from FluxDrive. Additionally, some changes have been made to the download function in IOUtils.js to support large files and handle download failures.


Cabecinha84 (changes requested):
When the backup is finished, let's have all files pre selected or people will not even know the possibility to upload to flux drive exist, unless they click on the file. This way seems like only download option exists. Makes sense?
image

Cabecinha84:
On my screen, with 100% zoom, I don't see the site options...
image

Cabecinha84:
image
Continues to not work...

Cabecinha84:
When I download the file from the drive I get a .gz file, where inside there is another zip file.
image
And inside that zip file it's the real content:
image
And when I download the backup file directly from fluxOs, I get a tar.gz with the content expected inside:
image

Cabecinha84:
Can we have a loading bar here when uploading to fluxdrive?
image

Cabecinha84:
Seems like even with the 3 retries it still fails...
image

Cabecinha84:
> Seems like even with the 3 retries it still fails... image

Pressed already 10 times the restore and always fails on download task.

alihm:
> When the backup is finished, let's have all files pre selected or people will not even know the possibility to upload to flux drive exist, unless they click on the file. This way seems like only download option exists. Makes sense? image

Improved visuals and added text so users know they need to select components to upload

alihm:
> On my screen, with 100% zoom, I don't see the site options... image
image

added ellipses for long links. urls are also clickable now.

Cabecinha84:
image
Instead of the message bellow the table I would prefer to see all elements of the table pre selected.

Cabecinha84:
Or leave the message, but put the part bellow showing even if nothing is selected, and the button to upload to fluxdrive only getting enable after you select some component.
image

Cabecinha84:
the restore process, needs to be adjusted to the screen.
Things only show up if I resize the browser to 90%, this is how it shows up if it is on 100%:
image

Cabecinha84:
Can we add a information on mouse ouver on FluxDrive button, explaining that there is a limit of 10GB max that can be used for free for each application and the backups done on this way are only available to download from FluxOS UI on restore page.
image

Cabecinha84 (approved):
Ack

PR #1293, opened 2024-05-03
Feature/dns management

MorningLightMountain713:
**What this pull does**

This is a small pull that adds additional dns servers and caching - appwide.

This has the added bonus of caching dns requests locally, so they don't have to go out to libuv and the OS.

This has no impact on system DNS.

We add 1.1.1.1, 8.8.8.8 and 9.9.9.9 (Cloudflare, Google and Quad9) in that order.

We prefer the original DNS, then cloudflare, then google, then quad9. 

If there is an issue with the original, it fails over to cloudflare, then resolves against that for any subsequent calls, until a timer expires, then it tries the primary dns again.

The additional servers only get added if they don't exist in the original servers.

**Why this is important**

In the current implemenatation, we rely on Operator / VPS DNS being set appropriately. This is not the most reliable and can cause issues. Some of the FluxOS code assumes that DNS is allways available, ie. the stats.runonflux.io endpoint etc.

- [x]  Manual testing
- [x]  Running on node testing
- [x]  Old NodeJS version testing

### DIFF
--- a/apiServer.js
+++ b/apiServer.js
@@ -25,16 +25,24 @@ const apiPortHttps = +apiPort + 1;
 let initialHash = hash(fs.readFileSync(path.join(__dirname, '/config/userconfig.js')));
 
 /**
- * This is solely for testing. The CacheableLookup()
+ * The Cacheable. So we only instantiate it once (and for testing)
  */
-let cacheable;
+let cacheable = null;
 
 /**
- * Gets the cacheable variable for testing
+ * Gets the cacheable CacheableLookup() for testing
  */
 function getCacheable() {
   return cacheable;
 }
+
+/**
+ * Gets the cacheable CacheableLookup() for testing
+ */
+function resetCacheable() {
+  cacheable = null;
+}
+
 /**
  * Adds extra servers to DNS, if they are not being used already. This is just
  * within the NodeJS process, not systemwide.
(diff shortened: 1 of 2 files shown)


MorningLightMountain713:
I've tested this on node `14.18.1` - works fine.

Of note, this adds dns to the global http agents. So if you're call new http.Agent(), it won't have the dns updated. We're only doing this in the code in one place at the moment (we should remove this). So it's not really an issue. I have created a ticket for the appsService for us to remove the new Agent() call.

If we do need to call `new Agent()` in the future, we can expose the functionality.

MorningLightMountain713:
Have had a look at the `cacheable-lookup` dependency required for this pull request.

Some info:

* It has 53M downloads per week.
* It has no dependencies.
* It is ~400 lines of code.
* It's under MIT license.

If we're worried about supply chain attacks, we could just inline this module under FluxOS (MIT license).

After looking at the code, it's not quite so simple as to create something similar and robust (Even without the caching).

In my opinion, we would get a lot of value out of this dependency.

Just browsing through the dependencies, there are some others that may be there for historic reasons that maybe we should look at removing?

* path                 - a copy of nodeJS path? Not sure why we have this.
* morgan            - an expressJS middlware logger, we could write our own middleware in a few lines.
* httpShutdown - this is used in app service, this one is easy to rewrite.
* inquirer            - Used to ask one question in init.js, we could rewrite this.
* node-cmd        - I'm in the process of removing.
* splitargs           - This one should really go.

Cabecinha84 (approved):
ack

PR #1294, opened 2024-05-04
Update repositories.json

dsearle:
Adding djsearle1977 (signalone's) repo to the whitelist so that I can use a custom image designed to work in the flux network for farcasterxyz/hubble

# Application whitelist

I'm going to run the farcasterxyz/hubble image with an added bootstrapper script so that it works with the flux network

Only locally

yes - look at farcasterxyz/hubble

# Checklist:

- [X ] Whitelist of application is only modifying repositories.json file
- [ X] repositories.json is still a valid JSON file
- [ X] Only whitelists single docker image organisation (one whitelist at a time, more whitelists, more PRs)
- [X ] No other whitelist has been deleted
- [X ] I agree with ToS https://cdn.runonflux.io/Flux_Terms_of_Service.pdf
- [X ] Application follows ToS - Application is not malicious. Application is not a scam. Application does what is meant to do and does not mislead in any way. Application does not do anything illegal. Application is not a mining application (not even bandwidth mining).
- [X ] In case application receives multiple reports, behaves maliciously, it will be blacklisted and removed from the network.

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -447,5 +447,6 @@
    "quay.io/invidious/invidious",
    "jpelton",
    "farcasterxyz",
-   "jpelton722"
+   "jpelton722",
+   "djsearle1977"
 ]


Cabecinha84 (approved):
Ack

PR #1295, opened 2024-05-04
Feature/stream chain

MorningLightMountain713:
Allows Node operators to stream the chain off a node, via the local network. Will not work over public IPs.

Means node owners can always pull an up to date chain, they can either take a small risk and do it live, uncompressed, or stop fluxd and pull it safely. 

This also has the advantage of not having to uzip the tarfile, which takes significant time.

Under no circumstance will it corrupt an existing chain. Worst case is if you do it live, you can corrupt the new chain (if a database compaction happens during the stream) This is a small risk, and just means you would have to download the chain again. In the dozen or times I tested this, I didn't get a corrupted chain.

*Mode matrix*

|             | compression        | no compression     |
|-------------|--------------------|--------------------|
| safe mode   | :white_check_mark: | :white_check_mark: |
| unsafe mode | :x:                | :white_check_mark: |
|             |                    |                    |

Comes with tests.

In the future, I can adapt it to work with our CDN. 

Here is the chain being streamed over the local network (uncompressed), it gets tarred on the fly:

Screenshot 2024-05-04 at 10 56 36 PM

Compared to a CDN download and untarring (with gzip):

Screenshot 2024-04-29 at 12 08 01 PM

Caveat:

Due to tarring on the fly, we are unable to set the response `content-length` header. This means that curl and others, don't know the size of the download in advance. So can't display "time remaining"

We could add it if fluxd is stopped as we know the file sizes. To get the tar size in advance, you just need to pad every file size with modulo 512, and add 512 bytes for every file + 1024 to signal eof (I had this working) but it seemed too complex for little value so I removed it.


MorningLightMountain713:
I've tested this on a node with Ubuntu 18 and NodeJS 14.18.1 - works fine.

All tests passing:

Screenshot 2024-06-21 at 9 34 51 AM

MorningLightMountain713:
I wrote a medium article so people know how it works when we release.

https://medium.com/@udingmyride/fluxnode-stream-chain-feature-49e6da0df38a

Cabecinha84 (approved):
Loved it! Great job!

Cabecinha84:
Only thing that I believe should be added on instruction is to pm2 stop both flux and watchdog before streaming and start them properly after.
FluxOS won't do anything with daemon shut down and watchdog will try to start the daemon when it's stopped.
Better safe than sorry.

PR #1297, opened 2024-05-08
Feature/load compose

MorningLightMountain713:
**What this pull does**

The template changes look gnarly for some reason, didn't change that much, mostly just lines 25-81. Main function that was added was `parseCompose` and a few other helper functions.

* Tested on Chrome, Safari, Firefox.
* Tested on a live node. Can just enter appname and description, upload spec and it is enough that registration message can be computed.

* Adds support for docker compose file uploads. Can drag and drop or use upload button. Supports compose deploy spec for resouces. https://github.com/compose-spec/compose-spec/blob/master/deploy.md
* Fixes bug where watcher was triggering when it shouldn't.
* Adds scroll to view for when specs are computed with a highlight / fadeout on the "sign with" as a visual cue that the user needs to interact with that element.
* Disabled the register app button unless there is some content in the signature box.

Here is the compose file I was testing with (renamed to json):

[compose.json](https://github.com/RunOnFlux/flux/files/15246900/compose.json)

### DIFF
--- a/HomeUI/src/services/Api.js
+++ b/HomeUI/src/services/Api.js
@@ -31,5 +31,6 @@ const sourceCancelToken = axios.CancelToken.source();
 export { sourceCancelToken };
 
 export default () => axios.create({
+
   baseURL: store.get('backendURL') || mybackend,
 });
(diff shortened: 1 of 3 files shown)


Cabecinha84 (approved):
ack

PR #1298, opened 2024-05-08
Feature - Sort components by docker compose "depends_on"

MorningLightMountain713:
**What this pull does**

* When importing a docker compose file, it takes into account the depends_on of the services.
* Sorts the components by depends_on. I.e. The containers that are "depended on" are first in the list, so they get installed first. I.e if `web` depends_on `mysql`, mysql will get installed first.
* Can specify both long form and short form `depends_on` however, only the keys are used for long_form. 
* Uses a DAG to resolve dependencies.
* Any cycles and the components are left in their original state. I.e. if you have `a depends_on b` and `b depends_on a`.
* Adds tests for topologicalSort.

### DIFF
--- a/HomeUI/src/services/Api.js
+++ b/HomeUI/src/services/Api.js
@@ -31,6 +31,5 @@ const sourceCancelToken = axios.CancelToken.source();
 export { sourceCancelToken };
 
 export default () => axios.create({
-
   baseURL: store.get('backendURL') || mybackend,
 });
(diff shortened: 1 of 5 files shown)


MorningLightMountain713:
All tests passing

Screenshot 2024-05-08 at 3 42 46 PM

Cabecinha84 (approved):
Ack, nice work!

PR #1300, opened 2024-05-14
v5.8.0

Cabecinha84:
- It's now possible to load .yaml files on registration page of FluxOs, making it more easier to move from tradicional docker to Flux network;
- Other UI Improvements/fixes;
- New Flux address for application payments, the new address will go live on block 1670000.


package.json
```diff
@@ -102,6 +102,7 @@
     "fullnode": "file:lib/fullnode",
     "http-shutdown": "~1.2.2",
     "inquirer": "~8.2.6",
+    "js-yaml": "~4.1.0",
```
TheTrunk, line 105: this should be dev dependency

PR #1302, opened 2024-05-16
Refactor/fluxapprequest

Cabecinha84:
This PR changes the way the node is asking for missing app messages:
- Introduces fluxapprequest version 2, where it sends a list of app hashes missing up to 500 hashes;
- Improves chaching on fluxapprequest message received processing;
- Adds missing validations to fluxapprequest message received before processing;
- The requests for missing app hashes is now done peer by peer, first we ask to a random outgoing peer, after to a random incoming peer and continues like this until 4 calls are made;
- We ask for all apps messages missing to first peer, and only ask to next peer the messages are still missing from the original batch;
- Should take few hours for one node to have all apps hashes messages and be able to start installing apps if required after install fluxnode;
- We continue to use the old way of calling all peers for a message before we classify the message as not found;
- Improve explorer block processing for apps messages.


ZelBack/src/services/fluxCommunicationMessagesSender.js
```diff
@@ -194,52 +299,73 @@ async function sendMessageToWS(message, ws) {
...
-        appSpecifications: appMessage.appSpecifications || appMessage.zelAppSpecifications,
-        hash: appMessage.hash,
-        timestamp: appMessage.timestamp,
-        signature: appMessage.signature,
-      };
-      myMessageCache.set(serviceHelper.ensureString(message), temporaryAppMessage);
-      sendMessageToWS(temporaryAppMessage, ws);
+
+    // eslint-disable-next-line no-restricted-syntax
+    for (const hash of appsMessages) {
```
TheTrunk, line 346: So this is a loop of up to 500 messages, Is it alright performance wise? Can we do just one search of database with all the messages avoiding this loop?

Cabecinha84: I think this way it's fine, I have added a small delay between iterations.
It's also important that this way we are sending the messages one by one, and not flooding the node that asked for the messages with all the messages at the same time.

TheTrunk (changes requested):
Concept OK
Correction of failing tests are needed plus writing of new tests

ZelBack/src/services/appsService.js
```diff
@@ -7856,7 +7876,7 @@ async function appHashHasMessageNotFound(hash) {
  * @param {number} height Block height.
  * @param {number} valueSat Satoshi denomination (100 millionth of 1 Flux).
  * @param {number} i Defaults to value of 0.
- * @returns {void} Return statement is only used here to interrupt the function and nothing is returned.
+ * @returns {boolean} Return true if app message is already present otherwise nothing is returned.
```
TheTrunk, line 7879: I do not like much the inconsistency in the return of the function and and that we do not know how long it will take to execute

ZelBack/src/services/appsService.js
```diff
@@ -8016,6 +8039,38 @@ async function checkAndRequestApp(hash, txid, height, valueSat, i = 0) {
...
+ */
+async function checkAndRequestMultipleApps(apps, incoming = false, i = 1) {
+  try {
+    await requestAppsMessage(apps, incoming);
+    await serviceHelper.delay(30 * 1000);
+    const appsToRemove = [];
+    // eslint-disable-next-line no-restricted-syntax
+    for (const app of apps) {
+      // eslint-disable-next-line no-await-in-loop
+      const messageReceived = await checkAndRequestApp(app.hash, app.txid, app.height, app.valueSat, 2);
```
TheTrunk, line 8066: We can be stuck in this loop for a long time

ZelBack/src/services/explorerService.js
```diff
@@ -450,6 +450,29 @@ async function processStandard(blockDataVerbose, database) {
...
+  if (appsTransactions.length > 0) {
+    const options = {
+      ordered: false, // If false, continue with remaining inserts when one fails.
+    };
+    await dbHelper.insertManyToDatabase(database, appsHashesCollection, appsTransactions, options);
+    if (blockDataVerbose.height >= config.fluxapps.fluxAppRequestV2) {
+      while (appsTransactions.length > 500) {
+        appsService.checkAndRequestMultipleApps(appsTransactions.splice(0, 500));
+        // eslint-disable-next-line no-await-in-loop
+        await serviceHelper.delay(60 + (Math.random() * 14) * 1000); // delay 60 and 75 seconds
```
TheTrunk, line 462: I think that the processing now will be actualy slower than before. As after every block there might be some significant delay.

ZelBack/src/services/appsService.js
Cabecinha84: No, as you can see we are calling the function with latest parameter 2, with latest parameter two we will just check if the message was received or not, nothing else is done.

ZelBack/src/services/appsService.js
Cabecinha84: inconsistency fixed.

ZelBack/src/services/explorerService.js
Cabecinha84: Have reduced the delay in case we are already on fluxAppRequestV2 to a max of 10 seconds between each 500 messages... getting 500 app messages on a block, hopefully will hapen soon but I don't see it slowing down the block processing any time soon :)
I have removed the await before the fluxAppRequestV2 as currently we also do not await in there.

TheTrunk (approved):

TheTrunk:
Adjusting height to 1674000 to activate in 3 weeks.

PR #1306, opened 2024-05-17
v5.8.1

TheTrunk:
- Fix application messages processing and inconsistencies


ZelBack/src/services/explorerService.js
```diff
@@ -1283,6 +1286,53 @@ async function reindexExplorer(req, res) {
   }
 }
 
+async function fixExplorer(height = 1637000, rescanApps = true) {
```
Cabecinha84, line 1289: maybe also change the default to 1633000?

Cabecinha84 (changes requested):
a small change before the ack

PR #1314, opened 2024-05-27
Waiting animation

XK4MiLX:
App management:
- control
- component control
- update/renew ( compute update message, update application)
- cancel subscription

Marketplace:
- register


Cabecinha84 (approved):
Ack

PR #1316, opened 2024-05-28
Update repositories.json

2ndtlmining:
Home.runonflux deploying via the new docker compose features needs the tag to match the docker compose, thus updated to match the docker compose tag

# Application whitelist

Sigmanaut mining pool front end. Just updated the tag

*Please describe what application(s) do you plan to run.*

https://github.com/marctheshark3/sigmanaut-mining-pool-ui

cloud provider

*Please provide a POC link to application if it is already running somewhere. (optional)*

yes provided above

*Please provide a source code (optional)*

# Checklist:

- [ ] Whitelist of application is only modifying repositories.json file
- [ ] repositories.json is still a valid JSON file
- [ ] Only whitelists single docker image organisation (one whitelist at a time, more whitelists, more PRs)
- [ ] No other whitelist has been deleted
- [ ] I agree with ToS https://cdn.runonflux.io/Flux_Terms_of_Service.pdf
- [ ] Application follows ToS - Application is not malicious. Application is not a scam. Application does what is meant to do and does not mislead in any way. Application does not do anything illegal. Application is not a mining application (not even bandwidth mining).
- [ ] In case application receives multiple reports, behaves maliciously, it will be blacklisted and removed from the network.

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -450,6 +450,6 @@
    "djsearle1977",
    "zquestz",
    "ziii",
-   "ghcr.io/marctheshark3/sigmanaut-mining-pool-ui",
+   "ghcr.io/marctheshark3/sigmanaut-mining-pool-ui:main",
    "garyotto"
 ]


MorningLightMountain713:
This won't fix the problem. There seems to be a bug here in relation to the ghcr registry. I will investigate

MorningLightMountain713:
Okay, so the reason this isn't working is because that repository is using a format we don't support (yet)

```
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:cf4140f9e5012afbccbf9f63cd1817ac6c5ba3f97e76cd94c3794a876b4b4f79",
      "size": 3152,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:d8e7b767b14a307fb99f50a51f95c2f2320b3c67f070d8a55cf3e6a2f8a03eb7",
      "size": 566,
      "annotations": {
        "vnd.docker.reference.digest": "sha256:cf4140f9e5012afbccbf9f63cd1817ac6c5ba3f97e76cd94c3794a876b4b4f79",
        "vnd.docker.reference.type": "attestation-manifest"
      },
      "platform": {
        "architecture": "unknown",
        "os": "unknown"
      }
    }
  ]
}
```

I will add support for OCI images next week. 

In the meantime, would suggest closing the PR.

2ndtlmining:
@MorningLightMountain713 i was wondering if it is working yet, still getting an error when registering the app via home.runonflux.io with docker compose file:

image

PR #1322, opened 2024-06-06
Feature: Add OCI Image support and refactor image testing

MorningLightMountain713:
**What this pull does**

* Adds support for OCI images. (yay!)
* Completely removes the docker hub api.
* Allows for significant technical debt removal in future.
* Splits out all image tag parsing, validation (size / arch) to separate class.
* Adds tests for image validation. (All tests aggregated under imageVerifier)
* Caches Flux whitelist for 10 minutes globally. (IMO this should be a message that gets transmitted across the network to save DDosing github but that's another discussion)
* Should reduce api calls as we are only testing images once now for arch / size instead of multiple times.
* Updates axios to latest (tbc, if this breaks anything on older nodes).

Conceptually there are 3 things we are testing for images, and they are all interrelated:
* That the image has an architecture that our network supports. (At least one of `['amd64', 'arm64']`)
* That the images for each arch that our network supports (above) are below that max configured size (2Gb)
* That an image has a supported arch for this specific node.

These are now all handled by the `ImageVerifier` class. Have modified the `appsService` to accomodate these changes. Small changes required for `dockerService` we now pass in the provider, instead of it having to parse the tag for it.

Of note, we are no longer using the `https://hub.docker.com/v2` api. This added a lot of unnecessary complexity to the code. We just use the docker registry api for everything now. (A lot more images are on other providers now anyway) This means the docker provider has changed from `hub.docker.com` to `registry-1.docker.io` - I have updated the code to reflect this. Have also tested pulling images with this provider with dockerode - works fine.

This is the workflow to validate an image now:

```javascript
  const imgVerifier = new imageVerifier.ImageVerifier(
    appSpecifications.repotag,
    { maxImageSize: config.fluxapps.maxImageSize, architecture, architectureSet: supportedArchitectures },
  );

  await imgVerifier.isWhitelisted();
  imgVerifier.throwIfError();

  await imgVerifier.verifyImage();
  imgVerifier.throwIfError();

  if (!imgVerifier.supported) {
    throw new Error(`Architecture ${architecture} not supported by ${appSpecifications.repotag}`);
  }
```

It's still a bit clunky as we are using errors everywhere to signal intent, which I would like to remove in the future. (The methods above do return values as well, we're just not using them)

I also managed to simplify the whitelisting right down to this:

```javascript
    const separators = ['/', ':'];

    const whitelisted = ImageVerifier.whitelistedImages.find(
      (otherTag) => {
        const len = otherTag.length;
        const thisTag = this.rawImageTag;

        return (
          thisTag === otherTag
          || (thisTag.slice(0, len) === otherTag
            && separators.includes(thisTag.slice(len, len + 1)))
        );
      },
    );
```

Basically, if the whitelisted tag matches our tag, and the next character in our tag is a separator, it means the whitelisted tag is an ancestor, and allowed.

**ToDo:**

- [x] Validate old vs new api calls to providers via `trySpawnGlobalApplication`
- [x] Test on ubuntu 18
- [x] Test on node 14.18.1
- [x] Test on ubuntu 22 with node 20.9.0
- [x] Spawn local app on ubuntu 22 with node 20.9.0


MorningLightMountain713:
I've tested this against the following:

private docker distribution - megachips/sneaky:latest
private docker distribution - europe-west2-docker.pkg.dev/chode-400710/mugawump/cv:latest
public oci index                   - ghcr.io/marctheshark3/sigmanaut-mining-pool-ui:main
public oci index                   - ghcr.io/lloesche/valheim-server:latest

MorningLightMountain713:
Here is Ubuntu 22.04 with node 14.18.1.

```
davew@chud:~/zelflux$ curl -s localhost:16187/flux/nodejsversions | jq .data.node
"14.18.1"
```

This is using the OCI image `ghcr.io/lloesche/valheim-server:latest`

Screenshot 2024-06-06 at 1 30 22 PM

Screenshot 2024-06-06 at 1 30 56 PM

This is using the docker distribution list image `phpmyadmin:latest`

Screenshot 2024-06-06 at 1 37 44 PM

Screenshot 2024-06-06 at 1 38 05 PM

MorningLightMountain713:
All tests passing:

Screenshot 2024-06-06 at 4 23 55 PM

MorningLightMountain713:
I moved the repository check right to the end `trySpawningGlobalApplication`, just before the `registerAppLocally` call.

This is so we do all our local checks first, and if they pass, then we reach out and pull registry info. This significantly reduces 3rd party api calls.

I'm pretty happy with this - I can see it pulls the images, I just haven't had an app spawn to completetion on my test node yet.

MorningLightMountain713:
An image got deployed on my test node which highlighted a gap in our approach, and a reasonable bug in production. Currently, any images that are deployed outside of dockerhub (via registry api) that are arm64 specific, and don't return a manifest list, could get deployed on any node, including amd64.

The current mpact of this bug is probably zero. I only found it as we're using the reigstry for dockerhub images now.

If an image manifest is returned instead of a list, it was assumed that the image was amd64. This is incorrect (as I just found out when this arm64 image got deployed on my node), It could be anything.

I've now patched this and updated the tests. If we receive an oci/distribution manifest as the first response, we go to the `/blobs` endpoint and look up the config sha. This has the arch in it, and we use that.

This is similar behavior as the docker cli - I took a peek at their code. 

I will run this for another week or so to make sure it's good.

MorningLightMountain713:
I finally got an app deployed on a node running this branch. I will rebase and move from draft

Screenshot 2024-07-03 at 4 48 26 PM

Cabecinha84 (approved):
Ack

Cabecinha84:
Great work David!

PR #1333, opened 2024-06-15
Feature: add debug machinery

MorningLightMountain713:
**What this pull does**

* Adds socket.io server class. This allows for better separation of each socket.io namespace handler.
* Adds a requestHistory class for easy storing of requests.
* Uses namespaces for each socket.io "functionality" i.e. debug and terminal.
* Adds info to Home page if daemon is uncontactable. (handy when testing)
* Easily extensible socket.io - just add more handlers.
* Adds a debug toggle to user config to enable / disable debug. Easily extensible in the future to have different debug flags. So you don't have to turn on all debug, just what you want to see. 
* Updates the `axiosGet` method. The old cancel token is deprecated. You no longer need to use a timeout as the native timeout handles connection timeouts too. Adds in the FluxController, for in the future we will be able to use this to cancel any running axios calls globally. (You can override the signal by passing a signal into axiosGet)
* Adds axios global config (basically just a 20 second timeout)
* Adds awaiable http server
* Adds global exception handling (this is a workaround for not being able to catch express error)
* Adds websocket server config to patch websocket issue.

If the debug flag is disabled, then the interceptors don't get updated and no request history is stored. When it's enabled, Axios will store every request for 1 hour.

The debug namespace is partitioned using rooms. Currently only using the `httpOutbound` room. We can emit per room, so we can have different rooms for different debugs, and the can be individually joined. 

I'll add tests before this gets merged.

**What it looks like**

This is on the Flux -> Debug page on the GUI. The debug namespace socket is behind auth. I.e. You need to be logged to see the page, and logged for the debug socket to connect.

_Each row is expandable. Here are the rows unexpanded:_

Screenshot 2024-06-15 at 1 27 14 PM

_Here is one row expanded:_

Screenshot 2024-06-15 at 1 26 18 PM

**Known issues**

Since we are using the same http server for both socket.io and websockets, there is an error on browser console for "Invalid frame header" as the websocket server tries to handle the connection for socket.io, which it can't. The solution here is to remove `express-ws` and separate the routes. Everything still works - just an error on the console. (This was present prior to this pull)

**Tasks to do after this pull**

* Move all axios calls to use the `axiosGet` method (and axiosPost, axiosInstance). Currently, it's a mix of both. While they work fine as is, if we want to change something, means we have to do it everywhere in the code. Similar to the `runCommand` thing - we want them all to go through the same place. That way they all get the defaults we have set in axios globally.
* Fix up some of the sync stuff in the apiServer module.
* Remove express-ws.
* Pretty up the UI, and add notifier if debug is disabled.


Cabecinha84 (approved):
ack

PR #1338, opened 2024-06-17
Bugfix - Outbound connections

MorningLightMountain713:
Fixes bug where outbound connections were being repeatedly tried when an existing connection already existed.

I have tested this on a node and it fixes the problem I was having. 

There are quite a few other occurences of this in the code in both `fluxCommunication` and `appsService`. I thought it better to do a quick bugfix for this specific issue, and evaluate the rest of the code to make sure it doesn't break anything.

Couple of things I would like to do here: 

1/ Move all the string splitting stuff for the ip / port into a function on `serviceHelper` so at least it's all in one place.

2/ Fix the root problem - Fluxbench should give the 16127 port regardless of if it's a upnp node or not. This would clean up significant tech debt. Overloading the meaning of an <ip:port> with the upnp status makes the code complicated down the line. If we want to show that a node is upnp, we should expose a flag for it.


Cabecinha84 (approved):
Ack

PR #1339, opened 2024-06-17
v5.13.0

Cabecinha84:
- Fix sending duplicates apps on fluxapprunning messages;
- Fix rebroadcasting of fluxapprunning messages;
- Add running since to all fluxapprunning messages, this helps the network to know where the app should be uninstalled from if a new instance not needed is spawned for any reason;
- Add mongodb gpg key check/update if needed;
- Improvement of connections between nodes on the network;
- New section on Administration -> Flux -> Debug page with Outbound Requests information.
- Other small improvements.

This version will be enforced on August 3rd.


ZelBack/src/lib/server.js
```diff
@@ -4,8 +4,15 @@ const cors = require('cors');
 const morgan = require('morgan');
 const compression = require('compression');
 
-const expressWs = eWS(express());
-const { app } = expressWs;
+// see https://github.com/HenningM/express-ws/issues/120
+const options = {
+  wsOptions: {
+    maxPayload: 65535,
```
TheTrunk, line 10: is this sufficient?

ZelBack/src/lib/server.js
```diff
@@ -4,8 +4,15 @@ const cors = require('cors');
...
+  },
+};
+
+const app = express();
+const expressWs = eWS(app, null, options);
```
TheTrunk, line 15: should not this be 
const { app } = expressWs(express(), undefined, options); ??

ZelBack/src/services/serviceHelper.js
```diff
@@ -156,29 +163,55 @@ function isDecimalLimit(value, decimals = 8) {
...
-      clearTimeout(id);
-      return res;
-    });
-};
+async function axiosGet(url, userOptions = {}) {
+  const options = { ...userOptions };
+
+  if (!options.signal) options.signal = shc.signal;
+
+  return axios.get(url, options);
```
TheTrunk, line 176: the cancel token is not needed? I remember we added it as sometimes it did not stop after the 20 seconds but maybe axios fixed that already?

TheTrunk: I am supposing it might be done through the abort controller but reading through it I can't figure it out from it.

ZelBack/src/services/systemService.js
```diff
@@ -567,6 +567,55 @@ async function mongoDBConfig() {
...
+    const { stdout, stderr, error } = await serviceHelper.runCommand('gpg', { runAsRoot: true, params: ['--show-keys', '/usr/share/keyrings/mongodb-archive-keyring.gpg'] });
+    if (error) {
+      throw new Error(`Executing gpg: ${error}`);
+    }
+    if (stderr) {
+      throw new Error(`gpg stderr: ${stderr}`);
+    }
+    const expiredMatch = stdout.match(/\[expired: (\d{4}-\d{2}-\d{2})\]/);
+    const versionMatch = stdout.match(/MongoDB (\d+\.\d+) Release Signing Key/);
+    if (expiredMatch) {
```
TheTrunk, line 583: this is only executed if the key is expired right. So we are not doing the network requests actually.
Are we ok on running this function on every restart?

apiServer.js
```diff
@@ -87,6 +96,57 @@ async function createDnsCache(userCache) {
   }
 }
 
+function setAxiosDefaults(socketIoServers) {
+  if (axiosDefaultsSet) return;
+
+  axiosDefaultsSet = true;
+
+  log.info('setting axios defaults');
+  axios.defaults.timeout = 20_000;
```
TheTrunk, line 105: is the numeric separator _ supported in nodejs 14?

TheTrunk (commented):
Nice work, just a few questions for clarification

ZelBack/src/services/systemService.js
XK4MiLX: The local key is checked on every restart, and the key is only updated if it has expired.

apiServer.js
MorningLightMountain713: Yes - have tested.

ZelBack/src/lib/server.js
MorningLightMountain713: It's the same thing. I personally find the above more readable. (the server option can be underfined or null)

ZelBack/src/lib/server.js
MorningLightMountain713: It may not be for the terminal. Worked fine during testing but I wasn't sending big payloads. Maybe we should set this to something like 1_048_576 * 16 (16Mb)

ZelBack/src/services/serviceHelper.js
MorningLightMountain713: It's not done through the abortController - that is more of a future thing, where we might want to reload the system gracefully.

There is a lot of incorrect info about axios on the internet, and the docs arern't that great, especially the "cancellation" doc here https://axios-http.com/docs/cancellation. I think it's very poorly worded, to the point of being misleading.

During testing, I wasn't able to get it to hang. I'm pretty sure (99%) that the timeout cancels all requests, including connection related issues, after the timeout. I think the docs are alluding that it will only cancel after the request timeout, if you want to cancel earlier, you need to use the abortController.

For example:

```javascript
const axios = require('axios').default

async function main() {
  await axios.get('http://3.4.5.6', { timeout: 5_000 }).catch(err => console.log(err.code))
}

main()
```

Will return `timeout of 5000ms exceeded` even though, obviously, it doesn't connect as the ip is not listening. 

I think the problem may have been that the timeout was set to 0, where it would hang indefinetely if you didn't have a way to cancel the request. However we are setting a fallback timeout globally now.

If you want, I can add a setTimeout with the abort controller, but I don't think it actually does anything. I probably need to read the Axios code and see what they're actually doing.

MorningLightMountain713:
Something I've just realized, once nodes update to this version, the "interactive terminal" won't work when connecting from 5.13 to a node with a version < 5.13 as the endpoint has changed from `/` to `/terminal`.

So there may be a few situations that don't work until this version gets enforced. If that's a problem I can put a check in `Management.vue` to fetch the remote flux version beforehand

ZelBack/src/lib/server.js
TheTrunk: Agreed lets try to set it to that higher value. Can you make a PR for it?

ZelBack/src/services/serviceHelper.js
MorningLightMountain713: Cool as, happy to add a timeout back in if we need it

yes - in my other pull that I'm still working through testing, I have updated axios to the latest version. Have tested in on node 14, works well

ZelBack/src/lib/server.js
MorningLightMountain713: Yup - will do this now

MorningLightMountain713: Done

TheTrunk (approved):

Issue #1090, opened 2023-08-19
Malicious docker container [BUG]


omega1582:
Docker image is running traffMonetizer.  As far as i've been aware, these type of apps(mining, bandwith mining...etc) are not allowed on Flux Network. Docker Hub Repo is holdroot/tft.  https://hub.docker.com/r/holdroot/tft.   
See atttache image.  

![traffMonetizer](https://github.com/RunOnFlux/flux/assets/63751693/5ccc2e51-b0d0-4293-8e03-89fb6566f5fc)

TheTrunk:
Thank you for the report. Our tests have shown the same.

Image is now blocked and nodes will start to remove the application.

0xJohnny21:
The same user now running a hidden proxy server along with the html5 game.
Suspicious...

PR #1340, opened 2024-06-18
Update blockedrepositories.json

0xJohnny21:
Hello,
I have found an user running hidden socks5 proxy server along with a HTML5 game App.
Dockerhub: https://hub.docker.com/r/holdroot/h5-game
Its been discovered that its been deployed on 600+ nodes on Flux Network.
Previously same user had found running a bandwidth Sharing app (TraffMonetizer) #1090 

below screenshot shows startup script tries to remove storage.json ( the same file found on TraffMonetizer, also it tries to figure out the IP type) why does this even included in a game?
Screenshot1

Container logs:
Screenshot2

I checked some of the Flux node IP which runs these DAPP on spur.us, they classified that IP is found on DataImpulse Proxy Pool, ( Owned by Softoria LLC) also owns TraffMonetizer
Screenshot3


TheTrunk:
Hi, 
Thank you for the PR. We currently allow proxy services, vpns. However we do not allow any sort of mining including bandwidth mining.

PR #1344, opened 2024-06-19
Feature: rework websockets

MorningLightMountain713:
**What this pull does**

* Completely removes `express-ws`
* Implements the `FluxServer` this is an http server with socket.io and native websockets.
* Greatly simplifies the `apiServer.js`
* Removes the old `server.js`
* Removes flux communication reliance on Wss.
* Moves all the websocket routes out of express.
* Uses the same (simplified) route matching mechanism for websockets as express uses under the hood.

I have this running on a live node now.

Can just do this now, and it sorts everything out, including all the express routes, websocket and socket.io stuff:

```javascript
  const httpServer = new fluxServer.FluxServer();
  const httpsServer = new fluxServer.FluxServer({
    mode: 'https', key, cert, expressApp: httpServer.app,
  });

  await httpServer.listen(apiPort);
  log.info(`Flux listening on port ${apiPort}!`);

  await httpsServer.listen(apiPortHttps);
  log.info(`Flux https listening on port ${apiPortHttps}!`);
```

This pull fixes an issue we were having where the native websocket server was greeding all the websocket upgrades, so socket.io was forced to use polling only.

Now we handle the http upgrade manually, and pass this on based on the path to the correct server.

This prevents these errors in the console:

Screenshot 2024-06-14 at 8 36 24 PM

ToDo:

- [x] Write tests
- [x] Make sure inbound / outbound connections still work
- [x] Interop testing with Ubuntu 18, Node 14 etc


MorningLightMountain713:
I've tested this on a node running ubuntu 18 and nodejs 14 - inbound / outbound connections work fine. 

I've tested socket.io with the debug page - works fine. 

I haven't tested the terminal page (no apps to connect to) - I assume this works if the debug does.

Cabecinha84 (approved):
Ack

PR #1345, opened 2024-06-20
Add Tinode Chat

flikites:
Tinode Chat

https://github.com/tinode/chat - no encryption yet, but planned.


TheTrunk:
Whitelist is no longer required. FluxOS is now completely open, however we are still merging submissions :)

PR #1355, opened 2024-06-24
stop broken app being installed.

MorningLightMountain713:
This is for that broken app `NiseSkateboards` - he redeployed so all is well, however this is the old broken app hash.

It's trying to install on all my nodes every few hours. It downloads all the images, so it must really be giving dockerhub a workout.

This has highlighted an issue that I will fix after my oci image PR - we should validate everything that needs validating before we pull any images. i.e. environment and secrets. Would stop this sort of thing happening.


MorningLightMountain713:
Screenshot 2024-06-24 at 2 51 00 PM

PR #1363, opened 2024-07-03
Feature - enable fluxd subscription endpoint

MorningLightMountain713:
**What this pull does**
* Removes the `fullnode` packge entirely
* Adds a `DaemonConfig` class to service utils to replace the fullnode package.
* Keeps fullnode folder as a workaround for the `file:xxx` hack that was in `package.json`
* All the file access stuff is now async 
* Leaves a backup config file in the fluxd config dir just in case
* Sets a .zmqEnabled hidden file in the fluxd config dir so the update is only run once
* Validates first fluxd then zelcash, instead of just assuming it's zelcash if flux doesn't exist
* Updates the fluxd config file parser to be much more robust and modern (we could just use an ini parser here)
* Adds full test suite for fluxd config file parsing
* Adds the `zeromq` package (5.3.0 is compatible with Ubuntu 18 and 20)
* Turns on the zmq pub / sub endpoint for fluxd on `localhost:16126`

This pull request is a precursor for one of my other pulls so I can make it smaller.

**Background**

Currently, we have to poll fluxd whenever we want updates. This means data is stale, and there is a lot of unnecessary cycles burnt looking for data, and we are using multiple levels of caching, which aren't really a good fit for what we are trying to do.

With the zmq endpoint enabled, we can subscribe to block updates, and know exactly when new information is ready to fetch. So we don't need to poll anymore.

This pub / sub endpoint is read only and only available on localhost.

**Testing**

I've tested this on Ubuntu 18 running Node 14.18.1 - the only caveat is that zeromq has to build from source on ubuntu 18, which takes roughly an extra 30 seconds to install on the first run.

I've tested it on an ubuntu 20 and 22 node - both update fine after the 20 minute timeout. (give fluxd time to warm up)

I've added quite a lot of tests and checks around the fluxd config file. Obviously we don't want to break fluxd. This is the process and the assumptions that have been made:

**Assumption made that all nodes have `zelcash.service` as the supervisor of fluxd.**
* Run the zmq updater on fluxOS start
* If node has already updated, then don't run the update
* fluxd must be running and must respond to `getblockcount` otherwise update is skipped
* fluxd config file is based off that which is created by `multitoolbox`
* The actual config file is parsed, then new options added, then written to a `flux.conf.new` file
* A backup file is written to `flux.conf.bak` of the exact config as read from the file.
* `flux-cli` is run with the `conf=` flag pointing to the flux.conf.new file to check the file syntax
* If the above command doesn't run properly, it means the config file is bad and we bail out and restore the original
* We then write the actual config file and restart the flux daemon
* If the daemon fails to start, we revert to the old config file, and restart again, then bail
* If all is well, we write the `.zmqEnabled` lock file so the updater only runs once.

I'd like to get this run on a bunch of nodes first, or feedback on if there are nodes out there that are running non standard config file locations and stuff.


MorningLightMountain713:
All tests passing. We've cracked 2k tests!

Screenshot 2024-07-03 at 4 22 12 PM

MorningLightMountain713:
I added another check - if ```sudo systemctl status zelcash.service``` returns an error, we skip enabling `zmq`

The above command will return an error if the service is not running, or doesn't exist.

Have also added another test around this.

Fixed up function name where I had a typo too.

If we get this merged, I'll get a pull in to update the default fluxd multitoolbox config

ZelBack/config/default.js
```diff
@@ -74,6 +74,7 @@ module.exports = {
     rpcport: 16124,
     porttestnet: 26125,
     rpcporttestnet: 26124,
+    zmqport: 16126,
```
Cabecinha84, line 77: This is the default api port.
Will it not cause problems?
Doesn't this needs to also be configured because of upnp nodes?

Cabecinha84 (commented):
Have one question before merging.

ZelBack/config/default.js
Cabecinha84: sorry not api, ui port

MorningLightMountain713: Yup, poor choice there. Have updated.

PR #1365, opened 2024-07-06
Feature - Add Flux RPC.

MorningLightMountain713:
**Forked from the `feature/zmq_enable` branch. So that branch getting merged into develop is a dependency for this branch.**

#1363

(It will look a lot smaller once that branch is merged)

**What this pull does**
* Removes the `bitcoin` package
* Remove the `daemonrpc` inline package that was reliant on bitcoin
* Refactor `daemonRpcUtils` to use asyncLock and small tidy up
* Fixes issue where every benchmark rpc call was synchronously checking filesysytem
* Creates our own JSON RPC client - `FluxRpc` Much simpler, and leverages existing `axios`
* Adds test suite for the rpc client, and updates other tests reliant on daemonrpc.
* Adds extra features to the `AsyncLock` You can now await ready with a timeout, or await and disable with timeout.

The main reason I needed this was so I can pass a `signal` into the rpc to cancel it, when necessary. It can take approx 500ms to pull the determ node list, if a block comes in when we are in the middle of this, it makes sense to cancel the pending request and start another.

It was also good to get rid of another couple of dependencies in `package.json`


MorningLightMountain713:
All tests passing

Screenshot 2024-07-06 at 7 56 50 PM

Cabecinha84:
Lots of changes, will review/merge after tomorrow release :)

Cabecinha84 (approved):
ack

PR #1366, opened 2024-07-07
Feature - Remove version check hack

MorningLightMountain713:
Small frontend update.

Now that every node is on > `5.13.0`, can remove the version check I added to handle that some nodes were on old versions that didn't use the new terminal endpoint.

now all nodes use `/terminal` for socket.io interactive terminal.


Cabecinha84 (approved):
Ack

PR #1369, opened 2024-07-10
Feature - markerCluster for flux maps

MorningLightMountain713:
**What this pull does**
* Adds a new `FluxMap` component.
* Implements dark mode for maps. (both light and dark, toggleable)
* Updates the dashboard map to use this new component.
* Adds map with nodes locations to `Applications -> Management -> My Active Apps`
* Adds markercluster dependency to dev deps.

Let me know if there are any other places you want me to add any new maps.

**markercluster background**

Markercluster is a map marker aggregation tool. It provides much more intuitive and easy to understand map displays when a large number of markers are used on a map.

Markercluster is great when working with large groups of markers like we are. Depending on the data format used, it is much faster when trying to paint 10k+ markers.

Markercluster also has sane defaults for multiple markers on the same spot, if this happens, it arranges the markers around the point with a "path" line to the origin. This means that we can keep the popup text to one per marker, instead of having to merge them manually.

**FluxMap component**

This component uses the data format from fluxstats as input. I.e. you can pass in a prop `nodes` with the data in the format:

```json
[{
  "geolocation": {
    "ip": "104.49.67.219",
    "continent": "North America",
    "continentCode": "NA",
    "country": "United States",
    "countryCode": "US",
    "region": "IN",
    "regionName": "Indiana",
    "lat": 41.4495,
    "lon": -87.4769,
    "org": "Private Customer - AT&T Internet Services"
  },
  "ip": "104.49.67.219:16157",
  "tier": "CUMULUS"
}]
```

it will then use this data to build a `geoJson` object, this is a standard mapping datatype used by many mapping softwares and is reuseable.

If you don't pass in any nodes, it will fetch the list from the fluxstats api - it will display a warning message (on the map) `Unable to fetch Node data. Try again later` if the api is unavailable

You can also pass in the `filterNodes` and `showAll` props, if you only want to show specific nodes from the `nodes` list. We use this on the myApps page.

Screenshots:

Dashboard page:

Screenshot 2024-07-10 at 9 08 34 AM

My Active Apps:

Screenshot 2024-07-10 at 9 07 14 AM

Markers on same spot:

Screenshot 2024-07-10 at 9 09 32 AM

GeoJson data format:

```javascript
[{
  type: 'FeatureCollection',
  crs: {
    type: 'name',
    properties: {
      name: 'urn:ogc:def:crs:OGC:1.3:CRS84',
    },
  },
  features: [],
}]
```

Where we add one feature per node, in the format:

```javascript
{
  type: 'Feature',
  properties: {
    ip: node.ip,
    tier: node.tier,
    org: node.geolocation.org,
  },
  geometry: {
    type: 'Point',
    coordinates: [node.geolocation.lon, node.geolocation.lat],
  },
}
```

### DIFF
--- a/HomeUI/src/views/apps/Management.vue
+++ b/HomeUI/src/views/apps/Management.vue
@@ -3428,6 +3428,9 @@
             </b-table>
           </b-col>
           <b-col cols="12">
+            <flux-map class="mb-0" :show-all="false" :filter-nodes="mapLocations" />
+          </b-col>
+          <b-col cols="12">
             <b-pagination
               v-model="instances.currentPage"
               :total-rows="instances.totalRows"
@@ -5775,6 +5778,7 @@ import { mapState } from 'vuex';
 import ToastificationContent from '@core/components/toastification/ToastificationContent.vue';
 import ConfirmDialog from '@/views/components/ConfirmDialog.vue';
 import ListEntry from '@/views/components/ListEntry.vue';
+import FluxMap from '@/views/components/FluxMap.vue';
 import JsonViewer from 'vue-json-viewer';
 import FileUpload from '@/views/components/FileUpload.vue';
 import { useClipboard } from '@vueuse/core';
@@ -5861,6 +5865,7 @@ export default {
     // eslint-disable-next-line vue/no-unused-components
     BProgressBar,
     ConfirmDialog,
+    FluxMap,
     ListEntry,
     // eslint-disable-next-line vue/no-unused-components
     ToastificationContent,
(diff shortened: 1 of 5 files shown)


Cabecinha84 (changes requested):
Can you also add the same changes on management page - sub menu running instances?
Thank you.

MorningLightMountain713:
I've added the map to running instances.

Of note, I don't have any apps with many instances to test to make sure they only display the apps per page. Pretty sure it should work though.

Screenshot 2024-07-16 at 8 51 27 AM

Cabecinha84 (approved):
ack

PR #1371, opened 2024-07-21
Feature - update myApps page (vue)

MorningLightMountain713:
Happy for this pull to wait a while, quite a few changes in the current release. I may have got carried away on this one, was originally just to fix the map bug that slipped through in the last release.

**What this pull does**
* Many bugfixes. Eg:
  * Api calls not being caught.
  * Maps not refreshing when row details collapsed / expanded.
  * Row details not resetting when back pressed then return to myApps page.
  * Various console warnings fixed.
  * Etc
* Splits out the 1600+ lines of code in `myApps` to separate components. The `MyAppsTab` component could be composed further, this is just a start.
* Makes templates more readable (this should be the goal of all frontend views etc).
* Removes old titan route that was causing a console warning.
* Removes deprecated metamask debug flag.
* Removes substantial code duplication. (and dead code).
* Reworks the data modelling to be more simple.
* Aligns the `My Active Apps` and `My Expired Apps` pages so they are visually the same. This removes about 600 lines of code as we can reuse the same component for both.

Screenshots:

**Prior "My Expired Apps":**

Screenshot 2024-07-21 at 2 52 20 PM

**New "My Expired Apps":**

_Adds the per page and filter features, adds the app count, and adds the application expired notifier_

Screenshot 2024-07-21 at 2 52 30 PM

**Prior "My Active Apps":**

Screenshot 2024-07-21 at 2 52 03 PM

**New "My Active Apps":**

_Updates the page count label to be the same size_

Screenshot 2024-07-21 at 2 52 10 PM

**Prior Logged out view:**

Screenshot 2024-07-21 at 2 52 50 PM

**New Logged out view:**

Screenshot 2024-07-21 at 2 55 31 PM


Cabecinha84 (approved):
ack

PR #1374, opened 2024-07-22
Bugfix - www-auth header not parsing correctly

MorningLightMountain713:
Fix for incorrect www-authenticate header parsing. This was causing any image that had an underscore to be rejected (on register app page) as it couldn't match the repo for the authentication request.

Was missing the allowed underscore character.

Adds a test for this as well.

Tested on a live node - works fine now.


Cabecinha84 (approved):
ack

PR #1376, opened 2024-07-23
Don't compress response if the x-no-compression header is present

sygem:
Several FluxOS endpoint use server-sent events to show progress to the user. For example '/apps/testappinstall/' details the install process as it goes.

I have been unable to get Flutter to read these streams, and managed to track the problem down to the express compression plugin.

This solution allows a client to send a "x-no-compression" header with the request to disable the response compression. If no such header is sent, compression will take place as normal. I am then able to read the SSE streams in Flux Cloud.

### DIFF
--- a/ZelBack/src/lib/fluxServer.js
+++ b/ZelBack/src/lib/fluxServer.js
@@ -23,7 +23,16 @@ class FluxServer {
   static supportedModes = ['http', 'https'];
 
   static defaultMiddlewares = [
-    compression(),
+    compression(compression({
+      filter: (req, res) => {
+        if (req.headers['x-no-compression']) {
+          // don't compress responses with this request header
+          return false;
+        }
+        // fallback to standard filter function
+        return compression.filter(req, res);
+      },
+    })),
     morgan('combined'),
     express.json(),
     cors(),


Cabecinha84 (approved):
ack

MorningLightMountain713:
We don't need to do this...

The compression middleware allows for this already:

This middleware will never compress responses that include a Cache-Control header with the [no-transform directive](https://tools.ietf.org/html/rfc7234#section-5.2.2.4), as compressing will transform the body.

Cabecinha84:
I have merged and tested on development node and I don't see any difference.

Cabecinha84:
@sygem can you check against node http://94.16.104.218:16127/ (dev node with the changes already running) if it makes any difference? If not I will revert it. Please let me know ASAP

sygem:
> @sygem @Cabecinha84
> 
> We don't need to do this...
> 
> The compression middleware allows for this already:
> 
> `This middleware will never compress responses that include a Cache-Control header with the [no-transform directive](https://tools.ietf.org/html/rfc7234#section-5.2.2.4), as compressing will transform the body.`

I tried that, but couldn't get it to work with my Flutter tests.

PR #1378, opened 2024-07-24
Flushing write calls to improve access to server-side event streams

sygem:
Unless immediately followed by a res.end(), all calls to res.write() are now followed by a call to res.flush().

These flush calls are guarded to prevent errors if the compression middleware is not present for some reason.


Cabecinha84 (approved):
Ack

Cabecinha84:
You can test in like 15m against a dev node. Cheers

PR #1379, opened 2024-07-25
Bugfix - download progress

MorningLightMountain713:
Axios > 0.27.2 moves the `target` property under the `event` property for download progress.

This pull updates every use of this property to the correct target.

Have tested on a live node - in conjunction with the `flush()` fix that was put into place - we get live updates again.


Cabecinha84 (approved):
Ack

PR #1382, opened 2024-08-08
Bugfix: Unhandled DNS lookup error and image parsing.

MorningLightMountain713:
Two things going on here.

First, the image parsing was missing a return statement if no match at all was made (edge case), which somehow allowed for an image starting with a / to be accepted. I'm not sure how this app even got registered, as I tested it on the home.runonflux.io register page, and it throws an error.

I will add some tests in for this tomorrow.

Second, when FluxOS goes to remove the app (the image doesn't exist) it's causing a dns lookup to fail somewhere for the hostname "images" I can't track this down at the moment, so have put a hack in place so we catch this error globally, and continue on.

With patch:

```
Uncaught DNS Lookup Error!!, swallowing.
Error: cacheableLookup ENOTFOUND images
    at CacheableLookup.lookupAsync (file:///home/davew/zelflux/node_modules/cacheable-lookup/source/index.js:202:18) {
code: 'ENOTFOUND',
hostname: 'images'
}
```

It has to be one of the axios calls as that is the only thing using cacheable lookups. I will try find this tomorrow - I can replicate the error by putting this blob back in the db:

```
  {
    name: 'gridprueba',
    compose: [
      {
        name: 'Webserver',
        description: 'grid',
        repotag: '/nginx:latest',
        ports: [ 34349 ],
        domains: [ '' ],
        environmentParameters: [],
        commands: [],
        containerPorts: [ 80 ],
        containerData: '/tmp',
        cpu: 0.5,
        ram: 500,
        hdd: 15,
        tiered: false,
        secrets: '',
        repoauth: ''
      }
    ],
    contacts: [],
    description: 'asd',
    expire: 5000,
    geolocation: [],
    hash: '6718b5e140301c815ef2f8f0fec1a6ad8fa853f68b85a69828cfaa8f75470910',
    height: 1702605,
    instances: 3,
    nodes: [],
    owner: '1Ltfpyy1PuUuqaX8QT59nZ6UD2WURedBSF',
    staticip: false,
    version: 7
  }
```


MorningLightMountain713:
Will redo against develop

PR #1384, opened 2024-08-10
Bugfix: iptables timing

MorningLightMountain713:
I have experienced this on one of my nodes (and seen it mentioned in discord by one person)

This has only affected my ubuntu 20.04 node, (I only have one) so don't know if it's based on the versions in use on this system or something else.

Screenshot 2024-08-10 at 11 38 49 AM

We were running the iptables rules and ufw rules at the same time - and ufw uses iptables under the hood so getting a lock on xtables.

This pull just runs these function synchronously, so they don't compete. Also moved the iptable deny rules before ufw (makes sense to deny host traffic first then clear ufw)

Tested on my Ubuntu 20.04 node - fixed the issue.

### DIFF
--- a/ZelBack/src/services/serviceManager.js
+++ b/ZelBack/src/services/serviceManager.js
@@ -110,9 +110,9 @@ async function startFluxFunctions() {
     }, 15 * 1000);
     setTimeout(async () => {
       log.info('Rechecking firewall app rules');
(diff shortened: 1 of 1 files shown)


Cabecinha84 (approved):
ack

PR #1391, opened 2024-08-23
Refactor/freeupdates

Cabecinha84:
- Adds to getAppFiatAndFluxPrice method a new response property, freeNetworkUpdate;
 freeNetworkUpdate will be true if, instances, cpu, ram and disk was not changed, if expire was not extended and if there were less than 5 updates in the last day.
- Updates UI to no extend subscription when it was not selected to do it;
- Removes verifyAppUpdateIsNotSpammingNetwork check, message can go to temporary apps but will only be free updated on the top conditions.

### DIFF
--- a/HomeUI/src/views/apps/Management.vue
+++ b/HomeUI/src/views/apps/Management.vue
@@ -5079,6 +5079,12 @@
                     </h4>
                   </div>
                 </b-card-text>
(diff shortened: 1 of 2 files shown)


MorningLightMountain713 (approved):
Ack - looks good

PR #1401, opened 2024-09-17
v5.26.0

Cabecinha84:
- Refactor logs section on application management;
- Changes on application subscription extension, now it adds the period to the current remaining subscription;
- Other small improvements.

This version will be enforced on October 2nd. (Needs to be enforced to make sure logs section works on all nodes)


Cabecinha84:
@XK4MiLX is it possible to make the error when entering app management and the node that we are connecting to is not updated to only show when pressing logs section?

Cabecinha84:
@XK4MiLX can we fix this warning?
image

TheTrunk (approved):

PR #1402, opened 2024-09-21
Refactor rebroadcast of websocket messages

Cabecinha84:
- Check if node already have already processed the message before sending it;
- This can reduce up to 30x times the bandwidth used on a FluxNode;
- This replaces the need of fluxapprunning messages refactor;


TheTrunk:
So now instead of sending a message, we first ask the peer if it has already processed the message. If yes, we do not send it, if no we do send it.

I believe we are already doing this on the peer side. If message arrives to us that we already have, we disregard it and don't process it again. So this feels like doubling it. Yes it saves a bit of bandwidth if we suppose there is a lots of duplicates but resource wise it's more heavy as we do api call instead of sending ws message. This would be quite a lot of api calls, responses, and awaiting. 

Currently we send the messages and letting the peer decide if they want the message or not. If we want to do check on the hash first which would be more resource heavy but save bandwidth, we shall do it over the already established ws connection. So sending hash over ws and upon that responding. This is kind of like sending headers first then blocks if requested.

Cabecinha84:
> So now instead of sending a message, we first ask the peer if it has already processed the message. If yes, we do not send it, if no we do send it.
> 
> I believe we are already doing this on the peer side. If message arrives to us that we already have, we disregard it and don't process it again. So this feels like doubling it. Yes it saves a bit of bandwidth if we suppose there is a lots of duplicates but resource wise it's more heavy as we do api call instead of sending ws message. This would be quite a lot of api calls, responses, and awaiting.
> 
> Currently we send the messages and letting the peer decide if they want the message or not. If we want to do check on the hash first which would be more resource heavy but save bandwidth, we shall do it over the already established ws connection. So sending hash over ws and upon that responding. This is kind of like sending headers first then blocks if requested.

Ok I think it's a good idea, I will try to redo it using the ws, like a ping pong message before broadcasting the message.

PR #1413, opened 2024-10-08
v5.29.0

Cabecinha84:
- Improvements on websocket messages compression;
- New service that monitors if node goes offline all apps are uninstalled and also check if there are apps locations present on db from nodes that are no longer confirmed on the network;
- Refactor monitoring menu on application management;
- If user session expires, gets automatic to home/login page;
- Change packages for compatiblity with nodes running nodeJS 14 and 16;
- Delete from UFW netscans deny riles as we have introduced them on iptables for the docker apps running on the node;
- Other small improvements.


TheTrunk:
Test suite fails are related to apps monitoring and statistics as of our rewrite. As its not critical we can correct the test suite later in a patch

PR #1416, opened 2024-10-23
Feature - add chain size header to streamchain

MorningLightMountain713:
**What this pull does**

* Updates dependencies to use the more commonly used `tar` package. It handles changing files better.
* Adds an `Approx-Content-Length` header. The reason this is approx is that if you do the stream using `unsafe` mode, the file sizes change during transfer. Also, as we are tarring the files on the fly, we have to account for the tar header (512 bytes for each file, with each file padded to 512 bytes, plus a 1024 byte suffix) The image uses this to display how long to download the files, as it is also untarring them on the fly.
* Adds a new test for the header, and tidies up some of the other tests.

All tests passing.

Screenshot 2024-10-23 at 3 37 08 PM


Cabecinha84 (approved):
ack

PR #1421, opened 2024-11-14
Refactor trySpawningGlobalApplication

Cabecinha84:
Currently the method picks randomly one app from the applist and see if should and try to install it on the node.
With the change we removed all the random stuff, we check what are the apps missing instances, for those we also check if geo is set if it matches or if enterprise is set, if ip matches.
We now run the method every 5m instead of running every few seconds.
This should speed up a lot installing apps on the network that are missing instances.

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -70,8 +70,8 @@ testingAppserver = httpShutdown(testingAppserver);
 
 const GlobalAppsSpawnLRUoptions = {
   max: 2000,
-  ttl: 1000 * 60 * 60 * 2, // 2 hours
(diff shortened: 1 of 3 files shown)


MorningLightMountain713:
Not sure if this is adding the index `db.zelappslocation.createIndex({name: 1});` on fluxos start?

As it will need it for fast queries

Cabecinha84:
> Not sure if this is adding the index `db.zelappslocation.createIndex({name: 1});` on fluxos start?
> 
> As it will need it for fast queries

It took few miliseconds to complete. the index was already created:
image

Cabecinha84:
I think I have put the wrong print screen :D
image

PR #1422, opened 2024-11-22
Flux Node Service

Cabecinha84:
Flux Node Service it's a service created that the docker containers have access.
Changes done:
- Improvements on the geolocation service;
- Adds a new service that is only available to call on flux docker containers.
The service for now only have one route hostinfo that returns 3 properties, host identifier, host public ip and host geolocation.
image

### DIFF
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -9,6 +9,7 @@ module.exports = {
   server: {
     allowedPorts: [16127, 16137, 16147, 16157, 16167, 16177, 16187, 16197],
     apiport: 16127, // homeport is -1, ssl port is +1
+    hostInfoServiceAddress: '169.254.43.43',
   },
   database: {
     url: '127.0.0.1',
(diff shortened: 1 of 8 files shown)


MorningLightMountain713 (approved):
Looks good to me.

However, we need to write tests at some point for this.

I'm a little concerned about the iptables rules. The rules themselves are fine, however the checkDenyRule is brittle as we are matching on an error string. I.e. "Bad rule". 

I've tested on Debian 12 and Ubuntu 20/22/24

I can't remember if iptables on Ubuntu 18 outputs "Bad rule" I think it does but I don't have a node to check at the moment.

Cabecinha84:
> Looks good to me.
> 
> However, we need to write tests at some point for this.
> 
> I'm a little concerned about the iptables rules. The rules themselves are fine, however the checkDenyRule is brittle as we are matching on an error string. I.e. "Bad rule".
> 
> I've tested on Debian 12 and Ubuntu 20/22/24
> 
> I can't remember if iptables on Ubuntu 18 outputs "Bad rule" I think it does but I don't have a node to check at the moment.
> 
> It's probably not a big deal, even if it doesn't match properly, it will spam iptables with rules on every fluxOS restart, until the node is rebooted, then they will get cleaned out. This will probably be moot when I rework the firewall anyway.

Soon "all" nodes will be on Ubuntu 24 :)

PR #1424, opened 2024-12-09
[WIP] FluxOS Image path integration

MorningLightMountain713:
Update paths for Arcane. 

We now use the following non hardcoded paths from `process.env` (If they don't exist it falls back to the old paths):

```
FLUXD_CONFIG_PATH
SYNCTHING_PATH
FLUXBENCH_PATH
FLUX_APPS_FOLDER
```

I have also moved the `zeromq` dep to optional. It's not actually being used in the code yet, but it's good to have installed for when we do. (All existing nodes have it already anyway)

I had to move this to optional as Ubuntu 24.04 will not run version 5.x due to GLIBC versions. Optional is good as it will only install it if it can. 

ToDo:
*  Live image testing on both old and new nodes.
* ~~Run the test suite and fix broken tests.~~

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -46,7 +46,9 @@ const { invalidMessages } = require('./invalidMessages');
 const fluxCommunicationUtils = require('./fluxCommunicationUtils');
 
 const fluxDirPath = path.join(__dirname, '../../../');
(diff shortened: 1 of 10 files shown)


MorningLightMountain713:
All tests passing

Screenshot 2024-12-10 at 9 19 21 AM

Cabecinha84 (approved):
ack

PR #1433, opened 2025-01-30
v5.38.1

Cabecinha84:
Fix storing apprunning messages;
Adds staticIP information to apprunning messages/db

This is a mandatory patch that will be enforced in two weeks


MorningLightMountain713 (approved):
Ack. Looks good to me

TheTrunk (approved):

PR #1435, opened 2025-01-30
Feature: cancel stream chain on slow transfers

MorningLightMountain713:
Due to the nature of streaming the chain, fluxd has to be stopped.

If a node operator has rate limited the node that the transfer is initiated from, it could potentially take several hours for the download to complete. This will most likely result in the node going offline.

As a solution to this, we now monitor the download for the first 35 seconds. If the throughput is below 200Mbps, the stream gets cancelled and the node will no longer accept requests to stream the chain. (Until FluxOS JS is restarted)

It would take approx 24 minutes to stream the chain at 200Mbps. However, It is highly unlikely that a node would only stream at 200Mbps, i.e. if they are getting 200Mbps they are probably getting more than 200Mbps, as it is not a default network speed (100Mbps, 1Gbps etc) and an operator most likely would rate limit lower than this.

35 seconds was chosen as this lets the preparation timer timeout first, and it's good to have a reasonble time to estimate speeds as TCP can take some time to ramp up. (due to windowing)

I've tested this on a live node.

This will interact well with Arcane, as Arcane will only download from one local node, if that fails - it goes straight to CDN.

### DIFF
--- a/ZelBack/src/services/fluxService.js
+++ b/ZelBack/src/services/fluxService.js
@@ -77,6 +77,20 @@ function lockStreamLock() {
 }
 
 /**
+ * For testing
+ */
+function disableStreaming() {
+  streamChainDisabled = true;
+}
+
+/**
+ * For testing
+ */
+function enableStreaming() {
+  streamChainDisabled = false;
+}
+
+/**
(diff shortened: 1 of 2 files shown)


Cabecinha84 (approved):
ack

PR #1436, opened 2025-02-05
v5.39.0

Cabecinha84:
- Cancel stream chain on slow transfers;
- Monitor sharedDB apps on FluxOS.


MorningLightMountain713 (approved):
Ack

PR #1437, opened 2025-02-07
Syslog Implementation

XK4MiLX:
feat(appDockerCreate): add conditional static IP assignment with syslog integration

- Implemented helper function `getNextAvailableIPForApp` using the ip library to inspect the Docker network (derived from the app name) and automatically determine the next available static IP.
- Updated `appDockerCreate` to conditionally include the `NetworkingConfig` section:
  - If an auto-assigned IP is available, it is set via `EndpointsConfig` using the object spread operator.
  - If no IP is available (i.e. the helper returns null), the `NetworkingConfig` is omitted so that Docker assigns a dynamic IP.
- Integrated syslog functionality:
  - Checks environment parameters for syslog roles (e.g., `LOG=SEND` and `LOG=COLLECT`).
  - If the container is a log sender and a syslog collector is identified, the collector’s IP is retrieved (via `getContainerIP`) and used to configure the container's `LogConfig` with a syslog address (e.g., `udp://<collectorIP>:514`).

### DIFF
--- a/ZelBack/src/services/dockerService.js
+++ b/ZelBack/src/services/dockerService.js
@@ -1,6 +1,7 @@
 const config = require('config');
 const stream = require('stream');
 const Docker = require('dockerode');
+const ipLib = require('ip');
 const path = require('path');
 const serviceHelper = require('./serviceHelper');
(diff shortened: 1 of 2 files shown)


Cabecinha84 (approved):
Looks good

PR #1438, opened 2025-02-07
Update repositories.json

adam-1312:
# Application whitelist

simple web app that generates mailto links and shows it as qr code

probably, depends on when you're going to check this PR... see https://scan2mail.up.railway.app/ for temporarily hosted version

yes, see https://github.com/adam-1312/scan2mail/tree/main

# Checklist:

- [D] Whitelist of application is only modifying repositories.json file
- [D] repositories.json is still a valid JSON file
- [D] Only whitelists single docker image organisation (one whitelist at a time, more whitelists, more PRs)
- [D] No other whitelist has been deleted
- [D] I agree with ToS https://cdn.runonflux.io/Flux_Terms_of_Service.pdf
- [D ] Application follows ToS - Application is not malicious. Application is not a scam. Application does what is meant to do and does not mislead in any way. Application does not do anything illegal. Application is not a mining application (not even bandwidth mining).
- [D] In case application receives multiple reports, behaves maliciously, it will be blacklisted and removed from the network.

### DIFF
--- a/helpers/repositories.json
+++ b/helpers/repositories.json
@@ -465,5 +465,6 @@
    "testnet32/flights_scraper",
    "kingsimpa69/50-50.io",
    "archivebox",
-   "jlesage/firefox"
+   "jlesage/firefox",
+   "adam1312"
 ]


Cabecinha84:
Whitelist is no longer required. You can deploy your app on the network if it is not against the TOS agreement you have to accept.

PR #1447, opened 2025-02-22
Bugfix: Move monaco editor out of table

MorningLightMountain713:
Fixes the issue with monaco editor.

The old code was adding one editor for every file in the directory.

This just moves the editor out of the table so there is only ever one.


MorningLightMountain713:
Tested and working

PR #1453, opened 2025-03-11
v5.44.0

Cabecinha84:
- Speeds up apps install on the network.
Apps that are forbidden on the node by repo or port, and apps that failed to verify repo are put on a longer cache for 1 week to not be checked again to install.


MorningLightMountain713 (approved):
Ack.

Looks good

PR #1455, opened 2025-03-13
v5.44.1

Cabecinha84:
- Enables getting on component who have log collect enable to also receive their own syslogs;
- Updates syslog-format to RFC 5424 on log stream functionality.


MorningLightMountain713 (approved):
Ack

PR #1462, opened 2025-04-02
Polyfill for Node 14.18.1 Promise.any

MorningLightMountain713:
Small patch for `fluxService` to allow for use of `Promise.any` of older NodeJS

We could probably just mode this to `apiServer` but not much point if all nodes will be off 14.18.1 soon.

This is the only place we are using `Promise.any`


Cabecinha84 (approved):
ack

PR #1465, opened 2025-04-09
v5.48.1

Cabecinha84:
This version fix FluxOS setting nodes as DOS with message "IP changes over the limit allowed, one in 20 hours" when the public ip did not changed.


MorningLightMountain713 (approved):
Ack

TheTrunk (approved):

PR #1466, opened 2025-04-11
Update ip fetches from fluxd to fluxbenchd

MorningLightMountain713:
This moves the IP fetches (via benchmark) from fluxd to fluxbenchd. This removes unnecessary complexity from the IP fetch chain.

Here is an overview:
* Fixes the broken `fluxNetworkHelper` tests - these wern't passing the userconfig correctly. This also removes proxquire, as it's not needed. It wasn't noticed before that these were broken as they were never run as standalone tests. (I run them standalone on docker)
* Uses fluxbenchd instead of fluxd for benchmarks
* Removes no longer needed json parsing.

I have left a couple of comments in the `getMyFluxIPandPort` function - I'm not sure we should be setting the ip to null on a failed rpc call?

I have this running on a test node now - I'd like to run it for at least a couple more hours though


Cabecinha84 (approved):
ack

PR #1469, opened 2025-04-21
v5.50.0

Cabecinha84:
- Fix not showing options to update daemon and benchmarks on UI for legacy nodes;
- Improvements on parsing updates on app specifications, not reinstalling when there were no components changes, example: extending app subscription;
- Temporary disable fluxbench synced call to upnp nodes for research on possible issues with ArcaneOS (Only on ArcaneOS nodes);
- Ohter small improvements.


MorningLightMountain713 (approved):
Ack

PR #1470, opened 2025-04-26
Feature - allow limiting listfluxnodes response

MorningLightMountain713:
This pull does the following:

* Allows limiting response size for `listfluxnodes` aka `viewdeterministiczelnodelist`.
* Tidies up the routes so that they all point at the same backend rpc.

I have a need on Arcane where I want to pull some nodes from the list, but I don't want to pull the entire list as it's ~7Mib and can be quite slow. Using a limit, I can get the first 100 nodes, and the response is a lot faster / smaller.

The backend RPCs are all the same in fluxd so we may as well just use one:

Screenshot 2025-04-26 at 9 24 18 AM

I've tested this on one of my nodes


Cabecinha84 (approved):
ack

PR #1473, opened 2025-05-11
[WIP]: Feature/keepupnpportsopen

Cabecinha84:
- Nodes under UPNP calls adjustFirewallForUPNP every 10m and 1 second;
- Nodes under UPNP start calling another random node on the network to askthem to make a connection to the ports my node have in use by fluxOS + apps to keep them open.

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -12259,6 +12259,7 @@ async function callOtherNodeToKeepUpnpPortsOpen() {
 
     const dataUPNP = {
       ip: myIP,
+      apiPort,
       ports,
       pubKey,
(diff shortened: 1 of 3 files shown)


MorningLightMountain713:
Have made some changes. I still have more testing to do, but seems to work.

* We no longer need to process the req to get the body - it's finalized by our middleware
* Respond immediately with either a 202 or 401.
* Moved the TCP connect and destroy into it's own function and added a timeout as the default is quite high
* Moved the firewall updates back to 60 minutes - this is only really there as a safety net if operators were to remove the rules and is quite intensive, in the future we shoudl refactor the way we handle firewall rules and upnp and let a "manager" handle that stuff so it's more efficient. Gravity uses a lot of CPU / memory currently.

PR #1481, opened 2025-05-26
Introducing fluxappinstalling message

Cabecinha84:
Reduce the amount of installs to cover just the instances missing on the network.

The ideia of this PR is to improve the current scenario:
- App instance missing on the network;
- Multiple nodes select at same time the app to be installed and pass all the cheks;
- Multiple nodes install the app and broadcast apprunning messages;
- App gets more instances that it is required and it's uninstalled after install;

New scenario:
- App instance missing on the network;
- Multiple nodes select at same time the app to be installed and pass all the cheks;
- The nodes broadcast the new message type fluxappinstalling. This message have a life cycle of 5 minutes,
- Awaits 30 seconds and check again now how many nodes are running + are installing the app, and if it's needed we install if not we don't do anything;

Changes:
- Adds new collection appsInstallingLocations to database globalzelapps;
- Adds two new apis to get app installing location of one app and another method to get all apps installing locations;
- Adds all the methods to process when a node receive this type of message;
- Change TrySpawnGlobalApp to work like explained on new scenario;
- Deleted method getRunningAppList that was duplicated of appLocation.

What happens while nodes don't update to latest fluxOS? It remains working as mention on current scenario, the message is broadcasted and the nodes that are not updated don't do anything with it.


ZelBack/src/services/appsService.js
```diff
@@ -10158,11 +10159,80 @@ async function trySpawningGlobalApplication() {
...
+      });
+      const index = runningAppList.findIndex((x) => x.ip.split(':')[0] === myIP.split(':')[0]);
+      log.info(`trySpawningGlobalApplication - Application ${appToRun} is already spawned or being installed on ${runningAppList.length} instances, my instance is number ${index + 1}`);
+      if (index + 1 > minInstances) {
+        const appRemovedMessage = {
+          type: 'fluxappremoved',
+          version: 1,
+          appName: appSpecifications.name,
+          ip: myIP,
+          broadcastedAt,
```
MorningLightMountain713, line 10213: This timestamp is stale

ZelBack/src/services/appsService.js
```diff
@@ -10158,11 +10159,80 @@ async function trySpawningGlobalApplication() {
...
+        }
+        if (a.runningSince < b.runningSince) {
+          return -1;
+        }
+        if (a.runningSince > b.runningSince) {
+          return 1;
+        }
+        return 0;
+      });
+      const index = runningAppList.findIndex((x) => x.ip.split(':')[0] === myIP.split(':')[0]);
```
MorningLightMountain713, line 10205: What if two nodes on the same IP spawn the application?

ZelBack/src/services/appsService.js
```diff
@@ -10158,11 +10159,80 @@ async function trySpawningGlobalApplication() {
...
+    // store it in local database first
+    // eslint-disable-next-line no-await-in-loop, no-use-before-define
+    await storeAppRunningMessage(newAppRunningMessage);
+    // broadcast messages about running apps to all peers
+    await fluxCommunicationMessagesSender.broadcastMessageToOutgoing(newAppRunningMessage);
+    await serviceHelper.delay(500);
+    await fluxCommunicationMessagesSender.broadcastMessageToIncoming(newAppRunningMessage);
+    // broadcast messages about running apps to all peers
+
+    await serviceHelper.delay(30 * 1000); // give it time so messages are propagated on the network
```
MorningLightMountain713, line 10185: This is just a question... have we done any modeling around how long it takes for a message to propagate through the network?

MorningLightMountain713 (changes requested):
I like the premise of this PR. Should save a lot of resources / network bandwidth etc.

Do you think we need another field in the appremoved message so we can differentiate between when an app is actually removed... and when an app was rejected as there was already enough instances?

ZelBack/src/services/appsService.js
Cabecinha84: I have done a manualy check long time ago, it would take couple of seconds. We haven't done any modeling/professional check :D

MorningLightMountain713:
This seems super complicated.

I agree that we should split the message types. We are overloading the meaning of a `fluxapprunning` message. It will be far more simple in the long run if we just split the types - one message should do one thing.

Also we are modifying the message but it's still the same message version, kind of defeats the point of the version.

I find it confusing that the messages that are stored in the database aren't actually what is sent.

For example, a node sends this:

Screenshot 2025-05-29 at 10 22 43 AM

Then it actually gets stored like this:

Screenshot 2025-05-29 at 10 23 18 AM

So really, runningSince, broadcastedAt, and InstallingAt are all the same. Why doesn't the origin node just set the values to what they actually are (i.e. a timestamp or null) including the validTill? Then the receiver just has to format the timestamps.

What does `installingAt` actually mean? That the node has begun installing the app or that it is installed? as if it's the later... it should be `installedAt` which I would assume is correct as it's a `fluxapprunning` message, i.e. the app is running

Cabecinha84:
> This seems super complicated.
> 
> I agree that we should split the message types. We are overloading the meaning of a `fluxapprunning` message. It will be far more simple in the long run if we just split the types - one message should do one thing.
> 
> Also adding new message types instead of modifying existing ones also has the advantage that nodes that haven't updated can just ignore the new message types.
> 
> Also we are modifying the message but it's still the same message version, kind of defeats the point of the version.
> 
> I find it confusing that the messages that are stored in the database aren't actually what is sent.
> 
> For example, a node sends this:
> 
> Screenshot 2025-05-29 at 10 22 43 AM
> 
> Then it actually gets stored like this:
> 
> Screenshot 2025-05-29 at 10 23 18 AM
> 
> So really, runningSince, broadcastedAt, and InstallingAt are all the same. Why doesn't the origin node just set the values to what they actually are (i.e. a timestamp or null) including the expireAt? Then the receiver just has to format the timestamps.
> 
> What does `installingAt` actually mean? That the node has begun installing the app or that it is installed? as if it's the later... it should be `installedAt` which I would assume is correct as it's a `fluxapprunning` message, i.e. the app is running

Nice feedback, I will create a specific message type, should be very easy to code. Await for asking for new feedback soon.

Cabecinha84:
Guys, I think this is it, now it is good! :D

ZelBack/src/services/fluxCommunication.js
```diff
@@ -219,6 +219,57 @@ async function handleAppRunningMessage(message, fromIP, port) {
...
+ * @param {string} fromIP Sender's IP address.
+ * @param {string} port Sender's node Api port.
+ */
+async function handleAppInstallingMessage(message, fromIP, port) {
+  try {
+    // check if we have it exactly like that in database and if not, update
+    // if not in database, rebroadcast to all connections
+    // do furtherVerification of message
+    // eslint-disable-next-line global-require
+    const appsService = require('./appsService');
```
MorningLightMountain713, line 234: I'm assuming this is done because of a dependency cycle - might be a bit of a perfomance impact importing this module on every installing message.

Can we move the `storeAppInstallingMessage` method to a common module, that both `fluxCommunication` and `appsService` import?

Seems like it would be a good fit on the `dbHelper` module?

MorningLightMountain713 (changes requested):
Much easier to follow now

ZelBack/src/services/fluxCommunication.js
MorningLightMountain713: I disagree, I think it should be done on this PR - however, you're the boss.

The reasons I think it should be done on this PR:
* The function that created the dependency cycle was created in this PR
* The appsService module is enormous - dynamically importing that will have a large memory / time cost each time one of these messages is processed
* It will take 2 minutes to move the function to another module and do it properly

Agree, we should go and fix all the other requires that are like this (in another pull)

ZelBack/src/services/appsService.js
```diff
@@ -3509,6 +3509,7 @@ async function installApplicationHard(appSpecifications, appName, isComponent, r
  * @param {object} componentSpecs Component specifications.
  * @param {object} res Response.
  * @param {boolean} test indicates if it is just to test the app install.
+ * @param {boolean} sendAppRunningMessage indicates if it should send the appRunningMessage after complete the install.
```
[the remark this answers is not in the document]
Cabecinha84: fixed

ZelBack/config/default.js
```diff
@@ -47,6 +47,7 @@ module.exports = {
         appsInformation: 'zelappsinformation', // stores actual state of flux app configuration info - initial state and its overwrites with update messages
         appsTemporaryMessages: 'zelappstemporarymessages', // storages for all flux apps messages that are not yet confirmed on the flux network
         appsLocations: 'zelappslocation', // stores location of flux apps as documents containing name, hash, ip, obtainedAt
+        appsInstallingLocations: 'appsInstallingLocations', // stores install location of flux apps as documents containing name, ip, obtainedAt
```
TheTrunk: currently we have a convention of having it all small letters. This is better for readability but just a note. I think in the future we want to make it in the camelcase all and get rid of the zel prefix

ZelBack/src/routes.js
```diff
@@ -355,6 +355,12 @@ module.exports = (app) => {
   app.get('/apps/locations', cache('30 seconds'), (req, res) => {
     appsService.getAppsLocations(req, res);
   });
+  app.get('/apps/installinglocation/:appname?', cache('30 seconds'), (req, res) => {
+    appsService.getAppInstallingLocation(req, res);
+  });
+  app.get('/apps/installlingocations', cache('30 seconds'), (req, res) => {
```
TheTrunk: typo in the path

ZelBack/src/services/appsService.js
```diff
@@ -6898,6 +6899,69 @@ async function storeAppRunningMessage(message) {
...
+  * @param broadcastedAt number
+  * @param name string
+  * @param ip string
+  */
+  if (!message || typeof message !== 'object' || typeof message.type !== 'string' || typeof message.version !== 'number'
+    || typeof message.broadcastedAt !== 'number' || typeof message.ip !== 'string' || typeof message.name !== 'string') {
+    return new Error('Invalid Flux App Installing message for storing');
+  }
+
+  if (message.version !== 1) {
```
TheTrunk, line 6973: can be more strict? But this is general issue throughout the codebase. Something for later

ZelBack/config/default.js
Cabecinha84: Done, changed the name of the collection to all lower case. Yeah, this one is new I didn't add the zel prefix on it... doesn't make sense to have prefixes.

ZelBack/src/services/appsService.js
```diff
@@ -10158,16 +10267,67 @@ async function trySpawningGlobalApplication() {
...
     if (!registerOk) {
+      broadcastedAt = Date.now();
       log.info('trySpawningGlobalApplication - Error on registerAppLocally');
-      const broadcastedAt = Date.now();
       const appRemovedMessage = {
         type: 'fluxappremoved',
         version: 1,
         appName: appSpecifications.name,
         ip: myIP,
+        cancel: 1,
```
TheTrunk: probably should not be here?

TheTrunk (changes requested):

ZelBack/src/routes.js
Cabecinha84: fixed

ZelBack/src/services/appsService.js
Cabecinha84: yeah if you thing that check should be different we should do it on a different PR and change all the places that are doing the same approach.

PR #1482, opened 2025-05-29
Improvements on app hashes sync

Cabecinha84:
Changes on this PR:
- Process in batch the inserts of transactions information on app hashes collections when explorer is not synced;
- Remove from proces explorer the ask for app hashes missing when explorer is not synced, that is done on continuousFluxAppHashesCheck, where we increase the performance of sending the max number of messages possible;
- Dont do daemon calls checking for apps transactions between block 699421 and block 862001 as there are none, to speed up the explorer sync;
- Add missing check that was being done to messages received from incoming connections to the outgoing connections;
- Fix appPricePerMonth that was missing block when some changes were activated and improve the code;
- Fix verifyAppHash to validate properly messages from apps version <=3, this was causing all new nodes syncing to have on app hashes collections these with message = false;
- Improve storeAppTemporaryMessage, method responsable for processing fluxappregister and fluxappupdate messages, we were wrongly passing the verifyAppSpecifications the current node daemon block instead of the block where the app was paid. Also VERY IMPORTANT we fix not rebroadcasting messages that the node was missing, currently if a node is asking for messages because it's syncing, when it receives it rebroadcast to entire network, this is a spamming fix;
- Improve checkAndRequestApp, we no longer do checks if the price was or not correct for apps that can no longer be running on the network, as it was used only update globalappspecifications. We are alo no longer inserting those apps that can no longer be running on the glovalappspecificaitons, so we don't need to be calling on expireglobalaplications on this method that was added here before as a bandage;
- Expire Global Aplications is now called from explorer, every two blocks when it's synced;
- Code remove of fluxAppRequestV2 that is already on the past and the checks no longer make sense;
- Increase rate limit of messages, as we can get more when node is syncing the app hashes;
- Other improvements and clean up.

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -10614,7 +10614,7 @@ async function expireGlobalApplications() {
       // eslint-disable-next-line no-await-in-loop
       await removeAppLocally(appName, null, false, true, true);
       // eslint-disable-next-line no-await-in-loop
-      await serviceHelper.delay(3 * 60 * 1000); // wait for 3 mins so we don't have more removals at the same time
+      await serviceHelper.delay(1 * 60 * 1000); // wait for 1 min
     }
   } catch (error) {
     log.error(error);
--- a/ZelBack/src/services/explorerService.js
+++ b/ZelBack/src/services/explorerService.js
@@ -565,6 +565,11 @@ async function processBlock(blockHeight, isInsightExplorer) {
     // this should run only when node is synced
     const isSynced = !(blockDataVerbose.confirmations >= 2);
     if (isSynced) {
+      if (blockHeight % 2 === 0) {
+        if (blockDataVerbose.height >= config.fluxapps.epochstart) {
+          await appsService.expireGlobalApplications();
+        }
+      }
       if (blockHeight % config.fluxapps.removeFluxAppsPeriod === 0) {
         if (blockDataVerbose.height >= config.fluxapps.epochstart) {
           appsService.checkAndRemoveApplicationInstance();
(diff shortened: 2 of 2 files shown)


MorningLightMountain713 (approved):
Nice PR description, and some good finds. Makes sense.

I've reviewed the code, looks good, but I guess the best way to validate is to get this in develop and running on some nodes (and resync etc)

Cabecinha84:
> Nice PR description, and some good finds. Makes sense.
> 
> I've reviewed the code, looks good, but I guess the best way to validate is to get this in develop and running on some nodes (and resync etc)

Was by reinstalling mongo and starting from scratch that I have found most of the things.
This node is currently running this branch for 3h, and I have wiped the db before switch.
http://89.58.52.65:16127/apps/hashes

TheTrunk:
Can we separate this huge PR into multiple small PRs?
This is a bit too complex for me to review properly. Would be nice to bring it to like small PRs, one PR one logic update

ZelBack/config/default.js
```diff
@@ -233,6 +232,7 @@ module.exports = {
     hddFileSystemMinimum: 10, // right now 10, to be decreased to a minimum of 5GB of free space on hdd for docker with v8 specs activation
     defaultSwap: 2, // 2gb swap memory minimum, this is in gb
     applyMinimumPriceOn3Instances: 1691000, // after this block we use the min. usd price on prices per 3 instances.
+    applyMinimumForExtraInstances: 1890000,
```
TheTrunk, line 235: its in the past - correct right?

TheTrunk (approved):

ZelBack/config/default.js
Cabecinha84: yes, already went live some time ago but I forgot to add it.

PR #1484, opened 2025-05-31
Improvements for v8 app specs

Cabecinha84:
List of changes:
- Add now property called arcaneSender to fluxappregister and fluxappupdate types that identifies if the node sending the message it's Arcane Node;
- Improve code to remove duplicated code on storeAppTemporaryMessage;
- If message is v8 enteprise check if sender is ArcaneOs if not do not accept message;
- If message is v8 enteprise, validate the app if the node receiving is ArcaneOS;
- Fix we were storing and sending v8 entperprise apps decrypted instead of encrypted;
- Fix decrypting app specs on try spawn global apps;
- Other fixes and code improvements.

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -9897,7 +9897,7 @@ async function trySpawningGlobalApplication() {
 
       log.info(`trySpawningGlobalApplication - Application ${appToRun} selected to try to spawn. Reported as been running in ${appToRunAux.actual} instances and ${appToRunAux.required} are required.`);
       if (appToRunAux.enterprise && !isArcane) {
-        log.info('trySpawningGlobalApplication - app missing one instance failed the 15% probability check to install');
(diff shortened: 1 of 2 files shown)


ZelBack/src/services/appsService.js
```diff
@@ -7837,9 +7855,8 @@ async function updateAppGlobalyApi(req, res) {
         hash: messageHASH,
         timestamp,
         signature,
+        arcaneSender: isEnterpriseApp,
```
TheTrunk, line 7858: I think we should put some special key into bench/sas and arcane node should be signing the message with the key and attach signature. 
As this approach can be still spoofed by standard nodes - the receiving node must pretty much verify that the sender is indeed arcane and that can only be done if the sender is signing a message signed by bench/sas

Cabecinha84: Edited...

TheTrunk (approved):

PR #1485, opened 2025-05-31
Get all app messages on one shot after explorer is synced for first time

Cabecinha84:
List of changes:
- Create new API to return if node explorer is synced;
- After explorer is synced for the first time on the continuousFluxAppHashesCheck function we call the new method checkAndSyncAppHashes;
- The new method will be executed only once;
- checkAndSyncAppHashes function, checks if we are missing over 95% of the app messages, if that the case we random connect to a outgoing peer and check if it has the explorer synced, if it is synced, we get all permanent app messages from it. After we order them by height and start processing them. The process is the same as receiving the message like it was sent over websocket, we call the method that does the validations and stores them on the temporary app messages and after we call the method that checks if that message was paid and moves it to permanent app messages and inserts it on globalapps if it is supposed to.

What we are doing here, is if the node is new, or mongodb was cleared, and it was syncing for the first time, instead of being spamming the network with messages, we get all the messages at once from one of the peer nodes, and process them. This should reduce a lot the bandwidth use and cpu usage across the network.

This is executed only once, after it starts using the normal mechanism.

This is a update over the PR done one the feature/syncmongo.

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -8911,7 +8911,7 @@ async function checkAndSyncAppHashes() {
           // eslint-disable-next-line no-continue
           continue;
         } */
-        log.info(`checkAndSyncAppHashes - Explorer is not synced on ${client.ip}:${client.port}`);
+        log.info(`checkAndSyncAppHashes - Explorer is synced on ${client.ip}:${client.port}`);
         axiosConfig = {
           timeout: 120000,
         };
@@ -8946,6 +8946,7 @@ async function checkAndSyncAppHashes() {
           }
         }
         finished = true;
+        log.info('checkAndSyncAppHashes - Process finished');
       }
     }
     checkAndSyncAppHashesWasEverExecuted = true;
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -8898,7 +8898,7 @@ async function checkAndSyncAppHashes() {
         let axiosConfig = {
           timeout: 5000,
         };
-        log.info(`checkAndSyncAppHashes - Getting explorer sync status from ${client.ip}:${client.port}`);
+        /* log.info(`checkAndSyncAppHashes - Getting explorer sync status from ${client.ip}:${client.port}`);
         // eslint-disable-next-line no-await-in-loop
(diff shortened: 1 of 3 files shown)


alihm:
Question, what happens if new messages arrive during this process? Like if we receive appremoved during or before the process, is it going to be applied after the process is done?

Cabecinha84:
> Question, what happens if new messages arrive during this process? Like if we receive appremoved during or before the process, is it going to be applied after the process is done?

All messages will be processed normally.

This processing is for to know, process and store on his database all app messages that existed. This is what in the end makes the node at the end know what apps are currently live on the network and show them on global app specifications.

Currently a new node joining the network takes around 8h to sync the explorer and after one to two days to get all app messages. This second part will now take seconds to retrieve with the api call and after about 1h30 to process all the messages.

App removed, app install messages are to store info on a different database, that shows where are all apps currently running.

Cheers

Cabecinha84:
image
image

ZelBack/src/services/appsService.js
```diff
@@ -8871,6 +8871,93 @@ async function rescanGlobalAppsInformationAPI(req, res) {
...
+      let i = 0;
+      while (!finished && i <= 5) {
+        i += 1;
+        const client = outgoingPeers[Math.floor(Math.random() * outgoingPeers.length)];
+        let axiosConfig = {
+          timeout: 5000,
+        };
+        log.info(`checkAndSyncAppHashes - Getting explorer sync status from ${client.ip}:${client.port}`);
+        // eslint-disable-next-line no-await-in-loop
+        const response = await serviceHelper.axiosGet(`http://${client.ip}:${client.port}/explorer/sync`, axiosConfig).catch((error) => log.error(error));
```
TheTrunk: likely should be the new endpoint `issynced` instead of `sync`

TheTrunk (changes requested):
`issynced` vs `sync` other than that ack

ZelBack/src/services/appsService.js
Cabecinha84: yeah I ahve changed after the name of the router and forgot to update it on the code. Nice catch, thanks. It's fixed

PR #1487, opened 2025-06-06
Adds more security for transmitting and receiving enterprise apps from UI

Cabecinha84:
Encrypted AES key format: (read right to left)

```
+--------+---------------------+---------+---------------+
| base64 | RSA Encrypted Bytes | base64  | AES Key Bytes |
+--------+---------------------+---------+---------------+
```

Enterprise field format:

```
+--------+---------------------------------------------------+
| base64 | Encrypted AES Key + IV + Cyphertext + Tag (bytes) |
+--------+---------------------------------------------------+
```

`Enterprise specs` (what is encrypted as the Cyphertext above) is defined as the json string of `{compose, contacts}` from the specs

There are a couple of things to note:
* The specs are secured by RSA - not AES. Even though the specs are encrypted with AES, the AES key is encrypted with RSA, and stored with the specs. (V9 will use full AES without storing the session key)
* It is the users responsibilty to dispose of the session key once the specs are encrypted. (If they choose) As this key will allways be able to decrypt the specs. (This will change in v9, once the specs are encrypted, the session key is unuseable)

On this PR The flow changes to this:
* Frontend requests RSA 2048 public key
* Frontend users RSA public key to encrypt the base64 encoded aes key
* Frontend uses a random `aes-256-gcm` key to encrypt the enterprise specs. (one time use) 
* Frontend concats the aes key + iv + cyphertext + tag then base64 encodes them all
* This base64 encoded blob is the "enterprise" key on the specs

Change methods on the backend:
* verifyAppRegistrationParameters
* verifyAppUpdateParameters
* registerAppGlobalyApi
* updateAppGlobalyApi
* getAppFiatAndFluxPrice

Modified methods on backend:
* decryptWithAesSession
* encryptWithAesSession
* decryptEnterpriseFromSession
* encryptEnterpriseWithAes (used in v9 specs)

New api endpoint - `/apps/appspecifications/:appname/update`. This was split out from the base endpoint as it simplifies the code greatly - we're using the routes to do the work instead of adding the logic in the code. This is to update an app specs to latest. This function is also futureproofed - you can pass in the `Enterprise-Key` header and update from v8 to v8+ (if enterprise). Note - this function just returns an updated app spec, it does not store it in the db etc. It's up to the frontend to then submit the new app specs.

New api endpoint `/apps/latestspecificationversion` - so clients can know what the latest spec version is (to check if they need to update specs)

- The changes are only for enterprise apps;

Change api method getApplicationSpecificationAPI to get enteprise app specs:
- Call api getpublickey that will return a RSA public key;
- Create a random AES 'AES-GCM', and encrypt the aes key information with the RSA public key;
- The AES object information encrypted with the RSA needs to be submited to the backendon req.headers;
- The specifications enteprise property will be returned encrypted with the AES key information passed to the backend so UI can decrypt it.

Fix get application price api that would not work for enterprise apps.

Other small fixes.

The following methods have been tested (up to at least the point the temp message would be broadcast):

verifyAppRegistrationParameters
verifyAppUpdateParameters
registerAppGlobalyApi
updateAppGlobalyApi
getAppFiatAndFluxPrice


MorningLightMountain713:
I'm going to make some changes.

The aes key needs to be encoded / decoded as base64

Cabecinha84:
> I've got a working POC to simplify the enterprise frontend / backend encryption. Means we don't have to pass the iv around. Still uses aes-256-gcm, however, uses webcrypto for the frontend and the standard crypto api for the backend.
> 
> Before the data is encoded, we insert the iv, so base64(bytes(iv + cyphertext + tag))
> 
> I'll update the code tomorrow

Perfect, thank you David!

MorningLightMountain713:
I just realised there is a bit of a flaw with the way this pull works.

So currently, version 7 and below, you don't need to call `verifyAppRegistrationParameters` before `regiaterAppGloballyApi` - if you want, you can just call` regiaterAppGloballyApi` directly.

However, due to the way that we are returning the backend encrypted specs from `verifyAppRegistrationParameters`, instead of the frontend encrypted, you now are forced to call  `verifyAppRegistrationParameters` in order to register an app.

The `verifyAppRegistrationParameters`. should return the frontend encrypted specs, then the frontend can register with those (or build them without calling verify first) and the `regiaterAppGloballyApi` method should rencrypt these for storage. 

That way it's the same as it is now

ZelBack/src/routes.js
```diff
@@ -337,7 +337,13 @@ module.exports = (app) => {
...
     appsService.getGlobalAppsSpecifications(req, res);
   });
-  app.get('/apps/appspecifications/:appname?/:update?/:decrypt?', cache('30 seconds'), (req, res) => {
+  app.get('/apps/latestspecificationversion', cache('5 minutes'), (req, res) => {
+    appsService.getlatestApplicationSpecificationAPI(req, res);
+  });
+  app.get('/apps/appspecifications/:appname/update', cache('30 seconds'), (req, res) => {
+    appsService.updateApplicationSpecificationAPI(req, res);
+  });
+  app.get('/apps/appspecifications/:appname/:decrypt?', cache('30 seconds'), (req, res) => {
```
TheTrunk, line 346: These path changes is a very big change. It is although very likely that it is unused. However we should not be modifying paths so quickly and without lots of notice and giving backwards support just by adding the function there too.

Further /apps/appspecifications/:appname/update is not something I would see oftenly. Usually adjustable parameters are at the end so /apps/updatetolatestspces/:appname makes more sense or something like that

Cabecinha84: The update was introduced on latest FluxOs and it's not yet on documentation. Not a issue to put it on a dedicated api call. Regarding the url for it, whatever you guys thinks make sense.

MorningLightMountain713: Here is a good article on routes: https://guides.rubyonrails.org/routing.html

Nested routes are quite common

PR #1488, opened 2025-06-09
Fix Decentralized login

Cabecinha84:
The PR fix the logins on the flux decentralized network. You can login on any node without first had to call loginphrase method on that specific node, if the message sent on login is valid and it's properly signed the node receiving the login will accept it.

### DIFF
--- a/ZelBack/src/services/idService.js
+++ b/ZelBack/src/services/idService.js
@@ -88,98 +88,101 @@ async function confirmNodeTierHardware() {
 }
 
 /**
+ * To return the message that should be signed if all is ok.
+ * @param {object} res Response.
+ * @returns {string} message that should be signed to login.
+ */
+let firstLoginPhraseExecution = true;
+async function getMessageToLogin(res, message = null) {
+  // check db
+  const db = dbHelper.databaseConnection();
+  const database = db.db(config.database.local.database);
+  const collection = config.database.local.collections.activeLoginPhrases;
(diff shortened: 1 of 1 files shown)


ZelBack/src/services/idService.js
```diff
@@ -279,7 +282,14 @@ async function verifyLogin(req, res) {
       const collection = config.database.local.collections.activeLoginPhrases;
       const query = { loginPhrase: message };
       const projection = {};
-      const result = await dbHelper.findOneInDatabase(database, collection, query, projection);
+      let result = await dbHelper.findOneInDatabase(database, collection, query, projection);
+      if (!result) {
+        const messageToSign = await getMessageToLogin(res, message);
+        if (!messageToSign) {
+          return;
```
TheTrunk, line 289: there should be some throw instead of return I guess

TheTrunk (changes requested):
This is not a good practice at all. This is basically getting rid of any verification and anyone can pretend that they now own some address. Anyone now knowing some combination of address/messsage/signature can easily pretend it - and since we are signing app specs its super easy to get too and node ops can see all the logging attempts, logged users what they signed. Further in transmission we are sending it as well

For login purposes, message should always be used just once and only on that particular node and only for this one time. Ideally any request thats being made should always come with new unique message and signing process (which is quite inconvenient). 

Conceptually I can't approve it as this would create flaws in the design and lessen the authentication purpose.

PR #1489, opened 2025-06-10
Detect broken apps on the network

Cabecinha84:
Currently if one application is broken and can't install/run, we have the entire network non stop trying to install the app, what causes lots of resources used lost.
This PR prevents that.
If one application fails to install/start, it broadcasts a new message to the network, fluxappinstalling error, with the error that happen.
If the nodes receive more than 5 errors in one hour, the messages are no longer deleted.
If one node broadcast app running for that app name and app hash, the information from that app is deleted from the app installing errors collection.

List of changes:
- Routes
-- Add two new api's to get app installing errors information;

- AppServices
-- Add a try catch specific around the call of installApplicationHard on register registerAppLocally, if there is a error send new message type fluxappinstallingerror to the network;
-- Changed trySpawningGlobalAplicationm on first execution, we call the method getPeerAppsInstallingErrorMessages, that will get from a peer with higher uptime the db info from the app installing error locations;
-- Method to store the new app message type and methods to give support to the new api's;
-- Changed updateAppSpecifications and expireGlobalAplications to delete info from app from the new db with errors;

- dbHelpder
-- Added bulkWriteInDatabase for bulk operations;
-- Added countInDatabase to get the count of documents on a collection.

- fluxCommunications
-- Added method handleAppInstallingErrorMessage and adapt other methods to process the new message type.

- fluxCommunicationUtils
-- Added a check to validate that the node sending the fluxappinstallingerror message match the ip sent on the message.

- serviceManager
-- Inserts new indexes on nodes

### DIFF
--- a/ZelBack/src/routes.js
+++ b/ZelBack/src/routes.js
@@ -358,13 +358,13 @@ module.exports = (app) => {
   app.get('/apps/installinglocation/:appname?', cache('30 seconds'), (req, res) => {
     appsService.getAppInstallingLocation(req, res);
   });
-  app.get('/apps/installlinglocations', cache('30 seconds'), (req, res) => {
+  app.get('/apps/installinglocations', cache('30 seconds'), (req, res) => {
     appsService.getAppsInstallingLocations(req, res);
   });
   app.get('/apps/installingerrorslocation/:appname?', cache('30 seconds'), (req, res) => {
     appsService.getAppInstallingErrorsLocation(req, res);
   });
-  app.get('/apps/installlingerrorslocations', cache('30 seconds'), (req, res) => {
+  app.get('/apps/installingerrorslocations', cache('30 seconds'), (req, res) => {
     appsService.getAppsInstallingErrorsLocations(req, res);
   });
   app.post('/apps/calculateprice', (req, res) => { // returns price in flux for both new registration of app and update of app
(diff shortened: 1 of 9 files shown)


ZelBack/src/routes.js
```diff
@@ -361,6 +361,12 @@ module.exports = (app) => {
   app.get('/apps/installlinglocations', cache('30 seconds'), (req, res) => {
     appsService.getAppsInstallingLocations(req, res);
   });
+  app.get('/apps/installingerrorslocation/:appname?', cache('30 seconds'), (req, res) => {
+    appsService.getAppInstallingErrorsLocation(req, res);
+  });
+  app.get('/apps/installlingerrorslocations', cache('30 seconds'), (req, res) => {
```
TheTrunk: typo and also in more places we have triple l

TheTrunk (changes requested):
I think we need to be very cautios about side effects. Node ops can send these malicious messages effectively stopping someone normally working app from running, installing. 
I think this carries a lot of potential risks and side effects.
Would be more better to not react to these messages at all. Just store them and observe how the network behaves, get the statistics, the data first, analyze it and come up how to handle it in next version. This is much safer approach and would give us valuable data insight to the network. Should also build internally some statistics/metrics/graphs on stats

ZelBack/src/routes.js
Cabecinha84: fixed

Cabecinha84:
Updated PR Topic and description

PR #1491, opened 2025-06-12
Brings v8 app specs to UI

Cabecinha84:
This PR brings v8 app specs to home.runonflux.io.

List of changes:
- App registrations;
- App management, registration and get app specifications decrypted if it is enterprise app.

Build will only be done on development.


TheTrunk (approved):
utack

PR #1492, opened 2025-06-12
Improve network issues detection on the nodes

Cabecinha84:
We have find out there are several nodes on the network configured with the original port forwarding configuration only allowing apps ports from range 30k to 40k.
This PR fix that by improving the checkMyAppsAvailability() function;

List of changes:
- On first run execute the test on port 10000;
- When a test fails, next we starting testing last failing port +1 until we have a success test, after we start testing original test failed -1 until success;
- After we have 100 ports failed the remaining we start randomly testing one of the 100 ports that have failed;
- Decrease the time between tests when it's failing and the amount of ports that have failed so far is bellow 100;
- Change DOS message for easy understand of the problem for the node operator.

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -13414,19 +13414,15 @@ async function checkMyAppsAvailability() {
     }
 
     log.info(`checkMyAppsAvailability - Testing port ${testingPort}.`);
-    const portNotWorking = portsNotWorking.includes(testingPort);
-    if (portNotWorking) {
-      log.info(`checkMyAppsAvailability - Testing port ${testingPort} is part of the list of ports not working on this node.`);
-      failedPort = null;
-      // skip this check, port is not possible to run on flux
-      await serviceHelper.delay(15 * 1000);
-      checkMyAppsAvailability();
-      return;
-    }
     let iBP = fluxNetworkHelper.isPortBanned(testingPort);
     if (iBP) {
       log.info(`checkMyAppsAvailability - Testing port ${testingPort} is banned.`);
       failedPort = null;
+      if (originalPortFailed && testingPort > originalPortFailed) {
+        portToTest = originalPortFailed - 1;
+      } else if (originalPortFailed) {
+        portToTest = null;
+      }
       // skip this check, port is not possible to run on flux
       await serviceHelper.delay(15 * 1000);
       checkMyAppsAvailability();
@@ -13437,6 +13433,11 @@ async function checkMyAppsAvailability() {
       if (iBP) {
         log.info(`checkMyAppsAvailability - Testing port ${testingPort} is UPNP banned.`);
         failedPort = null;
+        if (originalPortFailed && testingPort > originalPortFailed) {
+          portToTest = originalPortFailed - 1;
+        } else if (originalPortFailed) {
+          portToTest = null;
+        }
         // skip this check, port is not possible to run on flux
         await serviceHelper.delay(15 * 1000);
         checkMyAppsAvailability();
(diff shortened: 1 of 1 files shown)


MorningLightMountain713:
I've reviewed, but thought it easier to just write a comment
 
* We can just set testing port to 10k initially, and remove checkMyAppsAvailabilityFirstRun
* We are pushing "not working" ports to portsNotWorking. At no time are they ever retried if one fails. A port can fail due to the remote end. The portsNotWorking is also never cleared. So it's entirely possible (however unlikely) that we could fill up the portsNotWorking and it not be due to an error with this node.
* If a port fails - we should immediately verify if it is a local failure by trying another couple of nodes for consensus. (Total 3 nodes)
* All of the `portToTest = originalPortFailed - 1`, this doesn't take into account if we are incrementing up or down

Just want to make sure I understand the logic / intent here, as it is quite complex

* If node is not synced / confirmed / no public ip wait 4 minutes and try again
* We test starting at port 10k
* We then test one port per hour, incrementally (line 13599) unless I'm reading this wrong, but it seems excessively high?
* If a port fails (and less than 100 have failed in total) we start testing ports every 15 seconds
* If 25 ports fail, (i.e. we have incremented 0.4 each port and dosstate = 10), we set the dos message. However, on line 13607, if the portsnotworking are less than 100, we reset the dosstate back to 0, and remove the dosMessage
* If 100 total ports have failed, we set the dos message, and wait one minute 

I think it would be easier / error prone, and more testable, to determine the intent - then just rewrite this function entirely. If we want to make things testable, we should stop using recursive functions.(I appreciate that this function has been recursive before this PR) Recursive functions serve a purpose, but the way we are using them is just a basic loop, we should make the checkMyAppsAvailability function return a number (i.e. how long to delay for), and another function runs it in a loop. Then testing becomes a lot more simple, as you're just testing one iteration.

If we agree on the intent, I'm happy to do a rewrite and write tests for it, or we can work with what we have ¯\\_(ツ)_/¯

MorningLightMountain713:
I thought of a good algorithm to test ports. It's very simple

We're only testing that that the node operator has made the full range of ports available. We don't actually need to test all ports to determine this; as I believe the ports used for an application are validated at install (if they aren't, they should be)

We split the privileged ports 1-1023 and the rest 1024-65535. ISPs (generally) only block ports in the priviledged range, I.e. 445, 137 etc.

We then test test groups of ports seperately.

For the 1024-65535 range:

* Select a random port, then test it
* We then sample ports - i.e. test every thousandth port. So you add 1000 to the random port, then modulo with 65535. We then test that port, add 1000, modulo with 65535, and continue that in a loop for a max of 65 times.
* If a port fails, we test the port below and the port above. If all 3 fail, we start a "failure" counter.
* The reason we test the one below and above, is that it's a pretty reasonable assumption that if 3 ports in a row aren't working, none in that block of 1000 are.
* If there are 5 "failures" in a row (or some other arbitrary threshold, maybe even 3) - we have determined that ports aren't useable, and mark the node as DOS.
* The likelyhood of 5 failures in a row happening due to ISPs etc blocking ports is 0. The only way this could happen is if the ports aren't useable.
* If there are 3 (or another number below the threshold) failures in a row, and then success, we clear the failure counter back to 0. So it will only trigger if there are <threshold> failures, anything less is fine.
* Once the loop has finished (reached 65 iterations) We select another random port - then start again.

Doing this - we don't need to keep track of failed ports. (or, we keep track of them, and add them to excluded ports, so we don't try to install apps on them)

For the 1-1023 range.

I believe this range is only available for enterprise apps, therefore only ArcaneOS nodes would need to test this range. If this is not the case, then obviously everyone would need to test.

* Select a random port, then test it
* We then sample every 10th port. So add 10 to the random port, then modulo 1023.
* If it fails, we add it to the failed ports list.
* If we cross a threshold for failed ports, like 25-50, we mark the node DOS.
* Since we are starting with a random port, they will all get tested eventually.

By using a modulo, we only have to loop one way, and can start on a random port, and still test the whole range.

This would also be much faster to determine if a port range isn't available, as we are using samples.

Cabecinha84:
The "1-1023 range" is available for any app, not only enterprise, you just pay extra if your app uses them.

TheTrunk (approved):

PR #1493, opened 2025-06-17
v6.1.0

Cabecinha84:
- Brings v8 app specs to UI, the new version will appear automatic on application registration page on flux block 1932380;
- Adds more security for transmitting and receiving enterprise apps from UI (more information below);
- Improve network issues detection on the nodes, DMZ or UPNP requirement is mandatory; Node operators can specify up to 100 ports that they don't want FluxOS Gravity to use. If FluxOS detects other 100 ports not reachable from outside your node gets DOS;
- Detect broken apps on the network, the network will now transmit to other nodes if one aplication is failing to install or start and it was not supposed to, Flux Team can easily act this way over to block that app;
- Other small improvements.

Regarding enterprise apps, what was changed:

Frontend requests RSA 2048 public key
Frontend users RSA public key to encrypt the base64 encoded aes key
Frontend uses a random aes-256-gcm key to encrypt the enterprise specs. (one time use)
Frontend concats the aes key + iv + cyphertext + tag then base64 encodes them all
This base64 encoded blob is the "enterprise" key on the specs

Encrypted AES key format: (read right to left)

+--------+---------------------+---------+---------------+
| base64 | RSA Encrypted Bytes | base64  | AES Key Bytes |
+--------+---------------------+---------+---------------+
Enterprise field format:

+--------+---------------------------------------------------+
| base64 | Encrypted AES Key + IV + Cyphertext + Tag (bytes) |
+--------+---------------------------------------------------+
Enterprise specs (what is encrypted as the Cyphertext above) is defined as the json string of {compose, contacts} from the specs

There are a couple of things to note:

The specs are secured by RSA - not AES. Even though the specs are encrypted with AES, the AES key is encrypted with RSA.
It is the users responsibilty to dispose of the session key once the specs are encrypted.

You can check the changes on UI, to learn how to interact over API with enterprise apps.


MorningLightMountain713 (approved):
Ack

alihm (approved):
Ack

PR #1495, opened 2025-06-24
Feature/rework application port availability

MorningLightMountain713:
This PR reworks the `checkMyAppsAvailability` function.

Scope is just bugfix, readability improvements, and code commenting (so we can understand what is happening, and why we're doing it)

I isolated the function and tested against a dev http server acting as the `askingIP`. I am reasonably confident I have caught all the edge cases - but can't guarantee this as we don't have tests. This should be enough to get us validating ports again (fingers crossed).

As part of a future PR - I will do  up a scope document with architecture and algorithm design, where we can build a more robust / simple / faster / safer strategy for testing ports. This document will also clearly define the failure states, what determines a "failed port" and strategies for faster port group evaluation.

Once the design is complete, we can break out this functionality into it's own module, and use OOP concepts to contain this functionality; this will make building a test suite simple.

Some of the highlights:
* Fixes memory leak where we were not removing the event listeners off the http test server
* Moves the shutdown to a function so this can be more easily managed (and less code repetition)
* Moves the next port calculation to an arrow function to remove a bunch of duplicate code
* We now shutdown ALL test services on every path. This was not happening prior
* Removed the recursive loop inside the error handler - this was causing multiple instances to run
* Remove a bunch of the "global" throw / catches. It's better to deal with these up front - makes the control flow easier to follow. 
* Move the `portsNotWorking` to a set - more suitable to what we are doing / easier to use
* Fixes where we were setting the `portsNotWorking` to the failedPort - which was never set (always null)
* At no point will we recurse without having waited some time period. I.e. 15 seconds, 1 minute, 2 minutes etc

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -14455,8 +14455,15 @@ async function checkMyAppsAvailability() {
         setPortToTest = null;
         originalPortFailed = null;
       }
+    } else {
+      // we shouldn't get here, but just in case we clean up and retry
+      await handleTestShutdown(testingPort, {
+        skipFirewall: !firewallActive,
+        skipUpnp: !isUpnp,
+      });
+      await serviceHelper.delay(2 * 60 * 1000);
+      checkMyAppsAvailability();
     }
-    // there is no else statement here, there should be
 
     if (dosState > 10) {
       dosMessage = `Ports tested not reachable from outside, DMZ or UPNP required! All ports that have failed: ${JSON.stringify(
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -14040,16 +14040,65 @@ async function callOtherNodeToKeepUpnpPortsOpen() {
 }
 
 /**
+ *
+ * @param {*} testingPort The target port
+ * @param {{skipFirewall?: Boolean, skipUpnp?: Boolean, skipHttpServer?: Boolean}} options Options
+ */
+async function handleTestShutdown(testingPort, options = {}) {
+  const skipFirewall = options.skipFirewall || false;
+  const skipUpnp = options.skipUpnp || false;
+  const skipHttpServer = options.skipHttpServer || false;
+
+  const updateFirewall = skipFirewall ? false : await fluxNetworkHelper
+    .isFirewallActive()
+    .catch((e) => log.error(e));
+
+  if (updateFirewall) {
+    await fluxNetworkHelper
+      .deleteAllowPortRule(testingPort)
+      .catch((e) => log.error(e));
+  }
+
+  if (!skipUpnp) {
+    await upnpService
+      .removeMapUpnpPort(testingPort, 'Flux_Test_App')
+      .catch((e) => log.error(e));
+  }
+
+  if (!skipHttpServer) {
+    testingAppserver.removeAllListeners();
+    testingAppserver.shutdown((err) => {
+      if (err) {
+        log.error(`testingAppserver shutdown failed: ${err.message}`);
+      }
+    });
+  }
+}
+
+// any function that uses module level globals to keep state should be moved to
+//  a class.
+
+/**
  * Periodically check for our applications port range is available
-*/
-let failedPort;
-let testingPort;
-const portsNotWorking = [];
-let originalPortFailed;
-let setPortToTest = Math.floor(Math.random() * (25000 - 10000 + 1)) + 10000;
+ */
+// changed this to a set. Means we don't have to worry about duplicates etc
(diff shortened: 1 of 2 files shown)


ZelBack/src/services/appsService.js
```diff
@@ -14203,49 +14284,92 @@ async function checkMyAppsAvailability() {
...
     }
+
     if (myIP === askingIP) {
+      await handleTestShutdown(testingPort, {
+        skipFirewall: !firewallActive,
+        skipUpnp: !isUpnp,
+      });
+      // safer just to wait here, no harm in waiiting 15 seconds in unlikely event
+      // that we pull ourselves from the list (or we should just remove it first)
+      await serviceHelper.delay(15 * 60 * 1000);
```
Cabecinha84: this is awaiting 15 minutes, not 15 seconds

ZelBack/src/services/appsService.js
```diff
@@ -14203,49 +14284,92 @@ async function checkMyAppsAvailability() {
...
       return;
     }
+
     if (failedNodesTestPortsCache.has(askingIP)) {
+      await handleTestShutdown(testingPort, {
+        skipFirewall: !firewallActive,
+        skipUpnp: !isUpnp,
+      });
+      // same as above. This is unlikley, just wait the 15 seconds
+      await serviceHelper.delay(15 * 60 * 1000);
```
Cabecinha84: this is awaiting 15 minutes, not 15 seconds

ZelBack/src/services/appsService.js
```diff
@@ -14254,101 +14378,151 @@ async function checkMyAppsAvailability() {
...
+        && originalPortFailed >= testingPort
+        && originalPortFailed - 1 > 0
+      ) {
         setPortToTest = originalPortFailed - 1;
       } else {
         setPortToTest = null;
+        originalPortFailed = null;
       }
     }
+    // there is no else statement here, there should be
```
Cabecinha84: this comment is strange :)

ZelBack/src/services/appsService.js
MorningLightMountain713: lol. Yes. Will check it out tomorrow

ZelBack/src/services/appsService.js
MorningLightMountain713: Ha - yes, I will put an else in there and remove the comment

ZelBack/src/services/appsService.js
MorningLightMountain713: Will check out tomorrow

MorningLightMountain713: Good find, have updated to 15 seconds

ZelBack/src/services/appsService.js
MorningLightMountain713: Updated to 15 seconds

ZelBack/src/services/appsService.js
MorningLightMountain713: Added safety else statement

ZelBack/src/services/appsService.js
```diff
@@ -14254,101 +14378,158 @@ async function checkMyAppsAvailability() {
...
+        JSON.stringify(data),
+        axiosConfig,
+      )
+      .catch(() => {
+        log.error(
+          `checkMyAppsAvailability - ${askingIP} for app availability is not reachable`,
+        );
+        setPortToTest = testingPort;
+        failedNodesTestPortsCache.set(askingIP, askingIP);
+        return null;
```
Cabecinha84, line 14399: if we return null this will stop be executing, we need to call the function again.

Cabecinha84 (changes requested):
asked for a small fix

ZelBack/src/services/appsService.js
MorningLightMountain713: That is the return value from the error handler, it doesn't return from the `checkMyAppsAvailability`. It sets the `resMyAppAvailability` value to `null`

ZelBack/src/services/appsService.js
```diff
@@ -14254,101 +14378,158 @@ async function checkMyAppsAvailability() {
...
+
       portTestFailed = true;
       dosState += 0.4;
+
+      // this is what actually sets the port to test on the next iteration (if
+      // no broken ports have been found)
       if (!originalPortFailed) {
         originalPortFailed = testingPort;
-      } else if (testingPort >= originalPortFailed && testingPort + 1 <= 65535) {
+        setPortToTest = testingPort + 1;
```
Cabecinha84: if originalPortFailed is 65535 we will have issues here with +1

alihm (approved):
Ack

PR #1500, opened 2025-06-26
Improvements on explorer processing

Cabecinha84:
In this PR we are reducing the daemon calls that are heavy and can lead to daemon to crash.

We stop using the daemonServiceControlRpcs.getInfo(), and use the height from daemonServiceMiscRpcs.isDaemonSynced(). On this last method we are getting the value from the cache.

Also fixed this error throw new Error(daemonGetChainTips.data.message || daemonGetInfo.data);
where daemonGetInfo was wrongly there as it is not what is being checked/parser but the daemonGetChainTips instead.


MorningLightMountain713:
Updates look good

XK4MiLX (approved):

TheTrunk (approved):

PR #1501, opened 2025-06-26
Bugfix: pre 6.1.0 v7 signatures unable to be verified

MorningLightMountain713:
Version 6.1.0 introduced a subtle bug where the order of the specformatter options for secrets and repoauth got transposed. This means the signatures that are stored in the db were different from the data that was trying to be verified.

So now, for v7 specs, anything after 6.1.0 has the specs formatted a different way from pre 6.1.0.

This means that any node syncing after 6.1.0 is missing around 17200 app messages.

This pull retries with the specs formatted in the old way if it fails on the first attempt. Obviously this isn't great - so would be good to get all specs migrated to v8+ in the future so we can remove this hack.

In the future we could look at recursive alphanumeric sorting of keys so that we can avoid storing / formatting the specs in a particular order

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -4949,11 +4949,11 @@ async function verifyAppMessageUpdateSignature(type, version, appSpec, timestamp
       // try the old secrets / repoauth
       comp.secrets = secrets;
       comp.repoauth = repoauth;
-
-      const messageToVerifyC = type + version + JSON.stringify(appSpecsClone) + timestamp;
-      // we can just use the btc / eth verifier as v7 specs came out at 1688749251
-      isValidSignature = signatureVerifier.verifySignature(messageToVerifyC, appSpec.owner, signature);
     });
+
+    const messageToVerifyC = type + version + JSON.stringify(appSpecsClone) + timestamp;
+    // we can just use the btc / eth verifier as v7 specs came out at 1688749251
+    isValidSignature = signatureVerifier.verifySignature(messageToVerifyC, appSpec.owner, signature);
   }
   if (isValidSignature !== true) {
     log.debug(`${messageToVerify}, ${appOwner}, ${signature}`);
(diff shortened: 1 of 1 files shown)


ZelBack/src/services/appsService.js
```diff
@@ -4838,6 +4838,26 @@ async function verifyAppMessageSignature(type, version, appSpec, timestamp, sign
     if (timestamp > 1688947200000) {
       isValidSignature = signatureVerifier.verifySignature(messageToVerifyB, appSpec.owner, signature); // btc, eth
     }
+    // fix for repoauth / secrets order change for apps created after 1750273721000
+  } else if (isValidSignature !== true && appSpec.version === 7) {
```
Cabecinha84, line 4842: I think we should put version bellow 8, and check if the app specs have compose property. because we still have lots of apps on the network with older versions that have compose, so some might also need to be checked by this version.

ZelBack/src/services/appsService.js
```diff
@@ -4914,6 +4934,26 @@ async function verifyAppMessageUpdateSignature(type, version, appSpec, timestamp
     if (isValidSignature !== true && marketplaceApp) {
       isValidSignature = signatureVerifier.verifySignature(messageToVerifyB, fluxSupportTeamFluxID, signature); // btc, eth
     }
+    // fix for repoauth / secrets order change for apps created after 1750273721000
+  } else if (isValidSignature !== true && appSpec.version === 7) {
```
Cabecinha84, line 4938: Same here.

Cabecinha84 (changes requested):
I think this might not cover all messages. Asked for some changes.

Cabecinha84:
After writing the review I remembered, repoatuh and secrets are version 7 specific so the code is good

Cabecinha84 (approved):
ack

ZelBack/src/services/appsService.js
MorningLightMountain713: Isn't it only v7 that has secrets though?

MorningLightMountain713: I see your other comment

TheTrunk (approved):

PR #1502, opened 2025-06-27
Remove allow of inbound 53, reallow ssh (if apps configured)

MorningLightMountain713:
This PR is to mitigate the misterous operation that is removing ssh access in some situations.

We also remove the inbound 53 rule - this is quite a security risk, we should not be allowing inbound dns. This is how DNS Amplification attacks happen. (If people are running open resolvers on 53/udp) There is never a good reason to open 53 inbound (unless you are running public dns)

Just adds back the OpenSSH application (and the FluxadmSSH application)

Note - if nodes don't have the OpenSSH app (it is enabled by default) then this operation will fail silently.

Edit - I've tested this on a couple of Arcane nodes - works fine.


Cabecinha84 (approved):
ack

PR #1504, opened 2025-06-28
Fix Gravity api content types

MorningLightMountain713:
Currently, our express api is quite broken. You should be able to set a `content-type` header of either `application/json` or `application/x-www-form-urlencoded` and the data should be parsed as stated.

Our api does not work like that at all. Different endpoints act differently. 

Half the post endpoints are using the old event listener method of listening for data events on the request, and half are using the `express.json()` middleware.

This puts us in quite a bad spot. There are several things going on here which I'll explain.

### Hanging route handlers

All the event listener post methods, for example `/flux/checkappavailability'` - if you set a `Content-Type: applcation/json` header - they will hang indefinetely. This is because the body is parsed by the `express.json()` middleware before the route handler attaches it's listeners. So it sits there forever listening for data that has already been received.

### Incorrect content-type

Basically, we only use the `express.json()` middleware, when we should also be using the `express.urlencoded()` middleware. So what is happening, when an `application/x-www-form-urlencoded` content-type is set, and sent to one of the old style handlers, the data is getting parsed as a string, then we manually convert that to an object with the `ensureObject` method.

The problem with this is that axios uses urlencoded as a default. So in the frontend, we are stringifying the data and sending as urlencoded, which is totally wrong.

### How this PR addresses the problem.

This is messy. However it absoluely needs to be fixed. Some post endpoints accept json, and some don't.

* Enable both the json and urlencoded middlewares. These will parse the body for us.
* If no content-type header is set, we assume json.
* Remove all of the event listeners from the post route handlers.
* Update all the axios frontend calls to just post the objects (not stringify). Axios sets the correct application/json header.
* Update the checkMyAppsAvailability to remove the default content-type header. This will stop the middleware from parsing this call. This is to mitigate the fact that there will be old and new nodes trying to contact each other, new nodes will still call ensureObject so will be able to parse the result.

Need to wait before #1495 gets merged, so I can test the interoperability works etc.
I still need to test the frontend (and backend) to make sure I haven't missed any calls etc.

Would then probably want to rebase #1496 on top of this

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -14364,8 +14364,15 @@ async function checkMyAppsAvailability() {
     }
 
     const timeout = 30000;
+    // We have to set an emtpy content type here. As we've updated all the api endpoints to respect content
+    // type headers, when on the old version they need the data as a string (which by default sets the urlencoded
+    // header. However the new version has both json/urlencoded so we need to set the content-type header to emtpy
+    // so neither of the middlewares pick it up. We can remove this once all nodes are on the latest version)
     const axiosConfig = {
       timeout,
+      headers: {
+        'content-type': '',
+      },
     };
 
     const data = {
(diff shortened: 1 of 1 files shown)


TheTrunk (approved):
Concept ack
Just need to be careful. Sometime we were doing qs stringify for example. Shoud be more tested for any change

Cabecinha84:
@MorningLightMountain713 can you rebase this one also? Thank you.

MorningLightMountain713:
> @MorningLightMountain713 can you rebase this one also? Thank you.

I'll rebase this one last after everything else has been merged. We probably want to merge this after this weeks release so we have some time to make sure I haven't missed any endpoints etc

PR #1505, opened 2025-06-29
Add syntactic sugar for delay / sleeping

MorningLightMountain713:
I struggle reading all the delays everywhere. I.e. `3 * 60 * 1000` for 3 minutes. It's too much brainpower for me.

So I wrote a simple interval parsing function (and associated tests) so now you can just use plain english and understand at a glance how long you have to wait for

In the future, I would like to move highly used serviceHelper functions into the global namespace, a prime example is the delay function (and maybe rename to sleep)

You can now await a delay like this (whitespace is irrevelant):

```javascript
await serviceHelper.delay('3 minutes 30 seconds');
await serviceHelper.delay('1day');
await serviceHelper.delay('1 day');
await serviceHelper.delay('2 days 3 hrs 30 mins');
```

It takes a time interval as a number, and a unit. The unit can be any of the following:

```
's'
'sec'
'secs'
'second'
'seconds'
'm'
'min'
'mins'
'minute'
'minutes'
'h'
'hr'
'hrs'
'hour'
'hours'
'd'
'day'
'days'
```

This is a straight drop in, if you pass in a number, it will return the same number, so we can update the code as we go along.


MorningLightMountain713:
> Can there be some try catch with default fallback? As this is more prone to human error and can thus throw an error. Since we dot use typescript it would not get caught and it expects just any string, ideally should be strict requirement

It is a stricter requirement - that's what the regexes are for. It validates the input is what we want. I'm not a fan of "just in case" coding, it adds code bloat and makes code harder to read. We should just aim to get the code working as it should, with the tests to back it.

Some things I have added now, instead of the default 0ms timeout, I have changed this to 1 second. This is in case someone makes a mistake, and will stop the cpu going to 100% in case it is in a loop.

I have also added a bunch more test cases - for example, empty strings or an error, or objects etc.

All tests passing

TheTrunk (approved):
Ok

PR #1506, opened 2025-06-30
Fix app 8 registration

Cabecinha84:
List of changes/fixes:
- App 8 registration for non enterprise apps were failing saying enterprise field is mandatory, is now startup on ui as false and it's being added properly on the specificationFormatter method on appsService;
- Activates v8 app specs on UI, on last week when we had so many issues, I have put the activation on UI on block 194000:
- Small changes on Priority Nodes text;
- Fix route updatetolatestspecs where appname was not set as a parameter;
- Fix getting from db v8 enterprise apps. Because what is encrypted is what can be submited from ui, we need to pass the specificationFormatter method after decrypt, to make sure we move from strings to the real objects types properties expected;
- Remove from trySpawningGlobalApplication the checkAndDecryptAppSpecs call as that is already part of getting application specification.


alihm (approved):
Ack

XK4MiLX (approved):

TheTrunk (approved):
utACK

You can force push this to resolve the dist folder conflicts. Later rebuild the UI again on development branch.

It might be better to avoid compilation of UI on other branches and in the future we should probably do compilation only with CI

PR #1508, opened 2025-06-30
Improve storetemporarymessages processing

Cabecinha84:
We are currently validating everything from temporary app messages even if they are from old messages. What leads for several messages to not be accepted because we changed the validations mean while, example, port 22/53 is now banned because it was causing problems on the nodes.

What changes on this PR:
We only do all app spec validation for apps that are being received that can still be live on the network.

This will reduce the cpu usage when nodes are syncing receiving old messages and reduce the number of apps with message = to false on app hashes collection.

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -6858,7 +6858,7 @@ async function storeAppTemporaryMessage(message) {
     block = result.height;
   }
 
-  const appRegistraiton = message.type === 'zelappregister' || message.type === 'fluxappregister';
+  const appRegistration = message.type === 'zelappregister' || message.type === 'fluxappregister';
   
   if (daemonHeight - block + (appSpecFormatted.expire || 22000) >= 0) {
     if (appSpecFormatted.version >= 8 && appSpecFormatted.enterprise) {
@@ -6874,7 +6874,7 @@ async function storeAppTemporaryMessage(message) {
           { daemonHeight: block, owner: appSpecFormatted.owner },
         );
         await verifyAppSpecifications(appSpecFormattedDecrypted, block);
-        if (appRegistraiton) {
+        if (appRegistration) {
           await checkApplicationRegistrationNameConflicts(appSpecFormattedDecrypted, message.hash);
         } else {
           await checkApplicationUpdateNameRepositoryConflicts(appSpecFormattedDecrypted, messageTimestamp);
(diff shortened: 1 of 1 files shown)


MorningLightMountain713 (changes requested):
What you're saying makes sense about skipping verifying the specs, however I think we should still be verifying the hash and signature

Cabecinha84:
> What you're saying makes sense about skipping verifying the specs, however I think we should still be verifying the hash and signature

Agreed, changes done and pushed.

MorningLightMountain713 (approved):
Ack

TheTrunk (commented):
This feels like we are skipping verifications becuae we made accidental breaking changes. We shoudl fix the issues that were introduce, correct for those instead of just disabling verification. Yes more work and some may argue useless as those messages are old and won't cause any harm. I would argue it's now getting into more spaghetti and more side effects with bad long term impact.
A lot more cleanup is needed in the codebase, testing and fixing these issues.
I would not go against the group general consensus on lowering the requirements but just disagreement with it. We should build strong, resilient self governing network that can fully bootstrap validate itself, verify everything at any point of time. The moves to more controlled, governed model may seem nice, simpler but sacrificing decentralization would affect the resilience of the network long term.

Cabecinha84:
> This feels like we are skipping verifications becuae we made accidental breaking changes. We shoudl fix the issues that were introduce, correct for those instead of just disabling verification. Yes more work and some may argue useless as those messages are old and won't cause any harm. I would argue it's now getting into more spaghetti and more side effects with bad long term impact. A lot more cleanup is needed in the codebase, testing and fixing these issues. I would not go against the group general consensus on lowering the requirements but just disagreement with it. We should build strong, resilient self governing network that can fully bootstrap validate itself, verify everything at any point of time. The moves to more controlled, governed model may seem nice, simpler but sacrificing decentralization would affect the resilience of the network long term.

I understand what you are saying, but putting fix on the code just for those messages to be accepted would also be spagethi code, example, one of the app messages that nodes are rejecting is one app that the component name changed name. So we allow it at some time and because it was breaking stuff, we restricted that. So now, the nodes are constantly asking for that app message and some nodes have them as ok, reply back and the node reject the message because is no longer valid.

I don't think this change makes any centralization, the system remains completly decentralized, and doing all validations needed for the apps that now can be online.
Does it matter that 3 years ago, the network accepted a message that had something broken? That is what we need to ask... Time will never go to the past, so those apps that cannot be running now, have 0 value for the network currently.

PR #1510, opened 2025-07-03
Add port check logic, fix EADDRINUSE on servers, general clean up

MorningLightMountain713:
### This PR does the following:

* Adds logic to the port checker (and simplifies things) to handle failed state and sets a hysteresis to get back to a "normal" state
* Moves the test http server to a local variable and tidies up the testing server used by checkInstallingAppPortAvailable
* Moves the dosState to just use 100 - whatever the failure. Previous, it was 25 for failed ports and then it had to get to 100 to remove apps. Now it's just 100 everywhere.
* Removes all the serviceHelper delays - we just wait for the httpServer to be listening now (and don't need to wait for UPnP) there is an inbuilt delay anyway as we have to wait 1/2 RTT for the request to get to the remote end anyway. This was slowing down the total failure time significantly. (15 seconds per iteration * 125 iterations = over 30 minutes)
* Removes the janky call to netcat to test the port - we just use a socket now. (the netcat call was actually broken on some nodes, and was returning that the port was open even when it wasn't)
* Fixes the issue with the test http Server - this wasn't being caught properly if a port was already listening and was crashing the entire app.
* Fixes the EADDRINUSE error that was raised for the fluxServer http servers. We now handle this locally instead of on the process
* Removes the entire old / new portMin, portMax. This was broken. We were using the scanned height which is wrong. It should have been the current network block height (it's gone now anyway) Have updated it to PortMinLegacy etc, the only place this is now used is to verify correctness of apps  at the height they existed
* Totally removes the `http-shutdown` package. Just do this in house now. All it does is destroy all sockets when the server is closed. If we were on nodeJS 18.2.x - we could just use native node tools.

This means that if a node goes DOS for ports - apps will be uninstalled. The reason for this is that if the node goes DOS due to ports, it is more than likely that the app is broken anyway. I.e. it may have been fine when it was installed, but now it's not. Previously, this would have taken another 225 iterations before the apps were removed (several hours), where they're most likely broken anyway.

In most cases - this won't matter anyway as the node will go dos before any apps are installed.

Hopefully, this should be the last PR before I move the functionality to a class, update the algorithm, and write tests for it.

Here are the states:

    * Normal
          No broken ports, or broken ports less than 80 and a "good" port test
    * Normal - Rising edge
          I.e. broken ports increasing but threshold not reached. This state could
          also be considered normal, and could take many many hours to cross the threshold
    * Failed - Rising edge
          Threshold crossed. There are 100 ports in portsNotWorking. At this time the
          dosState starts rising 4 per fail. (takes 25 ports once in this state to DOS)
    * Failed - DOS
          There are 100 ports in the portsNotWorking array. 25 of those ports
          have failed a second time. Node is in DOS state.
    * Failed - Lowering edge
          Same as failed, however the node is now considered "working" and is removing
          ports from the portsNotWorking array. It will remain in this state until 20 ports
          have been removed from the portsNotWorking array. (hysteresis) Once this happens, the node
          is then considered in the "normal" state - and the portsNotWorking array is cleared

### Test methodology

Since we can't really write unit tests for this (recursion) without modifying the actual function logic, I did some integration testing.

I ran a test http server on one public ip, then on another public IP, ran the `checkMyAppsAvailability` function with the failure interval at one second, and the dos interval at 3 seconds. (so really fast)

The http server was set to fail the first 130 requests, which simulates that the port isn't open. This puts the node into DOS. Then after that http server was testing the ports were valid, and the node would go below the hysterisis value of 80 ports and go back to normal state

### DIFF
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -188,11 +188,8 @@ module.exports = {
     multisigAddressChange: 1670000,
     epochstart: 694000,
     publicepochstart: 705000,
-    portMin: 31000, // ports 30000 - 30999 are reserved for local applications
-    portMax: 39999,
-    portBlockheightChange: isDevelopment ? 1390000 : 1420000,
-    portMinNew: 1,
-    portMaxNew: 65535,
+    portMin: 1,
+    portMax: 65535,
     bannedPorts: ['16100-16299', '26100-26299', '30000-30099', 8384, 27017, 22, 23, 25, 3389, 5900, 5800, 161, 512, 513, 5901, 3388, 4444, 123, 53],
     enterprisePorts: ['0-1023', 8080, 8081, 8443, 6667],
     upnpBannedPorts: [],
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -188,6 +188,9 @@ module.exports = {
     multisigAddressChange: 1670000,
     epochstart: 694000,
     publicepochstart: 705000,
+    portMinLegacy: 31000, // ports 30000 - 30999 are reserved for local applications
+    portMaxLegacy: 39999,
+    portBlockheightChange: isDevelopment ? 1390000 : 1420000,
     portMin: 1,
     portMax: 65535,
     bannedPorts: ['16100-16299', '26100-26299', '30000-30099', 8384, 27017, 22, 23, 25, 3389, 5900, 5800, 161, 512, 513, 5901, 3388, 4444, 123, 53],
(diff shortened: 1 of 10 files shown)


MorningLightMountain713:
> Codewise seems good and nice improvement,despite not easy to follow on GitHub. Have you runner this on a node? We need this running on nodes for some time before releasing I would say as the change is quite big.

Yeah it's hard to see what is going on from the diff - easier to just look at the code. Would be good to test this for a week or two. I've had it running on one node for around 5 days - works fine.

However - I'll chunk this on 8 nodes and give it a good workout

MorningLightMountain713:
Will redo PR on new branch to include changes on development

PR #1512, opened 2025-07-07
Bugfix: fix v8 app messages

MorningLightMountain713:
Bug:

Prior v8 code was causing a registered app to expire immediately, as it had no height, so it was contiuously being installed on many nodes. Was also causing the apprunning message to not be stored as it had no hash.

This PR does the following:

* Fixes the v8 messages. Since the frontend is stringifying the compose / contacts properties, we are using the specificationFormatter to rehydrate these on a register or update message. When we do this, it was removing the height and hash - so we manually add these back in
* Fixes the getAppFolderSize on Arcane, the `/var/lib/docker` path was hardcoded
* Make sure we strip off any unencrypted data before storing in the db
* Strip the unencrypted data off api calls if decrypt not requested
* Add the port to the log if the checkMyFluxAvailabilty peer is uncontactable

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -1210,8 +1210,8 @@ async function appMonitorStream(req, res) {
  */
 async function getAppFolderSize(appName) {
   try {
-    const dirpath = path.join(__dirname, '../../../');
-    const directoryPath = `${dirpath}ZelApps/${appName}`;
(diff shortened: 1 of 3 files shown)


Cabecinha84 (approved):
Ack, great work!

XK4MiLX (approved):

PR #1517, opened 2025-07-18
Add Local specs to v8 enterprise apps

MorningLightMountain713:
### Background

Currently, when viewing a v8 enterprise app spec (as the owner) only the global specs are visible - it cannot fetch the local specs. The reason for this is that the `executeCommand` (the function that gets the specs off the actual nodes) was just pulling the raw specs - it wasn't decrypting / getting the compose / contacts properties.

### What this PR does

Being that the `Management.vue` file is almost 14k lines - I split the file; It's now split by template, script, and style. This saves vscode absoutely cooking every time it tries to save it etc. As now the script file is only 6k lines (only, lol).

We now fetch all the local specs, and if the enterprise field is different - we decrypt that locally, otherwise, we just use the fields from the global specs. This saves us having to fetch a bunch of stuff from all the nodes.

Of note - I have remove all the `enterprise` field property type mutation on the specs. I.e. where we were mutating a string to a boolean (I couldn't find any references in the backend) We should never be mutating a field's type. If we want a falsy value - set it to an empty string. (which is now what we are doing).

Have also fixed all the excessive api fetch calls we were doing. It was fetching some endpoints 3 or more times from the same node etc.

### Implementation

Being that the files are all new, here are the main changes in the `script.js` file:

* Add new getDecryptedEnterpriseFields method
* Modified getGlobalApplicationSpecifics to use new getDecryptedEnterpriseFields function
* Modified getInstalledApplicationSpecifics to use new getDecryptedEnterpriseFields function
* Updated the ipChanged watcher to not fire on inital load
* Only run the first case in updateManagementTab if we haven't already done so

### Testing

I have tested this on a local build with the following:
* Registering both v8 and non v8
* Displaying specs for v8 and non v8
* Moving around tabs etc

### Notes

When you go to the application to view the specs - it doesn't actually tell you anywhere that it is an enterprise app. I can't tell the difference between non enterprise and enterprise.

If needed, I can merge it all back to one file so you can see the diff, but I think we should still split it so it's more manageable.

### DIFF
--- a/HomeUI/src/views/apps/management/script.js
+++ b/HomeUI/src/views/apps/management/script.js
@@ -53,12 +53,12 @@ import DaemonService from '@/services/DaemonService';
 
 import SignClient from '@walletconnect/sign-client';
 import { MetaMaskSDK } from '@metamask/sdk';
-import 'xterm/css/xterm.css';
-import { Terminal } from 'xterm';
-import { FitAddon } from 'xterm-addon-fit';
-import { WebLinksAddon } from 'xterm-addon-web-links';
-import { Unicode11Addon } from 'xterm-addon-unicode11';
-import { SerializeAddon } from 'xterm-addon-serialize';
+import '@xterm/xterm/css/xterm.css';
+import { Terminal } from '@xterm/xterm';
+import { FitAddon } from '@xterm/addon-fit';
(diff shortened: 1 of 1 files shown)


Cabecinha84:
@MorningLightMountain713 can you rebase this PR/fix the conflicts? thank you.

PR #1519, opened 2025-07-21
Update fluxd getchaintips call

Cabecinha84:
getchaintips call was updated on latest daemon update to receive blockheight as parameter.
What was changed:
- We get the daemon version and cache it;
- If daemon version is higher than 7020050 means we are on daemon v8.0.0;
- Update to start calling getchaintips passing the last blockheight that was checked + 1;


MorningLightMountain713 (approved):
Ack

PR #1520, opened 2025-07-21
Improvements on trySpawningGlobalApplication

Cabecinha84:
List of changes:
- Remove probability to select one app to try to install it as now we are sending the app message fluxappinstalling to make sure we are not installing more instances than required;
- Get apps installed on top to filter from the apps that can be selected to install on the node;
- Adjust some logs;
- Remove some specific old code related with presearch when we were spawning the app on marketplace without port;
- Remove commented code so start up again checking if the app are failing to install across the network;
- Adjust all delays to run the function again after 5 minutes if it didn't finish with success.

### DIFF
--- a/ZelBack/src/services/appsService.js
+++ b/ZelBack/src/services/appsService.js
@@ -10969,7 +10969,7 @@ async function trySpawningGlobalApplication() {
         return;
       }
       if (appToRunAux.enterprise && !isArcane) {
-        log.info('trySpawningGlobalApplication - app can only install on ArcaneOS');
+        log.info(`trySpawningGlobalApplication - Application ${appToRun} can only install on ArcaneOS`);
         spawnErrorsLongerAppCache.set(appHash, '');
         await serviceHelper.delay(5 * 60 * 1000);
         trySpawningGlobalApplication();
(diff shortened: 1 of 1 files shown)


ZelBack/src/services/appsService.js
```diff
@@ -10969,41 +10978,22 @@ async function trySpawningGlobalApplication() {
...
     log.info(`trySpawningGlobalApplication - App ${appToRun} hash: ${appHash}`);
 
     const installingAppErrorsList = await appInstallingErrorsLocation(appToRun);
-    /* if (installingAppErrorsList.find((app) => !app.expireAt && app.hash === appHash)) {
+    if (installingAppErrorsList.find((app) => !app.expireAt && app.hash === appHash)) {
       spawnErrorsLongerAppCache.set(appHash, '');
       throw new Error(`trySpawningGlobalApplication - App ${appToRun} is marked as having errors on app installing errors locations.`);
-    } */
+    }
     if (installingAppErrorsList.length > 0) {
```
MorningLightMountain713: Does this need to be removed if we have the above uncommented?

MorningLightMountain713 (changes requested):
Looks good. Just a question if we need to remove some more code or not

ZelBack/src/services/appsService.js
Cabecinha84: removed.

PR #1521, opened 2025-07-23
v6.4.0

Cabecinha84:
- Memory optimizations;
- Start using new block parameter available on getchaintips flux daemon rcp call;
- Update backend so that messages to cancel applications to live just one block after the cancel message where currently it's 100;
- Improvements on management section for enterprise v8 apps;
- Other small improvements;


MorningLightMountain713 (approved):
Ack. However think UI needs rebuild

alihm (approved):
Ack

PR #1523, opened 2025-07-25
Bugfix: Syncthing cache and app installation

MorningLightMountain713:
#1521 introduced some issues for installing apps and running apps.

This patch fixes the following:

* Unable to spawn global apps as unable to verify the ports to install
* Syncthing apps stopping and restarting as the appId was being removed from the cache


Cabecinha84 (approved):
ack

PR #1524, opened 2025-07-25
Decrypt specs for app removal and port restore

MorningLightMountain713:
There is an issue with v8 Enterprise apps where it cannot remove the app as the specs need to be decrypted.

This also fixes the issue where the ports aren't "restored" for enterprise apps as the specs wern't decrypted.

I've tested the app removal for enterpriese - it works.
I've tested port restore for non enterpriese - it works.

* Decrypts specs for app removal.
* Decrypts specs for port restore for apps.

Screenshot 2025-07-25 at 3 47 18 PM


MorningLightMountain713:
I've tested this on a node again - works fine

Cabecinha84 (approved):
Ack

PR #1525, opened 2025-07-25
stop calling getDeviceID for own location

Cabecinha84:
We were calling getDeviceId for our own location that could cause timeout exception when wan did not accept the call:
image

So whe now exclude from locations the one with our ip, as it doesn't make sense to add as a peer on syncthing my one instance.


MorningLightMountain713 (approved):
This is running on a node that had the problem. Error log has gone away

alihm (approved):
Ack

PR #1526, opened 2025-07-26
v6.4.2

Cabecinha84:
- Fix not calling our own node getDeviceID from syncthing like it was a external node;
- Fix uninstalling v8 enterprise apps;
- Fix restoring v8 enterprise apps ports.


MorningLightMountain713 (approved):
Ack

TheTrunk (approved):

PR #1532, opened 2025-07-31
New nodeConnectivityService

Cabecinha84:
Summary

  This PR introduces a new Node Connectivity Service that automatically monitors and manages offline nodes in the Flux network. The service implements a robust mechanism
  to detect unreachable nodes and remove them from the network while preventing false positives.

  Key Features

  - Automated Monitoring: Runs every 10 minutes to check random node locations for connectivity
  - Retry Logic: Implements a 5-minute retry mechanism before marking nodes as down
  - Peer Verification: Added multi-layer verification where peers are asked to confirm node unreachability before removal
  - API Endpoint: New /flux/checkconnectivity endpoint allows nodes to verify connectivity of other nodes
  - Database Validation: Double-checks database state before final removal to prevent race conditions

  Implementation Details

  - Created nodeConnectivityService.js with TCP socket-based connectivity checks
  - Added helper function getRandomElements() in serviceHelper.js for peer selection
  - Integrated service into the main service manager for automatic startup
  - Added comprehensive unit tests for all new functionality

  Changes

  - New Service: ZelBack/src/services/nodeConnectivityService.js - Core connectivity monitoring logic
  - API Route: Added /flux/checkconnectivity endpoint in routes.js
  - Helper Functions: Added getRandomElements() utility in serviceHelper.js
  - Service Integration: Updated service manager and communication modules
  - Tests: Added unit tests for the new service and updated existing test suites

  Testing

  - Unit tests cover all major functions including connectivity checks, peer verification, and database operations
  - Integration tested with the existing Flux communication infrastructure
  - Verified retry logic and peer verification mechanisms work as expected

  This enhancement improves network reliability by automatically removing unresponsive nodes while minimizing false positives through peer consensus verification.


Cabecinha84:
have made some improvements to the service

MorningLightMountain713:
### The problem

My understanding of the problem this PR is attempting to solve:

App running messages have a lifetime of 125 minutes. If a node is running an app, and sends an apprunning message, then immediately goes down - it will take the network over 2 hours to figure this out.

Obviously - that is unacceptable and an issue we need to resolve. It is entirely possible that an app with 3 instances could lose all 3 instances within that timeframe and no new instances would be spawned.

### Initial thoughts

This issue is complex, it has a lot of edge cases and potential for corrupted views of the network. It's open to abuse; this is the sort of PR that we should have a detailed design and architecture document before any code is written.

We cannot, under any circumstance, (even if 2 peers have validated our findings) have the entire network send an HTTP request to one node to see if it is online, this is not scalable, unnecessary, and we are essentially DDoSing one of our own nodes if we suspect it's offline. (And it actually isn't, i.e. it may have intermittent connectivity, which is quite common)

This PR introduces an important distinction to how the network has operated before… nodes are making decisions about network state on other nodes; in real time, as opposed to waiting for a message to timeout - this could have unintended consequences and effects, and cause different views of the network. I.e. half the nodes have the location, and half don't.

### Mitigating the stampede effect

We can mitigate the need for the entire network to validate if a node is online by setting a timer on receipt of a nodedown message (or nodestate message)

* Set a scaled Random timer so one request is made every second. I.e. delay a random number between 0 and node count seconds. As this is random, there will be an even distribution between 0 and 9000 seconds, i.e. one every second.
* If x nodestate messages are received before the timer goes off - the node cancels the timer and takes action depending on what the other nodestate messages are. For a nodestate message to count, it must be from a unique node pubkey. There needs to be a 66% consensus for an action to be taken.
* If the timer is hit, it tests connectivity to the node, if it's good, you send a nodestate message with a field saying that it's up or if it's bad, a message saying its down.

This means at worst, only a handful of nodes must contact the node.

### Random Location "check" Timing

The app checks are random. I.e one location tested every 10 minutes per node. As of 2025/08/04 we have 6914 locations and 9785 nodes. That would mean we are checking the following:

9785 / 600 ‎ =  16.308 locations per second.
6914 / 16.308 ‎ =  423.964 Every location would get checked on average once every 423 seconds, ~7 minutes.

This is now an extra 16 api calls per second of network load, it would also take a minimum of 12 minutes to detect if a node was down. (7 minutes first check, then fail, then 5 minutes for second check. (We could do this a lot faster if we were using an app runner's peers to validate instead)

The problem with this is that as locations and node count changes - so do the above variables. It's very unpredictable, we can make the check timer a function of locations vs nodes to maintain a consistent location check time, but, there is a much easier way to do this without all the excessive network load.

### Abuse

In its current state, a rogue operator could peer with multiple nodes and spam node down messages. For each message, than is ~10k api calls, I.e. an amplification factor of 10000 - this is unworkable.

We could mitigate this by implementing the timer above, but that would still cause a DDoS effect, as there would still be around 8 api calls for each rogue call.

Until socket authentication is in place - this will always remain an attack vector; Even with the mitigation above in place (with no auth), if a rouge operator controlled enough pubkeys to trigger "node down" consensus - they could take down apps at will.

### Edge cases

There are quite a few edge cases which could corrupt a nodes view of locations. What if a node has intermittent connectivity? I.e. one node "checked" and it was down, and another node checked, and it was "up"?

### Socket authentication

IMO socket authentication MUST happen before this PR can be implemented safely, as we are making changes to the network state of other nodes.

Currently, anyone can peer with a node and listen in on the messages. Anyone with a node pubkey can send messages and have them accepted by the network. This is quite a problem. Especially how we are now looking at accepting state messages about a node from a 3rd party.

### Using peers as app monitors

IMO, there is a much easier, more scalable, and safer way to implement this.

Apps should be monitored by peers. (Once the socket ping stuff is fixed) Then you have a fixed amount of monitors and don't need any of the random application checks that are being done. This would save 16 api calls per second across the network. Each peer knows when you are running an app as they receive the apprunning message from you. Then can do a similar thing where they set a timer and the first 2 to trigger the timer start monitoring. 

If apps were monitored by peers - we wouldn't need to spam apprunning messages any more, we would just send one to start with, then either an appremoved message gets emitted, or peers send a nodestate changed message.

### Conclusion

In its current state, or without some serious algo changes and edge case handling -  I can't approve this PR.

Ideally, as this is a fundamental change to the way the network works - I would like to see network diagrams, a design, and some simulated testing with docker containers

MorningLightMountain713:
I agree with the intent of the PR, i.e. we need to detect when nodes are down, and take action.

What I don't agree with is the implementation, and the software development process. We shouldn't have gotten to this point without a design. The network is complex enough now that it isn't viable to make systemwide changes like this without considerable analysis.

I've made it clear there are fundamental issues with how this PR works, and things that can be implemented to mitigate. Again, we cannot have 10000 nodes sending api requests to one node. You just cannot do that. Surely this is obvious why it's a bad idea? Where do you think those 10000 api requests go? We have a distributed network, it should be able to communicate amongst itself to figure these things out.

It is important that the whole network has the same view of app locations. We can't have half the network saying an app is on a location, and the other half saying that it's not. (In the case of a node that is going up / down)

To reiterate, we need a solution for this problem. We just need to mitigate some of the main issues present here.

Where is the design?
Where is the impact / edge case analysis?
Where are the failure scenarios?
What is the expected behaviour for each failure scenario?
Where are the definitions / glossary?

What constitutes a node being "down"? It's not just so simple as a node being "up" or "down". A node transitions through states, I.e. "up", "down", "unavailable"

 Here is some example definitions from both the viewpoint of the network, and from the node itself. It also shows how defining one thing, leads to having to define what other things mean. Having this stuff defined is fundamental to building testable code, and being able to reason about how a system might work.

## Definitions (examples)

### Node unavailable (from network viewpoint)

A "Fluxnode" is considered unavailable when it has been "down" for a period of 7 minutes. This is the point where a nodestate message would be triggered

Note: 7 minutes, while arbitrary, was chosen as this is a reasonable timeframe for a node to reboot and apply updates, or if there was a short power cut. However, it may be that this timeframe gets shortened in the future as 7 minutes could also be considered too long a period from an apps perspective.

### Node unavailable (from node viewpoint)

The "Fluxnode" is considered unavailable when it is "down" for a period of 7 minutes.

Note: At this point the node will remove any apps that it is running. It also goes into a stand down period, where it needs to be "up" for a period of 130 minutes, so it can regain a full view of the network.

### Node down (from network viewpoint)

A "Fluxnode" is considered "down" when it is "unreachable" by its "peers"

### Node down (from node viewpoint)

The "Fluxnode" is considered "down" when it has lost all "peers", and cannot connect to other "Fluxnodes"

### Fluxnode definition

A compute resource running the Gravity orchestrator, Flux blockchain daemon, and Flux benchmarking service.

### Node unreachable definition:

A node is unreachable when it has missed 3 consecutive pings, or if the socket has been closed, and not accepting new connections.

### Peer definition

A "peer" is defined as any other "Fluxnode" on the "network" that is connected to this Fluxnode via web socket.

### Network definition

The aggregate collection of Fluxnodes

## Failure scenarios (limited examples)

### Node powered down for 30 minutes

In this timeframe, the nodes peers would transition the node to "state" down after 3 missed pings. After a further 7 minutes it would then transition to unavailable, where a random timer is set and the first node to activate the timer broadcasts a "nodestate" message to the network. All nodes would then remove any locations that that node has.

Note: the above would only work if sockets are authenticated. If not authenticated, then an additional quorum step would need to be performed with a random timer approach and the first 32 (or some arbitrary number) of nodes validate that the node is unavailable, then broadcast their judgement to the rest of the network, who then makes a quorum decision

### Node is flip flopping between offline and online

The node is disconnecting internet every 5 minutes

Blah blah blah

### Node goes down for 10 minutes, and then comes back online.

Blah blah blah

### Node is down for 7 minutes, but then comes back up just as the network validates

Blah blah blah

ZelBack/src/services/nodeConnectivityService.js
```diff
@@ -0,0 +1,368 @@
...
+  try {
+    const { ip, port = 16127, timeout = 5000 } = req.params.ip ? req.params : req.body;
+
+    if (!ip) {
+      const errMessage = messageHelper.createErrorMessage('Missing required parameter: ip');
+      return res.json(errMessage);
+    }
+
+    log.info(`nodeConnectivityService: API request to check connectivity for ${ip}:${port}`);
+    const isReachable = await checkNodeConnectivity(ip, port, timeout);
```
alihm, line 127: This can be used as a port scanner, I think we should validate that the ip belongs to the flux network, or check that port is in flux api range before calling checkNodeConnectivity

ZelBack/src/services/nodeConnectivityService.js
```diff
@@ -0,0 +1,368 @@
...
+          log.warn(`nodeConnectivityService: Invalid response from peer ${peer.ip}`);
+        }
+      } catch (error) {
+        log.warn(`nodeConnectivityService: Failed to contact peer ${peer.ip}: ${error.message}`);
+      }
+    }
+
+    if (!apiReachable) {
+      log.info(`nodeConnectivityService: All contacted peers were unreachable.`);
+      return true;
```
alihm, line 193: This should return false when unreachable

ZelBack/src/services/fluxCommunication.js
```diff
@@ -386,6 +386,86 @@ async function handleAppRemovedMessage(message, fromIP, port) {
...
+      return;
+    }
+
+    // Check if message is less than 5 minutes old
+    const currentTime = Date.now();
+    const messageAge = currentTime - broadcastAt;
+    const fiveMinutes = 5 * 60 * 1000; // 5 minutes in milliseconds
+
+    if (!broadcastAt || messageAge > fiveMinutes) {
+      log.warn(`handleNodeDownMessage: Ignoring old nodedown message from ${fromIP}:${port} - message age: ${messageAge}ms`);
```
alihm, line 406: The message is discarded if older than 5 min, but is later rebroadcast only when verifyTimestampInFluxBroadcast passes a 4 min(240000) window. A message between those is accepted yet never rebroadcast.

ZelBack/src/services/fluxCommunication.js
```diff
@@ -386,6 +386,86 @@ async function handleAppRemovedMessage(message, fromIP, port) {
...
+      return;
+    }
+
+    log.info(`handleNodeDownMessage: Received nodedown message for IP ${downNodeIP} from ${fromIP}:${port}`);
+
+    // Verify connectivity to the reported down node
+    const isReachable = await nodeConnectivityService.checkNodeConnectivity(downNodeIP, 16127, 10000);
+
```
alihm, line 414: Nodes running on non-default port will be incorrectly flagged unreachable.

Cabecinha84:
Closing as we will go on different route and we will discuss it better tomorrow.

PR #1535, opened 2025-08-03
[FIX] Redeploy spec import

XK4MiLX:
- Prevent import spec be overwrite by empty template
- Fix missing no-wrap style


MorningLightMountain713 (approved):
Ack

Edit: I just saw comment in Bitrix, assume this will end up getting closed

Cabecinha84 (approved):
ack

PR #1538, opened 2025-08-10
Feature: Add small delay for UPnP calls

MorningLightMountain713:
There seem to be a lot of routers out there that struggle with our upnp implementation.

The main reason is because `@runonflux/nat-upnp` goes and fetches all the details on every call, which puts the router under a lot of unnecessary strain.

As a bandaid to fix this (until I can fix the nat-upnp module) this PR just adds a half second delay after mappings. We add it after both a mapping and remove request, as there may be other requests after this one.

Here is an example of the behaviour:

Screenshot 2025-08-10 at 10 55 28 AM

I will do up a PR soon to check the router UPnP capability at boot and if we can set timeouts on UPnP mappings - can set a short timeout for test mappings and a timeout longer than our refresh time for other mappings. This will help when operators shut down nodes etc and restart on a new IP (at least the mappings will time out)

I've tested this on a node - works fine

### DIFF
--- a/ZelBack/src/services/upnpService.js
+++ b/ZelBack/src/services/upnpService.js
@@ -119,6 +119,8 @@ async function verifyUPNPsupport(apiport = config.server.apiport) {
     }
     // run test on apiport + 1
     await client.getPublicIp();
+
+    await serviceHelper.delay(500);
   } catch (error) {
     log.error(error);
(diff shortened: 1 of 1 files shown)


Cabecinha84 (approved):
ack

PR #1541, opened 2025-08-12
[FIX] Free app update service

Cabecinha84:
The service was getting the app specifications directly from database, where v8 enterprise apps are encrypted. Now instead we call the method to get global app specifications that decrypts the app specifications if it's a enterprise app.


alihm (approved):
Ack

PR #1547, opened 2025-08-17
Stagger multiport bench per IP

MorningLightMountain713:
Comes with tests. I've tested this on a node - working.

## Problem

Gravity benchmarks UPnP nodes at the same time across the network. This is a large problem for speedtesting as we are effectively DDoSing many servers as we are testing them all at the same time. Especially nodes with huge throughput rates.

The reason this is done is to check all nodes on the same hypervisor to make sure they aren't allocating resources they don't have. I.e. 8 Nimbus nodes sharing a 50Mbps connection.

## Solution

Use the IP of the node as a proxy for the Hypervisor (and upstream network connection). This isn't perfect as people can run multiple IPs on the same Hypervisor but it's pretty good.

Instead of running all nodes on the same block - we now group them by ip and run them at random times over 1 hour. I.e. If an operator is running 16 nodes on 2 different ips, all 8 nodes on IP 1 will run at a a random time between 0-3600 seconds, and all 8 nodes on IP B will also run at a random time between 0-3600.

The random time is the SAME for each of the nodes on the same ip. I.e. - the random timer is seeded.

The Potato block is still the trigger that starts the timers that run the multiport bench. So all nodes will have completed the multiport bench within 1 hour of the Potato block

## Random distribution

Here is the distribution across buckets (each bucket is 1000, i.e. 0-999, 1000-1999 etc) with 10000 random numbers:

```
Map(10) {
  0 => 983,
  1 => 992,
  2 => 996,
  3 => 995,
  4 => 1013,
  5 => 1027,
  6 => 1002,
  7 => 993,
  8 => 1020,
  9 => 979
}
```

## Bonus

This PR also fixes an issue we have if a node fails to verify UPnP support on initial Gravity start (for whatever reason) it was exiting with code 0. This means that process supervisors  would not restart the service (if set to restart on failure) as the process exited successfully. We now exit with error code 1, after waiting 2 minutes. I've tested this - working.

### DIFF
--- a/ZelBack/src/services/explorerService.js
+++ b/ZelBack/src/services/explorerService.js
@@ -627,9 +627,9 @@ async function processBlock(blockHeight, isInsightExplorer) {
       if (blockDataVerbose.height % config.fluxapps.benchUpnpPeriod === 0) {
         try {
           // every node behind the same ip will benchmark at the same time. I.e.
-          // we spread the network out (grouped by ip) over an hour so we don't
+          // we spread the network out (grouped by ip) over 4 hours so we don't
           // absolutely hammer the speedtest servers at the same time.
-          const maxBenchDelay = 3_600_000;
+          const maxBenchDelay = 4 * 3_600_000;
           const socketAddress = await fluxNetworkhelper.getMyFluxIPandPort();
 
           // socketAddress can be null. If it is, we just use an empty string. This
--- a/ZelBack/src/services/explorerService.js
+++ b/ZelBack/src/services/explorerService.js
@@ -643,7 +643,7 @@ async function processBlock(blockHeight, isInsightExplorer) {
           const benchDelayMs = serviceHelper.randomDelayMs(maxBenchDelay, { initializer });
           const benchDelayS = Math.round((benchDelayMs / 1000) * 100) / 100;
 
-          log.info(`Starting multiport bench in: ${benchDelayS}s`);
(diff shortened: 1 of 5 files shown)


ZelBack/src/services/explorerService.js
```diff
@@ -625,7 +626,26 @@ async function processBlock(blockHeight, isInsightExplorer) {
       }
       if (blockDataVerbose.height % config.fluxapps.benchUpnpPeriod === 0) {
         try {
-          benchmarkService.executeUpnpBench();
+          // every node behind the same ip will benchmark at the same time. I.e.
+          // we spread the network out (grouped by ip) over an hour so we don't
+          // absolutely hammer the speedtest servers at the same time.
+          const maxBenchDelay = 3_600_000;
```
Cabecinha84: instead of one hour  lets put the max value 4 hours.

ZelBack/src/services/serviceHelper.js
```diff
@@ -1,3 +1,5 @@
+/* eslint no-bitwise: 0 */
```
Cabecinha84, line 1: is this needed?

MorningLightMountain713: Yes. There are lots of bitwise operations in the seed generation

ZelBack/src/services/explorerService.js
MorningLightMountain713: Done

Cabecinha84 (approved):
ack

PR #1548, opened 2025-08-17
[WIP] Fix verify message hash and valueSat in permanent messages

MorningLightMountain713:
**Still need to test this (and write the code to fix up the permanent messages)

### Apps permanent messages

When a node first syncs, it can download the permanent messages off another node. There is a typo in the parsing, so the `valueSat` field ends up as `NaN` - which can manifest as a problem in different ways.

### Verifying message hashes

Since someone broke the ordering of the spec formatter for v7 (me) it causes issues with message hashing and verifying. A fix was put in place in the past for message verifying - but not for hashing.

## Solution

This PR just fixes the permanent message parsing on initial sync, and the `verifyAppHash` function, so that it retries verifying the hash with the old format.

There is also a cleanup function in `serviceManager` to tidy up the permanent messages using the value from the appsHashes collection. This is only run if any objects are found that have NaN for valueSat.


Cabecinha84 (approved):
ack

PR #1557, opened 2025-09-10
Added a new upnp boolean property to info.flux that indicates if the node is running under UPnP

Cabecinha84:


MorningLightMountain713 (approved):
Ack

PR #1574, opened 2025-10-21
Fix Syncthing state management and app deployment issues

Cabecinha84:
## Summary
  This PR addresses critical issues with Syncthing state management, app deployment lifecycle, and container orchestration that were causing data loss and deployment failures.

  ## Key Changes

  ### Syncthing State Management
  - **Preserve syncthing app cache during soft redeploy** to prevent unintended data deletion
  - Fixed syncthing apps logic to properly handle soft vs hard redeploy scenarios
  - Improved syncthingApps handling to avoid treating soft redeploys as new installations

  ### App Deployment Fixes
  - **Fixed v4+ app renewal process** that was failing to properly renew applications
  - **Fixed reinstallOldApplications functionality** - nodes were not properly calling this critical function
  - Improved reinstallOldApplications workflow with better error handling and state management
  - Ensured clean slate for app installation error tracking on each restart

  ### Docker Container Lifecycle Improvements
  - Optimized container management: don't call `docker stop` if container is already stopped
  - Improved restart logic: call `start` instead of `restart` for stopped containers
  - Reduced unnecessary Docker API calls and improved deployment reliability

  ### State Tracking Enhancements
  - Added global state flags `isSoftRedeploy` and `isHardRedeploy` to distinguish between deployment types
  - Better context awareness throughout the deployment pipeline

  ## Files Changed
  - `advancedWorkflows.js` - Core deployment logic improvements
  - `explorerService.js` - reinstallOldApplications fixes
  - `syncthingFolderStateMachine.js` - State management improvements
  - `globalState.js` - New deployment type flags
  - `serviceManager.js` - Integration of new state flags

  ## Impact
  - Prevents data loss during soft redeploys
  - Improves deployment reliability for v4+ applications
  - Reduces unnecessary Docker operations
  - Better error tracking and recovery

### DIFF
--- a/ZelBack/src/services/appLifecycle/advancedWorkflows.js
+++ b/ZelBack/src/services/appLifecycle/advancedWorkflows.js
@@ -2281,7 +2281,6 @@ async function reinstallOldApplications() {
       throw new Error('Failed to get installed Apps');
     }
     const appsInstalled = installedAppsRes.data;
-    globalState.reinstallationOfOldAppsInProgress = true;
     // eslint-disable-next-line no-restricted-syntax
     for (const installedApp of appsInstalled) {
       // get current app specifications for the app name
@@ -2295,6 +2294,7 @@ async function reinstallOldApplications() {
         // eslint-disable-next-line no-await-in-loop
         log.warn(`Application ${installedApp.name} version is obsolete.`);
         if (randomNumber === 0) {
+          globalState.reinstallationOfOldAppsInProgress = true;
           // check if the app spec was changed
           const auxAppSpecifications = JSON.parse(JSON.stringify(appSpecifications));
           const auxInstalledApp = JSON.parse(JSON.stringify(installedApp));
@@ -2322,7 +2322,7 @@ async function reinstallOldApplications() {
               upsert: true,
             };
             // eslint-disable-next-line no-await-in-loop
-            await dbHelper.updateOneInDatabase(appsDatabase, localAppsInformation, appsQuery, appSpecifications, options);
+            await dbHelper.updateOneInDatabase(appsDatabase, localAppsInformation, appsQuery, { $set: appSpecifications }, options);
             log.info(`Application ${installedApp.name} Database updated`);
             // eslint-disable-next-line no-continue
             continue;


ZelBack/src/services/appLifecycle/advancedWorkflows.js
```diff
@@ -2312,7 +2312,24 @@ async function reinstallOldApplications() {
           delete auxInstalledApp.instances;
           delete auxInstalledApp.owner;
 
-          if (JSON.stringify(auxAppSpecifications) === JSON.stringify(auxInstalledApp)) {
+          // Helper function to recursively sort object keys for consistent comparison
+          const sortKeys = (obj) => {
```
MorningLightMountain713: This is making a copy of the specs. So we can remove the earlier copy:

          const auxAppSpecifications = JSON.parse(JSON.stringify(appSpecifications));
          const auxInstalledApp = JSON.parse(JSON.stringify(installedApp));
          
   saves having to stringify and parse. (as long as the deletes are moved after the sort)

ZelBack/src/services/appLifecycle/advancedWorkflows.js
```diff
@@ -2632,12 +2860,21 @@ async function masterSlaveApps(globalStateParam, installedApps, listRunningApps,
...
-                  // eslint-disable-next-line no-await-in-loop
-                  const response = await axios.get(`http://${ipToCheckAppRunning}:${portToCheckAppRunning}/apps/listrunningapps`, { timeout, cancelToken: source.token });
-                  isResolved = true;
-                  const appsRunning = response.data.data;
-                  if (appsRunning.find((app) => app.Names[0].includes(installedApp.name))) {
-                    log.info(`masterSlaveApps: app:${installedApp.name} is not on fdm but previous master is running it at: ${url}`);
+                  let previousMasterStillRunning = false;
+                  try {
+                    // eslint-disable-next-line no-await-in-loop
+                    const response = await axios.get(`http://${ipToCheckAppRunning}:${portToCheckAppRunning}/apps/listrunningapps`, { timeout, cancelToken: source.token });
```
MorningLightMountain713, line 2849: We should be using https here.

Cancel tokens are deprecated, should be using AbortController (active since node v15.4.0) our min version on the network is 16.17.0

MorningLightMountain713 (changes requested):

ZelBack/src/services/appLifecycle/advancedWorkflows.js
Cabecinha84: I have reverted this change as it was not needed.

ZelBack/src/services/appLifecycle/advancedWorkflows.js
Cabecinha84: we are always using http when calling the nodes directly, not https.
We still have lots and lots of places on the code using the cancel token, I think both should be done in separate PR.

PR #1575, opened 2025-10-22
Enhance app validation and cleanup workflows

Cabecinha84:
## Summary
  - Allow hyphens in app names (excluding first and last character positions)
  - Enable apps to update their version information
  - Remove redundant database cleanup code from advanced workflows

  ## Changes

  ### App Name Validation Enhancement
  - Modified app name validation to permit hyphens within the name
  - Restriction: hyphens cannot appear as the first or last character
  - Improves flexibility for app naming conventions

  ### Version Update Support
  - Apps can now update their version information dynamically
  - Enables better version management for deployed applications

  ### Code Cleanup and Optimization
  - Removed duplicate database cleanup logic from `advancedWorkflows.js`
  - Cleanup is already handled by `appUninstaller.removeAppLocally`
  - Reduces code duplication and potential maintenance issues

  ## Files Changed
  - App validation logic (name and version constraints)
  - `ZelBack/src/services/appLifecycle/advancedWorkflows.js` (removed redundant cleanup)

  ## Impact
  - **Breaking Changes:** None
  - **Performance:** Minor improvement from reduced redundant operations
  - **Compatibility:** Backward compatible with existing apps

### DIFF
--- a/ZelBack/src/services/appRequirements/appValidator.js
+++ b/ZelBack/src/services/appRequirements/appValidator.js
@@ -548,8 +548,10 @@ function verifyRestrictionCorrectnessOfApp(appSpecifications, height) {
   if (appSpecifications.version !== 1 && appSpecifications.version !== 2 && appSpecifications.version !== 3 && appSpecifications.version !== 4 && appSpecifications.version !== 5 && appSpecifications.version !== 6 && appSpecifications.version !== 7 && appSpecifications.version !== 8) {
     throw new Error('Flux App message version specification is invalid');
   }
-  if (appSpecifications.name.length > 32) {
-    throw new Error('Flux App name is too long');
+  // Version 8+ allows up to 63 characters to align with FQDN label standards (RFC 1035)
+  const maxNameLength = appSpecifications.version >= 8 ? 63 : 32;
+  if (appSpecifications.name.length > maxNameLength) {
+    throw new Error(`Flux App name is too long. Maximum ${maxNameLength} characters allowed`);
   }
   // furthermore name cannot contain any special character
   if (!appSpecifications.name) {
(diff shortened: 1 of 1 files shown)


ZelBack/src/services/appLifecycle/advancedWorkflows.js
```diff
@@ -1916,10 +1916,10 @@ async function checkApplicationUpdateNameRepositoryConflicts(specifications, ver
         // v4 allows for changes of repotag
       });
     } else { // update is v4+ and current app have v1,2,3
-      throw new Error(`Flux App ${specifications.name} on update to different specifications is not possible`);
+      // node will perform hard redeploy of the app;
```
MorningLightMountain713: We should fix this function at the same time. It is ridiculously complex for no reason (and the variables are poorly named) Also the docstring doesn't say what the intent is.

ZelBack/src/services/appRequirements/appValidator.js
```diff
@@ -555,8 +555,15 @@ function verifyRestrictionCorrectnessOfApp(appSpecifications, height) {
   if (!appSpecifications.name) {
     throw new Error('Please provide a valid Flux App name');
   }
-  if (!appSpecifications.name.match(/^[a-zA-Z0-9]+$/)) {
-    throw new Error('Flux App name contains special characters. Only a-z, A-Z and 0-9 are allowed');
+  // Version 8+ allows hyphens in app names (but not as first or last character)
+  if (appSpecifications.version >= 8) {
+    if (!appSpecifications.name.match(/^[a-zA-Z0-9]([a-zA-Z0-9-]*[a-zA-Z0-9])?$/)) {
```
MorningLightMountain713: It's good practice to only use capture groups when we need them. i.e. `()` this should be a non capture group `(?:)`

i.e.:

`^[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$`

ZelBack/src/services/appLifecycle/advancedWorkflows.js
Cabecinha84: this should be fixed

ZelBack/src/services/appRequirements/appValidator.js
Cabecinha84: should be resolved

PR #1586, opened 2025-10-29
Network state update - Fix race condition and refactor

MorningLightMountain713:
Fixes issues with networkStateService since the fork.

The daemon can now produce many blocks very rapidly. On every block, we request the network state from the daemon. This can overwhelm the daemon. While it should handle it itself, this measure is put in place to protect it until the network settles.

The prior fix was only stopping the daemon fetch, it would still build the indexes - which was not necessary.

It also highlighted a race condition that would cause the index build to crash and log an error.

Screenshot 2025-10-29 at 9 52 47 AM


Cabecinha84 (approved):
ack

PR #1608, opened 2025-11-13
add payment service

TheTrunk:


TheTrunk:
@Cabecinha84 rechecked, the ai mentions are all fine, won't adjust.

PR #1618, opened 2025-11-22
Bugfix: test enterprise apps and verify arch

MorningLightMountain713:
ALL TESTS PASSING

The test app before deploy for v8 enterprise wasn't decrypting the specs - so it thought there were 0 components and would instantly pass.

This PR decrypts the specs so it can verify the app. Tested on v8 enterprise (haven't tested non enterprise to make sure it still works yet)

Also adds validation for the components image architectures BEFORE we register the app. This is where the user clicks the compute registration message. This is good as we catch user errors early.

It detects the following:

on v8 enterprise (Arcane) ALL images must support `amd64`

on v7 enterprise (and all other spec versions) ALL images must have a common architecture. I.e. they must either support `amd64` or `arm64`

It also fixes a bug when trying to test an `arm64` only install. Since all nodes via home.runonflux.com are Arcane, none of them can install these test images, and you couldn't get past the test part. In this edge case - we skip the install test

Also fixes a bug where we couldn't decrypt the specs if the app wasn't in permanentappmessages. I.e. temp apps. This was causing temp apps to not be able to be removed. We just use height 0 in this specific instance.

Eg common arch:

Screenshot 2025-11-22 at 9 58 05 AM

Eg v8 enterprise arch:

Screenshot 2025-11-22 at 9 46 59 AM

Skipping on `arm64` install:

Screenshot 2025-11-23 at 11 30 11 AM

### DIFF
--- a/ZelBack/src/services/utils/enterpriseHelper.js
+++ b/ZelBack/src/services/utils/enterpriseHelper.js
@@ -194,10 +194,12 @@ async function checkAndDecryptAppSpecs(appSpec, options = {}) {
     const lastUpdate = allPermanentAppMessage[allPermanentAppMessage.length - 1];
 
     if (!lastUpdate) {
-      throw new Error(`App: ${appSpecs.name} does not exist in global messages`);
+      // Not found in permanent messages, use height 0 (for test apps in temporary messages)
+      log.info(`App ${appSpecs.name} not found in permanent messages, using height 0 for decryption`);
+      daemonHeight = 0;
+    } else {
+      daemonHeight = lastUpdate.height;
     }
-
-    daemonHeight = lastUpdate.height;
   }
 
   const enterprise = await decryptEnterpriseFromSession(


Cabecinha84:
Critical Issues

  1. Breaking API Change Without Backward Compatibility

  Location: imageManager.js:verifyRepository()

  The PR changes verifyRepository() return type from boolean to object:
  // Before: return true;
  // After: return { verified: true, supportedArchitectures: [...] }

  Problem: This breaks existing callers that expect boolean. I found at least 3 call sites:
  - appValidator.js:1247
  - appValidator.js:1262
  - appSpawner.js:522

  Recommendation: The PR should either:
  - Return the object AND ensure all callers are updated in the same PR (test files were updated but check
  production code)
  - Add backward compatibility by checking how the return value is used

  2. Missing findCommonArchitectures Helper Function

  Issue: The PR references findCommonArchitectures() but I couldn't find this function in the current codebase,
  suggesting it's added in the PR but likely duplicated.

  Recommendation: Create a single, well-tested utility function:
  // Suggested location: ZelBack/src/services/utils/appUtilities.js
  /**
   * Find architectures common to all component arrays
   * @param {Array<Array<string>>} componentArchitectures
   * @returns {Array<string>} Common architectures
   */
  function findCommonArchitectures(componentArchitectures) {
    if (!componentArchitectures || componentArchitectures.length === 0) {
      return [];
    }

    return componentArchitectures.reduce((common, archs) =>
      common.filter(arch => archs.includes(arch))
    );
  }

  3. Inconsistent Error Handling for Architecture Mismatch

  Location: appInstaller.js:testAppInstall()

  The PR makes test installations "silently skip" with a 200 status when architecture doesn't match:
  if (!commonArchitectures.includes(localArch)) {
    res.write(serviceHelper.ensureString(skipMessage));
    res.end();
    return; // Returns 200 OK
  }

  Problem: HTTP 200 with "skipped" message is ambiguous - clients can't distinguish success from skip.

  Recommendation:
  - Return a specific status code or structured response indicating the skip reason
  - Or throw an error with a clear message that this is expected behavior on this node

  4. Missing Null/Empty Validation

  Location: Both appValidator.js and appInstaller.js

  Problem: No validation before processing componentArchitectures array:
  const commonArchitectures = findCommonArchitectures(componentArchitectures);
  // What if componentArchitectures is empty or contains empty arrays?

  Recommendation: Add validation:
  if (!componentArchitectures || componentArchitectures.length === 0) {
    throw new Error('No component architectures found to validate');
  }

  // Filter out empty arrays before processing
  const validArchs = componentArchitectures.filter(archs => archs && archs.length > 0);
  if (validArchs.length === 0) {
    throw new Error('No valid architectures found in components');
  }

  Non-Critical Issues

  5. Code Duplication: Architecture Validation Logic

  The same architecture checking logic appears in both:
  - appValidator.js (registration-time validation)
  - appInstaller.js (test installation-time check)

  Recommendation: Extract to a shared validation function:
  // ZelBack/src/services/appRequirements/architectureValidator.js
  function validateAppArchitectures(componentArchitectures, isEnterpriseArcane) {
    const commonArchs = findCommonArchitectures(componentArchitectures);

    if (isEnterpriseArcane) {
      // Enterprise requires amd64 on ALL components
      const missingAmd64 = componentArchitectures
        .filter(archs => !archs.includes('amd64'));
      if (missingAmd64.length > 0) {
        throw new Error('Enterprise apps require amd64 support on all components');
      }
    } else {
      // Non-enterprise requires at least ONE common architecture
      if (commonArchs.length === 0) {
        throw new Error('No common architecture across all components');
      }
    }

    return commonArchs;
  }

  6. ImageVerifier Needs Getter for Supported Architectures

  Location: imageVerifier.js

  The PR adds a supportedArchitectures getter but I don't see where #supportedArchitectures is populated in the
  current code.

  Recommendation: Verify the PR includes logic in #evaluateImageManifest() to populate this array:
  // In evaluateMultipleImages()
  const supportedArchs = images.map(img => img.platform.architecture);
  this.#supportedArchitectures = [...new Set(supportedArchs)];

  7. Insufficient Test Coverage

  The PR shows 89.47% patch coverage with 6 uncovered lines.

  Recommendation: Add tests for:
  - Empty componentArchitectures array
  - Mixed architecture scenarios (some components amd64-only, others multi-arch)
  - Enterprise apps with arm64-only components (should fail)
  - Error handling when verifyRepository() fails

  8. Missing JSDoc Updates

  Location: imageManager.js:verifyRepository()

  Recommendation: Update JSDoc to reflect new return type:
  /**
   * Verify repository and image compliance
   * @param {string} repotag - Repository tag to verify
   * @param {object} options - Verification options
   * @returns {Promise<{verified: boolean, supportedArchitectures: string[]}>} Verification result with supported
  architectures
   */

  Minor Improvements

  9. Magic Constant for Enterprise Architecture Requirement

  Good: The PR adds enterpriseRequiredArchitectures = ['amd64'] constant
  Better: Reference this constant everywhere instead of hardcoding 'amd64' checks

  10. Error Messages Could Be More Specific

  When architecture validation fails, include which components failed and why:
  throw new Error(
    `Architecture validation failed for enterprise app. ` +
    `Components without amd64 support: ${missingComponents.join(', ')}`
  );

  Summary

  The PR implements important functionality but needs refinement:

  Must Fix:
  1. Ensure backward compatibility for verifyRepository() API change
  2. Add null/empty array validation
  3. Create centralized findCommonArchitectures() utility
  4. Clarify test installation skip behavior (HTTP status)

  Should Fix:
  5. Reduce code duplication with shared architecture validator
  6. Improve test coverage for edge cases
  7. Update JSDoc comments

  Nice to Have:
  8. More descriptive error messages
  9. Use constants consistently

  The core logic is sound (separating enterprise/non-enterprise validation, checking at registration AND test time),
   but code quality and error handling need improvement before ACK.

PR #1624, opened 2025-12-12
fix: Improve network diversity and syncthing app distribution across subnets

Cabecinha84:
## Summary

  - Fix incorrect index wraparound calculation in `fluxDiscovery` deterministic peer selection
  - Add /16 subnet diversity for random peer connections to improve network resilience
  - Improve syncthing app distribution to prevent multiple instances on same IP subnet
  - Optimize app spawning timing for ArcaneOS nodes and target IP scenarios
  - Fix not allowing to update app versions to latest one if they were on version <=3

  ## Changes

  ### FluxCommunication - Peer Discovery Improvements

  **Bug Fix: Deterministic peer wraparound calculation**
  Fixed incorrect index calculation when iterating backwards through the sorted node list for incoming deterministic connections. The previous formula `sortedNodeList.length - fluxNodeIndex - i` was incorrect; changed to `sortedNodeList.length + fluxNodeIndex - i`.

  **Subnet Diversity for Random Connections**
  - Added `getIpSubnet()` helper function to extract /16 subnet prefix (first 2 octets)
  - Random outgoing/incoming connections now skip nodes that:
    - Are in the same /16 subnet as the local node
    - Are in a /16 subnet that already has an existing connection
  - This improves network resilience by distributing connections across different ISPs/datacenters

  ### AppSpawner - Syncthing & Timing Improvements

  **Improved Subnet Check for Syncthing Apps**
  - Changed IP range matching from `includes()` to `startsWith()` for more accurate subnet detection
  - Moved `ipPrefix` calculation outside the syncthing block for reuse
  - Added a second syncthing subnet check after the delay logic to catch race conditions where another node in the same subnet started installing during the delay period

  **Optimized Spawning Delays**
  - Reduced non-enterprise app check delay on ArcaneOS from ~1 hour to ~30 minutes
  - Apps with explicit target IPs (`nodes` array containing this node) now skip the delay entirely

### DIFF
--- a/ZelBack/src/services/appLifecycle/advancedWorkflows.js
+++ b/ZelBack/src/services/appLifecycle/advancedWorkflows.js
@@ -2821,7 +2821,7 @@ async function updateAppGlobalyApi(req, res) {
       if (!appInfo) {
         throw new Error('Flux App update received but application to update does not exist!');
       }
-      if (appInfo.version <= 3 && appInfo.repotag !== appSpecFormatted.repotag) { // this is OK. <= v3 cannot change, v4 can but does not have this in specifications as its compose
+      if (appInfo.version <= 3 && appSpecFormatted.version <= 3 && appInfo.repotag !== appSpecFormatted.repotag) { // this is OK. <= v3 cannot change, v4 can but does not have this in specifications as its compose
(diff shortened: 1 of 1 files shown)


MorningLightMountain713 (approved):
Ack

PR #1628, opened 2025-12-20
feat: Add graceful SIGTERM handling with fluxnodesigterm broadcast for faster app respawning

Cabecinha84:
## Summary
  - Implements graceful SIGTERM signal handling for FluxOS
  - Introduces new `fluxnodesigterm` message type to notify peers when a node is shutting down
  - Enables faster app respawning by reducing TTL on app location records when a node signals shutdown

  ## Problem
  When a node operator shuts down or reboots their computer, FluxOS receives a SIGTERM signal but previously had no handler for it. This meant:
  - The node would terminate immediately without notifying peers
  - App location records would remain in the network database for up to 2 hours (TTL of 7500 seconds)
  - Apps wouldn't respawn on new nodes until the old records expired naturally

  ## Solution
  This PR adds SIGTERM handling that:
  1. Tracks running apps in a lightweight cache (Set of app names)
  2. On SIGTERM, broadcasts a `fluxnodesigterm` message to all peers if the node was running apps
  3. Peers receiving the message update the app location records to expire in 7 minutes
  4. Reduced initial `fluxapprunning` broadcast delay from 2 minutes to 1 minute to ensure rebooted nodes re-register before records expire

  ## Changes

  ### New Files/Features
  - **`fluxnodesigterm` message type**: New broadcast message containing `ip` and `broadcastedAt`

  ### Modified Files

  **`apiServer.js`**
  - Added SIGTERM signal handler
  - Broadcasts `fluxnodesigterm` to peers if node was running apps

  **`ZelBack/src/services/utils/globalState.js`**
  - Added `runningAppsCache` (Set) to track broadcasted running app names

  **`ZelBack/src/services/appMessaging/peerNotification.js`**
  - Updates `runningAppsCache` when broadcasting `fluxapprunning` messages

  **`ZelBack/src/services/appLifecycle/appUninstaller.js`**
  - Removes app from `runningAppsCache` when sending `fluxappremoved`

  **`ZelBack/src/services/appLifecycle/advancedWorkflows.js`**
  - Removes app from `runningAppsCache` when sending `fluxappremoved` (2 locations)

  **`ZelBack/src/services/fluxCommunicationUtils.js`**
  - Added validation case for `fluxnodesigterm` message type

  **`ZelBack/src/services/fluxCommunication.js`**
  - Added `handleNodeSigtermMessage()` function that:
    - Validates the message timestamp (4-minute window)
    - Checks if the IP has apps in `globalAppsLocations` database
    - If found, updates `broadcastedAt` to make records expire 7 minutes after sigterm
    - Rebroadcasts to other peers
  - Added message handlers in both incoming and outgoing connection handlers

  **`ZelBack/src/services/serviceManager.js`**
  - Changed first `fluxapprunning` broadcast from 2 minutes to 1 minute after startup

  ## Timing Design
  - SIGTERM at time T → app location records set to expire at T + 7 minutes
  - Node reboots and FluxOS restarts
  - First `fluxapprunning` broadcast after 1 minute of startup
  - **6 minutes available** for complete reboot cycle (shutdown + POST + OS boot + service start)

### DIFF
--- a/apiServer.js
+++ b/apiServer.js
@@ -13,7 +13,6 @@ const fs = require('node:fs');
 const http = require('node:http');
 const https = require('node:https');
 const path = require('node:path');
-const { execSync } = require('node:child_process');
 const { watch } = require('node:fs/promises');
 
 const axios = require('axios').default;
@@ -163,16 +162,16 @@ function setAxiosDefaults(socketIoServers) {
  * @param {{delay?: number, exitCode?: number}} options
  */
 async function logErrorAndExit(msg, options = {}) {
-  const delay = options.delay || 1_000;
+  const delayMs = options.delay || 1_000;
   const exitCode = options.exitCode || 0;
 
   if (msg) log.error(msg);
 
-  const delayS = Math.round((delay / 1000) * 100) / 100;
+  const delayS = Math.round((delayMs / 1000) * 100) / 100;
 
   log.info(`Waiting: ${delayS}s, before exiting with code: ${exitCode}`);
 
-  await new Promise((r) => { setTimeout(r, delay); });
+  await serviceHelper.delay(delayMs);
   process.exit(exitCode);
 }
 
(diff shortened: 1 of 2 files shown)


MorningLightMountain713 (approved):
Ack

PR #1629, opened 2025-12-23
Bugfix: Token provider auth

MorningLightMountain713:
This PR solves several issues we were having with authentication for token providers to pull images. Primarily we wern't passing the repoauth to testapp, so it would fail.

There was also an issue where the namespace can be null for token provider image tags.

Last issue with Azure uses a different auth scheme that ECR, and GAR, so we now account for that in the auth scheme.

Registered the following apps and made sure they can all "test the app" successfully:

non enterprise, using nginx:latest
enterprise using nginx:latest
enterprise using a dockerhub private repo with username:docker pat
enterprise using AWS
enterprise using Google
enterprise using Azure

I'm hoping these are the last changes we have to do here for docker authentication and specification changes in general, until we get a proper testing environment (I will build it) as testing changes is no longer practical, and completely error prone.

### DIFF
--- a/ZelBack/src/services/utils/imageVerifier.js
+++ b/ZelBack/src/services/utils/imageVerifier.js
@@ -305,6 +305,20 @@ class ImageVerifier {
       return;
     }
 
+    // For Bearer auth with pre-obtained tokens (Azure ACR, Google GAR)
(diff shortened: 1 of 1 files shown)


Cabecinha84 (approved):
Ack

PR #1632, opened 2025-12-24
v8.1.1

Cabecinha84:
---
  Summary

  This release includes bug fixes and improvements to enterprise app handling, app pricing calculations, and hardware discount policies.

  Changes

  1. Enterprise App Removal on Non-Arcane Nodes (#1631)

  Problem: Enterprise apps (v8+) were being processed through the full redeployment flow before being checked for arcaneOS compatibility, leading to unnecessary processing and duplicate removal code in multiple locations.

  Solution:
  - Moved the enterprise/arcaneOS compatibility check to occur before any redeployment processing begins
  - Replaced manual broadcast logic with the existing appUninstaller.removeAppLocally() function with proper flags for force removal and peer notification
  - Removed duplicate arcaneOS checks that existed in two separate code paths
  - Result: Cleaner code, reduced duplication (-76 lines), and more efficient enterprise app handling

  2. Fix appPricePerMonth Calculation

  Problem: The pricing calculation for v4+ apps had inconsistencies and the v8 per-instance pricing was not correctly implemented.

  Changes:
  - Simplified per-instance pricing for v4+ apps:
    - Base price is calculated for resources, then divided by 3 to get pricePerInstance
    - Final price = pricePerInstance × number of instances
    - Additional instances surcharge: each instance beyond the first adds pricePerInstance to the total
  - Fixed cheap app threshold:
    - Changed threshold from $1.50 to $0.50 for the flat-rate additional instance surcharge
    - Flat $0.50/instance surcharge only applies when pricePerInstance < $0.50 AND more than 2 additional instances
  - Removed incorrect divisions for v3 tiered apps:
    - Removed erroneous /3 divisions on CPU, RAM, and HDD totals for tiered v3 apps
  - Removed instance surcharges for v3 and below:
    - Apps v3 and below don't have configurable instances, so removed the additional instance surcharge logic from those code paths

  3. Hardware Discount Policy Update

  Problem: Hardware tier discounts (20% for small apps, 10% for medium apps) were being applied to all apps regardless of instance count, allowing large-scale deployments to receive discounts intended for small deployments.

  Solution: Hardware discounts now only apply when:
  - App version ≤ 3, OR
  - App instances < 5

  This ensures that apps with 5+ instances pay the full resource-based price without hardware tier discounts.

  Pricing Examples (v8 app, 0.2 cores, 500MB RAM, 1GB HDD, no syncthing)

  | Instances | Old Behavior        | New Behavior   |
  |-----------|---------------------|----------------|
  | 1         | $0.99 (min)         | $0.99 (min)    |
  | 3         | $0.99 (min)         | $0.99 (min)    |
  | 4         | ~$1.35              | $1.35          |
  | 5+        | HW discount applied | No HW discount |
  | 100       | ~$31.80             | $49.69         |

  Files Changed

  - ZelBack/src/services/appLifecycle/advancedWorkflows.js - Enterprise app removal refactor
  - ZelBack/src/services/utils/appUtilities.js - Pricing calculation fixes
  - ZelBack/src/services/utils/appSpecHelpers.js - HW discount policy update
  - package.json - Version bump to 8.1.1


alihm (approved):
Ack

PR #1633, opened 2025-12-26
Bugfix/app uninstall

MorningLightMountain713:
For apps that have been upgraded from non enterprise to enterprise - we have non Arcane nodes that now have encrypted apps in their local apps db, as can be seen here:

`http://149.154.178.30:16127/apps/installedapps`

Screenshot 2025-12-26 at 9 50 15 AM

This node can't remove the app as the spec decryption fails. It fails in both `checkAndRemoveEnterpriseAppsOnNonArcane` and the `reinstallOldApplications` methods.

In order to remove these apps, we need to restore the previous app spec; this PR does that for these nodes. We also move the check before the db is updated, so this won't happen in the future. 

~~I'll see if I can test this against one of my non Arcane nodes~~

Edit: I replicated this on one of my non Arcane nodes - verified that this fix works.

### DIFF
--- a/ZelBack/src/services/appLifecycle/advancedWorkflows.js
+++ b/ZelBack/src/services/appLifecycle/advancedWorkflows.js
@@ -100,6 +100,77 @@ async function getStrictApplicationSpecifications(appName) {
 }
 
 /**
+ * Find and restore non-enterprise app specifications for proper removal.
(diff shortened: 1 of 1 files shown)


Cabecinha84:
like it is, the code will only be called if there is an update on those apps. If I udnerstood properly.

Cabecinha84 (approved):
ack

PR #1635, opened 2025-12-30
Feature/security

Cabecinha84:


ZelBack/config/default.js
```diff
@@ -10,6 +10,16 @@ module.exports = {
     allowedPorts: [16127, 16137, 16147, 16157, 16167, 16177, 16187, 16197],
     apiport: 16127, // homeport is -1, ssl port is +1
     fluxNodeServiceAddress: '169.254.43.43',
+    // CORS configuration (SEC-04 fix)
+    // Set allowedOrigins to an array of specific origins to restrict cross-origin requests
+    // Use '*' only for development/testing (insecure for production)
+    // Example: ['https://dashboard.runonflux.io', 'https://home.runonflux.io']
+    cors: {
+      // In development, allow all origins. In production, this should be restricted.
+      allowedOrigins: isDevelopment ? '*' : ['https://dashboard.runonflux.io', 'https://home.runonflux.io', 'https://jetpack.runonflux.io'],
```
MorningLightMountain713, line 19: This is missing .com and cloud

ZelBack/src/services/IOUtils.js
```diff
@@ -6,11 +6,159 @@ const log = require('../lib/log');
...
+    if (parsedUrl.protocol !== 'http:' && parsedUrl.protocol !== 'https:') {
+      return { valid: false, error: `Invalid protocol: ${parsedUrl.protocol}` };
+    }
+
+    // Block common metadata endpoints
+    const blockedHostnames = [
+      'metadata.google.internal',
+      'metadata.gcp.internal',
+      '169.254.169.254', // AWS/GCP/Azure metadata
+      '169.254.170.2', // AWS ECS metadata
```
MorningLightMountain713, line 87: This needs our internal 169.254.43.43 nodeinfo address

ZelBack/src/services/IOUtils.js
```diff
@@ -6,11 +6,159 @@ const log = require('../lib/log');
...
+
+/**
+ * Validates a URL to prevent SSRF attacks.
+ * Blocks private IP ranges, localhost, and metadata endpoints.
+ * @param {string} urlString - The URL to validate.
+ * @param {object} options - Validation options.
+ * @param {boolean} options.allowHttp - Whether to allow HTTP (default: false, HTTPS only).
+ * @returns {Promise<{valid: boolean, error?: string}>} - Validation result.
+ */
+async function validateUrl(urlString, options = {}) {
```
MorningLightMountain713, line 68: This entire function is unnecessary / it's not the right way to go about this.

The only things that use this are downloadFileFromUrl and getRemoteFileSize in (IOUtils), which I would argue shouldn't exist. They are specific to the backup and restore functionality.

We should validate at the source point where it is needed (we know what endpoints are allowed as they are our endpoints)

Much more simple

PR #1636, opened 2025-12-30
Feature/security

Cabecinha84:


ZelBack/src/services/appSystem/fileSystemManager.js
```diff
@@ -87,16 +89,18 @@ async function renameAppsObject(req, res) {
...
       }
       const fileURIArray = fileURI.split('%2F');
       fileURIArray.pop();
       if (fileURIArray.length > 0) {
         const renamingFolder = fileURIArray.join('/');
-        newfullpath = `${appVolumePath[0].mount}/${renamingFolder}/${newname}`;
+        // Sanitize the combined path as well
+        newfullpath = sanitizePath(`${renamingFolder}/${newname}`, appVolumePath[0].mount);
       }
       const cmd = `sudo mv -T "${oldfullpath}" "${newfullpath}"`;
```
alihm: We can use `serviceHelper.runCommand('mv', { runAsRoot: true, params: ['-T', oldfullpath, newfullpath] })`  to further harden this.

ZelBack/src/services/backupRestoreService.js
alihm: same here, safer to use `await serviceHelper.runCommand('chmod', { runAsRoot: true,params: ['777', filepath]});`

ZelBack/src/services/utils/pathSecurity.js
```diff
@@ -0,0 +1,354 @@
...
+ * to prevent bypass attempts.
+ *
+ * @param {string} inputPath - Path to normalize
+ * @returns {string} Path with all backslashes converted to forward slashes
+ */
+function normalizePathSeparators(inputPath) {
+  if (!inputPath || typeof inputPath !== 'string') {
+    return inputPath;
+  }
+  return inputPath.replace(/\\/g, '/');
```
MorningLightMountain713: I don't agree with this. We are linux only. If people are using backslashes in paths... we should reject the path completey

ZelBack/src/services/utils/pathSecurity.js
```diff
@@ -0,0 +1,354 @@
...
+ * // Throws error
+ * sanitizePath('..', '/apps/myapp')  // Error: directory traversal
+ * sanitizePath('../other', '/apps/myapp')  // Error: directory traversal
+ * sanitizePath('foo\0bar', '/apps/myapp')  // Error: null byte
+ */
+function sanitizePath(userPath, basePath, options = {}) {
+  const { strict = true } = options;
+
+  // If no user path provided, return the base path
+  if (!userPath || typeof userPath !== 'string') {
```
MorningLightMountain713: If userPath is defined and not a string, this should raise

ZelBack/src/services/utils/pathSecurity.js
```diff
@@ -0,0 +1,354 @@
...
+    return basePath;
+  }
+
+  // === LAYER 1: Blocklist checks (fast rejection of known-bad patterns) ===
+
+  // Check for null bytes (can bypass security checks in some systems)
+  if (userPath.includes('\0')) {
```
MorningLightMountain713, line 188: This should just check for backslashes, then raise if found (and we delete normalizePathSeparators)

ZelBack/src/services/utils/pathSecurity.js
```diff
@@ -0,0 +1,354 @@
...
+    throw new Error('Invalid path: null bytes not allowed');
+  }
+
+  // Normalize backslashes to forward slashes BEFORE any path operations
+  // This ensures consistent handling and prevents bypass via backslashes
+  const normalizedUserPath = normalizePathSeparators(userPath);
+
+  // Check for traversal patterns in the normalized path
+  // We check each component separately to catch '..' as a path component
+  const components = normalizedUserPath.split('/');
```
MorningLightMountain713: The only check here that really matters is this one:

resolvedPath !== normalizedBase && !resolvedPath.startsWith(normalizedBase + path.sep

The rest is just code bloat

Assuming that we have already rejected if it contains backslashes. We could also use `path.normalize` instead of reinventing the wheel.

Can also do something as simple as this:

const relative = path.relative(parent, dir);
const isRelative = relative && !relative.startsWith('..') && !path.isAbsolute(relative);

ZelBack/src/services/backupRestoreService.js
```diff
@@ -18,10 +19,22 @@ const appsFolder = `${appsFolderPath}/`;
  * @returns {boolean} - True if the filepath is valid, otherwise false.
  */
 function pathValidation(filepath) {
+  // Security: Check for directory traversal attempts
+  if (!filepath || filepath.includes('..') || filepath.includes('\0')) {
```
MorningLightMountain713: This shouldn't be checking this manually here - it's part of the pathSecurity module

ZelBack/src/services/backupRestoreService.js
```diff
@@ -18,10 +19,22 @@ const appsFolder = `${appsFolderPath}/`;
...
+    return false;
+  }
+
   const pathStart = filepath.startsWith(appsFolder);
   let uploadType = null;
 
   if (pathStart) {
+    // Security: Verify the resolved path stays within appsFolder
+    const resolvedPath = path.resolve(filepath);
+    const resolvedAppsFolder = path.resolve(appsFolder);
```
MorningLightMountain713: This should also use the pathSecurity module

Cabecinha84:
Plus added SSRF protection

alihm (approved):
Ack

PR #1641, opened 2026-01-06
eat: Allow designated users to extend app subscriptions on behalf of app owners

Cabecinha84:
---
  PR Description

  Summary

  - Add support for designated addresses that can extend application subscriptions on behalf of app owners
  - Implement expire-only update verification to ensure these users can only extend subscriptions, not modify app configurations

  Changes

  Configuration (ZelBack/config/default.js)
  - Added new usersToExtend array in fluxapps config containing authorized addresses
  - Initial authorized address: 196GJWyLxzAw3MirTT7Bqs2iGpUQio29GH

  Message Verification (ZelBack/src/services/appMessaging/messageVerifier.js)
  - Added isExpireOnlyUpdate() helper function that compares app specs to verify only the expire property changed (ignoring system-added fields like height and hash)
  - Extended verifyAppMessageUpdateSignature() to check signatures from usersToExtend addresses when the app owner signature fails
  - Restricted usersToExtend functionality to expire-only updates for security

  How It Works

  1. When an app update signature verification fails for the app owner, the system checks if the signature matches any address in usersToExtend
  2. If a match is found, it validates that the update ONLY changes the expire property
  3. If both conditions pass, the update is accepted and logged

  Security Considerations

  - Users in usersToExtend can only extend subscriptions - they cannot modify any other app specifications
  - The comparison uses deep JSON equality to ensure no other properties are changed
  - Per the comment, addresses can be added to the list over time but should not be removed to maintain backwards compatibility

### DIFF
--- a/ZelBack/src/services/appMessaging/messageVerifier.js
+++ b/ZelBack/src/services/appMessaging/messageVerifier.js
@@ -283,34 +283,27 @@ async function verifyAppMessageUpdateSignature(type, version, appSpec, timestamp
   }
 
   // Check if usersToExtend can sign this update (only for expire-only changes)
-  if (isValidSignature !== true) {
-    const usersToExtend = config.fluxapps.usersToExtend || [];
-    if (usersToExtend.length > 0) {
+  const usersToExtend = config.fluxapps.usersToExtend || [];
+  if (isValidSignature !== true && usersToExtend.length > 0) {
+    // eslint-disable-next-line global-require
+    const registryManager = require('../appDatabase/registryManager');
+    const existingSpec = await registryManager.getApplicationGlobalSpecifications(appSpec.name);
+    if (existingSpec) {
+      // For v8+ enterprise apps, we need to decrypt the new spec before comparing
+      let newSpecToCompare = appSpec;
+      if (appSpec.version >= 8 && appSpec.enterprise) {
+        // eslint-disable-next-line global-require
+        const { checkAndDecryptAppSpecs } = require('../utils/enterpriseHelper');
+        newSpecToCompare = await checkAndDecryptAppSpecs(appSpec, { daemonHeight, owner: existingSpec.owner });
+      }
       // Check if signature matches any of the usersToExtend addresses
       // eslint-disable-next-line no-restricted-syntax
       for (const userToExtend of usersToExtend) {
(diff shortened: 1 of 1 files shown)


ZelBack/src/services/appMessaging/messageVerifier.js
```diff
@@ -254,6 +281,41 @@ async function verifyAppMessageUpdateSignature(type, version, appSpec, timestamp
...
+      // Check if signature matches any of the usersToExtend addresses
+      // eslint-disable-next-line no-restricted-syntax
+      for (const userToExtend of usersToExtend) {
+        const isValidUserToExtendSignature = signatureVerifier.verifySignature(messageToVerify, userToExtend, signature);
+        if (isValidUserToExtendSignature === true) {
+          // Verify this is an expire-only update by fetching existing app specs
+          // eslint-disable-next-line global-require
+          const registryManager = require('../appDatabase/registryManager');
+          // eslint-disable-next-line no-await-in-loop
+          const existingSpec = await registryManager.getApplicationGlobalSpecifications(appSpec.name);
```
MorningLightMountain713: This should be outside the for loop

ZelBack/src/services/appMessaging/messageVerifier.js
```diff
@@ -254,6 +281,41 @@ async function verifyAppMessageUpdateSignature(type, version, appSpec, timestamp
     // we can just use the btc / eth verifier as v7 specs came out at 1688749251
     isValidSignature = signatureVerifier.verifySignature(messageToVerifyC, appOwner, signature);
   }
+
+  // Check if usersToExtend can sign this update (only for expire-only changes)
+  if (isValidSignature !== true) {
+    const usersToExtend = config.fluxapps.usersToExtend || [];
+    if (usersToExtend.length > 0) {
```
MorningLightMountain713: Return early instead of nesting inside another if (cyclomatic complexity)

ZelBack/src/services/appMessaging/messageVerifier.js
Cabecinha84: should be resolved

ZelBack/src/services/appMessaging/messageVerifier.js
Cabecinha84: should be resolved

MorningLightMountain713 (approved):
Ack

PR #1651, opened 2026-01-14
feat: Add CloudUI auto-update service for legacy nodes without watchdog

Cabecinha84:
## Summary

  - Add `cloudUIUpdateService.js` to automatically check and update CloudUI on FluxOS startup
  - Integrate the service into `serviceManager.js` to run during node initialization
  - Ensures legacy nodes not running ArcaneOS/watchdog receive UI updates

  ## Problem

  Legacy Flux nodes that are not running ArcaneOS do not have the watchdog service running. As a result, these nodes are not picking up new CloudUI releases and remain stuck on outdated UI versions. This creates an inconsistent user experience across the network.

  ## Solution

  Implemented a CloudUI update check service that runs on FluxOS startup (in `serviceManager.js`). This mirrors the update logic used in the watchdog project but integrates directly into FluxOS, ensuring all nodes—regardless of whether they have watchdog running—will receive UI updates.

  ## Implementation Details

  ### New Service: `ZelBack/src/services/cloudUIUpdateService.js`

  The service performs the following checks on startup:

  1. **CloudUI folder missing or empty** → Runs `npm run update:cloudui` to install
  2. **Version file missing** → Runs update to ensure proper installation
  3. **Version file exists** → Fetches latest release from GitHub API and compares SHA256 hashes

  Version comparison logic (same as watchdog):
  - Fetches latest release from `https://api.github.com/repos/RunOnFlux/fluxos-frontend/releases/latest`
  - Only proceeds if release targets the `master` branch
  - Extracts SHA256 hash from `dist.tar.gz` asset digest
  - Compares with local hash stored in `CloudUI/version`
  - If hashes differ, executes `npm run update:cloudui`
  - Verifies update success by re-checking hash after update

  ### Integration: `ZelBack/src/services/serviceManager.js`

  - Added import for `cloudUIUpdateService`
  - Calls `checkAndUpdateCloudUI()` early in `startFluxFunctions()`, after port validation but before heavy service initialization

  ## Testing

  - Tested with missing CloudUI folder → Successfully installs UI
  - Tested with existing CloudUI and matching version → Skips update correctly
  - Tested with outdated version hash → Successfully detects and updates

  ## Related

  This implementation is based on the update mechanism in the watchdog project, ensuring consistency across both update paths.

### DIFF
--- a/ZelBack/src/services/cloudUIUpdateService.js
+++ b/ZelBack/src/services/cloudUIUpdateService.js
@@ -10,6 +10,9 @@ const PROJECT_ROOT = path.join(__dirname, '..', '..', '..');
 const CLOUDUI_DIR = path.join(PROJECT_ROOT, 'CloudUI');
 const VERSION_FILE = path.join(CLOUDUI_DIR, 'version');
 
+// ArcaneOS nodes have watchdog handling CloudUI updates
+const isArcaneOS = Boolean(process.env.FLUXOS_PATH);
+
 /**
  * Checks if CloudUI folder exists and has content
  * @returns {boolean}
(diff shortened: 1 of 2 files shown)


MorningLightMountain713 (changes requested):
This needs to only run on legacy.

What happens if a legacy node is running watchdog... and they both try to update at the same time? It should probably detect if watchdog is active on the system and only run if not

MorningLightMountain713 (approved):
Ack

PR #1653, opened 2026-01-14
set restart always for user

Cabecinha84:
Set docker restart policy to always for a specific user. This will be here until we have new app specs allowing different selections.

### DIFF
--- a/ZelBack/src/services/utils/volumeConstructor.js
+++ b/ZelBack/src/services/utils/volumeConstructor.js
@@ -8,6 +8,7 @@
 const log = require('../../lib/log');
 const { MountType } = require('./mountParser');
 const { appsFolder } = require('./appConstants');
+const config = require('../../../config/default');
 
 /**
  * Get app identifier with proper flux prefix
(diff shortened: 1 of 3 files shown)


ZelBack/src/services/utils/volumeConstructor.js
```diff
@@ -249,15 +249,20 @@ function constructVolumes(parsedMounts, identifier, appName, fullAppSpecs, appSp
...
  * @returns {string} Docker restart policy
  */
-function getRestartPolicy(flags) {
+function getRestartPolicy(flags, owner) {
   // 'g' flag (primary/standby) requires 'no' restart policy
   if (flags.includes('g')) {
     return 'no';
   }
+  // Specific owner gets 'always' restart policy
+  if (owner === '16mzUh6byiQr7rnYQxKraDbeBPsEHYpSTW') {
```
MorningLightMountain713: We should extract this out into the config (as an array), so at least we can update them if needed without having to hardcode. (until we update the app specs)

MorningLightMountain713 (changes requested):

ZelBack/src/services/utils/volumeConstructor.js
Cabecinha84: fixed

PR #1660, opened 2026-01-19
Enhanced geolocation service with static IP detection, datacenter support, and IP change tracking

Cabecinha84:
## Summary                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                       
  This PR significantly enhances the Flux geolocation service with improved static IP detection, datacenter node identification, and IP stability tracking. It also introduces datacenter requirements for v8 apps with enterprise-only access control.                                                                
                                                                                                                                                                                                                                                                                                                       
  ## Key Features                                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                       
  ### 1. Enhanced Static IP Detection                                                                                                                                                                                                                                                                                  
  - **Network Interface Check**: Detects if node has a public IP directly configured on its network interface (strong indicator of static IP)                                                                                                                                                                          
  - **IP Stability Tracking**: Tracks `lastIpChangeDate` to determine static IP status through behavioral patterns                                                                                                                                                                                                     
  - **Graceful Migration**: Existing nodes without IP change history are not penalized and are immediately considered static if they have a public IP on their interface                                                                                                                                               
                                                                                                                                                                                                                                                                                                                       
  ### 2. Datacenter Node Identification                                                                                                                                                                                                                                                                                
  - Adds `dataCenter` boolean field based on ip-api.com's `hosting` data                                                                                                                                                                                                                                               
  - Falls back to checking organization name against known datacenter providers (Hetzner, OVH, Contabo, etc.)                                                                                                                                                                                                          
  - Persists datacenter status to database for consistent identification                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                       
  ### 3. Datacenter App Requirements (v8+)                                                                                                                                                                                                                                                                             
  - New `datacenter` field in app specs (v8+): `null`, `true`, or `false`                                                                                                                                                                                                                                              
  - Apps with `datacenter: true` will only run on datacenter nodes                                                                                                                                                                                                                                                     
  - **Enterprise-only restriction**: `datacenter: true` is only accepted from app owners in the configurable `enterpriseAppOwners` list                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                       
  ### 4. Intelligent App Spawning Priority                                                                                                                                                                                                                                                                             
  - Static IP nodes delay spawning non-static-IP apps by ~27 minutes, prioritizing apps that need static IP                                                                                                                                                                                                            
  - Datacenter nodes delay spawning non-datacenter apps by ~27 minutes, prioritizing apps that require datacenter resources                                                                                                                                                                                            
  - Ensures specialized node resources are allocated to apps that specifically need them                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                       
  ## Changes                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                       
  | File | Description |                                                                                                                                                                                                                                                                                               
  |------|-------------|                                                                                                                                                                                                                                                                                               
  | `ZelBack/src/services/geolocationService.js` | IP stability tracking, datacenter detection, private IP helpers |                                                                                                                                                                                                   
  | `ZelBack/src/services/appRequirements/appValidator.js` | Datacenter field validation, enterprise owner check |                                                                                                                                                                                                     
  | `ZelBack/src/services/appRequirements/hwRequirements.js` | Datacenter requirement check at runtime |                                                                                                                                                                                                               
  | `ZelBack/src/services/appLifecycle/appSpawner.js` | Priority spawning for static IP and datacenter nodes |                                                                                                                                                                                                         
  | `ZelBack/src/services/fluxNetworkHelper.js` | Geolocation updates on IP changes |                                                                                                                                                                                                                                  
  | `ZelBack/config/default.js` | New `enterpriseAppOwners` configuration list |                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                       
  ## Database Schema Updates                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                       
  New fields in `geolocation` collection:                                                                                                                                                                                                                                                                              
  - `dataCenter` (boolean): Whether node is in a datacenter                                                                                                                                                                                                                                                            
  - `lastIpChangeDate` (timestamp): Last time the node's IP changed                                                                                                                                                                                                                                                    
  - `staticIp` (boolean): Whether node has a static IP                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                       
  ## Configuration                                                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                       
  New config option in `default.js`:                                                                                                                                                                                                                                                                                   
  ```javascript                                                                                                                                                                                                                                                                                                        
  enterpriseAppOwners: [                                                                                                                                                                                                                                                                                               
    // Flux IDs allowed to deploy apps with datacenter=true                                                                                                                                                                                                                                                            
  ]                                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                       
  Test Plan                                                                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                       
  - Verify static IP detection works for nodes with public IP on interface                                                                                                                                                                                                                                             
  - Verify datacenter detection correctly identifies hosting providers                                                                                                                                                                                                                                                 
  - Verify apps with datacenter: true only spawn on datacenter nodes                                                                                                                                                                                                                                                   
  - Verify non-enterprise owners cannot set datacenter: true                                                                                                                                                                                                                                                           
  - Verify datacenter nodes delay spawning non-datacenter apps                                                                                                                                                                                                                                                         
  - Verify static IP nodes delay spawning non-static-IP apps

### DIFF
--- a/ZelBack/src/services/fluxNetworkHelper.js
+++ b/ZelBack/src/services/fluxNetworkHelper.js
@@ -157,21 +157,6 @@ async function hasPublicIpOnInterface() {
     return false;
   } catch (error) {
     log.error(`Failed to check network interfaces via routing table: ${error.message}`);
-    // Fallback to simple interface enumeration
-    try {
-      const interfaces = os.networkInterfaces();
-      for (const name of Object.keys(interfaces)) {
-        for (const iface of interfaces[name]) {
-          if (iface.family === 'IPv4' && !iface.internal) {
-            if (!serviceHelper.isNonRoutableAddress(iface.address)) {
-              return true;
-            }
-          }
-        }
-      }
-    } catch (fallbackError) {
-      log.error(`Fallback interface check also failed: ${fallbackError.message}`);
-    }
     return false;
   }
 }


ZelBack/src/services/fluxNetworkHelper.js
```diff
@@ -817,12 +817,34 @@ async function adjustExternalIP(ip) {
...
       if (apps.status === 'success' && apps.data.length > 0) {
         apps = apps.data;
         let appsRemoved = 0;
         // eslint-disable-next-line no-restricted-syntax
         for (const app of apps) {
+          // Decrypt enterprise app specs if needed (v8+ with enterprise field)
+          let appSpecs = app;
+          try {
+            // eslint-disable-next-line no-await-in-loop
+            appSpecs = await enterpriseHelper.checkAndDecryptAppSpecs(app);
```
MorningLightMountain713: This should get moved inside the if statement. We only need to decrypt if we are removing. Then the if statement can just use `app`

MorningLightMountain713: (then the decrypt error should be a hard fail)

ZelBack/src/services/geolocationService.js
```diff
@@ -9,15 +10,72 @@ const { geolocation: geolocationCollection } = config.database.local.collections
...
 let execution = 1;
 const staticIpOrgs = ['hetzner', 'ovh', 'netcup', 'hostnodes', 'contabo', 'hostslim', 'zayo', 'cogent', 'lumen'];
+const staticIpStabilityDays = 10;
+
+/**
+ * Checks if an IP address is in a private range
+ * @param {string} ip - The IP address to check
+ * @returns {boolean} True if the IP is private
+ */
+function isPrivateIp(ip) {
```
MorningLightMountain713: I'm sure this is defined somewhere else already... if not it should be and then we import it

ZelBack/src/services/geolocationService.js
```diff
@@ -9,15 +10,72 @@ const { geolocation: geolocationCollection } = config.database.local.collections
...
+
+  return false;
+}
+
+/**
+ * Checks if the node has a public IP directly configured on a network interface
+ * This is a strong indicator of a static IP (data center/VPS/dedicated server)
+ * @returns {boolean} True if a public IP is configured on an interface
+ */
+function hasPublicIpOnInterface() {
```
MorningLightMountain713: This should be in the networkHelper, not geolocationService

ZelBack/src/services/geolocationService.js
```diff
@@ -9,15 +10,72 @@ const { geolocation: geolocationCollection } = config.database.local.collections
...
+  if (parts[0] === 100 && parts[1] >= 64 && parts[1] <= 127) return true;
+
+  return false;
+}
+
+/**
+ * Checks if the node has a public IP directly configured on a network interface
+ * This is a strong indicator of a static IP (data center/VPS/dedicated server)
+ * @returns {boolean} True if a public IP is configured on an interface
+ */
```
MorningLightMountain713: We want to make this more robust. This is only checking that ANY external interface has a public ip. Not necessarily the one we are routing over.

Here is some code that gets the interface via the routing table, makes sure the interface is up, the gets the ip for that interface. You can then check it's a public ip. This is safe on linux.

This should all go in our network helpers

```
const fs = require('node:fs/promises');
const os = require('node:os');

/**
 * Checks if a network interface is operationally up by reading its state from sysfs.
 * This is important because static routes can remain in the routing table even when
 * their interface is down.
 *
 * @param {string} interfaceName - The name of the network interface (e.g., 'eth0', 'wlan0')
 * @returns {Promise<boolean>} True if the interface is up, false otherwise
 */
async function isInterfaceUp(interfaceName) {
  try {
    const operstate = await fs.readFile(`/sys/class/net/${interfaceName}/operstate`, 'utf8');
    return operstate.trim() === 'up';
  } catch (error) {
    return false;
  }
}

/**
 * @typedef {Object} RouteInfo
 * @property {string} interface - The network interface name (e.g., 'eth0')
 * @property {string|null} gateway - The gateway IP address in dotted decimal notation
 * @property {number} metric - The route metric (lower values have higher priority)
 */

/**
 * Finds the default route with the lowest metric by parsing /proc/net/route.
 * Only returns routes that are marked as UP and whose interface is actually operational.
 * If multiple default routes exist, returns the one with the lowest metric.
 *
 * @returns {Promise<RouteInfo|null>} The default route information, or null if no valid default route exists
 * @throws {Error} If /proc/net/route cannot be read (not on Linux or permission denied)
 */
async function getDefaultRouteInterface() {
  let routeTable;

  try {
    routeTable = await fs.readFile('/proc/net/route', 'utf8');
  } catch (error) {
    if (error.code === 'ENOENT') {
      throw new Error('/proc/net/route not found - are you running on Linux?');
    } else if (error.code === 'EACCES') {
      throw new Error('Permission denied reading /proc/net/route');
    }
    throw error;
  }

  const lines = routeTable.split('\n');

  let bestRoute = null;
  let lowestMetric = Infinity;

  // Skip header line
  for (let i = 1; i < lines.length; i++) {
    const line = lines[i].trim();
    if (!line) continue;

    const fields = line.split(/\s+/);
    if (fields.length < 8) continue;

    const iface = fields[0];
    const destination = fields[1];
    const gateway = fields[2];
    const flagsHex = fields[3];
    const mask = fields[7];
    const metricStr = fields[6];

    // Validate hex values
    if (!/^[0-9A-Fa-f]+$/.test(destination) ||
      !/^[0-9A-Fa-f]+$/.test(mask) ||
      !/^[0-9A-Fa-f]+$/.test(flagsHex)) {
      continue;
    }

    // Check if this is a default route
    if (destination !== '00000000' || mask !== '00000000') {
      continue;
    }

    const flags = parseInt(flagsHex, 16);
    const metric = parseInt(metricStr, 10);

    // Validate parsed numbers
    if (isNaN(flags) || isNaN(metric)) {
      continue;
    }

    // Check if route is UP (RTF_UP = 0x0001)
    if (!(flags & 0x0001)) {
      continue;
    }

    // Check if interface is actually up (important for static routes)
    const interfaceUp = await isInterfaceUp(iface);
    if (!interfaceUp) {
      continue;
    }

    // Prefer route with lowest metric
    if (metric < lowestMetric) {
      lowestMetric = metric;
      bestRoute = {
        interface: iface,
        gateway: hexToIP(gateway),
        metric: metric
      };
    }
  }

  return bestRoute;
}

/**
 * Converts a hexadecimal IP address from /proc/net/route format to dotted decimal notation.
 * The hex format is little-endian (reversed byte order).
 *
 * @param {string} hex - An 8-character hexadecimal string representing an IPv4 address
 * @returns {string|null} The IP address in dotted decimal notation (e.g., '192.168.1.1'),
 *                        or null if the hex string is invalid or represents 0.0.0.0
 * @example
 * hexToIP('0100007F') // Returns '127.0.0.1'
 * hexToIP('00000000') // Returns null
 */
function hexToIP(hex) {
  if (!hex || hex === '00000000') return null;

  // Validate hex string
  if (!/^[0-9A-Fa-f]{8}$/.test(hex)) {
    return null;
  }

  const bytes = [];
  for (let i = 0; i < 8; i += 2) {
    bytes.push(parseInt(hex.substr(i, 2), 16));
  }
  return bytes.reverse().join('.');
}

/**
 * Gets the IPv4 address assigned to a network interface.
 *
 * @param {string} interfaceName - The name of the network interface (e.g., 'eth0', 'wlan0')
 * @returns {string|null} The IPv4 address in dotted decimal notation, or null if the interface
 *                        doesn't exist or has no IPv4 address assigned
 */
function getInterfaceIP(interfaceName) {
  const interfaces = os.networkInterfaces();

  if (!interfaces[interfaceName]) {
    return null;
  }

  const ipv4 = interfaces[interfaceName].find(addr => addr.family === 'IPv4');
  return ipv4 ? ipv4.address : null;
}

// Usage
(async () => {
  try {
    const route = await getDefaultRouteInterface();

    if (!route) {
      console.log('No default route found');
      return;
    }

    console.log('Default interface:', route.interface);
    console.log('Gateway:', route.gateway);
    console.log('Metric:', route.metric);

    const ip = getInterfaceIP(route.interface);
    console.log('Interface IP:', ip);
  } catch (error) {
    console.error('Error:', error.message);
  }
})();
```

MorningLightMountain713 (changes requested):

ZelBack/src/services/fluxNetworkHelper.js
Cabecinha84: should be fixed

ZelBack/src/services/geolocationService.js
Cabecinha84: should be fixed

ZelBack/src/services/geolocationService.js
Cabecinha84: should be fixed

ZelBack/src/services/geolocationService.js
Cabecinha84: should be fixed

ZelBack/src/services/fluxNetworkHelper.js
```diff
@@ -41,6 +41,141 @@ const lruRateCache = cacheManager.rateLimitCache;
...
+          log.info(`Public IP ${ip} found on default route interface ${route.iface}`);
+          return true;
+        }
+      }
+    }
+
+    return false;
+  } catch (error) {
+    log.error(`Failed to check network interfaces via routing table: ${error.message}`);
+    // Fallback to simple interface enumeration
```
MorningLightMountain713: Don't fallback. This defeats the whole point of the above checks

Cabecinha84: should be fixed

PR #1711, opened 2026-04-01
fix: skip remote availability check when IP not in confirmed list

MorningLightMountain713:
## Summary
- Fixes a chicken-and-egg problem introduced by a367ec0db: when a node's IP changes, remote nodes reject the availability check because the new IP isn't in the confirmed list yet — but the node can't get confirmed at the new IP because the availability check always fails.
- Uses `socketAddressInFluxList` to detect the IP mismatch state and skip the remote probe, same pattern as the existing non-CONFIRMED skip (66882f75e).
- Adds test coverage for the new IP-not-in-list scenario.

## Test plan
- [x] Unit tests pass (`checkDeterministicNodesCollisions` — 8 passing)
- [x] Verified on live node: log shows `Our IP 47.206.56.185:16137 is not in the confirmed flux list. Skipping remote availability check.`

### DIFF
--- a/ZelBack/src/services/fluxNetworkHelper.js
+++ b/ZelBack/src/services/fluxNetworkHelper.js
@@ -1427,11 +1427,17 @@ async function checkDeterministicNodesCollisions() {
           }
         }
       }
-      // If this node is not CONFIRMED, remote nodes will reject the availability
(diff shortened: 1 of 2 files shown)


Cabecinha84 (approved):
ack

PR #1712, opened 2026-04-05
Fix ghost fields in zelappsinformation from $set accumulation

MorningLightMountain713:
## Summary

- `updateAppSpecifications` and `updateAppSpecsForRescanReindex` in `registryManager.js` used MongoDB `$set` to upsert into `zelappsinformation`. When an app upgraded across spec versions (e.g. v1→v4), `$set` only added new fields without removing old ones, leaving ghost flat fields (`repotag`, `cpu`, `ram`, `hdd`, `tiered`, etc.) on v4+ documents that should only have `compose`.
- Changed both functions to use `replaceOne` so the document always matches exactly what the latest message contains.
- Added ghost field detection to `isReindexAppsInformationRequired` in `dbHelper.js` — checks for v4+ docs with a `repotag` field. Triggers a one-time reindex on first startup after deploy, then never fires again.

## Impact

- 4 apps affected per node: `DiBiFetch`, `Minecraft`, `PokerTH`, `EthereumNodeLight`
- All ~8000 nodes have the same replicated data
- `zelappsmessages` (permanent messages) are unaffected — they were always clean

## Test plan

- [x] Verified on chud: ghost detection triggered reindex on startup
- [x] Reindex completed in ~34 seconds
- [x] 638 docs before and after — no data loss
- [x] Per-version counts identical
- [x] All 4 affected apps now have correct shape (ghost fields removed)
- [x] DiBiFetch v5 matches other v5 specs exactly

### DIFF
--- a/ZelBack/src/services/appDatabase/registryManager.js
+++ b/ZelBack/src/services/appDatabase/registryManager.js
@@ -862,10 +862,10 @@ async function updateAppSpecsForRescanReindex(appSpecs) {
       // replaceOne instead of $set to avoid accumulating ghost fields
       // from prior spec versions (e.g. flat fields from v1-v3 lingering
       // after upgrade to v4+ compose format)
-      await database.collection(globalAppsInformation).replaceOne(query, appSpecs, options);
+      await dbHelper.replaceOneInDatabase(database, globalAppsInformation, query, appSpecs, options);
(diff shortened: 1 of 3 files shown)


Cabecinha84 (approved):
ack

PR #1713, opened 2026-04-08
Fix Flux app message synchronization: 3 distinct issues

MorningLightMountain713:
## Summary

Three independent bugs in the app message sync pipeline are causing slow but progressive state divergence across the network — every node holds a slightly different (and stale) view of which apps exist, which messages have been applied, and which have expired.

This PR fixes all three. After deploying to a test node, message counts climbed from `56,375 → 57,196` (within 26 of the best-maintained production node) in under an hour.

## Network state evidence

Sample of 23 well-maintained "steel raven" nodes (latest production release) queried for `/apps/hashes`. All have the same total hash count (57,303) — the chain is consistent. The divergence is in *which message bodies each node managed to fetch and store*:

| Node | Permanent | Missing | messageNotFound |
|---|---|---|---|
| node A (and 9 sibling cluster nodes) | 57,222 | 81 | 81 |
| node B | 57,221 | 82 | 82 |
| node C | 57,211 | 92 | 92 |
| node D | 57,209 | 94 | 94 |
| node E, node F | 57,206 | 97 | 97 |
| node G | 57,183 | 120 | 120 |
| node H | 57,179 | 124 | 124 |
| node I | 57,108 | 195 | 195 |
| node J | 57,002 | 301 | 301 |
| node K | 56,968 | 335 | 335 |
| node L | 56,949 | 354 | 238 (mid-sync) |
| node M | 56,673 | 630 | 630 |
| node N | 56,514 | **789** | 789 |
| **test node (this branch, post-fix)** | **57,196** | 107 | 28 |

`missing == messageNotFound` on every node — the retry budget is exhausted everywhere. The 81-message baseline shared by the best-maintained nodes is genuine ancient loss. What this PR fixes is the *additional* divergence — the gap between 81 and 789 across nodes running the same release.

Concrete example: `tiktokss` was retired by its owner via an `expire=100` update at h:2,490,157. Some nodes received that gossip live and correctly removed the app. Three nodes (incl. one of the steel ravens) never received it and are still running zombie containers — their `zelappshashes` has the chain commitment with `messageNotFound: true`, the body is unrecoverable on production code, and the app continues broadcasting `fluxapprunning`.

## Fix 1: Version upgrade policy blocking historical messages

`validateApplicationUpdateCompatibility` enforced a current-era policy ("version changes are only allowed when updating to v8") inside `storeAppTemporaryMessage`. This blocked historical v5→v6 / v6→v7 updates being replayed during hash sync — messages valid when originally accepted but illegal under current rules. ~22,000 "Version changes are only allowed" errors in the test node's logs prior to the fix.

Example: `GFSixtyFive` has 9 messages on healthy nodes (v5 → v6 → v7 → v7 → v7). Resyncing nodes stopped at the v5→v6 transition (h:1,317,088), only ever storing 5 of the 9.

**Fix:**
1. Removed the version policy check from `validateApplicationUpdateCompatibility` (now pure structural compatibility — components, names, repotag).
2. Removed the version policy check from `storeAppTemporaryMessage`.
3. Moved enforcement to the API submission boundary: `validateAppUpdate()` (appValidator.js) and `updateAppGlobaly()` (advancedWorkflows.js), with thin HTTP wrappers `verifyAppUpdateApi()` and `updateAppGlobalyApi()` left as the request/response surface.
4. Added `config.fluxapps.latestSupportedSpecVersion` instead of hardcoding 8.

Hash sync and gossip can now replay historical messages without policy interference; fresh user submissions still get the full check at the API layer.

## Fix 2: Ownership change race condition

When a user submits two updates in quick succession where the first changes ownership from X to Y, both messages pass intake verification against permanent storage as it stood when each arrived. After the first is promoted (changing the owner to Y), `checkAndRequestApp` promotes the second without re-verifying the signature — the original code comment said "temp message means its all ok."

On a fresh resync, messages are processed in height order. The second message fails its signature check (previous owner is now Y; it's signed by X) and is permanently unverifiable from a clean state.

Example: `wordpress1735018430692` h:1,880,961. The h:1,880,959 update changed ownership. h:1,880,961 is signed by the old owner `1GTMhsaa55...`, but `getPreviousAppSpecifications` correctly returns the new owner `1GygtXKccaXPbJfB...`.

**Fix:** Re-verify the signature against current permanent state at promotion time. Block processing is deterministic, so every node reaches the same accept/reject decision. Failed re-verifications fall through to the normal retry path; they don't pollute permanent storage.

## Fix 3: arcaneSender silently drops enterprise messages on hash sync

This is the largest of the three by impact. The check was added with good intent — to let non-arcane nodes know whether an enterprise message originated from a node that could properly validate it before they store it blind. But the implementation can't deliver that property.

`arcaneSender` is security theatre. It's a plain field in the message body — no signature over it, no proof of origin, no way for the receiver to verify the claim. Any node — arcane, non-arcane, or actively malicious — can set `arcaneSender: true` and the gate opens. As a security mechanism it has zero effective adversaries: anyone reading the source can bypass it. The only nodes it actually blocks are honest ones doing legitimate hash-sync.

This will be fixed on the next app specs version where we use attestations. (signed messages where every node has the attestation pubkey, similar to how the signed Arcane updates work)

The lifecycle:

1. `arcaneSender` is set on the temp message at submission time and travels with the body through gossip. Receivers store it in temp DB along with the message.
2. When the message is promoted from temp to permanent, the field is **dropped** — `checkAndRequestApp` constructs the permanent message without it. Confirmed empirically against production nodes:

```
mongo > db.zelappsmessages.findOne({"appSpecifications.version":8, type:"fluxappregister", arcaneSender:{$exists:true}})
NO permanent message has arcaneSender field
```

3. When a fresh node hash-syncs, it sends `fluxapprequest` to a peer, the peer reads from permanent storage (no `arcaneSender`), and `respondWithAppMessage` constructs the response with `arcaneSender: appMessage.arcaneSender || false` → always `false`.
4. The receiving node's intake check fires: `if (!message.arcaneSender) return new Error('...')`. This is `return new Error`, not `throw`. `handleAppMessages` checks `if (rebroadcastToPeers === true)` and ignores anything else. **No log entry. No retry suppression. The message silently disappears.** After 10 retries the hash gets `messageNotFound: true` and is never retried again until the next reconstruct cycle (~5 days), which runs the same broken sync against the same peers and gets the same result.

**Impact.** This explains the bulk of the network state divergence in the table above. Long-running nodes that were online during the original gossip window for an enterprise app caught it and stored it. Any node that missed that ~30-second pre-promotion window — for any reason: brief restart, network blip, peer churn, packet loss — could never recover the message via hash sync. Every node accumulates a different set of permanently-missing enterprise messages.

Confirmed on a steel raven node: `zelappshashes` for `tiktokss` h:2,490,157 has `message: false, messageNotFound: true`. The node has the chain commitment for the message and *cannot get it* on the current production code. It runs the obsolete version of the app indefinitely.

**Fix.** Removed the `arcaneSender` intake check from `storeAppTemporaryMessage`. Real security for app messages comes from mechanisms that *can't* be forged:

1. `verifyAppHash` — content must hash to the on-chain committed value
2. `verifyAppMessageSignature` / `verifyAppMessageUpdateSignature` — must be signed by the app owner's key (or, for updates, the current owner per permanent state)
3. The on-chain transaction itself — someone paid `minPrice` flux for the hash commitment
4. On arcane nodes: actual decryption + spec validation via `checkAndDecryptAppSpecs`

None of these depend on `arcaneSender`. The check was removed. The permanent message structure is unchanged (we did NOT add `arcaneSender` to permanent storage — that would just hide the underlying problem behind a schema change).

**Backwards compatibility.** Pre-upgrade nodes still have the intake check. To keep them functioning during the rollout window:

- `respondWithAppMessage` defaults to `arcaneSender: true` when responding from permanent storage (`?? true` instead of `|| false`). Old nodes accept the response; new nodes don't care.
- All submission paths still set `arcaneSender: isArcane` on outbound messages, so old nodes accept broadcasts from arcane new nodes.

The field can be removed entirely from submission paths and the temp DB schema in a future release after the network has rolled past this one.

## End-to-end validation

The fix node was wiped, given the fix branch, and observed across multiple sync cycles:

| Metric | Before fix | After fix (1h) | After fix (full sync) |
|---|---|---|---|
| Permanent messages | 56,375 | 56,980 | 57,196 |
| Total missing | 921 | 323 | 107 |
| messageNotFound | 900 | 28 | 28 |
| Pending (retrying) | 0 | 267 | 50 |

Subsequent test: 487 messages were manually deleted from the test node's permanent storage and reset to `message: false` in the hashes collection — including 30 enterprise updates, 20 random non-enterprise updates, 20 registrations, and the entire history of 3 apps with 5+ messages each (`mc`, `TestAppB`, `rustserver`). On restart, the fix branch successfully re-fetched all 487 via the ordinary sync loop. A second test node was wiped completely (databases dropped) and is currently bulk-syncing from cold to verify the fresh-sync path.

## Test plan

- [x] Unit tests passing (`env NODE_CONFIG_DIR=$PWD/tests/unit/globalconfig npx mocha tests/unit/messageStore.test.js tests/unit/messageVerifier.test.js tests/unit/appHashSyncService.test.js --exit` → 43 passing)
- [x] Test node 1: partial-state recovery (487 messages deleted, all re-fetched successfully)
- [x] Test node 2: cold-start full sync (databases dropped, bulk-syncing from chain rescan — in progress)

## Rollout

No wire-format breaks. Old nodes continue to gossip and respond to hash-sync requests as before. Upgraded nodes set the same `arcaneSender` field on outbound submissions and respond to hash-sync requests with `arcaneSender: true` from the compatibility shim, so old nodes don't reject anything during the rollout window.

Self-healing: each upgraded node will recover its previously-stuck `messageNotFound` hashes on its next `reconstructAppMessagesHashCollection` cycle (~5 days post-fork) without any operator intervention.

## What's NOT in this PR

A few things came up during investigation that we know about and are tracked separately:

1. **Promoted temp messages stay as zombies** in `zelappstemporarymessages` until the 60-min mongo TTL sweeps them. The promotion code writes to permanent and updates the hash flag but doesn't delete the temp row. Harmless to correctness, will be fixed in the next refactor.
2. **`getPreviousAppSpecifications` reads only from permanent**, which means hash-sync batches that deliver children before parents can race at intake-time verification. Fully fixing this requires reorganizing the temp-vs-permanent lookup logic, which is part of the planned message-pipeline refactor.
3. **The block scanner promotion path** still uses `checkAndRequestApp` (which conflates "is this stored?" with "promote it from temp"). The redesign splits these into explicit `promoteTempToPermanent` and `fetchFromPeerAndStorePermanent` functions. Out of scope for this PR.
4. **Full removal of `arcaneSender`** from submission paths and the temp DB schema. Phase 2, gated on the network having rolled past this PR.


MorningLightMountain713:
I've tested this on 3 nodes (resyncing)

Cabecinha84 (approved):
ack

PR #1714, opened 2026-04-08
Fix CPU burst correctness: kernel rule, ground-truth gating, label-based eligibility

MorningLightMountain713:
## Summary

Fixes correctness bugs and architectural drift in the CPU burst feature merged in #1696. The original PR shipped with a formula that produces kernel-invalid burst values for any `burstMultiplier > 2.0`, a silent-failure path that left enterprise containers in a worst-of-both-worlds state on misconfigured hosts, and three different eligibility checks across three call sites that could (and did) drift out of sync.

This branch fixes those bugs and tightens the architecture so the same class of bug can't recur.

### Bug 1 — `calculateBurstParams` produces kernel-invalid values
The kernel's hard rule for cgroup-v2 `cpu.max.burst` is `0 ≤ burst ≤ quota` (cgroup-v2 docs: "The burst in the range [0, $quota]"). Empirically verified on Linux 6.17: writing `quota+1us` returns `EINVAL`.

The previous formula produced `burstUs = quotaUs * (burstMultiplier - 1)`. With the default `burstMultiplier=2.0`, that lands exactly at the kernel ceiling (`burstUs == quotaUs`). With anything higher, the kernel silently rejects the write and the caller swallows the failure. There is no operational reason to set burst lower than the kernel maximum either — burst is "free" headroom that only fires when banked idle time exists.

### Bug 2 — `setCpuBurst` failure was silently discarded
`dockerService.appDockerStart` called `await cpuBurstHelper.setCpuBurst(...)` and threw away the return value. On a node where the write failed (wrong cgroup driver, EINVAL from a bad config, etc.), the container started normally with no burst, no metric, no operator-visible signal — only a `log.warn` deep in the helper.

### Bug 3 — Throttle-bypass and burst capability could drift apart
`appInspector.checkApplicationsCpuUSage` gated its throttle-skip on `isEnterpriseOwner(app.owner)` only. It did NOT verify that burst had actually been applied. So on a node where `setCpuBurst` silently failed (Bug 2), an enterprise app got the throttle bypass *without* the burst capability — the worst of both worlds. The container could hit its hard `NanoCpus` cap with no bursting and no manual reduction either.

### Bug 4 — Two-place eligibility decision in dockerService
`getCpuHostConfig` (deleted in this branch) decided eligibility at create time and emitted `CpuPeriod`/`CpuQuota` instead of `NanoCpus` for enterprise containers. `appDockerStart` then re-decided by checking `if (hostCfg.CpuPeriod > 0 && hostCfg.CpuQuota > 0)` as a proxy for "burst was requested." Two places, two checks, ripe for drift. And the create-time mutation produced no runtime benefit — `NanoCpus` and `CpuPeriod`/`CpuQuota` translate to the same kernel cgroup state.

### `cpuBurstHelper.js` — simplified and hardened
- `calculateBurstParams` now starts at the kernel maximum (`burstUs = quotaUs`) and applies a per-container host-fairness cap so a single container's peak (quota + burst) never exceeds `(hostCpus - reservedCores) * period`. This bounds any one app from being able to consume the whole host during a burst window.
- Drop `burstMultiplier` from config — the kernel rule already binds burst at `quota`, so the multiplier can never produce anything other than 1.0. Keep `reservedCores: 1` — it controls the host-fairness cap.
- `getCgroupBurstPath` now resolves the cgroup file via `/proc/<pid>/cgroup` instead of guessing from two hardcoded path candidates. Works on any cgroup-v2 driver/runtime, not just docker-systemd. (Same approach used by Koordinator and Alibaba ACK's CPU burst implementations.)
- `setCpuBurst` takes `pid` instead of `containerId` (matches the new lookup).
- New `isBurstActive(pid)` reads `cpu.max.burst` directly as ground truth.
- New `applyBurst(pid, cores, identifier)` wraps support check + calculate + setCpuBurst + a loud `log.error` on failure.

### `dockerService.js` — label-based eligibility, decided once
- **Delete `getCpuHostConfig` entirely.** `appDockerCreate` uses `NanoCpus` for every container, identical to pre-#1696 behavior.
- Eligibility is decided **once**, in `appDockerCreate`, where the full app spec (with owner) is in scope. The decision is stamped onto the container as docker labels:
  - `flux.burst.eligible=true`
  - `flux.burst.cores=<n>`
- `appDockerStart` reads the labels post-start and calls `applyBurst` with the container's pid + cores. Every start path (initial install, soft redeploy, hard redeploy, recovery, peer-triggered restart, controller start) reapplies burst automatically — no per-caller plumbing through the 12+ call sites of `appDockerStart`.

### `appInspector.js` — gate throttle bypass on cgroup ground truth
- Replace `isEnterpriseOwner(app.owner)` with `isBurstActive(pid)`. The skip fires iff burst is structurally present on the container right now, so the two behaviors (burst + throttle bypass) can no longer drift apart.
- For v8/compose apps, the check is per-component — components of the same app can have independent burst state if some `setCpuBurst` calls failed.

## Per-container cap behavior across flux node tiers

With `reservedCores=1`:

| Tier | Host cores | Uncapped (`burst = quota`) | Capped (`burst < quota`) | No burst (`burst = 0`) |
|---|---|---|---|---|
| Cumulus | 4 | cpu ≤ 1.5 | cpu = 2 → burst=100k, peak=3 | cpu ≥ 3 |
| Nimbus | 8 | cpu ≤ 3.5 | cpu = 4 → burst=300k, peak=7 | cpu ≥ 7 |
| Stratus | 16 | cpu ≤ 7.5 | cpu = 8 → burst=700k, peak=15<br>cpu = 12 → burst=300k, peak=15 | cpu ≥ 15 |

When the cap reduces `burstUs` to 0, the cgroup file is written with 0 (default kernel behavior, no banking), `isBurstActive` returns false, and the throttle loop falls through to its normal throttling path — exactly the same behavior as a non-enterprise app. No special-casing; the ground-truth gating handles it for free.

This is a per-container cap, not a host-aggregate budget. Multiple burstable containers on the same host can collectively oversubscribe during simultaneous spikes; CFS handles contention proportionally. Real CPU isolation (`isolcpus` / cgroup affinity) is the right answer if a guaranteed reservation is needed and is independent of burst.

## Architecture: why ground-truth gating

The owner whitelist is a *prediction* of "should this container have burst." The cgroup file is *fact*: "cpu.max.burst has value N right now." Predictions can drift from reality; facts can't. The new gating asks the question that actually matters ("does this container have burst right now?") in the only place that knows ("the kernel"), so the bypass and the burst capability cannot disagree.

Cost is negligible: `isBurstActive` does 3 reads against `procfs`/`cgroupfs` (kernel virtual filesystems, no disk I/O, sub-microsecond each). The same loop already calls `dockerService.dockerContainerInspect` per app per cycle, which is orders of magnitude more expensive. Throttle loop fires every ~15 minutes.

## Validation

Every code path was validated end-to-end against a real Linux 6.17 cgroups-v2 host running this branch, with a real flux app (`burstwiretest`, `polinux/stress` with an idle/spike workload) installed via the normal global registration path.

| Layer | Validation |
|---|---|
| Helper functions (pure) | 35 unit tests, all green (incl. 9 new per-tier cap cases) |
| Helper functions (real kernel) | Direct exercise on the test node — `isCpuBurstSupported`, `getCgroupBurstPath`, `setCpuBurst` (incl. EINVAL rejection of `burst > quota`), `isBurstActive`, `applyBurst` |
| `appDockerStart` label-gated burst (positive) | Synthetic container with labels — `cpu.max.burst` written |
| `appDockerStart` label-gated burst (negative) | Synthetic container without labels — `cpu.max.burst` stays at 0 |
| `appDockerCreate` label stamping | Real flux app installed via global registration — `flux.burst.eligible=true`, `flux.burst.cores=1` confirmed via `docker inspect` |
| `applyBurst` from real `appDockerStart` | Real install — `cpu.max.burst=100000` matches quota |
| `appInspector` ground-truth gating | Three consecutive throttle-loop fires logged: `checkApplicationsCpuUSage burstcomp_burstwiretest burst-active, skipping CPU throttling` |
| Kernel actually fires burst | `cpu.stat` shows `nr_bursts: 126`, `burst_usec: 12_600_000` after 3 minutes runtime — definitive proof the kernel honors the write and is delivering burst capacity to a real workload |


Cabecinha84 (approved):
ack

PR #1562, opened 2025-09-27
Major Refactoring: Modularization of appsService

Cabecinha84:
This PR represents a significant architectural refactoring of the Flux node application management system. The monolithic appsService.js file (originally ~16,000 lines) has been decomposed into a well-organized
  modular architecture, reducing it to ~3,000 lines while maintaining all original functionality.

  Key Changes

  📦 New Modular Structure

  The refactoring introduces 8 new domain-specific directories with 13 specialized modules plus 10 utility modules:

  Core Modules:
  - appLifecycle/ - Application lifecycle management
    - appInstaller.js - Application installation logic
    - appUninstaller.js - Application removal and cleanup
    - advancedWorkflows.js - Complex workflows (soft redeploy, reinstallation, etc.)
  - appManagement/ - Application control and monitoring
    - appController.js - Application start/stop/restart operations
    - appInspector.js - Application monitoring and health checks
  - appRequirements/ - Resource and validation management
    - appValidator.js - Application specification validation
    - hwRequirements.js - Hardware requirements checking
  - appNetwork/ - Network layer
    - portManager.js - Port allocation and management
  - appMessaging/ - Inter-app communication
    - messageStore.js - Message storage and retrieval
    - messageVerifier.js - Message verification and validation
  - appSecurity/ - Security layer
    - imageManager.js - Docker image security and verification
  - appDatabase/ - Database operations
    - registryManager.js - Application registry management
  - appSystem/ - System integration
    - systemIntegration.js - OS-level integration (cron, monitoring, etc.)

  Utility Modules:
  - appConstants.js - Centralized configuration and constants
  - appSpecHelpers.js - Application specification helpers
  - appUtilities.js - Shared utility functions
  - chainUtilities.js - Blockchain-related utilities
  - enterpriseHelper.js - Enterprise feature support
  - enterpriseEncryption.js - Encryption utilities
  - globalState.js - Centralized state management
  - appProgressManager.js - Progress tracking
  - registrationHelper.js - Registration utilities
  - imageVerifier.js - Image verification enhancements

  📊 Statistics

  - 42 files changed
  - 18,450 insertions, 15,550 deletions (net +2,900 lines)
  - 189 commits of iterative refinement
  - Main appsService.js reduced by 12,218 lines (80% reduction)
  - New modular code: ~14,300 lines across specialized modules

  🔧 Technical Improvements

  1. Circular Dependency Resolution - Fixed multiple circular dependency issues through careful module design and dependency injection patterns
[...]

PR #1715, opened 2026-04-08
perf: restore bulk reindexGlobalAppsInformation in dbHelper

MorningLightMountain713:
## Results

Measured on fluxnode (full mainnet corpus: 57,235 messages, 672 alive apps, systemd `MemoryMax=1.5G`). RSS sampled from `/proc/<MainPID>/status` at ~40k samples/sec across the full boot/reindex window. Both runs forced a reindex by dropping `globalzelapps.zelappsinformation` before `systemctl restart fluxos.service`.

| Metric | Old path (`master`) | New path (`fix/bulk-reindex`) | Δ |
|---|---|---|---|
| Reindex duration | 38.3 s | 274 ms | **140× faster** |
| RSS before reindex | 145 MB | 145 MB | — |
| Peak RSS during reindex | 451 MB | 158 MB | **−293 MB** |
| RSS delta during reindex | +306 MB | +13 MB | **−293 MB** |
| `Expiring application` log lines | 19,545 | 0 | — |
| Post-reindex RSS (+30s after end) | 451–478 MB | 158–186 MB | ~−290 MB |

The `appsToRemove` path was separately verified end-to-end by inserting a ghost entry into `localzelapps.zelappsinformation` with a name not present in any global message, dropping `globalzelapps.zelappsinformation`, and restarting. The logs show the full chain firing:

```
validateAppsInformation reindexRequired: true
Reindexing of global applications finished. Local apps to be removed: ["claude_ghost_test_..."]
Application claude_ghost_test_... is expired, removing
REMOVAL REASON: App expired - claude_ghost_test_... reached expiration date (serviceManager)
APP REMOVAL TRIGGERED: claude_ghost_test_... | caller: at Immediate.appRemover [as _onImmediate] (ZelBack/src/services/serviceManager.js:198:30)
Error removing app claude_ghost_test_...: Flux App not found
```

The stack trace pins the call to `serviceManager.js:198`, which is exactly the line that was being fed an empty list before this PR. (The final `Flux App not found` error is expected — the ghost has no real container, but the code path was fully exercised.)

## Background

The original `reindexGlobalAppsInformation` in `dbHelper.js` was written specifically as a **memory optimization**. The previous (and reverted-to) implementation buffers the entire `globalAppsMessages` collection into a JS array via `.toArray()` before doing any work — on a fully-synced node that's ~57k message documents, each containing the full `appSpecifications` object (and for v8 enterprise apps, a multi-KB encrypted blob). In practice this puts hundreds of MB onto V8 heap during reindex, all of which is GC pressure on a process that's also running the explorer, p2p, and the API. On smaller / busier nodes this is a real source of memory pressure and pause times.

The original PR replaced that with a streaming design:

- A single mongo aggregation cursor (`returnArray: false`) — no `.toArray()`, no full materialisation
- `$sort` + `$group` to take the latest message per app
- `$match` to filter alive apps inside the aggregation
- `$replaceWith` + `$mergeObjects` to project the final spec shape directly
- Cursor consumed by `syncAppsInformationCollection`, which inserts in chunks of 500 via `insertManyToDatabase`

Peak working set during reindex is bounded by the chunk size (~500 docs), regardless of how big the messages collection grows. As a side benefit it's also dramatically faster because the heavy lifting (sort + group + filter) happens server-side in mongo instead of in 57k sequential round-trips from a JS for-loop.

## What got removed

In commit `74d8071e3` ("fix") inside RunOnFlux/flux#1562 (Modularization of appsService), the bulk path was deleted from `dbHelper.js` (138 lines removed) and replaced with a one-line `require('./appDatabase/registryManager')` that calls a different `reindexGlobalAppsInformation` living in `registryManager.js`.

That replacement does:

1. `findInDatabase(globalAppsMessages, {})` — calls `.toArray()` internally, **buffering the entire messages collection (hundreds of MB) in JS heap**
2. `for (const message of results)` loop, ~57k iterations
3. Per-iteration: `findOneInDatabase` + conditional `replaceOneInDatabase` (so ~114k sequential awaits)
4. Then calls `expireGlobalApplications` — another per-app `for` loop with `findOneAndDeleteInDatabase` per expired app, logging `Expiring application X` per app

In addition, `validateAppsInformation` started discarding the `appsToRemove` return value (`response.appsToRemove = []` literal), so locally-installed apps that no longer exist in the live spec set were no longer being removed by `serviceManager`.

### `dbHelper.js`

- New `expireHeightExpr(heightField, expireField)` helper — pure function returning a mongo aggregation expression that computes actual expiration height with full **PON fork rate adjustment**:
  - Pre-fork (`height < daemonPONFork`) chain runs at 1×, post-fork at 4×
  - Apps registered post-fork: `expireHeight = height + expire`
  - Apps registered pre-fork whose original expiration crosses the fork: post-fork tail is multiplied by 4 so the wall-clock lifetime the user paid for is preserved (`PON_FORK + 4 * (origExp - PON_FORK)`)
  - Default expire fallback is height-aware: `22000` pre-fork, `88000` post-fork
- `isReindexAppsInformationRequired` pipelines now use `expireHeightExpr` so the count comparison and the actual rebuild apply identical alive-or-dead logic
- New `reindexGlobalAppsInformation(...)` function — drop, create indexes, single aggregation, stream cursor through `syncAppsInformationCollection` for chunked bulk inserts, then bulk-clear `globalAppsInstallingErrorsLocations`. Returns `appsToRemove`.
- `validateAppsInformation` propagates the real `appsToRemove` list back to `serviceManager` (which already iterates it and calls `removeAppLocally` per entry — that code path was just being fed an empty list before)

### `registryManager.js`

- `reindexGlobalAppsInformation()` (the no-arg one called from `explorerService` and the API endpoint) is now a thin wrapper that gathers `appsGlobalDb` / `appsLocalDb` / `daemonDb`, fetches `scannedHeight`, and delegates to `dbHelper.reindexGlobalAppsInformation`. Same external no-arg signature, so no caller needs to change. The `reindexRunning` re-entry guard is preserved.

## What we deliberately didn't touch

- **`expireGlobalApplications`** still exists in `registryManager.js` and is still called from `explorerService.js` (twice), `appSpawner.js`, and `appHashSyncService.js`. Those are non-reindex contexts where you want a standalone expire pass. Could be made bulk in a follow-up, but it's a separate concern.
- **`updateAppSpecsForRescanReindex`** is still used by `rescanGlobalAppsInformation` (the height-bounded API variant), so it's kept.
- **All test stubs** (`tests/unit/appValidator.test.js`, `tests/unit/appUninstaller.test.js`) only stub the function name with `sinon.stub().resolves()` — the wrapper signature is unchanged so they continue to work.

## Impact

- **Memory** (the primary motivation): peak working set during reindex drops from hundreds of MB held in JS heap (the entire messages collection) to ~500 docs. Bounded by chunk size, not by collection size — so the win grows as the network accumulates more apps.
- **Speed**: on a node with ~57k permanent messages and ~20k unique apps the slow per-message path took 30+ seconds and emitted spammy `Expiring application X` log lines. The bulk path completes in well under a second.

## Testing checklist

- [x] Existing unit tests pass (`tests/unit/appValidator.test.js`, `tests/unit/appUninstaller.test.js`, `tests/unit/dbHelper.test.js` if present)
- [x] Deploy to a test node with a wiped `globalzelapps` and confirm reindex completes in seconds, not minutes (**274 ms on fluxnode, full corpus**)
- [x] **Observe peak RSS / V8 heap during reindex on a node with the full message corpus** — should be flat compared to baseline, not the hundreds-of-MB spike the slow path produces. (**+13 MB vs +306 MB, see Results table**)
- [x] After reindex, confirm `alive(messages) == alive(information)` (672 = 672 on fluxnode; aggregation count hand-computed with `expireHeightExpr` matches the rebuilt collection exactly)
- [x] Spot-check a pre-fork app whose original expiration straddles `daemonPONFork` — implicit in the exact count match above (the straddle branch is the only way the two sides could disagree)
- [x] Confirm `globalAppsInstallingErrorsLocations` is empty after a reindex (0 on both fluxnodes post-reindex)
- [x] Trigger `validateAppsInformation` with a locally-installed app that's no longer in the live spec set — verify `serviceManager` actually invokes `removeAppLocally` for it (ghost-app test, see Results section)

## Related

Restores the design from the original PR that introduced the bulk path, which was unintentionally undone by the modularization in #1562.

### DIFF
--- a/ZelBack/src/services/appDatabase/registryManager.js
+++ b/ZelBack/src/services/appDatabase/registryManager.js
@@ -1241,8 +1241,14 @@ async function updateAppSpecifications(appSpecs) {
 }
 
 /**
- * Rebuild the global apps information collection from messages collection
- * @returns {Promise<string>} Success message
+ * Rebuild the global apps information collection from messages collection.
+ *
+ * Thin wrapper around dbHelper.reindexGlobalAppsInformation, which does the
+ * heavy lifting in a single mongo aggregation + chunked bulk inserts. The
+ * dbHelper version filters expired apps inside the aggregation (full PON fork
+ * rate adjustment), so no separate expire pass is needed here.
+ *
+ * @returns {Promise<boolean>} True on success
  */
 async function reindexGlobalAppsInformation() {
   try {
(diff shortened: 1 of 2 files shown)


Cabecinha84 (approved):
ack

PR #1716, opened 2026-04-09
v8.10.0

Cabecinha84:
---                                                                                                                                                                                                                                                                                       
  Summary
                                                                                                                                                                                                                                                                                            
  Release v8.10.0 — fixes app message synchronization, hardens CPU burst management, and restores bulk reindex performance.
                                                                                                                                                                                                                                                                                            
  App Message Sync Fixes                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                            
  - Version compatibility during historic sync: Move version upgrade policy (must target latest spec version) from storeAppTemporaryMessage to the API layer, so historical permanent messages (e.g. v5→v6 updates) are no longer rejected during hash sync on fresh nodes                  
  - Enterprise app sync (arcaneSender): Remove the unauthenticated arcaneSender intake gate that blocked ~800 enterprise messages from syncing via hash sync; default to true in sync responses for backward compatibility with pre-upgrade nodes
  - Ghost field cleanup: Replace $set with replaceOne in updateAppSpecifications to prevent stale flat fields (repotag, cpu, ram, hdd) from accumulating on v4+ compose-era app specs; includes one-time reindex detection                                                                  
  - Re-verify signatures at promotion time: Prevent stale-state promotion when pending updates change ownership between temp and permanent storage                                                                                                                                          
  - Fix log.warning → log.warn typo in isReindexAppsInformationRequired                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                            
  CPU Burst Correctness                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                            
  - Kernel rule compliance: Set burstUs = quotaUs (the kernel maximum) instead of the previous formula that produced invalid values for multipliers > 2.0                                                                                                                                   
  - Host-fairness cap: Restore per-container cap preventing large apps from consuming the entire host during burst windows (e.g. 12-core app on 16-core stratus)
  - Label-based eligibility: Stamp flux.burst.eligible and flux.burst.cores as Docker labels at create time; all start paths automatically reapply burst from labels                                                                                                                        
  - Ground-truth throttle gating: Gate throttle bypass on actual cpu.max.burst cgroup state via isBurstActive(pid) instead of owner check, so burst and throttle bypass can never drift apart                                                                                               
  - Cgroup path resolution: Use /proc/<pid>/cgroup instead of hardcoded path candidates                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                            
  Performance                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                            
  - Bulk reindex restoration: Restore aggregation-based reindexGlobalAppsInformation (removed in PR #1562) replacing a per-message for-loop that made ~57k individual DB calls                                                                                                              
  
  Other                                                                                                                                                                                                                                                                                     
                                                                             
  - Skip remote availability check when node IP is not yet in the confirmed flux list (reduces log spam after IP changes)                                                                                                                                                                   
  - Refactor storeAppTemporaryMessage to use options object instead of positional booleans
  - Add replaceOneInDatabase to dbHelper                                                                                                                                                                                                                                                    
  - Add config.fluxapps.latestSupportedSpecVersion instead of hardcoding


MorningLightMountain713 (approved):
Ack

alihm (approved):
Ack

PR #1717, opened 2026-04-10
refactor: remove all zelAppSpecifications legacy references

MorningLightMountain713:
## Summary
- Remove all `zelAppSpecifications` fallbacks, duplicate queries, and dead indexes across 9 files
- Add one-time startup migration that renames any `zelAppSpecifications` → `appSpecifications` in `zelappsmessages` and drops legacy indexes — safe for nodes that have been running since before Feb 2021 without a resync
- Migration is idempotent (matches zero docs after first run) and runs once per boot via module-level flag
- Verified on live node (chud): 57,253 records in `zelappsmessages`, 0 with `zelAppSpecifications` — all records use `appSpecifications` exclusively
- Field name is not part of signed data, so no signature impact

## Cleanup
Once all nodes on the network have been upgraded to this version, `migrateZelAppSpecifications` and the `zelAppSpecsMigrationDone` flag in `explorerService.js` can be removed.

## Testing
- [x] Deploy to test node (cabbage)
- [x] Full sync from scratch completes successfully (57,468 messages synced, matches network)
- [x] Migration renames test `zelAppSpecifications` records and drops legacy indexes on boot
- [x] App owner queries return correct results (verified against 3 network nodes)
- [x] Global app specs and permanent messages counts match network

### DIFF
--- a/ZelBack/src/services/appDatabase/registryManager.js
+++ b/ZelBack/src/services/appDatabase/registryManager.js
@@ -1640,7 +1640,7 @@ async function rescanGlobalAppsInformation(height = 0, removeLastInformation = f
 
     // eslint-disable-next-line no-restricted-syntax
     for (const message of results) {
-      const updateForSpecifications = message.appSpecifications || message.zelAppSpecifications;
+      const updateForSpecifications = message.appSpecifications;
       updateForSpecifications.hash = message.hash;
       updateForSpecifications.height = message.height;
       // eslint-disable-next-line no-await-in-loop
--- a/ZelBack/src/services/appLifecycle/advancedWorkflows.js
+++ b/ZelBack/src/services/appLifecycle/advancedWorkflows.js
@@ -415,33 +415,10 @@ async function getPreviousAppSpecifications(specifications, verificationTimestam
       }
     }
   });
-  // some early app have zelAppSepcifications
-  const appsQueryB = {
-    'zelAppSpecifications.name': specifications.name,
-  };
-  const permanentAppMessageB = await dbHelper.findInDatabase(database, globalAppsMessages, appsQueryB, projection);
-  permanentAppMessageB.forEach((foundMessage) => {
-    // has to be registration message
-    const validTypes = ['zelappregister', 'fluxappregister', 'zelappupdate', 'fluxappupdate'];
-    if (validTypes.includes(foundMessage.type)) {
-      if (!latestPermanentRegistrationMessage && foundMessage.timestamp <= verificationTimestamp) {
-        // no message and found message is not newer than our message
-        latestPermanentRegistrationMessage = foundMessage;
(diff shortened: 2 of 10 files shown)


Cabecinha84 (changes requested):
Minor observations (non-blocking)                               
                                                                                                                                                       
  - Fresh install edge case: col.indexes() behavior on a never-created collection is driver/version dependent. In practice the collection is almost    
  always pre-existing by this point in initiateBlockProcessor, but a defensive try/catch around migrateZelAppSpecifications (or an early
  listCollections check) would make fresh-node boots bulletproof.                                                                                      
  - Peer wire compatibility: the PR assumes every node on the network now emits appSpecifications. messageStore.storeAppTemporaryMessage will now
  reject any inbound message with only zelAppSpecifications. The author's justification ("since Feb 2021") is reasonable, but worth an explicit        
  confirmation that no gossip path still carries the legacy field.
  - Cleanup note in PR body is good — flag and migration function can be deleted in a follow-up once the fleet is upgraded.                            
                                                                                                                                                       
  Recommendation
                                                                                                                                                       
  Ack. Optionally ask the author to wrap the migration call in a try/catch with a log line so a malformed fresh-sync DB can't stall block-processor    
  init.

MorningLightMountain713:
> Minor observations (non-blocking)
> 
> * Fresh install edge case: col.indexes() behavior on a never-created collection is driver/version dependent. In practice the collection is almost
>   always pre-existing by this point in initiateBlockProcessor, but a defensive try/catch around migrateZelAppSpecifications (or an early
>   listCollections check) would make fresh-node boots bulletproof.
> * Peer wire compatibility: the PR assumes every node on the network now emits appSpecifications. messageStore.storeAppTemporaryMessage will now
>   reject any inbound message with only zelAppSpecifications. The author's justification ("since Feb 2021") is reasonable, but worth an explicit
>   confirmation that no gossip path still carries the legacy field.
> * Cleanup note in PR body is good — flag and migration function can be deleted in a follow-up once the fleet is upgraded.
> 
> Recommendation
> 
> Ack. Optionally ask the author to wrap the migration call in a try/catch with a log line so a malformed fresh-sync DB can't stall block-processor init.

Fixed. Wrapped in try / catch

MorningLightMountain713:
Superseded by #1757.

This branch was based before the `appMessaging` refactor (the `messageStore` / `messageVerifier` / `appHashSyncService` split and `getPreviousAppSpecifications` moving into `registryManager`), so rebasing it onto `development` conflicted against code that had since been rewritten. #1757 rebuilds the same change fresh on current `development` — full hard cutoff of the `zelAppSpecifications` fallbacks plus the one-time startup migration — and folds in the review feedback from here (try/catch around the migration, wire-compatibility confirmation). Closing in favour of #1757.

PR #1718, opened 2026-04-17
feat: app tampering detection, frequent-restart tracking, and txhash-based DOS enforcement

Cabecinha84:
## Summary                                                      

  Adds observability for node-level tampering and an operator-driven path to                                                                                                                                              
  take a misbehaving (non-ArcaneOS) node offline until manual intervention.
                                                                                                                                                                                                                          
  - **`appTamperingDetectionService`** — records events into a new                                                                                                                                                        
    `apptamperingevents` collection (30-day TTL). Event types:                                                                                                                                                            
    `container_vanished`, `network_pruned`, `mount_vanished`, `crontab_wiped`,                                                                                                                                            
    `recreation_failed`, and `frequent_restart`. Each call inserts a new row so                                                                                                                                           
    the full history per app is preserved.                                                                                                                                                                                
    - Public read endpoint: `GET /apps/tamperingevents/:appname?` (newest first)                                                                                                                                          
    - `frequent_restart` fires when FluxOS starts less than one hour after its                                                                                                                                            
      previous start, stored under a synthetic `__system__` app name. Previous                                                                                                                                            
      startup time is tracked in a single-doc `nodestartuptracker` collection.                                                                                                                                            
                                                                                                                                                                                                                          
  - **`appTamperingBlocklistService`** — on startup (after daemon sync) and                                                                                                                                               
    every 12h thereafter, fetches a manually-curated list of node txhashes from                                                                                                                                           
    `helpers/tamperingblockednodes.json` on the RunOnFlux master branch. If                                                                                                                                               
    **this node's collateral txhash is on the list** AND it has **more than 10                                                                                                                                            
    tampering events**, the service sets a sticky DOS message so the node                                                                                                                                                 
    starts failing benchmarks and exits the network. On the next 12h tick, if                                                                                                                                             
    the txhash has been removed from the list OR events have dropped to ≤10,                                                                                                                                              
    the DOS is cleared.                                                                                                                                                                                                   
                                                                                                                                                                                                                          
  ## Operator workflow                                                                                                                                                                                                    
                                                                  
  1. Review the network via `GET /apps/tamperingevents` (or                                                                                                                                                               
     `GET /apps/tamperingevents/__system__` for restart history) on nodes of
     interest.                                                                                                                                                                                                            
  2. If a node is clearly misbehaving, add its collateral txhash to
     `helpers/tamperingblockednodes.json` on master.                                                                                                                                                                      
  3. Within 12h (or on next reboot) the flagged node will DOS itself and exit                                                                                                                                             
     the network. Fixing the underlying issue and waiting for events to drop,                                                                                                                                             
     or removing the txhash from the file, clears the DOS on the next tick.                                                                                                                                               
                                                                                                                                                                                                                          
  ## Implementation notes                                                                                                                                                                                                 
                                                                  
  - **ArcaneOS gating.** Blocklist enforcement is a no-op on ArcaneOS nodes                                                                                                                                               
    (detected via `process.env.FLUXOS_PATH`, the canonical check used
    throughout the codebase). Event recording continues on all nodes for                                                                                                                                                  
    observability; only legacy nodes can be DOSed via this path.  
  - **Sticky DOS** lives in `fluxNetworkHelper` via a parallel                                                                                                                                                            
    `stickyDosMessage`/`stickyDosState` slot. `getDosMessage()` and                                                                                                                                                       
    `getDOSState()` prefer the sticky slot when set. Regular                                                                                                                                                              
    `setDosMessage(null)` calls from other checks cannot clear it — only the                                                                                                                                              
    blocklist enforcer's own 12h tick can.                                                                                                                                                                                
  - Indexes on `apptamperingevents`: TTL on `detectedAt` (30 days) and                                                                                                                                                    
    `{ appName: 1, detectedAt: -1 }` for per-app queries. Created via a new                                                                                                                                               
    `ensureIndex` helper in `serviceManager` that tolerates                                                                                                                                                               
    `IndexOptionsConflict` by dropping and recreating.                                                                                                                                                                    
  - `isNetworkMissingError` helper centralizes detection of "docker network is                                                                                                                                            
    missing" errors (requires both `"network"` and one of `"not found"` /                                                                                                                                                 
    `"no such"` / `"does not exist"` / `"missing"`), replacing a loose                                                                                                                                                    
    substring match.                                                                                                                                                                                                      
  - `frequent_restart` check ignores negative time deltas to stay safe against                                                                                                                                            
    clock skew.                                                                                                                                                                                                           
  - Adds `tamperingBlocklistCache` (6h TTL) mirroring                                                                                                                                                                     
    `blockedRepositoriesCache`.                                                                                                                                                                                           
                                                                                                                                                                                                                          
  ## Commits                                                                                                                                                                                                              
                                                                                                                                                                                                                          
  1. `feat: app tampering detection and txhash-based DOS enforcement`
  2. `feat: add frequent_restart tampering event`
  3. `feat: skip tampering DOS enforcement on ArcaneOS nodes`                                                                                                                                                             
   
  ## Test plan                                                                                                                                                                                                            
                                                                  
  - [x] `tests/unit/appTamperingDetectionService.test.js` — 25 tests covering                                                                                                                                             
        `isNetworkMissingError`, `recordEvent` (insert-per-call, DB
        unavailable, error swallow), `getEvents` (filter variants, sort, error                                                                                                                                            
        path), and `checkFrequentRestart` (under/over/at-threshold, first-ever                                                                                                                                            
        boot, clock skew, marker upsert, DB unavailable, error swallow).                                                                                                                                                  
  - [x] `tests/unit/appTamperingBlocklistService.test.js` — 21 tests covering                                                                                                                                             
        `fetchBlocklist` (cache/axios/bad-shape), `countTamperingEvents`,                                                                                                                                                 
        `getMyTxhash`, the full `enforceBlocklist` decision matrix including                                                                                                                                              
        sticky set/clear transitions and orphaned-state recovery, and ArcaneOS                                                                                                                                            
        gating (no-op enforcement, no blocklist fetch, no interval install).                                                                                                                                              
  - [x] `tests/unit/fluxNetworkHelper.test.js` — 8 new tests for sticky DOS                                                                                                                                               
        helpers, precedence rules, and `setDosMessage(null)` non-interference.                                                                                                                                            
  - [x] All previously-passing tests for affected files continue to pass —                                                                                                                                                
        242 total green across the affected and new suites.


ZelBack/src/services/appTamperingBlocklistService.js
```diff
@@ -0,0 +1,184 @@
...
+  if (shouldDos) {
+    const message = `${DOS_MESSAGE_PREFIX}: ${eventCount} events, txhash ${myTxhash}`;
+    fluxNetworkHelper.setStickyDosMessage(message);
+    fluxNetworkHelper.setStickyDosStateValue(100);
+    ourDosActive = true;
+    log.error(message);
+    return;
+  }
+
+  if (ourDosActive || fluxNetworkHelper.getStickyDosMessage()) {
```
alihm: The condition fluxNetworkHelper.getStickyDosMessage() checks if any sticky DOS message exists, not just one set by this service. If another module sets a sticky DOS for a different reason, this service will incorrectly clear it when its own condition (shouldDos) is false.

The clearing logic should only target DOS messages owned by this service.

ZelBack/src/services/appTamperingBlocklistService.js
```diff
@@ -0,0 +1,184 @@
...
+    return null;
+  }
+}
+
+/**
+ * Block until the daemon reports synced. Polls every SYNC_POLL_MS.
+ */
+async function waitForDaemonSynced() {
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
```
alihm: If stop() is called while waitForDaemonSynced() is polling, the wait continues indefinitely, and after daemon syncs, the service proceeds to run enforceBlocklist() and schedule intervals despite being "stopped". Consider adding a cancellation mechanism.

ZelBack/src/services/appMessaging/peerNotification.js
```diff
@@ -254,6 +259,8 @@ async function checkAndNotifyPeersOfRunningApps(
 
               if (!containerExists) {
                 log.warn(`Container for ${stoppedApp} doesn't exist, recreating immediately...`);
+                // eslint-disable-next-line no-await-in-loop
+                await appTamperingDetectionService.recordEvent(mainAppName, 'container_vanished', `Container ${stoppedApp} missing, not found in Docker`);
```
alihm, line 263: Needs wrapping in a try-catch, If recordEvent throws here, the outer catch block will trigger removeAppLocally.

Cabecinha84: Not valid as stated. The premise is wrong. recordEvent already has an internal try/catch that swallows every error and logs it

ZelBack/src/services/appTamperingBlocklistService.js
Cabecinha84: should be fixed

ZelBack/src/services/appTamperingBlocklistService.js
Cabecinha84: should be fixed

ZelBack/src/services/appMonitoring/syncthingFolderStateMachine.js
```diff
@@ -81,6 +82,7 @@ async function verifyFolderMountSafety(appId, folderPath) {
       result.isSafe = false;
       result.reason = 'base_directory_missing';
       log.warn(`verifyFolderMountSafety - ${appId} base directory does not exist: ${baseDir}`);
+      appTamperingDetectionService.recordEvent(appId, 'mount_vanished', `Base directory missing: ${baseDir}`);
```
MorningLightMountain713: This should await for consistency

ZelBack/src/services/appTamperingBlocklistService.js
```diff
@@ -0,0 +1,222 @@
...
+  return typeof msg === 'string' && msg.startsWith(DOS_MESSAGE_PREFIX);
+}
+
+/**
+ * Fetch the manually-curated txhash blocklist from the RunOnFlux repo.
+ * Returns [] on any failure so the caller never crashes the enforcer loop.
+ */
+async function fetchBlocklist() {
+  try {
+    const cached = fluxCaching.tamperingBlocklistCache.get('tamperingBlocklist');
```
MorningLightMountain713: This is dead code. We call it once at startup then every 12 hours. Will always miss the cache... should just remove the cache entirely

ZelBack/src/services/appTamperingBlocklistService.js
```diff
@@ -0,0 +1,222 @@
...
+const generalService = require('./generalService');
+const daemonServiceMiscRpcs = require('./daemonService/daemonServiceMiscRpcs');
+const fluxCaching = require('./utils/cacheManager').default;
+
+const BLOCKLIST_URL = 'https://raw.githubusercontent.com/RunOnFlux/flux/master/helpers/tamperingblockednodes.json';
+const CHECK_INTERVAL_MS = 12 * 60 * 60 * 1000; // 12 hours
+const SYNC_POLL_MS = 60 * 1000; // 60s while waiting for daemon sync
+const TAMPERING_EVENT_THRESHOLD = 10;
+const DOS_MESSAGE_PREFIX = 'Node flagged via tampering blocklist';
+const isArcane = Boolean(process.env.FLUXOS_PATH);
```
MorningLightMountain713: process.env.FLUXOS_PATH` is trivially spoofable — any legacy operator can `export FLUXOS_PATH=1` to bypass enforcement entirely.

We already have `fluxService.isSystemSecure()` and reads `systemsecure` from the benchmark response.

However, `isSystemSecure()` currently returns `false` both when fluxbenchd says "not secure" and when fluxbenchd is unreachable. If we use it as-is, an ArcaneOS node with fluxbenchd temporarily down would get falsely DOSed.

Suggestion: use a three-state check — `true` (ArcaneOS confirmed, skip enforcement), `false` (explicitly not ArcaneOS, enforce), `null` (fluxbenchd unreachable, skip this tick). Remove the module-level `const isArcane` and call this dynamically in `enforceBlocklist()` and `start()` each tick. Skipping a tick on `null` is harmless since we re-check every 12h, and it means we never falsely DOS a real ArcaneOS node.

MorningLightMountain713 (changes requested):

ZelBack/src/services/appMonitoring/syncthingFolderStateMachine.js
Cabecinha84: Should be fixed

ZelBack/src/services/appTamperingBlocklistService.js
Cabecinha84: Should be fixed

ZelBack/src/services/appTamperingBlocklistService.js
Cabecinha84: Should be fixed

MorningLightMountain713 (approved):
Ack

PR #1719, opened 2026-04-19
feat: opt-in enterprise network — gate app spawning and ownership split by node pubkey 

Cabecinha84:
## Summary                                                      
                                                                                                                                                                                
  Adds an **opt-in enterprise network** of fluxnodes identified by pubkeys in a new
  `enterpriseNodesPublicKeys` config list. Splits the ownership space between enterprise                                                                                        
  and non-enterprise nodes and adds spawn-path gating plus a one-shot boot-time cleanup.                                                                                        
                                                                                                                                                                                
  ## Behavior                                                                                                                                                                   
                                                                                                                                                                                
  | Situation | Enterprise-network node | Every other node |                                                                                                                    
  |---|---|---|                                                   
  | App owner ∈ `enterpriseAppOwners` | spawn / keep | skip / uninstall + broadcast `fluxappremoved` |
  | App owner ∉ `enterpriseAppOwners` | skip / uninstall + broadcast `fluxappremoved` | spawn / keep |                                                                          
  | `enterpriseNodesPublicKeys` empty | — | — |
                                                                                                                                                                                
  Ships with an empty `enterpriseNodesPublicKeys` list, so the feature is a **no-op on                                                                                          
  unconfigured nodes** and rolls out safely.                                                                                                                                    
                                                                                                                                                                                
  Enterprise nodes additionally get:                              
                                                                                                                                                                                
  - **Tight spawn cadence** inside `trySpawningGlobalApplication`: `shortDelayTime = 30s`,                                                                                      
    `delayTime = 60s`, and the terminal `delay(delayTime)` before the self-recurse is
    skipped. The mid-function dynamic delay reassignments are gated so these values stick                                                                                       
    for the whole run.                                                                                                                                                          
  - **>4 vCore burst headroom** required before install, via the new exported                                                                                                   
    `hwRequirements.checkAppCpuBurstHeadroom(appSpecs)`. Rejects apps whose installation                                                                                        
    would leave `cpuCores − lockedSystemResources.cpu/10 − appsCpusLocked − appHWrequirements.cpu ≤ 4`.                                                                         
                                                                                                                                                                                
  ## Safety: never act on an unknown identity                                                                                                                                   
                                                                                                                                                                                
  `isEnterpriseNode()` throws when this node's own pubkey cannot be resolved                                                                                                    
  (`fluxNetworkHelper.getFluxNodePublicKey` returns an Error because daemon/benchmark
  is still coming up). On that throw:                                                                                                                                           
                                                                                                                                                                                
  - The **spawner** lets it propagate to the existing outer `catch`, which retries after                                                                                        
    `shortDelayTime`.                                                                                                                                                           
  - The **boot-time cleanup** lets it propagate to a self-rescheduling `setTimeout` that                                                                                        
    retries every 5 minutes until a run completes without throwing.                                                                                                             
                                                                                                                                                                                
  Cache semantics:                                                                                                                                                              
                                                                                                                                                                                
  - Result is cached only on successful resolution, so transient daemon outages don't                                                                                           
    stick as `false`.                                             
  - Empty `enterpriseNodesPublicKeys` short-circuits to `false` without any network call.                                                                                       
  - `resetEnterpriseNodeCache()` is exported for future hot-reload integration with                                                                                             
    `configManager`.                                                                                                                                                            
                                                                                                                                                                                
  Per-app `try/catch` was removed from the cleanup loop so an uninstall failure also                                                                                            
  bubbles up and triggers a retry; partial progress is safe because subsequent runs                                                                                             
  re-query `localAppsInformation` and only see remaining offenders.                                                                                                             
                                                                                                                                                                                
  ## Files                                                                                                                                                                      
                                                                                                                                                                                
  - **New** `ZelBack/src/services/utils/enterpriseNetwork.js` — `isEnterpriseNode`,                                                                                             
    `isEnterpriseAppOwner`, `getEnterpriseAppOwners`, `getEnterpriseNodesPublicKeys`,
    `resetEnterpriseNodeCache`, `cleanupOwnershipViolations`.                                                                                                                   
  - `ZelBack/config/default.js` — new empty `enterpriseNodesPublicKeys: []`.                                                                                                    
  - `tests/unit/globalconfig/default.js` — mirrors the new field for the test harness.                                                                                          
  - `ZelBack/src/services/appLifecycle/appSpawner.js` — resolves `isEnterprise` at the top                                                                                      
    of `trySpawningGlobalApplication`; adds `owner` to the aggregation `$project`;                                                                                              
    symmetric owner filter; enterprise-only burst-headroom check; enterprise-only tight                                                                                         
    delays.                                                                                                                                                                     
  - `ZelBack/src/services/appRequirements/hwRequirements.js` — new exported                                                                                                     
    `checkAppCpuBurstHeadroom`.                                                                                                                                                 
  - `ZelBack/src/services/serviceManager.js` — self-rescheduling boot cleanup
    (`scheduleEnterpriseCleanup`) fired at T+5m after startup.                                                                                                                  
                                                                                                                                                                                
  ## Tests                                                                                                                                                                      
                                                                                                                                                                                
  - **New** `tests/unit/enterpriseNetwork.test.js` — 18 tests covering every helper,                                                                                            
    cache semantics, throw-on-unresolved-pubkey, and both directions of the cleanup
    (enterprise vs. non-enterprise offender selection, no-op when clean, throw                                                                                                  
    propagation from uninstall failure).                                                                                                                                        
  - `tests/unit/hwRequirements.test.js` — 6 new tests for `checkAppCpuBurstHeadroom`:                                                                                           
    boundary at 4 free cores (throws), 5 (passes), negative (throws), `appsResources`                                                                                           
    failure, v4+ compose-sum behavior. Export check updated.                                                                                                                    
                                                                                                                                                                                
  Full unit suite: **3594 passing, 5 pending, 1 failing** — the failure is the pre-existing                                                                                     
  `syncthingHealthMonitor > should remove app after remove threshold` case on this branch,                                                                                      
  unchanged by this PR.                                                                                                                                                         
                                                                  
  ## Test plan                                                                                                                                                                  
                                                                  
  - [ ] CI unit suite green (`npm run test:zelback:unit` / container compose variant).                                                                                          
  - [ ] On a real node with empty `enterpriseNodesPublicKeys`: no behavior change
        (the spawner's `isEnterpriseNode()` resolves to `false` without a network call,                                                                                         
        so no additional overhead per iteration; cleanup is a no-op).                                                                                                           
  - [ ] On a node added to `enterpriseNodesPublicKeys` with an existing non-enterprise                                                                                          
        app installed: within ~5 min the app is uninstalled and `fluxappremoved` is                                                                                             
        broadcast; subsequent spawns only pick from `enterpriseAppOwners`; reject paths                                                                                         
        show tight `delayTime=60s` cadence in logs.                                                                                                                             
  - [ ] On a node **not** in `enterpriseNodesPublicKeys` with an enterprise-owned app                                                                                           
        installed: within ~5 min the app is uninstalled and `fluxappremoved` is broadcast.                                                                                      
  - [ ] Daemon-down simulation at boot: cleanup logs "retrying in 5 min" and only acts                                                                                          
        once `getFluxNodePublicKey` resolves.                                                                                                                                   
  - [ ] Enterprise node at CPU ceiling: `checkAppCpuBurstHeadroom` rejects an app that                                                                                          
        would leave ≤ 4 free vCores; spawner caches the hash and retries later.

### DIFF
--- a/ZelBack/src/services/appLifecycle/appSpawner.js
+++ b/ZelBack/src/services/appLifecycle/appSpawner.js
@@ -17,6 +17,7 @@ const portManager = require('../appNetwork/portManager');
 const appUtilities = require('../utils/appUtilities');
 const systemIntegration = require('../appSystem/systemIntegration');
 const globalState = require('../utils/globalState');
+const enterpriseNetwork = require('../utils/enterpriseNetwork');
 const { FluxCacheManager } = require('../utils/cacheManager');
 // const advancedWorkflows = require('./advancedWorkflows'); // Moved to dynamic require to avoid circular dependency
 
@@ -46,10 +47,18 @@ function initialize(deps) {
  * @returns {Promise<void>}
  */
 async function trySpawningGlobalApplication() {
+  let isEnterprise = false;
   let shortDelayTime = 5 * 60 * 1000; // Default 5 minutes
   let delayTime = 30 * 60 * 1000; // Default 30 minutes
   let appHash = null; // Declare outside try block to be accessible in catch
   try {
+    // Throws if fluxnode pubkey can't be resolved (daemon/benchmark down);
+    // the outer catch handles the retry so we never act on an unknown identity.
+    isEnterprise = await enterpriseNetwork.isEnterpriseNode();
+    if (isEnterprise) {
+      shortDelayTime = 30 * 1000;
+      delayTime = 60 * 1000;
+    }
     // how do we continue with this function?
     // we have globalapplication specifics list
     // check if we are synced
@@ -149,6 +158,7 @@ async function trySpawningGlobalApplication() {
           hash: '$hash',
           version: '$version',
           enterprise: '$enterprise',
+          owner: '$owner',
         },
       },
       { $sort: { name: 1 } },
@@ -211,9 +221,19 @@ async function trySpawningGlobalApplication() {
       globalAppNamesLocation = globalAppNamesLocation.filter((app) => (app.geolocation.length === 0 || app.geolocation.filter((loc) => loc.startsWith('a!c')).length === 0 || !app.geolocation.find((loc) => loc.startsWith('a!c') && `a!c${myNodeLocation}`.startsWith(loc.replace('_NONE', '')))));
       // filter apps that dont have geolocation or have and match my node geolocation
       globalAppNamesLocation = globalAppNamesLocation.filter((app) => (app.geolocation.length === 0 || app.geolocation.filter((loc) => loc.startsWith('ac')).length === 0 || app.geolocation.find((loc) => loc.startsWith('ac') && `ac${myNodeLocation}`.startsWith(loc))));
+      // enterprise network gating:
+      //   enterprise-network nodes install ONLY apps owned by enterpriseAppOwners
+      //   every other node NEVER installs apps owned by enterpriseAppOwners
+      if (isEnterprise) {
+        globalAppNamesLocation = globalAppNamesLocation.filter((app) => enterpriseNetwork.isEnterpriseAppOwner(app.owner));
+      } else {
+        globalAppNamesLocation = globalAppNamesLocation.filter((app) => !enterpriseNetwork.isEnterpriseAppOwner(app.owner));
+      }
 
       appsCountAvailableToInstallOnMyNode = globalAppNamesLocation.length + appsSyncthingToBeCheckedLater.length + appsToBeCheckedLater.length;
-      shortDelayTime = appsCountAvailableToInstallOnMyNode > 1 ? 60 * 1000 : 5 * 60 * 1000;
+      if (!isEnterprise) {
+        shortDelayTime = appsCountAvailableToInstallOnMyNode > 1 ? 60 * 1000 : 5 * 60 * 1000;
+      }
 
       if (globalAppNamesLocation.length === 0) {
         log.info('trySpawningGlobalApplication - No app currently to be processed');
@@ -254,8 +274,10 @@ async function trySpawningGlobalApplication() {
     }
 
     // If there are multiple apps to process, use shorter delays
-    delayTime = appsCountAvailableToInstallOnMyNode > 1 ? 60 * 1000 : 30 * 60 * 1000;
-    shortDelayTime = appsCountAvailableToInstallOnMyNode > 1 ? 60 * 1000 : 5 * 60 * 1000;
+    if (!isEnterprise) {
+      delayTime = appsCountAvailableToInstallOnMyNode > 1 ? 60 * 1000 : 30 * 60 * 1000;
+      shortDelayTime = appsCountAvailableToInstallOnMyNode > 1 ? 60 * 1000 : 5 * 60 * 1000;
+    }
 
     globalState.trySpawningGlobalAppCache.set(appHash, '');
     log.info(`trySpawningGlobalApplication - App ${appToRun} hash: ${appHash}`);
@@ -347,6 +369,10 @@ async function trySpawningGlobalApplication() {
 
     // verify requirements
     await hwRequirements.checkAppRequirements(appSpecifications);
+    // enterprise network nodes: reserve >4 vCores of burst headroom (automatic CPU burst)
+    if (isEnterprise) {
+      await hwRequirements.checkAppCpuBurstHeadroom(appSpecifications);
+    }
 
     // ensure ports unused
     // Get apps running specifically on this IP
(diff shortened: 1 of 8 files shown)


ZelBack/src/services/appLifecycle/appSpawner.js
```diff
@@ -46,10 +47,18 @@ function initialize(deps) {
...
  */
 async function trySpawningGlobalApplication() {
+  let isEnterprise = false;
   let shortDelayTime = 5 * 60 * 1000; // Default 5 minutes
   let delayTime = 30 * 60 * 1000; // Default 30 minutes
   let appHash = null; // Declare outside try block to be accessible in catch
   try {
+    // Throws if fluxnode pubkey can't be resolved (daemon/benchmark down);
+    // the outer catch handles the retry so we never act on an unknown identity.
+    isEnterprise = await enterpriseNetwork.isEnterpriseNode();
```
MorningLightMountain713: This should all be outside of this function, set it once

ZelBack/src/services/appLifecycle/appSpawner.js
```diff
@@ -211,9 +221,19 @@ async function trySpawningGlobalApplication() {
       globalAppNamesLocation = globalAppNamesLocation.filter((app) => (app.geolocation.length === 0 || app.geolocation.filter((loc) => loc.startsWith('a!c')).length === 0 || !app.geolocation.find((loc) => loc.startsWith('a!c') && `a!c${myNodeLocation}`.startsWith(loc.replace('_NONE', '')))));
       // filter apps that dont have geolocation or have and match my node geolocation
       globalAppNamesLocation = globalAppNamesLocation.filter((app) => (app.geolocation.length === 0 || app.geolocation.filter((loc) => loc.startsWith('ac')).length === 0 || app.geolocation.find((loc) => loc.startsWith('ac') && `ac${myNodeLocation}`.startsWith(loc))));
+      // enterprise network gating:
+      //   enterprise-network nodes install ONLY apps owned by enterpriseAppOwners
+      //   every other node NEVER installs apps owned by enterpriseAppOwners
+      if (isEnterprise) {
```
MorningLightMountain713: This function is already insanely complex. I don't think it's a good idea adding even more branching logic here.

We should split out this logic into a helper function, something like:

```
  const spawnConfig = await resolveSpawnConfig();
  // { shortDelay, delay, skipTrailingDelay, appFilter, extraChecks }
```

MorningLightMountain713 (changes requested):

ZelBack/src/services/appLifecycle/appSpawner.js
Cabecinha84: answer from AI: Two things keep this inside the function rather than hoisted to module-level / a one-shot set:
  ▎                                                                                                         
  ▎ 1. It already is "set once" in practice — isEnterpriseNode() caches internally on first successful      
  ▎ resolution (cachedIsEnterpriseNode in enterpriseNetwork.js), so subsequent spawner ticks are a cached   
  ▎ boolean read, not a network call. Short-circuits to false without any call when                         
  ▎ enterpriseNodesPublicKeys is empty.                                      
  ▎ 2. The throw-and-retry path depends on being called inside this try/catch. isEnterpriseNode() 
  ▎ deliberately throws when getFluxNodePublicKey returns an Error (daemon/benchmark still coming up). The  
  ▎ outer catch here is what turns that into a shortDelayTime retry, so we never act on an unknown identity.
  ▎  Hoisting to module load would resolve before the daemon is guaranteed up and would need its own retry  
  ▎ scaffolding to be equivalent.                                            
  ▎ 3. Hot-reload. resetEnterpriseNodeCache() is exported specifically so configManager can invalidate the 
  ▎ cached role if enterpriseNodesPublicKeys is edited at runtime. A module-level one-shot capture loses    
  ▎ that.

ZelBack/src/services/appLifecycle/appSpawner.js
Cabecinha84: should be fixed

ZelBack/src/services/appLifecycle/appSpawner.js
MorningLightMountain713: Sure, having it here is easier short term, but architecturally it's wrong and mixing concerns in this function.

1/ Nope. Doesn't matter  - shouldn't be here
2/ Throw and retry is a terrible pattern
3/ This is a side effect, if it's a issue, should be handled up front

However, I'll just let it go

MorningLightMountain713 (approved):
Ack

XK4MiLX (approved):

ZelBack/src/services/appLifecycle/appSpawner.js
Cabecinha84: No, don't let it go, AI agreed and did the changes. Can you check again?

alihm (approved):
Ack

MorningLightMountain713 (approved):
Nice one :) Ack

PR #1725, opened 2026-05-01
allow bigger upload and timeout

TheTrunk:


ZelBack/src/lib/fluxServer.js
```diff
@@ -82,6 +82,12 @@ class FluxServer {
       this.expressApp,
     );
 
+    // Node's default requestTimeout is 5 min, which kills large file uploads
+    // (app backups, fluxshare) on slow connections. headersTimeout (60s default)
+    // still defends against slowloris on the request line/headers, and
+    // formidable caps body size at 10GB.
+    server.requestTimeout = 12 * 60 * 60 * 1000;
```
MorningLightMountain713, line 89: This is a good reason why we should have a reverse proxy (nginx) in front of Gravity - It's quite difficult to set per request socket timeouts in express, and setting the default across the entire server is a pretty blunt instrument. (There are other good reasons as well)

This patch is now opening unauthenticated POST endpoints to allow 12 hour connections. For our use case, think 2 hours would be more appropriate as an upper level to allow for a backup upload / download?

Note, requestTimeout only applies for file uploads, requestTimeout is for the request body, i.e. client -> server.

Could maybe add something like this as well... so all other routes are 5 minutes, and only the uploads use the 2 hour?

```javascript
  // after the requireHttps function (line 71), add:

  const uploadRoutes = [
    /^\/ioutils\/fileupload(\/|$)/,
    /^\/apps\/fluxshare\/uploadfile(\/|$)/,
  ];

  function requestTimeout(req, res, next) {
    if (uploadRoutes.some((r) => r.test(req.path))) return next();
    const timer = setTimeout(() => req.socket.destroy(), 5 * 60 *
  1000);
    res.on('close', () => clearTimeout(timer));
    return next();
  }

  // at the top of module.exports, before the first route:

  module.exports = (app) => {
    app.use(requestTimeout);
```

Edit: This may destroy websockets... let me check

Websockets are fine, but not sure how it would interact with socket.io (could filter `socket.io` from the path but that's getting messy.)

I'd be happy if we just changed it to 2 hours... for a 5Gb file over 2 hours... that would be 5.4Mbps.

MorningLightMountain713 (changes requested):

PR #1728, opened 2026-05-07
Handle g: syncthing mode per component, not per app

Cabecinha84:
## Summary

  Currently FluxOS treats `g:` syncthing mode as an app-level property: if any   
  single component of a compose app uses `g:`, the entire app is treated as
  master/slave-managed. This means non-syncthing components in mixed compose apps
   are left stopped on boot, blocked from auto-restart on crash, and held back by
   a 30-minute install grace they don't need.

  This PR makes the decision **per component** instead.                          
   
  ## Behavior change                                                             
                                                                  
  | Scenario | Before | After |
  |---|---|---|
  | Boot on a slave node with `frontend` (no g:) + `db` (g:) | Both stay stopped 
  indefinitely | `frontend` starts; `db` waits for master/slave election |       
  | `frontend` crashes mid-session | Stays dead until next reboot | Auto-restarts
   on the next peer-check cycle |                                                
  | 30 minutes after install | `frontend` held back by syncthing grace |
  `frontend` runs immediately |                                                  
  | Master/slave election for `db` | Works as before | Unchanged |
  | All-`g:` compose app | Skipped entirely | Skipped entirely (unchanged) |     
  | All-non-`g:` app | Starts normally | Starts normally (unchanged) |           
                                                                                 
  ## Files changed                                                               
                                                                  
  - `ZelBack/src/services/appLifecycle/stoppedAppsRecovery.js` — new helper      
  `getNonGComponentIdentifiers()`; boot recovery now partitions components and
  starts each non-g one individually via                                         
  `advancedWorkflows.appDockerStart('comp_app')`. New `appsPartiallyStarted`
  result bucket for mixed compose apps.
  - `ZelBack/src/services/appMessaging/peerNotification.js` — auto-restart
  routing inside `checkAndNotifyPeersOfRunningApps` now classifies the specific  
  stopped component (`stoppedCompIsG` / `stoppedCompIsRorG`) instead of the whole
   app. The 30-minute install grace is also gated per component. App-level       
  broadcast logic (`masterSlaveAppsInstalled`) is preserved.      
  - `tests/unit/stoppedAppsRecovery.test.js` — 9 new tests (mixed compose, all-g
  compose, helper coverage).                                                     
  - `tests/unit/peerNotification.test.js` — 4 new tests (non-g auto-start, g:
  routing, grace inheritance, r: grace still applies).                           
                                                                  
  ## What's intentionally NOT in this PR (potential follow-ups)                  
                                                                  
  1. `advancedWorkflows.appDockerRestartWithPermissionsFix` step 4 still calls   
  `appDockerRestart(appname)`, restarting all components when this node becomes
  master. Cosmetic 5-10s blip, not breaking.                                     
  2. The restore path at `advancedWorkflows.js:2473` still restarts all
  components after restore; the backup path already does the per-component       
  partition. Inconsistency only.
                                                                                 
  ## Risk to flag for app authors                                                
  
  If a non-g component depends on its g: sibling being local (e.g. a frontend    
  hardcoded to `localhost:5432` against a local db container), running the
  frontend on a slave where the db is intentionally stopped will now start a     
  broken half. Apps that route through service mesh / FDM-style discovery are
  unaffected. Worth communicating before this ships.

  ## Test plan

  - [x] `npx mocha tests/unit/stoppedAppsRecovery.test.js                        
  tests/unit/peerNotification.test.js --exit` — 35 passing, including 13 new
  - [x] `node --check` clean on both source files                                
  - [x] No regression: all 22 pre-existing tests in these files still pass       
  - [ ] Manual: deploy a mixed-compose app (frontend + g: db) to a 3-node        
  cluster, reboot a slave, confirm frontend comes up and db stays slave-stopped  
  - [ ] Manual: kill the frontend container on a slave, wait one peer-check cycle
   (~60min in prod, can shorten for testing), confirm it auto-restarts           
  - [ ] Manual: install a fresh mixed app, confirm frontend starts immediately
  (not after 30 min)                                                             
  - [ ] Manual: confirm pure-g: and pure-non-g apps behave exactly as before

### DIFF
--- a/ZelBack/src/services/appLifecycle/stoppedAppsRecovery.js
+++ b/ZelBack/src/services/appLifecycle/stoppedAppsRecovery.js
@@ -2,9 +2,10 @@
  * Stopped Apps Recovery Service
  *
  * This module handles the recovery of Flux applications that are stopped on boot.
- * It checks for installed apps that have stopped containers and starts them,
- * but only if they do NOT use g: syncthing mode (master/slave mode).
- * Apps using g: syncthing mode are managed by the masterSlaveApps service.
+ * It iterates installed apps with stopped containers and starts every component
+ * that is NOT in g: syncthing mode. Components in g: mode are left stopped so the
+ * masterSlaveApps service can elect a primary. For mixed compose apps (some
+ * components g:, some not), only the non-g components are started here.
  */
 
 const config = require('config');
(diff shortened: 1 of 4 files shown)


MorningLightMountain713 (approved):
Ack

PR #1732, opened 2026-05-14
feat: integration test harness for app lifecycle verification

MorningLightMountain713:
## Why this exists

FluxOS manages thousands of applications across a decentralized network. When a node boots, it must sync app hashes, discover peers, register apps, and spawn containers — all through interleaved async operations that depend on daemon state, peer availability, blockchain confirmations, and Docker health. Bugs in this lifecycle (cancelled app re-installations, spawner races, hash sync failures) are invisible until nodes are live on the network, and reproducing them requires a multi-node environment with coordinated state.

We had no way to test any of this. Unit tests verify individual functions in isolation, but the real bugs live in the interactions: what happens when a node boots unconfirmed? When a peer disconnects mid-sync? When the daemon returns an error during a status check? The only "test" was deploying to production and watching logs.

This PR adds a full integration test harness that boots real FluxOS nodes in Docker containers, connects them through a real peer network, and verifies the complete app lifecycle from registration through to Docker container installation. 97 tests across 12 suites, running in ~13 minutes.

## The core innovation: FluxEventBus

The fundamental challenge of integration testing FluxOS is **knowing when something has happened**. In production, you watch logs. In tests, the naive approach is `sleep(30); checkState()` — but this is slow, flaky, and tells you nothing about *what* happened, only what the state looks like at an arbitrary moment.

FluxEventBus solves this. It's a lightweight Server-Sent Events (SSE) stream built into FluxOS that emits structured lifecycle events as they occur:

```
event: peers:added
data: {"outbound":4,"inbound":3}

event: app:installed  
data: {"name":"myapp","ip":"198.18.2.0"}

event: node:statusChecked
data: {"confirmed":true,"synced":true}
```

Tests subscribe to these events and wait for specific conditions:

```javascript
// Instead of: sleep(30); expect(peers.length).gte(4)
await waitForEvent('peers:added', d => d.outbound >= 4);

// Instead of: sleep(120); expect(app.status).to.equal('installed')
await waitForEvent('app:installed', d => d.name === appName);
```

This makes tests **deterministic** — they complete as soon as the condition is met, not after an arbitrary timeout. A peering test that took 54 seconds with polling now completes in 24 seconds because it reacts to the exact event instead of sleeping past it.

**Critically, FluxEventBus is disabled in production.** The SSE endpoint (`GET /flux/eventstream`) is gated on `config.testEventStream` which defaults to `false`. In production, no SSE connections are accepted, no events are buffered, and the event emission calls are no-ops. The bus adds zero overhead to production nodes.

Events are emitted from natural points in the existing code — where FluxOS already logs state changes. The bus doesn't add new logic or change control flow. It simply makes internal state transitions observable from outside the process.

## Why config had to be separated from logic

FluxOS has hundreds of hardcoded timing values scattered across the codebase: cache TTLs, retry intervals, spawn delays, discovery timers, monitor periods. In production these are fine — they're tuned for a network with thousands of nodes and 2-minute block times. But they make integration testing impossible:

- A 60-second API cache means tests can't verify state changes without arbitrary sleeps
- A 2-minute spawn delay means a single install test takes 3+ minutes
- A 30-minute hash sync interval means sync tests would need to run for an hour
- A 120-second connection backoff means peering tests timeout waiting for retries

The solution was extracting these values into `config/default.js` where they keep their exact production values, while the test environment overlays compressed timings (1-second caches, 5-second delays). Every extracted value matches the original hardcoded constant exactly — the config keys are new, but the runtime behavior is identical.

## What's in the harness

**Test infrastructure** (`test-infra/`):
- **Daemon stub** — simulates fluxd/benchmarkd RPC responses with a control API for test manipulation (block advancement, node status overrides, deterministic list manipulation, RPC failure injection). Requires `FLUX_TEST_HARNESS=true` env guard — exits immediately without it.
- **Syncthing stub** — minimal health endpoint for availability checks
- **Docker-in-Docker FluxOS containers** — each runs a real FluxOS instance with its own MongoDB database, Docker daemon, and peer identity. Docker named volumes provide real block device storage for FluxFS volume creation.
- **Testcontainers orchestration** — per-suite isolated environments with static IPs on a /16 network, automatic cleanup via Ryuk

**Test framework** (`test-infra/runner/`):
- **SSE client** — connects to each node's FluxEventBus, buffers events, provides `waitForEvent(type, predicate, timeout)` with both buffer-check and live-listener to eliminate race windows
- **DB client** — direct MongoDB queries for hash counts, app specs, permanent messages. Verifies database state independently of API responses.
- **Daemon control** — programmatic block advancement, ticker start/stop, node status manipulation
- **Network manipulation** — disconnect/reconnect individual nodes from the Docker network for partition testing

**12 test suites, 97 tests:**

| Suite | What it verifies |
|---|---|
| 01 Auth | Admin/owner/team/user authentication, signature validation |
| 02 Boot prerequisites | Daemon readiness, IP detection, geolocation, benchmark polling, discovery start |
| 03 Explorer sync | Block processing, scanned height advancement |
| 04 Confirmation state | Confirmed boot, unconfirmed boot, runtime confirmation loss |
| 05 Peering | Peer discovery, outbound/inbound connection establishment |
| 06 Daemon interaction | RPC routing, status monitoring, deterministic list, failure recovery |
| 07 Hash sync | Late-joining node sync, network partition recovery, stale state recovery |
| 08 App registration | Spec submission, temp message propagation, blockchain confirmation |
| 09 App spawning | Spec propagation, spawner decision, full Docker container installation |
| 10 App removal | Confirmation loss blocks spawner, DOS boundary behavior |
| 11 DOS state | DOS threshold boundaries via loginPhrase, dosMessage blocking, recovery |
| 12 Ticker control | Block advancement, height setting, status overrides, RPC failure injection |

## Production code changes

All production changes serve testability without altering runtime behavior:

- **Config extraction** — 26 timing/interval constants moved to `config/default.js` with identical production values
- **FluxEventBus** — new SSE endpoint gated on `config.testEventStream` (default `false`). Zero overhead in production — no connections accepted, no events buffered, emission calls are no-ops
- **FluxRpc error propagation** — `run()` now checks the `error` field on JSON-RPC 200 responses. Previously `{ result: null, error: { code: -28 } }` was treated as success with null data. All 156 callers already guard on `status === 'success'`.
- **Route cache config** — 193 `cache()` calls in `routes.js` reference config keys instead of hardcoded strings. Production values unchanged.
- **Discovery control** — `discoveryAutostart` config (default `true`) and `GET /flux/startdiscovery` API (fluxteam auth) for test-controlled peer discovery timing
- **DOS state gate** — `POST /flux/dosstate` checks `config.testEventStream` before processing (404 in production)
- **`app:specStored` event** — emitted from `registryManager.js` when app specs are written, enabling event-driven test waits

## Security

- All test keys are generated via `randomBytes(32)` in `generate-keys.js` — no real keys anywhere in test fixtures
- Test flux team zelid differs from production
- Daemon stub requires `FLUX_TEST_HARNESS=true` env guard — won't start without it
- `POST /flux/dosstate` and FluxEventBus SSE gated on `config.testEventStream` — disabled in production
- All config changes are in git-tracked `default.js` — uniform across all nodes on the same version

## What's next

- Deploy to 2-3 production nodes to verify the config extraction doesn't affect real-world behavior
- Run the harness against the `feature/event-log-running-state` branch (signed sync requests, ephemeral state sync, orchestrator state machine)
- Add suites for the remaining lifecycle paths: boot reconciliation, daemon staleness thresholds, message capability gating

## Test plan

- [x] All 97 integration tests passing (12 suites, ~13 minutes)
- [x] Unit tests unaffected (226 passing)
- [x] Security audit: no real keys, credentials, or production addresses in test fixtures
- [ ] Deploy to 2-3 confirmed nodes and verify normal operation (spawning, hash sync, peering) with the config extraction in place


MorningLightMountain713:
Recreating from upstream branch for CI

PR #1735, opened 2026-05-20
v8.12.1

Cabecinha84:
---
  v8.12.1

  Patch release bundling two fixes.

  🐛 Fixes

  Preserve app height during enterprise image update redeploy

  decryptEnterpriseApps defaults to formatSpecs: true, which runs specificationFormatter and strips the height field. When imageUpdateService triggered
   a soft redeploy, the spec — now missing height — was written back to the local DB. The next expireGlobalApplications check then treated the missing
  height as expired and force-removed the app mid-reinstall.

  checkForImageUpdates now calls decryptEnterpriseApps with { formatSpecs: false } so the raw spec (including height) is preserved.

  Return 404 for file-like URLs in SPA fallback

  The catch-all SPA route served index.html for every unmatched route, so requests for missing static assets (e.g. /llms.txt, /ads.txt) returned a
  "false 200" containing the SPA shell instead of a proper 404.

  The fallback in both app.js and homeServer.js now returns a plain-text 404 for any URL with a file extension, while extensionless routes continue to
  serve the SPA shell.

  📦 Other

  - Bump version 8.12.0 → 8.12.1

  Changes

  ┌────────────────────────────────────────────┬─────────┐
  │                    File                    │   +/−   │
  ├────────────────────────────────────────────┼─────────┤
  │ ZelBack/src/services/imageUpdateService.js │ +1 / −1 │
  ├────────────────────────────────────────────┼─────────┤
  │ app.js                                     │ +8 / −1 │
  ├────────────────────────────────────────────┼─────────┤
  │ homeServer.js                              │ +8 / −1 │
  ├────────────────────────────────────────────┼─────────┤
  │ package.json                               │ +1 / −1 │
  └────────────────────────────────────────────┴─────────┘


MorningLightMountain713 (approved):
Ack

PR #1736, opened 2026-05-22
feat: app-to-app network linking via networkWith description token

Cabecinha84:
## Summary

Lets an app share a private network with other apps and reach their
containers — as if both apps were one. The owner opts in by embedding a
token in the app **description** text:

```
networkWith:[appA,appB]
```

Brackets are required, quotes are optional, the `networkWith` key is matched
case-insensitively, and names are comma-separated. An app whose description
has no (or a malformed) token behaves exactly as before.

This is **purely node-local behaviour** — no app specification field, no
validation change, no network/consensus impact. The `description` field
already exists and is signed; the node simply parses it.

A second, related capability rides on top of the same link: a `LOG=SEND`
component in app B can now ship to a `LOG=COLLECT` component in a linked
app A (previously only same-app sender→collector worked).

## Behaviour

When the token is present:

- **Before install / soft redeploy / hard redeploy**, the node verifies every
  named app is installed locally **and owned by the same owner** — otherwise
  the operation fails.
- Each of the app's component containers is attached to the private docker
  network of every linked app (`fluxDockerNetwork_<linked>`), so it can reach
  that app's components by their docker DNS name `flux<component>_<linkedApp>`.
- When a linked-to app is itself (re)deployed, every locally installed app
  networked with it is reconnected to its network (auto-heals the gap a hard
  redeploy would otherwise leave).

If a linked app is uninstalled, the dependent keeps running and is
automatically reconnected when the linked app returns.

### Cross-app log shipping

If a `LOG=SEND` component is being installed in an app whose own compose has
no `LOG=COLLECT` component, the node walks every `networkWith`-linked app in
declaration order and ships to the first linked app exposing a collector. The
container's Docker GELF log driver is wired to the linked collector's IP at
create time; reachability is provided by the network linking above.

Fallbacks (warning logged, container drops to `json-file` logging):
- the linked collector container exists but isn't reachable at install time;
- the linked app is enterprise with `compose` blanked in the local DB and
  cannot be decrypted on this node.

## Implementation

- **New module `appLifecycle/appNetworkLinker.js`** — token parser, install
  gate, forward/reverse network wiring, and cross-app log-collector
  resolution (`findLinkedAppLogCollector`).
- **`dockerService` additions** — `appDockerNetworkConnect` (idempotent
  attach via inspect-first), `getAppContainerNames` /
  `getAppContainerObjects` (anchored regex match for v4+ component
  containers plus literal match for legacy v≤3 single-component containers;
  also consumed by `getNextAvailableIPForApp` to eliminate the duplicated
  inline filter).
- `appDockerCreate` falls back to `findLinkedAppLogCollector` for `LOG=SEND`
  components when no in-compose collector exists.
- The gate and forward wiring run inside `installApplicationHard` /
  `installApplicationSoft` (the only callers of `appDockerCreate`), so **every
  container-creation path is covered**, including direct callers that bypass
  `registerAppLocally` (container health recovery, legacy v≤3 redeploys). The
  gate also runs early in `registerAppLocally` / `softRegisterAppLocally` to
  fail fast before any side effects.
- Reverse wiring runs in `registerAppLocally` / `softRegisterAppLocally`; a
  boot-time reconcile sweep in `appStartupManager` re-applies all links.
- Validation-side regex constants (`APP_NAME_REGEX`, `APP_NAME_REGEX_LEGACY`)
  centralised in `utils/appConstants.js` and consumed by both `appValidator`
  and `appNetworkLinker`.

`appUninstaller`, `appSpawner`, `appValidator` (behaviour) and routes are
unchanged.

## Testing

- `tests/unit/appNetworkLinker.test.js` — 30 cases (parser edge cases,
  install gate, forward/reverse wiring, boot reconcile, cross-app log
  collector resolution including the legacy `enviromentParameters` typo and
  blanked-compose enterprise fallback).
- `tests/unit/dockerService.test.js` — 6 cases for `appDockerNetworkConnect`
  (inspect-first / already-attached / inspect-fails / race-window /
  generic-404 / overloaded-403) plus 2 cases for `getAppContainerNames`
  (multi+legacy match anchored to flux/zel; regex-metachar escape).
- All affected suites baseline-compared: zero regressions across 380 unit
  tests in the touched modules.

## Notes for reviewers

- Apps must be **co-located** for network linking to work — placement is
  passive: an app simply fails to install on nodes lacking its linked apps
  and the spawn loop retries elsewhere.
- A network link may only target an app with the **same owner**.
- The declaration is one-directional (app B writes `networkWith:[A]`; A
  declares nothing), but the resulting reachability is mutual.
- Cross-app traffic stays within the existing `172.23.0.0/16` range already
  permitted by the `DOCKER-USER` firewall rules.
- Cross-app log shipping inherits the same co-location and same-owner
  guarantees from the network link it rides on. The linked collector must be
  running when the sender installs; if it isn't, the sender silently drops
  to `json-file` (a warning is logged) and there is no auto-rewire when the
  collector later starts. Acceptable for v1; a reactive rewire can follow.

### DIFF
--- a/ZelBack/src/services/appLifecycle/appNetworkLinker.js
+++ b/ZelBack/src/services/appLifecycle/appNetworkLinker.js
@@ -24,10 +24,7 @@ const config = require('config');
 const dbHelper = require('../dbHelper');
 const dockerService = require('../dockerService');
 const log = require('../../lib/log');
-const { localAppsInformation } = require('../utils/appConstants');
-
-// Flux app name syntax (version 8 superset — alphanumerics with internal hyphens).
-const appNameRegex = /^[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$/;
+const { localAppsInformation, APP_NAME_REGEX } = require('../utils/appConstants');
 
 /**
  * Parses the `networkWith:[...]` token out of an app description.
@@ -48,7 +45,7 @@ function parseNetworkWith(description) {
   match[1].split(',').forEach((raw) => {
     const name = raw.trim().replace(/^["']+|["']+$/g, '').trim();
     const key = name.toLowerCase();
-    if (name && appNameRegex.test(name) && !seen.has(key)) {
+    if (name && APP_NAME_REGEX.test(name) && !seen.has(key)) {
       seen.add(key);
       names.push(name);
     }
@@ -69,31 +66,6 @@ function getLinkedApps(appSpecs) {
   }
   const selfName = String(appSpecs.name).toLowerCase();
   return parseNetworkWith(appSpecs.description).filter((linked) => linked.toLowerCase() !== selfName);
-}
-
-/**
- * Resolves the docker container names belonging to an installed app by
- * inspecting docker directly. This is robust for enterprise apps, whose stored
- * `compose` array is blanked in the local database.
- *
- * @param {string} appName - application name
- * @returns {Promise<string[]>} docker container names (without leading slash)
- */
-async function getAppContainerNames(appName) {
-  const containers = await dockerService.dockerListContainers(true);
-  const singleComponentName = dockerService.getAppIdentifier(appName);
-  const names = [];
-  (containers || []).forEach((container) => {
-    (container.Names || []).forEach((rawName) => {
-      const name = rawName.replace(/^\//, '');
-      // component container: flux<component>_<appName>; single-component app: flux<appName> / zel<appName>
-      const belongsToApp = name === singleComponentName || name.endsWith(`_${appName}`);
-      if (belongsToApp && !names.includes(name)) {
-        names.push(name);
-      }
-    });
-  });
-  return names;
 }
 
 /**
@@ -187,7 +159,7 @@ async function reconnectLinkedApps(appName) {
     }
     try {
       // eslint-disable-next-line no-await-in-loop
-      const containerNames = await getAppContainerNames(app.name);
+      const containerNames = await dockerService.getAppContainerNames(app.name);
       // eslint-disable-next-line no-restricted-syntax
       for (const containerName of containerNames) {
         // eslint-disable-next-line no-await-in-loop
@@ -226,7 +198,7 @@ async function reconcileAllAppNetworkLinks() {
     }
     try {
       // eslint-disable-next-line no-await-in-loop
-      const containerNames = await getAppContainerNames(app.name);
+      const containerNames = await dockerService.getAppContainerNames(app.name);
       // eslint-disable-next-line no-restricted-syntax
       for (const linkedApp of linkedApps) {
         const networkName = `fluxDockerNetwork_${linkedApp}`;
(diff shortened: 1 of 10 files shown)


ZelBack/src/services/appLifecycle/appNetworkLinker.js
```diff
@@ -0,0 +1,255 @@
...
+ */
+
+const config = require('config');
+const dbHelper = require('../dbHelper');
+const dockerService = require('../dockerService');
+const log = require('../../lib/log');
+const { localAppsInformation } = require('../utils/appConstants');
+
+// Flux app name syntax (version 8 superset — alphanumerics with internal hyphens).
+const appNameRegex = /^[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$/;
```
MorningLightMountain713: This is the same regex as in appValidator.js, should move it to appConstants and use it in both (All regexes codebase wide should be in there)

ZelBack/src/services/appLifecycle/appNetworkLinker.js
```diff
@@ -0,0 +1,255 @@
...
+
+/**
+ * Resolves the docker container names belonging to an installed app by
+ * inspecting docker directly. This is robust for enterprise apps, whose stored
+ * `compose` array is blanked in the local database.
+ *
+ * @param {string} appName - application name
+ * @returns {Promise<string[]>} docker container names (without leading slash)
+ */
+async function getAppContainerNames(appName) {
```
MorningLightMountain713: This is a duplicate. Use the one from dockerService

MorningLightMountain713: The endsWIth is a bit dodgey too... it's safe for now as app names don't have underscores, but would be safer with a full regex like `/^(?:flux|zel)[a-zA-Z0-9]+_${appName}$/`

ZelBack/src/services/dockerService.js
```diff
@@ -1424,6 +1424,27 @@ async function forceRemoveFluxAppDockerNetwork(appname) {
...
+ * @param {string} networkName - target docker network name
+ * @returns {Promise<void>}
+ */
+async function appDockerNetworkConnect(containerIdOrName, networkName) {
+  const network = docker.getNetwork(networkName);
+  try {
+    await network.connect({ Container: containerIdOrName });
+  } catch (error) {
+    // 403 - endpoint already exists in network (container already connected)
+    if (error.statusCode === 403 || /already exists|already connected/i.test(error.message || '')) {
```
MorningLightMountain713: This is brittle. Maybe a better approach would be to inspect the container and only connect the network if it is not already connected. dockerService already has the tools for this

MorningLightMountain713 (changes requested):
Looks good. Small changes

ZelBack/src/services/appLifecycle/appNetworkLinker.js
Cabecinha84: should be fixed

ZelBack/src/services/appLifecycle/appNetworkLinker.js
Cabecinha84: should be fixed

ZelBack/src/services/dockerService.js
Cabecinha84: should be fixed

MorningLightMountain713 (approved):
Ack

PR #1737, opened 2026-05-24
fix: standardize socket address handling for ip:port transition

MorningLightMountain713:
## Problem

FluxOS has a systemic fragility around IP address handling. The root cause is that fluxbench historically omits the API port for nodes running on the default port (16127), reporting just `1.2.3.4` instead of `1.2.3.4:16127`. This forces FluxOS to compensate — every file that touches node IPs has to decide whether to strip a port, add a port, or compare with `.split(':')[0]`. Over time this led to ~30 files each implementing their own ad-hoc parsing: `ip.split(':')[0]`, `ip.split(':')[1] || '16127'`, `myIP === node.ip`, and various conditional port-stripping patterns — all slightly different, all easy to break.

A companion fluxbench change will always attach the API port, but deploying it against the current codebase would cause nodes to go DOS — remote nodes would fail to match `1.2.3.4:16127` against `1.2.3.4` in their node lists and report the node as unreachable. This PR must be deployed network-wide first so all nodes can handle both formats during the transition.

## Approach

Rather than patching individual comparison sites, this PR consolidates all socket address logic into a single utility module (`socketAddressUtils.js`) and refactors every consumer to use it:

- **`normalizeSocketAddress(addr)`** — ensures ip:port format, defaults to :16127
- **`extractIp(addr)`** — replaces every `.split(':')[0]`
- **`extractPort(addr)`** — replaces every `.split(':')[1] || '16127'` (always returns a number, falls back to DEFAULT_API_PORT for malformed input)
- **`parseSocketAddress(raw)`** — validates and parses user input with strict IP and port regex (moved from `serviceHelper.normalizeNodeIpApiPort`)
- **`socketAddressesMatch(a, b)`** — format-agnostic comparison, replaces all `=== myIP` checks

### Variable naming cleanup

The codebase used `myFluxIP`, `myIP`, `ip` etc. interchangeably for both bare IPs and socket addresses. This made it impossible to tell at a glance what format a variable held. This PR establishes a naming convention:

| Old | New | Holds |
|-----|-----|-------|
| `myFluxIP` / `myIP` | `localSocketAddr` | socket address (ip:port) |
| `getMyFluxIPandPort()` | `getLocalSocketAddress()` | returns normalized ip:port |
| `setMyFluxIp()` | `setLocalSocketAddress()` | normalizes on set |
| `myUrl` / `myIpWithoutPort` | `localIp` | bare IP only |

### Manual benchmark fetch consolidation

Six files were doing their own `benchmarkService.getBenchmarks()` call and extracting `myIP` from the response instead of using the centralized `getLocalSocketAddress()`. These have been refactored to use the single path:

- `appInstaller.js`, `systemIntegration.js`, `fluxService.js`, `peerNotification.js`, `hwRequirements.js` — replaced manual fetch with `getLocalSocketAddress()`
- `containerHealthMonitor.js`, `appUtilities.js` — renamed `myIP` parameter to `localSocketAddr`
- `appSpecs.nodes.includes(myIP)` replaced with `.find((node) => socketAddressesMatch(node, localSocketAddr))` matching the pattern already in `appSpawner.js`

### What was removed

- `normalizeNodeIpApiPort` from `serviceHelper.js` — socket address validation doesn't belong in a general helper. Replaced by `parseSocketAddress` in `socketAddressUtils`.
- `defaultPortAltAddress` from `networkStateService.js` — was a duplicate of `normalizeSocketAddress` added as a quick compatibility fix. Now uses the shared utility, gated on `DEFAULT_API_PORT` to prevent false positives for non-default-port lookups.
- Every manual `.split(':')` pattern for socket addresses across the codebase (replaced with `extractIp`/`extractPort`).
- All hardcoded `16127` magic numbers in service code replaced with `DEFAULT_API_PORT` from `socketAddressUtils`.

### Hardening

- `extractPort` returns `DEFAULT_API_PORT` for malformed input (trailing colon, non-numeric port) instead of 0 or NaN
- `parseSocketAddress` uses strict port regex (1-65535 decimal only — rejects hex, leading zeros, trailing garbage)
- `socketAddressesMatch(null, null)` returns `false` (not `true` via the `a === b` fast path)
- `sortAndFilterLocations` normalizes addresses before sorting to ensure deterministic order with mixed formats
- API response port fields preserve string type to maintain backward compatibility

## Rollout

This is one half of a two-part change. The sequence matters:

1. **This PR rolls out network-wide first** — all nodes gain `socketAddressesMatch()` so they can handle both formats
2. **Then** the fluxbench change can be deployed — nodes will accept the new ip:port format without going DOS

Once all nodes run the new fluxbench, `normalizeSocketAddress` becomes a no-op and `socketAddressesMatch` becomes `===`. At that point the module can be simplified further.

## Files changed (45)

- **New**: `socketAddressUtils.js`, `socketAddressUtils.test.js`
- **Core refactor** (30 service files): renamed functions/variables, replaced comparisons, split patterns, and manual benchmark fetches
- **Tests updated** (13 test files): mock renames + 46 new utility tests + format-transition tests + mixed-format sort/filter tests

## Deployment verification

Deployed to chalk, charlie, and chud. All three nodes:
- Found on valid indexes in the deterministic node list
- Confirmed with `messageCapable=true`
- Geolocation resolving correctly
- No errors from the refactor

## Test plan

- [x] 460 unit tests passing across 13 test files
- [x] `socketAddressUtils.test.js` — 46 tests covering all functions and edge cases
- [x] Format-transition tests verify bare IP → ip:port normalization
- [x] Mixed-format sort and filter tests for syncthing peer ordering
- [x] Live deployment on 3 nodes — healthy startup confirmed
- [ ] Monitor deployed nodes through 2+ benchmark cycles
- [ ] Verify no DOS state changes over 24h

### DIFF
--- a/ZelBack/src/services/analyticsService.js
+++ b/ZelBack/src/services/analyticsService.js
@@ -9,7 +9,7 @@ let bufferTotal = 0;
 let isFlushing = false;
 let consecutiveFailures = 0;
 let lastFailureTime = 0;
-let cachedNodeIp = null;
+let cachedNodeSocketAddress = null;
 let flushTimer = null;
 let analyticsUrlCached = '';
 let initialized = false;
@@ -26,9 +26,9 @@ const NODE_IP_REFRESH_INTERVAL = 600000; // 10 minutes
  */
 async function refreshNodeIp() {
   try {
-    const ip = await fluxNetworkHelper.getMyFluxIPandPort();
-    if (ip) {
-      cachedNodeIp = ip;
+    const localSocketAddr = await fluxNetworkHelper.getLocalSocketAddress();
+    if (localSocketAddr) {
+      cachedNodeSocketAddress = localSocketAddr;
     }
   } catch (error) {
     log.warn(`Analytics: failed to get node IP: ${error.message}`);
@@ -194,7 +194,7 @@ function analyticsMiddleware(req, res, next) {
         responseCode: res.statusCode,
         responseStatus: res.statusCode < 400 ? 'success' : 'error',
         responseTimeMs: Date.now() - startTime,
-        nodeIp: cachedNodeIp,
+        nodeIp: cachedNodeSocketAddress,
         ipAddress: clientIp,
         timestamp: new Date().toISOString(),
       };
@@ -236,7 +236,7 @@ function trackTerminalSession(zelidauth, appName, action, ipAddress, component) 
     responseCode: 200,
     responseStatus: 'success',
     responseTimeMs: 0,
-    nodeIp: cachedNodeIp,
+    nodeIp: cachedNodeSocketAddress,
     ipAddress: ipAddress || null,
     timestamp: new Date().toISOString(),
   };
--- a/ZelBack/src/services/appLifecycle/advancedWorkflows.js
+++ b/ZelBack/src/services/appLifecycle/advancedWorkflows.js
@@ -13,7 +13,7 @@ const dockerService = require('../dockerService');
 const verificationHelper = require('../verificationHelper');
 const daemonServiceMiscRpcs = require('../daemonService/daemonServiceMiscRpcs');
 const fluxNetworkHelper = require('../fluxNetworkHelper');
-const { extractIp, extractPort, socketAddressesMatch } = require('../utils/socketAddressUtils');
+const { DEFAULT_API_PORT, extractIp, extractPort, socketAddressesMatch } = require('../utils/socketAddressUtils');
 const generalService = require('../generalService');
 // eslint-disable-next-line no-unused-vars
 const upnpService = require('../upnpService');
@@ -3910,7 +3910,7 @@ async function masterSlaveApps(globalStateParam, installedApps, listRunningApps,
                   // Look up the correct port from runningAppList since FDM API returns IP without port
                   const previousMasterNode = runningAppList.find((x) => socketAddressesMatch(x.ip, previousMasterIp));
                   const ipToCheckAppRunning = extractIp(previousMasterIp);
-                  const portToCheckAppRunning = previousMasterNode ? extractPort(previousMasterNode.ip) : 16127;
+                  const portToCheckAppRunning = previousMasterNode ? extractPort(previousMasterNode.ip) : DEFAULT_API_PORT;
                   let previousMasterStillRunning = false;
                   try {
                     // eslint-disable-next-line no-await-in-loop
--- a/ZelBack/src/services/appLifecycle/advancedWorkflows.js
+++ b/ZelBack/src/services/appLifecycle/advancedWorkflows.js
@@ -13,6 +13,7 @@ const dockerService = require('../dockerService');
 const verificationHelper = require('../verificationHelper');
 const daemonServiceMiscRpcs = require('../daemonService/daemonServiceMiscRpcs');
 const fluxNetworkHelper = require('../fluxNetworkHelper');
+const { extractIp, extractPort, socketAddressesMatch } = require('../utils/socketAddressUtils');
 const generalService = require('../generalService');
 // eslint-disable-next-line no-unused-vars
 const upnpService = require('../upnpService');
@@ -3048,9 +3049,9 @@ async function checkAndRemoveApplicationInstance() {
             return 0;
           });
           // eslint-disable-next-line no-await-in-loop
-          const myIP = await fluxNetworkHelper.getMyFluxIPandPort();
-          if (myIP) {
-            const index = runningAppList.findIndex((x) => x.ip === myIP);
+          const localSocketAddr = await fluxNetworkHelper.getLocalSocketAddress();
+          if (localSocketAddr) {
+            const index = runningAppList.findIndex((x) => socketAddressesMatch(x.ip, localSocketAddr));
             if (index === 0) {
               log.info(`Application ${installedApp.name} going to be removed from node as it was the latest one running it to install it..`);
               log.warn(`REMOVAL REASON: Too many instances - ${installedApp.name} running on ${runningAppList.length} instances (max: ${minInstances}) - This node is the newest instance`);
@@ -3549,8 +3550,8 @@ async function forceAppRemovals() {
     const appUninstaller = require('./appUninstaller');
 
     // Get current node's IP for checking app locations
-    const myIP = await fluxNetworkHelper.getMyFluxIPandPort();
-    if (!myIP) {
+    const localSocketAddr = await fluxNetworkHelper.getLocalSocketAddress();
+    if (!localSocketAddr) {
       log.warn('Unable to get node IP, skipping forceAppRemovals');
       return;
     }
@@ -3587,15 +3588,15 @@ async function forceAppRemovals() {
         // Check if this app is registered in locations for this node's IP
         let shouldBroadcast = false;
         try {
-          const locationQuery = { name: dApp, ip: myIP };
+          const locationQuery = { name: dApp, ip: localSocketAddr };
           const locationProjection = { projection: { _id: 0 } };
           // eslint-disable-next-line no-await-in-loop
           const appLocation = await dbHelper.findOneInDatabase(database, globalAppsLocations, locationQuery, locationProjection);
           if (appLocation) {
             shouldBroadcast = true;
-            log.info(`${dApp} found in locations for this IP (${myIP}), will broadcast removal`);
+            log.info(`${dApp} found in locations for this IP (${localSocketAddr}), will broadcast removal`);
           } else {
-            log.info(`${dApp} not found in locations for this IP (${myIP}), skipping broadcast`);
+            log.info(`${dApp} not found in locations for this IP (${localSocketAddr}), skipping broadcast`);
           }
         } catch (locationError) {
           log.error(`Error checking app location for ${dApp}: ${locationError.message}`);
@@ -3769,19 +3770,16 @@ async function masterSlaveApps(globalStateParam, installedApps, listRunningApps,
           // no ip means there was no row with ip on fdm
           // down means there was a row ip with status down
           // eslint-disable-next-line no-await-in-loop
-          let myIP;
+          let localSocketAddr;
           try {
             // eslint-disable-next-line no-await-in-loop
-            myIP = await fluxNetworkHelper.getMyFluxIPandPort();
+            localSocketAddr = await fluxNetworkHelper.getLocalSocketAddress();
           } catch (error) {
             log.error(`masterSlaveApps: Failed to get my IP for app:${installedApp.name}, error: ${error.message}`);
             // eslint-disable-next-line no-continue
             continue;
           }
-          if (myIP) {
-            if (myIP.indexOf(':') < 0) {
-              myIP += ':16127';
-            }
+          if (localSocketAddr) {
             // Validate ip is a string if it exists
             if (ip && typeof ip !== 'string') {
               log.error(`masterSlaveApps: Invalid IP type from FDM for app:${installedApp.name}, got: ${typeof ip}`);
@@ -3851,7 +3849,7 @@ async function masterSlaveApps(globalStateParam, installedApps, listRunningApps,
                   }
                   return 0;
                 });
-                const index = runningAppList.findIndex((x) => x.ip.split(':')[0] === myIP.split(':')[0]);
+                const index = runningAppList.findIndex((x) => socketAddressesMatch(x.ip, localSocketAddr));
 
                 // Helper function to check if any lower-index nodes are running the app
                 const checkLowerIndexNodesRunning = async () => {
@@ -3865,8 +3863,8 @@ async function masterSlaveApps(globalStateParam, installedApps, listRunningApps,
                     const nodeToCheck = runningAppList[i];
                     if (!nodeToCheck) continue;
 
-                    const ipToCheck = nodeToCheck.ip.split(':')[0];
-                    const portToCheck = nodeToCheck.ip.split(':')[1] || '16127';
+                    const ipToCheck = extractIp(nodeToCheck.ip);
+                    const portToCheck = extractPort(nodeToCheck.ip);
                     const source = CancelToken.source();
                     let isResolved = false;
 
@@ -3898,7 +3896,7 @@ async function masterSlaveApps(globalStateParam, installedApps, listRunningApps,
                   // Index 0: Start immediately if no history
                   appDockerRestartWithPermissionsFix(installedApp.name, appId);
                   log.info(`masterSlaveApps: starting docker app:${installedApp.name} index: ${index}`);
-                } else if (!timeTostartNewMasterApp.has(identifier) && mastersRunningGSyncthingApps.has(identifier) && mastersRunningGSyncthingApps.get(identifier) !== myIP) {
+                } else if (!timeTostartNewMasterApp.has(identifier) && mastersRunningGSyncthingApps.has(identifier) && !socketAddressesMatch(mastersRunningGSyncthingApps.get(identifier), localSocketAddr)) {
                   // There was a previous master (not me), and it's no longer on FDM
                   const { CancelToken } = axios;
                   const source = CancelToken.source();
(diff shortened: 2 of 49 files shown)


Cabecinha84 (commented):
Deep Analysis: PR #1737 — Socket Address Standardization

  I went through every changed file plus called audits across the whole repo. Here's what I found, ordered by severity.
  
  ---
  🔴 BUG #1 — apiServer.js calls removed function (will crash at shutdown)
  
  Severity: Critical — silent runtime failure on graceful shutdown.
  
  apiServer.js:392 still calls fluxNetworkHelper.getMyFluxIPandPort():
  
  const ip = await fluxNetworkHelper.getMyFluxIPandPort();
  
  That function was renamed to getLocalSocketAddress and removed from the module's exports (fluxNetworkHelper.js:2221,
  :2251). At runtime, ip will be await undefined → undefined → the sigterm broadcast in the shutdown handler is skipped
  silently. Peers won't get the fluxnodesigterm event when the node shuts down, so app re-spawn elsewhere is delayed until
   TTL.
  
  Why missed: apiServer.js isn't in the PR's diff. The author didn't grep for stale call sites outside
  ZelBack/src/services/. ESLint can't catch dynamic property accesses.
  
  Fix: rename the call to getLocalSocketAddress().

  ---
  🔴 BUG #2 — appSpawner.js uses extractPort without importing it (will crash on syncthing app spawn)
  
  Severity: Critical — ReferenceError thrown in real code path.

  appSpawner.js:11:
  const { normalizeSocketAddress, extractIp, socketAddressesMatch } = require('../utils/socketAddressUtils');

  But lines 470 and 489 use extractPort(node.ip) — not in the import. ESLint reports this directly:
  470:24  error  'extractPort' is not defined  no-undef
  489:24  error  'extractPort' is not defined  no-undef
  
  These lines are inside trySpawningGlobalApplication → the syncthing connectivity check loop. So any spawn of a syncthing
   app (containerData with g:/r:/s:) that has an existing running/installing peer will throw → caught by the outer
  try/catch → spawn fails. Node-wide syncthing app spawning is broken.
  
  Why missed: Tests don't run trySpawningGlobalApplication in a way that reaches that loop, and npm run lint likely wasn't
   run (or was ignored).

  Fix: add extractPort to the import.
  
  ---
  🔴 BUG #3 — socketAddressInNetworkState widened alt-format fallback beyond default-port (false positives for 
  non-default-port lookups)

  Severity: High — semantic regression introduced by the refactor commit d6650e95.
  
  The original defaultPortAltAddress (commit 939de5d1) was correct — it returned null for non-default-port addresses, so
  no fallback ran:
  
  function defaultPortAltAddress(socketAddress) {
    if (socketAddress.includes(':')) {
      const [ip, port] = socketAddress.split(':');
      if (+port === DEFAULT_API_PORT) return ip;   // only fallback for default port
    } else {
      return `${socketAddress}:${DEFAULT_API_PORT}`;
    }
    return null;
  }

  The refactor replaced it with (networkStateService.js:147-152):
  
  const normalized = normalizeSocketAddress(socketAddress);
  const bareIp = extractIp(socketAddress);
  const alt = normalized === socketAddress ? bareIp : normalized;
  if (alt) return stateManager.includes(alt, 'socketAddress');
  
  Trace "1.2.3.4:16137":
  - normalized = "1.2.3.4:16137" → equals input → alt = bareIp = "1.2.3.4"
  - Falls back to looking up bare "1.2.3.4"
  
  If a different node sits at "1.2.3.4" (port 16127), the lookup returns true for an entirely unrelated non-default-port 
  socket address. This affects socketAddressInFluxList → cascades into isFluxAvailable, checkDeterministicNodesCollisions,
   and nodeStatusMonitor. Worst case: a non-default-port node gets falsely treated as confirmed, or a DOS check is
  short-circuited based on a different node's record.
  
  Fix: gate the fallback on extractPort(socketAddress) === DEFAULT_API_PORT. Bring back the spirit of
  defaultPortAltAddress.
  
  Test gap: networkStateService.test.js:145-146 doesn't exercise this edge case. Add a test where '1.2.3.4' is in state
  but lookup is for '1.2.3.4:16137' — should be false.

  ---
  🟡 ISSUE #4 — sortAndFilterLocations still uses strict !== for socket addresses

  Severity: Medium — pre-existing latent bug that this PR's mission was supposed to eliminate, but missed.
  
  syncthingMonitorHelpers.js:74:
  .filter((loc) => loc.ip !== myIP);

  Caller in syncthingMonitor.js:150 now passes localSocketAddr (normalized ip:port). If loc.ip is bare (old fluxbench
  format), this filter doesn't strip the current node's location → current node appears in its own peer list → wasted
  syncthing config / leader-election skew.
  
  Fix: use socketAddressesMatch(loc.ip, localSocketAddr). Same pattern that was applied everywhere else in the PR — this
  one got missed.

  ---
  🟡 ISSUE #5 — Manual normalization left in place in appHashSyncService.js

  Severity: Low — cleanup miss, currently functional.
  
  appHashSyncService.js:501:
  const localSocketAddress = await fluxNetworkHelper.getLocalSocketAddress();
  const selfKey = localSocketAddress.includes(':') ? localSocketAddress : `${localSocketAddress}:16127`;
  
  getLocalSocketAddress() already returns normalized form, so the ternary is redundant. Should be const selfKey = 
  localSocketAddress; or normalizeSocketAddress(localSocketAddress). Also, no null check — if getLocalSocketAddress()
  returns null, the next line throws. Pre-existing, but contradicts the PR's "consolidate everywhere" theme.
  
  ---
  🟡 ISSUE #6 — parseSocketAddress loosened validation vs. normalizeNodeIpApiPort

  Severity: Low — affects HTTP endpoints addPeer, removePeer, removeIncomingPeer.
  
  The old normalizeNodeIpApiPort used a strict port regex ^[1-9][0-9]{0,4}...$ (digits only). The new parseSocketAddress
  uses +parts[1] which accepts:
  - "1.2.3.4:0x10" → {ip: '1.2.3.4', port: 16} (hex)
  - "1.2.3.4:16127:foo" → {ip: '1.2.3.4', port: 16127} (trailing garbage silently dropped)

  The IP regex is also unchanged from the old code, so no IP-side regression. But the looser port handling is a small
  validation reduction at the HTTP boundary. Probably never abused in practice; worth a one-line tightening.
  
  ---
  🟡 ISSUE #7 — Type change in API responses (string → number)

  Severity: Low — externally visible response shape changes.
  
  Several places now produce numeric ports where the old code produced strings:
  - getUnstableNodes response: port: 16127 (was '16127') at fluxCommunication.js:1485
  - data.port sent to /flux/checkappavailability: number, was string
  - dockerService label host_ip always normalized to ip:port (was bare for default port)

  For the inter-node checkappavailability flow, the signature verification re-serializes whatever was received, so
  cross-version signatures still verify. But any external consumer (monitoring, dashboards, third-party tooling) reading
  these fields/labels needs to handle the change.
  
  Worth noting in the PR description and migration notes; not a code change needed.

  ---
  🟢 ACK — Things that look right
  
  - socketAddressUtils.js is well-scoped with 39 tests, clear TODO for post-rollout cleanup. Module is good.
  - fluxNetworkHelper rename + normalization: setLocalSocketAddress correctly normalizes on set. getLocalSocketAddress
  returns normalized.
  - fluxCommunication.js: peer key generation (${ip}:${port}) still produces consistent strings even with number ports.
  Compatibility maintained.
  - appLifecycle/advancedWorkflows.js, appController, availabilityChecker, nodeStatusMonitor, portManager, dockerService,
  explorerService, fluxNodeService, geolocationService, nodeConfirmationService — all consistently use the new utilities;
  imports match uses; comparisons via socketAddressesMatch.
  - Test updates correctly rename stubs and the new format-transition tests (fluxNetworkHelper.test.js:271+) explicitly
  cover bare → normalized transition.
  - Rollout sequencing logic (this PR before the fluxbench port-always-attached change) is sound and well-explained in the
   PR body.
  
  ---
  Recommended actions before merge
  
  1. Must fix: Add extractPort to the import in appSpawner.js.
  2. Must fix: Update apiServer.js:392 to call getLocalSocketAddress.
  3. Must fix: Restore the default-port gate in networkStateService.socketAddressInNetworkState, or this PR ships with a
  worse correctness story than the standalone defaultPortAltAddress it was replacing. Add a regression test.
  4. Should fix: Convert sortAndFilterLocations's loc.ip !== myIP to use socketAddressesMatch.
  5. Should clean up: Remove the dead manual normalization in appHashSyncService.js:501.
  6. Run lint (npm run lint) before resubmit — it catches BUG #2.
  7. Mention type changes in PR body (response/label shape changes).

MorningLightMountain713:
Thanks for the thorough review. All issues addressed — here's the response to each:

---

> 🔴 BUG #1 — apiServer.js calls removed function (will crash at shutdown)

Fixed. Renamed to `getLocalSocketAddress()`. Good catch — `apiServer.js` sits above `ZelBack/` and was missed by the original grep scope.

---

> 🔴 BUG #2 — appSpawner.js uses extractPort without importing it (will crash on syncthing app spawn)

Fixed. Added `extractPort` to the destructured import on line 11.

---

> 🔴 BUG #3 — socketAddressInNetworkState widened alt-format fallback beyond default-port

Fixed. The alt-format lookup is now gated on `extractPort(socketAddress) === DEFAULT_API_PORT`. Non-default-port addresses no longer fall back to bare IP lookup, which was the correct behavior of the original `defaultPortAltAddress`.

---

> 🟡 ISSUE #4 — sortAndFilterLocations still uses strict !== for socket addresses

Fixed. Replaced `loc.ip !== myIP` with `!socketAddressesMatch(loc.ip, localSocketAddr)`. Also renamed the `myIP` parameter to `localSocketAddr` and normalized addresses before sorting to ensure deterministic order with mixed formats.

---

> 🟡 ISSUE #5 — Manual normalization left in place in appHashSyncService.js

Fixed. Removed the manual `.includes(':')` ternaries, replaced with `normalizeSocketAddress()` from socketAddressUtils, and added a null guard for `getLocalSocketAddress()`.

---

> 🟡 ISSUE #6 — parseSocketAddress loosened validation vs. normalizeNodeIpApiPort

Fixed. `parseSocketAddress` now uses a strict port regex (`/^([1-9]\d{0,3}|[1-5]\d{4}|6[0-4]\d{3}|65[0-4]\d{2}|655[0-2]\d|6553[0-5])$/`) and rejects inputs with more than one colon. Hex, leading zeros, and trailing garbage are all rejected. Additionally, `extractPort` now returns `DEFAULT_API_PORT` for malformed input (trailing colon, non-numeric port) instead of 0 or NaN.

---

> 🟡 ISSUE #7 — Type change in API responses (string → number)

Fixed. `String(extractPort(...))` is now used at the two API response boundaries (`getUnstableNodes` and `checkappavailability` data payload) to preserve the original string type.

---

Additionally, the review exposed that the refactor scope was too narrow. Six more files were doing their own manual `benchmarkService.getBenchmarks()` call with `myIP` variables instead of using the centralized `getLocalSocketAddress()`. These have all been refactored: `appInstaller.js`, `systemIntegration.js`, `fluxService.js`, `peerNotification.js`, `hwRequirements.js`, `containerHealthMonitor.js`, and `appUtilities.js`. All hardcoded `16127` magic numbers in service code have been replaced with `DEFAULT_API_PORT`.

All changes deployed and verified clean on chalk, charlie, and chud.

Cabecinha84 (commented):
Solid refactor, but reviewers found 4 real correctness concerns worth pushing back on before ack. No blockers in the utility module itself — it's well-designed, well-tested (46 passing tests), and the mechanical
  renames are faithful. The issues are at the seams.

  Must-fix before ack

  1. advancedWorkflows.js:3911, 3936, 3989, 3992 — masterSlave logic breaks for non-default-port nodes
  Original code compared bare IPs only (myIP.split(':')[0] !== ip.split(':')[0]). FDM returns bare IPs (no port). New code uses socketAddressesMatch(localSocketAddr, ip), which normalizes the bare FDM IP to ip:16127. A
  node running on :16137 will never match its own FDM entry → it'll stop its app and try to start it on a "non-master" node. Should compare bare IPs explicitly (extractIp(localSocketAddr) === ip) since FDM doesn't carry
   the port.

  2. networkStateService.js:187 — getFluxnodeBySocketAddress missing the alt-format fallback
  The sibling socketAddressInNetworkState got hardened with a bare-IP/normalized fallback for the transition, but this lookup didn't. Callers — nodeConfirmationService.poll() (gates newMessageCapable) and
  fluxNetworkHelper.js:1350 (gates the DoS-reset / uptime-check branch) — will silently miss self/peers if the network state stores them in the other format. The asymmetry is the bug; push the fallback down into the
  store.

  3. appUtilities.js:1092 appHasValidLocationOnNode — strict DB equality during a mixed-format window
  Query is { ip: localSocketAddr }. globalAppsLocations is populated by peers that may broadcast either format. A miss returns false, and appStartupManager.reconcileAppsOnBoot (185) treats false as "expired" → local app
   uninstall. Switch to ip: { $in: [bareIp, normalized] } or normalize-on-write across the message store.

  4. syncthingFolderStateMachine.js:91-92, 217-218 — leader election sorts raw IPs
  isDesignatedLeader and sortRunningAppList still sort by a.ip < b.ip un-normalized. sortAndFilterLocations was explicitly fixed for the same reason elsewhere in this PR. During the transition some peers will store
  1.2.3.4 and others 1.2.3.4:16127; lexicographically they sort differently, so peers can disagree on which node is the leader. Apply the same normalization here.

MorningLightMountain713:
All four findings addressed in e9cc7d53a, plus additional issues discovered during the sweep. Responding to each:

---

> **1. advancedWorkflows.js:3911, 3936, 3989, 3992 — masterSlave logic breaks for non-default-port nodes**
> Original code compared bare IPs only (myIP.split(':')[0] !== ip.split(':')[0]). FDM returns bare IPs (no port). New code uses socketAddressesMatch(localSocketAddr, ip), which normalizes the bare FDM IP to ip:16127. A node running on :16137 will never match its own FDM entry → it'll stop its app and try to start it on a "non-master" node.

Valid. Fixed lines 3989/3992 to use `extractIp(localSocketAddr) !== ip` / `=== ip` since FDM always returns bare IPs via `extractIp(ips[0])` in `getMasterIpFromFdm`.

Also found and fixed three more comparisons against `mastersRunningGSyncthingApps` which stores the same bare FDM IP — lines 3905, 3918, 3943 all had the same bug. Changed to `extractIp` comparisons.

---

> **2. networkStateService.js:187 — getFluxnodeBySocketAddress missing the alt-format fallback**
> The sibling socketAddressInNetworkState got hardened with a bare-IP/normalized fallback for the transition, but this lookup didn't.

Valid. Added the same alt-format fallback to `getFluxnodeBySocketAddress`. Also fixed both functions' fallback paths which were missing `await` on the async stateManager calls (assigned to const, then return).

---

> **3. appUtilities.js:1092 appHasValidLocationOnNode — strict DB equality during a mixed-format window**
> Query is { ip: localSocketAddr }. globalAppsLocations is populated by peers that may broadcast either format. A miss returns false, and appStartupManager.reconcileAppsOnBoot (185) treats false as "expired" → local app uninstall.

Valid. Extracted a `socketAddressVariants(address)` helper into `socketAddressUtils.js` that returns `[addr, bareIp, normalized]` for default-port nodes, used with `{ $in: variants }` in DB queries. Applied to `appUtilities.js` and also found three more locations with the same issue:
- `advancedWorkflows.js` `forceAppRemovals` location query
- `peerNotification.js:110` `checkAndNotifyPeersOfRunningApps` 
- `containerHealthMonitor.js:179` `monitorAndRecoverApps`

---

> **4. syncthingFolderStateMachine.js:91-92, 217-218 — leader election sorts raw IPs**
> isDesignatedLeader and sortRunningAppList still sort by a.ip < b.ip un-normalized.

Valid. Extracted a `lexSortBySocketAddress(a, b)` comparator into `socketAddressUtils.js` and replaced all 5 inline sort patterns:
- `syncthingMonitorHelpers.js` — `sortAndFilterLocations` and `sortRunningAppList`
- `syncthingFolderStateMachine.js` — `isDesignatedLeader`
- `advancedWorkflows.js` — `checkAndRemoveApplicationInstance` and `masterSlaveApps` (two additional un-normalized sorts found during sweep)

---

### Additional issues found and fixed

**networkStateManager.js:287 — `getRandomSocketAddress` direct `===` comparison**
Used `localSocketAddress === socketAddress` instead of `socketAddressesMatch`. During the transition, this could return the local node's own address when it should be filtered out. Fixed.

**networkStateManager.js:292 — crash when only node is local node**
`getRandomSocketAddress` would crash with `TypeError: Cannot read properties of undefined` when the randomly selected node was the local node and it was the first (or only) element — `iterator.next().value.ip` throws when the iterator is exhausted. Fixed to return `null` gracefully.

---

17 new tests added covering mixed-format sort consistency, `socketAddressVariants`, `lexSortBySocketAddress`, FDM bare-IP matching, networkStateManager fallback, and networkStateService alt-format lookups.

Cabecinha84 (approved):
ack

PR #1726, opened 2026-05-04
feat: event-driven app state sync with event log

MorningLightMountain713:
## Summary

Replaces the timer-based spawner startup (125-minute fixed wait) with an event-driven architecture that syncs all ephemeral app state from peers in seconds. Rewrites the hash sync pipeline for correctness and performance (63 minutes → 47 seconds). Fixes zombie apps caused by stale hash sync state and a fire-and-forget race condition. Also fixes a broken TTL on the install errors collection. Offloads ECDSA signature verification to a worker thread pool, bypasses the gossip pipeline for sync responses, and adds sender-side backpressure to prevent WebSocket pong timeouts during large syncs.

Additionally, reworks the app lifecycle startup to give FluxOS full control of container management. Replaces Docker's `unless-stopped` restart policy with `no` so containers never auto-start — FluxOS decides what to run based on boot context (machine reboot detection, downtime, shutdown reason) and DB readiness. Fixes a reindex race condition where `reindexGlobalAppsInformation` dropped `zelappsinformation` while concurrent cursors were reading it.

### Key changes
- **App state event log**: Single append-only collection (`appstateevents`) replaces `fluxapprunningbroadcasts`. `zelappslocation` remains as a materialized cache. Current running state derived via aggregation — no mutations, no deletes, no orphan risk
- **AppSyncOrchestrator**: State machine coordinating ephemeral sync, hash sync, DB rebuild, and spawner startup via events instead of timers
- **Signed broadcast sync protocol**: Four binary message types (0x20-0x23) for syncing temp messages, apprunning, appinstalling, and install errors from peers over WebSocket
- **Hash sync rewrite**: Streaming sequential fetch with backpressure from up to 3 peers, batch verification, batch insert. 58k messages in ~2 minutes vs ~63 minutes on master
- **Zombie app recovery**: Version-upgrade `messageNotFound` reset + `hashesChanged` event for periodic recovery
- **`updateAppSpecifications` split**: Separated insert (registration, upsert) from update (no upsert, awaited) to prevent cancelled app resurrection
- **Sync pipeline bypass + worker threads**: Sync responses routed directly in `FluxPeerSocket.onmessage`, bypassing the gossip pipeline (no hash, no cache, no random delay). ECDSA verification offloaded to `cpus-1` worker threads via `verifyPool.js`. Per-peer FIFO chunk ordering prevents premature sync completion
[...]

PR #1739, opened 2026-05-27
feat: instant crash recovery via Docker event stream

MorningLightMountain713:
## Summary

Instant crash recovery for Docker containers via event stream subscription, replacing the 1-2 hour polling gap left after PR #1726 removed `restart: always`.

### What it does
- Subscribes to Docker `die` events for real-time crash detection
- Restarts crashed containers (non-zero exit) with exponential backoff
- Suppresses restart for intentional FluxOS stops (uninstall, redeploy, backup, API stop)
- Queues events during boot and drains after `bootContainerStateSettled`
- Reconciles exited containers on FluxOS-only restarts (previously skipped)

### Backoff schedule

| Attempt | Delay | Cumulative |
|---------|-------|------------|
| 1st | immediate | 0s |
| 2nd | 30s | 30s |
| 3rd | 5 min | 5m 30s |
| 4th | 15 min | 20m 30s |
| 5th+ | 30 min (cap) | 50m 30s+ |

Backoff resets after 10 minutes of stable running — if the container ran for 10+ minutes since the last restart, the next crash starts fresh at immediate. Never gives up — permanently broken containers are retried at the 30m cap indefinitely until the operator uninstalls.

### Suppression mechanism

`appDockerStop`/`Kill` add the container to `globalState.stoppingContainers`. `appDockerStart`/`Remove`/`ForceRemove` clear it. The crash recovery handler checks and consumes the entry — no TTL, no leaks.

| Scenario | Behaviour |
|----------|-----------|
| Container crashes | Restarted with backoff |
| `docker stop` clean exit (code 0) | Filtered by exit code, suppression consumed |
| `docker stop` dirty exit (code 137) | Filtered by suppression set |
| Container removed (uninstall) | Suppression cleared on remove |
| Crash during boot | Queued, drained after boot settles |
| Container handled during backoff wait | Checked after delay, skipped |

### Files changed

| File | Change |
|------|--------|
| `containerCrashRecovery.js` (new) | Docker event stream listener, backoff, boot queue, NDJSON line buffer |
| `dockerService.js` | Suppression in stop/kill/restart, cleanup in start/remove, filter passthrough for `getEvents` |
| `appStartupManager.js` | Remove `machineRebooted` gate so FluxOS restarts reconcile exited containers |
| `serviceManager.js` | Wire up crash recovery on startup |
| `globalState.js` | `stoppingContainers` Set |

## Test plan

Verified on live Arcane nodes (46.225.195.244, 119.246.12.164):

| Test | Verified | Evidence |
|------|----------|----------|
| External `docker kill` → immediate restart | Yes | Log: `crashed (exit 137), restarting` → `restarted successfully` |
| Second crash → 30s backoff | Yes | Log: `waiting 30s before restarting` → restarted 30s later |
| FluxOS API stop, clean exit (0) → not restarted | Yes | Container stayed exited, no crash recovery log |
| FluxOS API stop, dirty exit (137) → suppressed | Yes | Log: `was intentionally stopped (exit 137), skipping` |
| Boot reconciliation, expired location → removed | Yes | Log: `no longer has a valid location record, removing locally` |
| Boot reconciliation, no stopped containers → no-op | Yes | Log: `No stopped containers found` |
| Boot queue drain | Yes | Log: `draining 1 queued event(s)` |
| 20 unit tests | Yes | All passing |
| 65 dockerService tests | Yes | All passing |
| 33 appStartupManager tests | Yes | All passing |

Written against bc64c97e7.

### DIFF
--- a/ZelBack/src/services/appLifecycle/appStartupManager.js
+++ b/ZelBack/src/services/appLifecycle/appStartupManager.js
@@ -287,7 +287,7 @@ async function manageAppsOnBoot(bootContext) {
       }
     }
 
-    // Machine rebooted, locations still valid — wait for daemon + sync then reconcile.
+    // Locations still valid — wait for daemon + sync then reconcile.
     const DAEMON_TIMEOUT_MS = config.system.bootDaemonTimeoutMs ?? 300000;
     try {
       await Promise.race([
--- a/ZelBack/src/services/appLifecycle/appStartupManager.js
+++ b/ZelBack/src/services/appLifecycle/appStartupManager.js
@@ -274,11 +274,6 @@ async function removeAllApps(reason) {
 
 async function manageAppsOnBoot(bootContext) {
   try {
-    if (!bootContext.machineRebooted) {
-      log.info('appStartupManager - FluxOS-only restart, containers already running');
-      return;
-    }
-
     if (bootContext.firstBoot) {
       log.info('appStartupManager - First boot (no heartbeat history), waiting for sync');
     } else {
--- a/ZelBack/src/services/appMonitoring/containerCrashRecovery.js
+++ b/ZelBack/src/services/appMonitoring/containerCrashRecovery.js
@@ -7,10 +7,14 @@ const appInspector = require('../appManagement/appInspector');
 const BACKOFF_DELAYS_MS = [0, 30 * 1000, 5 * 60 * 1000, 15 * 60 * 1000, 30 * 60 * 1000];
 const BACKOFF_RESET_MS = 60 * 60 * 1000;
 
+// containerName -> [monotonicMs, ...]
+const restartHistory = new Map();
+
 const bootQueue = [];
 
 let eventStream = null;
 let stopped = false;
+let lineBuf = '';
 
 function isFluxContainer(name) {
   return name.startsWith('flux') || name.startsWith('zel');
@@ -28,22 +32,22 @@ function nowMs() {
 
 function getBackoffDelay(containerName) {
   const now = nowMs();
-  const history = globalState.containerRestartHistory.get(containerName);
+  const history = restartHistory.get(containerName);
   if (!history) return 0;
   const recent = history.filter((ts) => now - ts < BACKOFF_RESET_MS);
   if (recent.length === 0) {
-    globalState.containerRestartHistory.delete(containerName);
+    restartHistory.delete(containerName);
     return 0;
   }
-  globalState.containerRestartHistory.set(containerName, recent);
+  restartHistory.set(containerName, recent);
   const index = Math.min(recent.length, BACKOFF_DELAYS_MS.length - 1);
   return BACKOFF_DELAYS_MS[index];
 }
 
 function recordRestart(containerName) {
-  const history = globalState.containerRestartHistory.get(containerName) || [];
+  const history = restartHistory.get(containerName) || [];
   history.push(nowMs());
-  globalState.containerRestartHistory.set(containerName, history);
+  restartHistory.set(containerName, history);
 }
 
 async function handleContainerDie(event) {
@@ -53,13 +57,12 @@ async function handleContainerDie(event) {
 
   if (!containerName || !isFluxContainer(containerName)) return;
 
-  if (exitCode === 0) return;
-
   if (globalState.stoppingContainers.has(containerName)) {
     globalState.stoppingContainers.delete(containerName);
-    log.info(`containerCrashRecovery - ${containerName} was intentionally stopped, skipping`);
     return;
   }
+
+  if (exitCode === 0) return;
 
   if (!globalState.bootContainerStateSettled) {
     log.info(`containerCrashRecovery - ${containerName} died (exit ${exitCode}) during boot, queuing`);
@@ -108,9 +111,9 @@ async function waitForBootAndDrain() {
   if (!stopped) await drainBootQueue();
 }
 
-async function start() {
+async function subscribe() {
   if (eventStream) return;
-  stopped = false;
+  lineBuf = '';
 
   try {
     eventStream = await dockerService.dockerGetEvents({
@@ -119,13 +122,19 @@ async function start() {
 
     eventStream.on('data', (buf) => {
       if (stopped) return;
-      try {
-        const event = JSON.parse(buf.toString());
-        handleContainerDie(event).catch((err) => {
-          log.error(`containerCrashRecovery - event handler error: ${err.message}`);
-        });
-      } catch (parseErr) {
-        log.error(`containerCrashRecovery - failed to parse docker event: ${parseErr.message}`);
+      lineBuf += buf.toString();
+      const lines = lineBuf.split('\n');
+      lineBuf = lines.pop();
+      for (const line of lines) {
+        if (!line.trim()) continue;
+        try {
+          const event = JSON.parse(line);
+          handleContainerDie(event).catch((err) => {
+            log.error(`containerCrashRecovery - event handler error: ${err.message}`);
+          });
+        } catch (parseErr) {
+          log.error(`containerCrashRecovery - failed to parse docker event: ${parseErr.message}`);
+        }
       }
     });
 
@@ -133,7 +142,7 @@ async function start() {
       log.error(`containerCrashRecovery - event stream error: ${err.message}`);
       eventStream = null;
       if (!stopped) {
-        setTimeout(() => start(), 10000);
+        setTimeout(() => subscribe(), 10000);
       }
     });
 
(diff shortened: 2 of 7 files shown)


Cabecinha84 (commented):
Deep review of PR #1739 — feat: instant crash recovery via Docker event stream
  
  Scope — 485+/18-, 7 files, 6 commits, base development. Adds containerCrashRecovery.js, instruments dockerService.js with a stop-suppression set, drops the FluxOS-restart short-circuit in appStartupManager.js, wires
  the recovery into serviceManager, and adds two globalState collections.

  Architecture — sound

  The high-level design is correct: subscribe to filtered Docker die events, distinguish intentional stops (via globalState.stoppingContainers, set/cleared inside the appDockerStop/Kill/Restart/Start/Remove/ForceRemove
  helpers) from real crashes, apply exponential backoff per container, queue events that arrive before bootContainerStateSettled, and drain after settle. The suppression set is bounded by the same appDocker* lifecycle
  calls and covers every uninstall path in appUninstaller.js (verified: lines 254/263/290/301/402/411/438/449/608/629/694/715 all funnel through these helpers).

  The removal of the !machineRebooted early-return in appStartupManager.manageAppsOnBoot is justified — with restart: always already removed (#1726), a FluxOS-only restart could leave containers stuck in exited until
  the hourly health monitor ran. Now reconciliation runs every boot. For the healthy FluxOS-restart case it's a no-op because getStoppedFluxContainers() returns empty. Acceptable cost.

  Critical bug — exponential backoff does not work in production

  dockerService.js:appDockerStart (diff lines 218–224) clears both the suppression entry and the restart history unconditionally at the top of the function:

  const dockerName = getDockerName(idOrName);
  globalState.stoppingContainers.delete(dockerName);
  globalState.containerRestartHistory.delete(dockerName);  // <— problem
  await dockerContainer.start();

  containerCrashRecovery.js:handleContainerDie (lines 108–115) restarts via appDockerStart:

  try {
    recordRestart(containerName);            // history := [t1]
    await dockerService.appDockerStart(identifier);  // history := [] (cleared inside)
    ...
  }

  getAppIdentifier() (dockerService.js:42) maps 'www_Osmosis' → 'fluxwww_Osmosis', and getDockerName() strips the leading /. So containerCrashRecovery records under key fluxwww_Osmosis, then calls
  appDockerStart('www_Osmosis'), which computes the same fluxwww_Osmosis key and deletes it.

  Net effect: after every crash recovery, the restart history is empty. The next die event observes history.length === 0, returns delay 0, and restarts immediately. The backoff schedule (30s / 5m / 15m / 30m) is
  unreachable from the crash-recovery path. A genuinely crash-looping container will be restarted every time the kernel can spawn the process — exactly the behavior the backoff was supposed to prevent.

  The commit message for 29203afef explicitly says "clear it when appDockerStart is called … from any other path." The author did not notice that crashRecovery itself is "any other path."

  Why the unit test doesn't catch it: tests/unit/containerCrashRecovery.test.js stubs dockerServiceStub.appDockerStart = sinon.stub().resolves('started') — the stub has no side effects on
  globalStateStub.containerRestartHistory. The "should apply backoff delay on second crash" test passes against the stub but the same scenario fails against real dockerService. Validation on live nodes presumably saw
  "ordine" crashloop slow down only because the health monitor's poll-based restart cleared history at a different cadence, masking the bug.

  Fix sketch (one of several):
  - Move containerRestartHistory.delete(dockerName) out of appDockerStart and into a new explicit "successful run" signal (e.g., clear in the health monitor after it confirms the container has stayed up, or in
  appDockerStart only when called from non-recovery paths via a flag).
  - Or: in containerCrashRecovery, restart via dockerService.getDockerContainerOnly(identifier).start() directly (bypassing appDockerStart), and apply the CFS-burst logic manually — uglier, but keeps the "any-other-path
   clears backoff" semantic intact.
  - Or: move recordRestart to after appDockerStart returns, and have appDockerStart skip the history clear if called from recovery (worst option — ordering still races on parallel events).

  Secondary issues

  1. appDockerRestart leaks stoppingContainers on throw (diff lines 254–258)
  globalState.stoppingContainers.add(dockerName);
  await dockerContainer.restart();      // no try/catch
  globalState.stoppingContainers.delete(dockerName);
  If restart() rejects after Docker has emitted the die event, the die handler will consume the entry — fine. But if restart() rejects before touching the container (auth error, daemon hiccup), the entry is orphaned in
  memory until the next appDocker* call for that container. In-memory leak only; bounded per process lifetime. Compare with appDockerStop/Kill which both have proper try/catch.
  
  2. Exit-code-0 path doesn't consume stoppingContainers (containerCrashRecovery.js:78)
  if (exitCode === 0) return;                                 // returns before...
  if (globalState.stoppingContainers.has(containerName)) {    // ...this check
  If appDockerStop is called on a container that exits cleanly (code 0), the stoppingContainers entry is set but never consumed. Same in-memory leak. Not a correctness bug (clean exits are ignored anyway), but
  inconsistent with the "no TTL, no leaks" claim in the PR description.
  
  3. NDJSON parsing is fragile (containerCrashRecovery.js:142–151)
  eventStream.on('data', (buf) => {
    const event = JSON.parse(buf.toString());
    ...
  });
  Confirmed from node_modules/dockerode/lib/docker.js:1464–1494: getEvents returns a raw HTTP stream with isStream: true and no line buffering. Docker's events API is newline-delimited JSON. With the die-only filter,
  single-event chunks are the common case, but the moment two die events arrive in the same TCP segment (likely during mass shutdown — e.g., removeAllApps), or one event is split across two chunks, JSON.parse throws and
   the events are dropped. Health monitor catches up eventually, so the failure mode is "delayed restart" rather than "lost forever" — but it undermines the "instant" claim under load. A StringDecoder + line splitter
  (the pattern used in 5b4e1a94d for getEvents in another context, per recent commits) would harden this.
  
  4. Stale comment in appStartupManager.js:295
  // Machine rebooted, locations still valid — wait for daemon + sync then reconcile. — this branch now also handles FluxOS-only restarts. The serviceManager.js comment was updated, this one wasn't. Trivial.

  5. bootQueue is module-scoped and not reset on stop() (line 32). If start() is ever invoked, stop()ed, and re-start()ed within the same process (the error/end handlers do this on disconnect), queued events from the
  prior session survive. Not exploitable in current wiring (only serviceManager.startFluxFunctions calls start()), but a latent footgun.

  6. Double-drain on reconnect. If eventStream.on('error') fires and setTimeout(start, 10000) succeeds, waitForBootAndDrain() is invoked again. With bootContainerStateSettledGate already open, both promises resolve and
  call drainBootQueue() concurrently. The while(length>0) shift() interleaves safely, but it's noisier than necessary.

  Verdict — do not ack as-is

MorningLightMountain713:
Thanks for the thorough review @Cabecinha84. All issues addressed in `e98a715ef`, `a4ca17b52`, and `bc64c97e7`. Responses below:

---

> **Critical bug — exponential backoff does not work in production**
>
> `appDockerStart` clears `containerRestartHistory` unconditionally, so after every crash recovery restart the history is empty and the backoff schedule is unreachable.

Confirmed. Root cause: commit `29203afef` moved `restartHistory` to `globalState` and added `containerRestartHistory.delete(dockerName)` inside `appDockerStart` — which crash recovery itself calls. Fix: removed the `delete` from `appDockerStart` entirely. The restart history is now module-local in `containerCrashRecovery.js` and resets after 10 minutes of stable running (if the container ran 10+ minutes since the last restart, the next crash starts fresh at immediate). Verified backoff works on live node (30s delay on second crash).

---

> **1. appDockerRestart leaks stoppingContainers on throw**
>
> No try/catch around `dockerContainer.restart()`.

Fixed. Wrapped in `try/finally` so `stoppingContainers.delete` runs on both success and failure.

---

> **2. Exit-code-0 path doesn't consume stoppingContainers**
>
> `if (exitCode === 0) return` exits before the suppression check.

Fixed. Moved the `stoppingContainers` check above the exit code filter so it always consumes the entry regardless of exit code.

---

> **3. NDJSON parsing is fragile**
>
> Raw `JSON.parse(buf.toString())` on data events — breaks if two events arrive in one TCP segment or one event is split across chunks.

Fixed. Added a `lineBuf` that accumulates data and splits on `\n`. Incomplete lines are held until the next chunk. Buffer is reset on reconnect so partial data from a dead stream doesn't contaminate the new one. Added two unit tests: "two events in one chunk" and "event split across two chunks".

---

> **4. Stale comment in appStartupManager.js:295**
>
> "Machine rebooted, locations still valid" — this branch now handles all restarts.

Fixed. Updated to "Locations still valid".

---

> **5. bootQueue is module-scoped and not reset on stop()**

Fixed. `stop()` now calls `bootQueue.splice(0)`. Added unit test "should clear boot queue on stop()".

---

> **6. Double-drain on reconnect**
>
> `waitForBootAndDrain()` is invoked again on every reconnect via `setTimeout(start, 10000)`.

Fixed. Split `start()` into `start()` (called once — subscribes + kicks off drain) and `subscribe()` (called on reconnect — only re-establishes the event stream). Reconnects no longer trigger `waitForBootAndDrain()`.

---

### Additional improvement (bc64c97e7)

Backoff now resets after 10 minutes of stable running instead of a 1-hour passive age-out. If a container ran for 10+ minutes since the last restart, the next crash starts fresh at immediate — the container was stable enough that the new crash is a new problem, not a continuation of a loop.

---

### Live node verification (46.225.195.244)

All features verified on commit `bc64c97e7`:

| Test | Verified | Evidence |
|------|----------|----------|
| External `docker kill` → immediate restart | Yes | Log: `crashed (exit 137), restarting` → `restarted successfully` |
| Second crash → 30s backoff | Yes | Log: `waiting 30s before restarting` → restarted 30s later |
| FluxOS API stop, clean exit (0) → not restarted | Yes | Container stayed exited, no crash recovery log |
| FluxOS API stop, dirty exit (137) → suppressed | Yes | Log: `was intentionally stopped (exit 137), skipping` |
| Boot reconciliation, expired location → removed | Yes | Log: `no longer has a valid location record, removing locally` |
| Boot reconciliation, no stopped containers → no-op | Yes | Log: `No stopped containers found` |
| Boot queue drain | Yes | Log: `draining 1 queued event(s)` |
| 20 unit tests | Yes | All passing |
| 65 dockerService tests | Yes | All passing |
| 33 appStartupManager tests | Yes | All passing |

Cabecinha84 (approved):
ack

PR #1740, opened 2026-05-27
v8.13.1

Cabecinha84:
## Summary

  Release **v8.13.1** — bundles PR #1739 (`feat/docker-event-crash-recovery`) into master.

  Replaces the hourly health-monitor polling cycle with a Docker event-stream subscription so crashed app containers are restarted within seconds instead of up to an hour later.

  ## What's included

  ### Instant crash recovery via Docker events
  - New `containerCrashRecovery` service subscribes to Docker `die` events and restarts crashed containers immediately.
  - Only non-zero exits are treated as crashes; clean exits (code 0) from FluxOS-initiated stops are ignored.
  - Events received during boot are queued and drained after startup settles (not discarded).
  - NDJSON line buffer handles split/merged chunks from the Docker event stream.
  - `subscribe()` is separated from `start()` to avoid double-drain on reconnect.

  ### Crash-loop protection
  - Exponential backoff schedule: immediate → 30s → 5m → 15m → 30m (capped). Never gives up, just slows down.
  - Backoff resets to immediate after **10 minutes of stable running** — a fresh crash after a stable period is treated as a new problem, not a continuation of a loop.
  - Uses monotonic time (`process.hrtime.bigint`) and prunes stale history entries.
  - Re-checks container state after the backoff delay in case the health monitor or uninstaller already handled it.

  ### Intentional-stop suppression
  - New `stoppingContainers` set in `globalState`. `dockerService` adds entries on `appDockerStop`/`appDockerKill` and removes on `appDockerStart`/`Remove`/`ForceRemove`.
  - Crash recovery checks the set before restarting and consumes the entry, so user/system-initiated stops never trigger a restart.
  - `appDockerRestart` is wrapped in `try/finally` to prevent leaks into the set.
  - Suppressed die events are logged for observability.

  ### Reconcile on FluxOS restart
  - Removed the `machineRebooted` gate in `appStartupManager` so `reconcileAppsOnBoot` runs on every FluxOS startup.
  - Closes the gap where a container that crashed before a FluxOS-only restart was never recovered.

  ## Files changed
  - `ZelBack/src/services/appMonitoring/containerCrashRecovery.js` (new, +187)
  - `ZelBack/src/services/dockerService.js` (+42 / −8)
  - `ZelBack/src/services/appLifecycle/appStartupManager.js` (+1 / −6)
  - `ZelBack/src/services/serviceManager.js` (+6 / −2)
  - `ZelBack/src/services/utils/globalState.js` (+5)
  - `tests/unit/containerCrashRecovery.test.js` (new, +357 — 19 tests)
  - `tests/unit/appStartupManager.test.js` (+8 / −4)
  - `package.json` — version bump to 8.13.1

  ## Test plan
  - [ ] Unit tests pass (`containerCrashRecovery.test.js`, `appStartupManager.test.js`)
  - [ ] Crash a running app container (`docker kill --signal=SIGSEGV <id>`) and confirm it restarts within seconds without waiting for the hourly cycle
  - [ ] Stop a container via FluxOS (`appDockerStop`) and confirm crash recovery does NOT restart it
  - [ ] Trigger repeated crashes and confirm the backoff schedule (immediate → 30s → 5m → 15m → 30m)
  - [ ] Let a container run for 10+ minutes after a crash, crash it again, and confirm backoff resets to immediate
  - [ ] Restart FluxOS (without rebooting host) while an app container is exited and confirm it's reconciled on boot
  - [ ] Verify Docker event stream reconnects cleanly (kill `dockerd`, restart it) with no double-processing


MorningLightMountain713 (approved):
Ack

Issue #4, opened 2019-10-06
Testsuite


TheTrunk:
Tests are needed especially for the Backend part

PR #1741, opened 2026-05-28
Enterprise: GitHub-synced per-node app-owner allowlists

Cabecinha84:
## Summary

Moves the enterprise allowlists out of static `config/default.js` into a single GitHub-synced JSON file, and turns them into a **per-node mapping** so each enterprise node declares exactly which app owners may run on it. Because the data is fetched from GitHub on an interval, the owner↔node relationships can be changed live **without a FluxOS release**.

### 1. GitHub-synced enterprise config — `enterpriseConfig.js`
- New `helpers/enterprisenodes.json`: a map of enterprise **node pubkey → array of app-owner addresses** allowed to install on that node (many-to-many — an owner may appear under several nodes).
- Replaces the former static `enterpriseAppOwners` / `enterpriseNodesPublicKeys` arrays in `default.js`.
- Seeded from the on-disk file synchronously at load (consumers always have data), then re-synced from `…/helpers/enterprisenodes.json` every 6h. A failed/invalid fetch keeps the current value (disk on the first run, last good thereafter).
- Derives the global owner set (deduped union of all values), the node pubkey set (keys), and per-node allowed owners from the one map.

### 2. Per-node ownership enforcement — `enterpriseNetwork.js`
- Enterprise nodes install **only** apps whose owner is mapped to that specific node; a node mapped to an empty list hosts nothing.
- Non-enterprise nodes still never host apps owned by **any** enterprise owner (the union).
- Node-agnostic checks are unchanged (datacenter validation, CPU-burst eligibility, excluding enterprise apps from public nodes) — they use the union.
- Only the node pubkey + enterprise identity are cached; the allowed-owner list is read **live** from the current map, so owner changes take effect within the sync interval without a node restart.

### 3. Strict node-IP targeting for enterprise apps — `appSpawner.js`
- In `trySpawningGlobalApplication`, an enterprise-owned app that lists node IPs in its spec may install **only** on a node whose IP is in that list, regardless of app version. Previously the `version >= 8` path let any v8+ node install such an app even when the spec targeted specific IPs.

## Deployment note
`helpers/enterprisenodes.json` must be published to `RunOnFlux/flux` under `master/helpers/` for the live sync to take effect (the sync URL points at `master`). Until then, nodes use the on-disk copy shipped in this PR — identical to the previous `default.js` values, so behavior is unchanged at launch.

## Test plan
- [ ] `npm run test:zelback:unit` — new/updated suites (`enterpriseConfig`, `enterpriseNetwork`, `cpuBurstHelper`, `appSpawner`) pass locally
- [ ] A node whose pubkey is absent from the map behaves as a normal public node
- [ ] Editing `enterprisenodes.json` re-targets owners within the sync interval (no restart)
- [ ] An enterprise app that targets specific node IPs only spawns on a matching-IP node


MorningLightMountain713:
## Review

### Bugs

1. **Stale enterprise identity cache** — `cachedIsEnterpriseNode` in `enterpriseNetwork.js` is set once at boot and never re-evaluated when the GitHub map re-syncs. A node removed from the map becomes a zombie (hosts nothing, uninstalls everything). A node added after boot stays non-enterprise until restart. Nothing calls `resetEnterpriseNodeCache()` after sync.

2. **No value-level validation on synced JSON** — `enterpriseConfig.js:47` only checks `isPlainObject`. A key mapped to a non-array value (e.g. `null`) silently causes `getAllowedOwnersForNode()` to return `[]`, triggering mass app uninstall on that node with no warning logged.

3. **Three stale comments reference nonexistent files** — `default.js:384` says `enterpriseappowners.json`, `default.js:406` says `enterprisenodespublickeys.json`, `enterpriseNetwork.js:30` says `enterprisenodespublickeys.json`. All should say `enterprisenodes.json`.

### Architecture

4. **Synchronous disk I/O** — `enterpriseConfig.js:30` uses `fs.readFileSync`. Synchronous I/O should never be used in a Node.js server. The disk read should be async (`fs.promises.readFile`).

5. **Side effect on module load** — `enterpriseConfig.js:41-42` reads from disk and mutates module state as a side effect of `require()`. Module loading should be pure; initialization belongs in an explicit `startSync()` or `init()` call. The async disk read (from fix #4) should move into `startSync()`, which should then be `await`ed in `startFluxFunctions()`. The GitHub fetch should use an explicit 10s timeout so boot is bounded.

### Performance

6. **`getEnterpriseAppOwners()` rebuilds Set on every call** — `enterpriseConfig.js:99-101` allocates `Object.values().filter().flat()` + `new Set()` + spread per invocation. Called per-app via `isEnterpriseAppOwner()` in two filter passes during the spawn loop (~2N reconstructions per iteration). Should memoize, invalidated when `nodeOwnerMap` is replaced.

### Operational

7. **Version-enforce after rollout** — During rolling deploy, old-code nodes use the flat owner list while new-code nodes use per-node mapping. Once fully rolled out, the old code path should be version-gated so the fleet can't diverge if the JSON is updated mid-rollout.

8. **No integrity verification on synced JSON** — Plain HTTPS to `raw.githubusercontent.com`, no signature. A GitHub account compromise could reassign enterprise ownership within 6h. Longer-term concern — consider signing the JSON with a dedicated key and verifying before accepting the payload.

### Test Gaps

9. **`cleanupOwnershipViolations` per-node scoping** — No test verifies that an app owned by a valid enterprise owner but not mapped to *this* node gets removed. This is the PR's core new behavior.

10. **Malformed map values** — No test for non-array values in the map (core of bug #2).

11. **`filterAppsByOwnership` when identity is null** — No test for the silent drop-all-apps edge case.

12. **Enterprise-owned app + no nodes** — Not explicitly tested in the appSpawner suite.

MorningLightMountain713 (changes requested):
See comment

Cabecinha84:
> ## Review
> ### Bugs
> 1. **Stale enterprise identity cache** — `cachedIsEnterpriseNode` in `enterpriseNetwork.js` is set once at boot and never re-evaluated when the GitHub map re-syncs. A node removed from the map becomes a zombie (hosts nothing, uninstalls everything). A node added after boot stays non-enterprise until restart. Nothing calls `resetEnterpriseNodeCache()` after sync.
> 2. **No value-level validation on synced JSON** — `enterpriseConfig.js:47` only checks `isPlainObject`. A key mapped to a non-array value (e.g. `null`) silently causes `getAllowedOwnersForNode()` to return `[]`, triggering mass app uninstall on that node with no warning logged.
> 3. **Three stale comments reference nonexistent files** — `default.js:384` says `enterpriseappowners.json`, `default.js:406` says `enterprisenodespublickeys.json`, `enterpriseNetwork.js:30` says `enterprisenodespublickeys.json`. All should say `enterprisenodes.json`.
> 
> ### Architecture
> 4. **Synchronous disk I/O** — `enterpriseConfig.js:30` uses `fs.readFileSync`. Synchronous I/O should never be used in a Node.js server. The disk read should be async (`fs.promises.readFile`).
> 5. **Side effect on module load** — `enterpriseConfig.js:41-42` reads from disk and mutates module state as a side effect of `require()`. Module loading should be pure; initialization belongs in an explicit `startSync()` or `init()` call. The async disk read (from fix [Testsuite #4](https://github.com/RunOnFlux/flux/issues/4)) should move into `startSync()`, which should then be `await`ed in `startFluxFunctions()`. The GitHub fetch should use an explicit 10s timeout so boot is bounded.
> 
> ### Performance
> 6. **`getEnterpriseAppOwners()` rebuilds Set on every call** — `enterpriseConfig.js:99-101` allocates `Object.values().filter().flat()` + `new Set()` + spread per invocation. Called per-app via `isEnterpriseAppOwner()` in two filter passes during the spawn loop (~2N reconstructions per iteration). Should memoize, invalidated when `nodeOwnerMap` is replaced.
> 
> ### Operational
> 7. **Version-enforce after rollout** — During rolling deploy, old-code nodes use the flat owner list while new-code nodes use per-node mapping. Once fully rolled out, the old code path should be version-gated so the fleet can't diverge if the JSON is updated mid-rollout.
> 8. **No integrity verification on synced JSON** — Plain HTTPS to `raw.githubusercontent.com`, no signature. A GitHub account compromise could reassign enterprise ownership within 6h. Longer-term concern — consider signing the JSON with a dedicated key and verifying before accepting the payload.
> 
> ### Test Gaps
> 9. **`cleanupOwnershipViolations` per-node scoping** — No test verifies that an app owned by a valid enterprise owner but not mapped to _this_ node gets removed. This is the PR's core new behavior.
> 10. **Malformed map values** — No test for non-array values in the map (core of bug [fix zelnoeservice #2](https://github.com/RunOnFlux/flux/pull/2)).
> 11. **`filterAppsByOwnership` when identity is null** — No test for the silent drop-all-apps edge case.
> 12. **Enterprise-owned app + no nodes** — Not explicitly tested in the appSpawner suite.

Should be resolved. Can you re-review?

MorningLightMountain713 (approved):
Ack

PR #1745, opened 2026-06-03
Handle g: master/slave per component in masterSlaveApps + restore #1728 recovery behavior

Cabecinha84:
## Summary

Completes the per-component `g:` syncthing work started in #1728, in two parts:

### 1. Regression fix: restore #1728 behavior in container recovery (`containerHealthMonitor.js`)

#1728 (merged May 7) made runtime auto-restart decisions per component in `peerNotification.js`. The May 14 refactor (0636fc23c) moved that logic into `containerHealthMonitor.js` but started from the pre-#1728 version of the code, silently reverting it to app-level checks. (The boot-recovery half of #1728 survived the refactor — it lives in `appStartupManager.js` / `appUtilities.getNonGComponentIdentifiers`.)

Restored at the new location:
- A stopped component is only treated as master/slave-managed if **that component** uses `g:` (`stoppedCompIsG`), not if any sibling does
- The 30-minute install grace applies only to `r:`/`g:` components, not to non-syncthing siblings
- v≤3 single-component apps handled via the `containerData` fallback

### 2. `masterSlaveApps` stops/starts only the g: component, not the whole app

`masterSlaveApps` already tracks everything per component (`identifier`, `appId`, `mastersRunningGSyncthingApps`), but the docker stop/start calls passed `installedApp.name`, which expands to **every** component of a compose app. On standby nodes this stopped all components — including non-g siblings that must run on every instance.

- 6 stop/start call sites now pass the component `identifier`
- The two remote "is it running there" checks (lower-index node, previous master) now match on the component identifier, so non-g siblings running on every node are not mistaken for the master/slave component being active remotely

### Why this matters

A mixed compose app like **n8n (`g:`) + flux-pg-cluster (HA postgres)** is impossible with whole-app behavior: the pg cluster components get stopped on standby nodes, so etcd never has quorum, replication never happens, and a failover promotes a node with an empty data directory. With both fixes, the `g:` component fails over master/slave style while its siblings keep running on all instances.

### Backward compatibility

- v≤3 apps: `identifier === app name` — unchanged behavior
- Single-component compose apps: stopping/starting the only component ≡ stopping/starting the app
- Multi-component `g:` apps: only the `g:` component fails over (intended post-#1728 behavior)

## Tests

- Ported the #1728 per-component test cases to `containerHealthMonitor.test.js` (7 tests) — the two regression cases **fail against current development** and pass with this PR
- Added standby-stop and steady-state `masterSlaveApps` cases with a mixed n8n+pgcluster spec (2 tests)
- All existing `masterSlaveApps`, `peerNotification`, `appStartupManager`, `containerCrashRecovery` tests still pass

### DIFF
--- a/ZelBack/src/services/appLifecycle/advancedWorkflows.js
+++ b/ZelBack/src/services/appLifecycle/advancedWorkflows.js
@@ -3892,8 +3892,11 @@ async function masterSlaveApps(globalStateParam, installedApps, listRunningApps,
                       const response = await axios.get(`http://${ipToCheck}:${portToCheck}/apps/listrunningapps`, { timeout, cancelToken: source.token });
                       isResolved = true;
                       const appsRunning = response.data.data;
-                      if (appsRunning.find((app) => app.Names[0].includes(installedApp.name))) {
-                        log.info(`masterSlaveApps: app:${installedApp.name} is running on lower-index node (index ${i}) at ${ipToCheck}, will not start`);
+                      // Match on the g: component identifier, not the app name: non-g siblings
+                      // (e.g. a DB cluster component) run on every node and must not be mistaken
+                      // for the master/slave component being active there.
+                      if (appsRunning.find((app) => app.Names[0].includes(identifier))) {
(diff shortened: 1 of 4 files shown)


MorningLightMountain713 (approved):
Ack

PR #1746, opened 2026-06-03
v8.14.0

Cabecinha84:
## Summary

Release **v8.14.0** — bundles PR #1737 (`fix/socket-address-default-port-compat`), PR #1741 (`enterprise-changes`), and PR #1745 (`fixgsyncthingpercomponent`) into master, plus minimum-version bumps for FluxBench/FluxOS.

### Socket address standardization for the ip:port transition (#1737)
Prepares FluxOS to handle both the old fluxbench format (bare IP for default-port nodes) and the new format (always `ip:port`). Must be rolled out network-wide before the fluxbench always-attach-port change.

- New `socketAddressUtils.js` module: `normalizeSocketAddress`, `extractIp`, `extractPort`, `parseSocketAddress`, `socketAddressesMatch`.
- All `===` comparisons on socket addresses replaced with `socketAddressesMatch()` so `ip` vs `ip:16127` entries match regardless of which fluxbench version produced them.
- All `.split(':')[0]` / `.split(':')[1] || '16127'` patterns replaced with `extractIp()` / `extractPort()`; hardcoded `16127` replaced with `DEFAULT_API_PORT`.
- Consistent naming throughout: `myFluxIP` → `localSocketAddress`, `getMyFluxIPandPort` → `getLocalSocketAddress`, socket addresses named `socketAddr`, bare IPs named `ip`.
- `networkStateService` lookups check both `ip` and `ip:16127` formats (compat layer, marked TODO for removal once all nodes run updated fluxbench).
- Hardened parsing: strict port regex, malformed input falls back to `DEFAULT_API_PORT`, deterministic sort order with mixed address formats.
- 39 new tests in `socketAddressUtils.test.js` plus format-transition tests.

### Enterprise: GitHub-synced per-node app-owner allowlists (#1741)
Enterprise node and app-owner configuration moves out of the static config into a GitHub-synced file, so changes apply network-wide without a release.

- New `helpers/enterprisenodes.json`: maps each enterprise node pubkey to the app-owner addresses allowed to install on it (many-to-many), re-synced from GitHub every 6h (mirroring the blockedrepositories pattern).
- New `enterpriseConfig` service: seeds from the on-disk file at boot (bounded 10s GitHub fetch timeout), refreshes on a 6h interval; failed fetches keep the last good value.
- Enterprise nodes now install only apps owned by the owners mapped to that specific node; node-agnostic checks (datacenter validation, CPU burst, excluding enterprise apps from public nodes) use the deduped union of all owners.
- Enterprise membership and allowed owners are evaluated live against the current map — node/owner changes take effect within the sync interval without a restart.
- Enterprise-owned apps that list node IPs in their spec may only install on those nodes, regardless of app version (closes the v8+ bypass).
- Validation rejects malformed payloads wholesale (keeps last good map); 350+ lines of new tests.

### Per-component g: master/slave handling (#1745)
Fixes mixed compose apps (e.g. n8n with `g:` + HA postgres cluster) so non-syncthing siblings keep running on standby nodes.

- `masterSlaveApps` docker stop/start now targets the `g:` component identifier instead of the whole app name — on standby nodes only the `g:` component is stopped, siblings (e.g. an HA postgres cluster needing etcd quorum) keep running.
- Remote "is it running there" checks match on the component identifier, so siblings running everywhere are not mistaken for the master being active on a remote node.
- Restores per-component recovery logic in `containerHealthMonitor` (regression from the May refactor that reverted #1728): a stopped component is only treated as master/slave-managed if *that* component uses `g:`, and the 30-minute install grace applies only to `r:`/`g:` components.
- Backward compatible with v≤3 apps and single-component compose apps.
- Ported #1728 per-component test cases plus new standby-stop/steady-state cases.

### Version bumps
- Minimum allowed FluxBench: 5.0.0 → **6.3.0**
- Minimum allowed FluxOS: 8.0.0 → **8.13.1**
- benchmarkinfo helper: 5.4.0 → 6.3.1; daemoninfo helper: 9000350 → 9010050 (protocol 170020 → 170021)


MorningLightMountain713 (approved):
Ack

alihm (approved):
Ack

XK4MiLX (approved):

PR #1747, opened 2026-06-03
fix: broadcast running apps after releasing the install lock

MorningLightMountain713:
## Problem
A freshly-installed multi-component **syncthing** app (e.g. shared-db WordPress) is excluded from its own post-install "app running" broadcast, so it doesn't propagate to the network until the next hourly periodic broadcast — and if it churns/redeploys before then, it never broadcasts at all.

## Root cause
In `registerAppLocally` (`appInstaller.js`), the post-install broadcast — `onInstallComplete()` → `peerNotification.checkAndNotifyPeersOfRunningApps()` — was invoked **while the install lock was still held** (`globalState.installationInProgress === true`), before the lock was cleared.

With the lock held, `containerHealthMonitor.monitorAndRecoverApps()` bails immediately (`if (globalState.isOperationInProgress()) return …`, logging *"Stopped application checks not running…"*) and returns empty `masterSlaveAppsInstalled` / `startedApps`. Those arrays are the **only** path that force-includes a syncthing app whose components aren't all simultaneously in `docker ps` at the broadcast instant (e.g. `wp` mid receive-only resync). So the strict gate in `peerNotification.js` drops it and the broadcast goes out without the just-installed app. Observed in production: the cycle that should have announced `wordpress…` carried only `globalping-probe`.

## Fix
Move the `onInstallComplete()` call to **after** `globalState.installationInProgress = false`, so the recovery/force-include path is active when the broadcast runs. `checkAndNotifyPeersOfRunningApps` catches internally and never throws, so running it after `res.end()` is safe.

## Test
Adds a regression test in `appInstaller.test.js` that drives a successful install and asserts the broadcast fires only after the lock is released. Confirmed it **fails on the previous ordering** and passes with the fix; full file passes 17/17.

## Why existing tests missed it
Every app in the integration suite (`test-infra/runner`) and unit tests is single-component, non-syncthing (`containerData: '/tmp'`); such apps pass the broadcast's strict gate directly and never exercise the force-include path that the held lock disabled. The unit tests also stubbed the seam (`containerHealthMonitor` always stubbed `isOperationInProgress → false`).

### DIFF
--- a/ZelBack/src/services/appLifecycle/appInstaller.js
+++ b/ZelBack/src/services/appLifecycle/appInstaller.js
@@ -607,11 +607,6 @@ async function registerAppLocally(appSpecs, componentSpecs, res, test = false, s
 
     log.info(`Flux App: ${appName} is test install: ${test}`);
 
-    if (!test && onInstallComplete) {
-      await onInstallComplete();
-      fluxEventBus.publish('app:installed', { name: appSpecifications.name, hash: appSpecifications.hash });
-    }
-
     // Reconnect any locally installed apps that are networked with this app —
     // its private network was (re)created during this install. Guarded on
     // appComponent (the unmutated entry value) since isComponent is flipped to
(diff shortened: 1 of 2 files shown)


Cabecinha84 (approved):
ack

PR #1748, opened 2026-06-04
fix: default-port nodes never become message-capable → boot wipes all apps

MorningLightMountain713:
## Problem

On default-port (16127) FluxNodes, `nodeConfirmationService.canSendMessages()` is permanently `false` — the self-lookup `getFluxnodeBySocketAddress()` never resolves the node against its own pubkey. That stalls `AppSyncOrchestrator`: `dbReady` only opens via the ~2h block-timer fallback instead of the fast hash-sync path, so `manageAppsOnBoot()` hits `removeAllApps('Sync timeout')` after 300s on every restart, **wiping all installed apps**.

## Root cause

Since #1737, `getLocalSocketAddress()` returns the normalized `ip:16127` form. But the `networkStateManager` socketAddress index is keyed by `node.ip` exactly as the daemon lists it — which is **bare** for default-port nodes. So `getFluxnodeBySocketAddress("ip:16127")` does an exact `Map.get()` against the bare key, misses, and returns `null`.

\#1737 added a bare↔`:16127` fallback to `socketAddressInNetworkState` but not to `getFluxnodeBySocketAddress`, so the message-capability path stayed broken. Non-default (UPnP) ports are unaffected — they're always explicit.

## Fix

Canonicalize at the `networkStateManager` boundary: normalize the socketAddress index key on build, and normalize the lookup filter in `search()` / `includes()`. `normalizeSocketAddress` only appends the default port to a bare IP, so explicit (UPnP) ports pass through unchanged. Removes the now-redundant per-consumer fallback in `socketAddressInNetworkState`.

## Impact

Sweep of 677 reachable default-port nodes: **0 of 667** on the current release hold any apps, vs 33–50% on pre-#1737 versions. Public/default-port nodes are ~27% of the network, so this silently removed a large share of durable app-hosting capacity (and feeds directly into apps being unable to reach their target instance counts).

## Testing

- Added regression tests covering bare↔`:16127` resolution and that a bare query does not match a non-default-port node (networkStateManager suite, all green).
- **Validated end-to-end on a live node.** Before the fix the node held **0 apps** with a permanently gated spawner. After deploying the fix and restarting:
  - `Message capability gained` now fires (it never did before) and stays stable.
  - `DB ready` opens in **~53s** (previously ~2h via the block-timer fallback) — well inside the 300s boot timeout, so `removeAllApps('Sync timeout')` cannot trigger.
  - The spawner becomes active and goes all the way through: **selects → installs → and is now running a real app container** (a healthy `yurinnick/folding-at-home` instance, `Up` and serving).
  - **0** `Sync timeout` / `removeAllApps` / removals since boot.

### DIFF
--- a/ZelBack/src/services/networkStateService.js
+++ b/ZelBack/src/services/networkStateService.js
@@ -1,6 +1,5 @@
 const daemonServiceFluxnodeRpcs = require('./daemonService/daemonServiceFluxnodeRpcs');
 const networkStateManager = require('./utils/networkStateManager');
-const { normalizeSocketAddress, extractIp, extractPort, DEFAULT_API_PORT } = require('./utils/socketAddressUtils');
 
 /**
  * @typedef {import('./utils/networkStateManager').Fluxnode} Fluxnode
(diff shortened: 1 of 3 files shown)


Cabecinha84 (approved):
ack

PR #1750, opened 2026-06-05
fix: IP-granular matching for FDM master selection (UPnP master flap)

MorningLightMountain713:
> **Note:** this branch has two commits. The first (`keep FDM master socket port`) was the original socket-based approach; the second (`use IP-granular matching / ipsMatch`) **supersedes** it. The description below reflects the final state. The net diff is the `ipsMatch` approach.

## Problem

A master/slave app whose primary lands on a **UPnP node** never stabilizes. The primary container starts, loads, then is stopped ~30s later — repeatedly. On a live Valheim master (`valheim1777035136949` on `90.228.196.203:16157`) we observed **268 starts / 268 stops in one day**; the container's own logs showed it stopping itself:

```
masterSlaveApps: stopping docker component:...valheim1777035136949
  it's running on ip:90.228.196.203 and localSocketAddr is: 90.228.196.203:16157
```

The FDM domain points at that master, so users can't connect while it's down.

## Root cause

`masterSlaveApps` asks FDM `GET /appips/<app>` for the current primary and compares it against the node's own socket (`localSocketAddr`). FDM returns a **bare IP** (it tracks an app's master per-IP).

PR #1737's `ip:port` standardization sweep mechanically converted the deliberately **IP-only** comparisons here (`x.ip.split(':')[0] === myIP.split(':')[0]`) into `socketAddressesMatch`. That helper normalizes a bare IP to the **default API port `16127`**:

```
socketAddressesMatch("90.228.196.203:16157", "90.228.196.203" -> "90.228.196.203:16127")
  → 16157 ≠ 16127 → false
```

So a UPnP master fails to recognise **itself** as primary, stops its own container, and flaps forever. Default-port masters accidentally matched (their real port *is* 16127), so only UPnP-hosted masters were affected — a large share of the fleet.

## Fix

Restore IP-granular comparison. Add `ipsMatch(a, b)` to `socketAddressUtils` (compares on IP only, ignoring the port) and use it at the six master/slave comparison sites. `getMasterIpFromFdm` is left untouched.

**Why IP-granular is correct here:** FDM master identity is **per-IP** — only one instance of an app can run on a given IP (enforced by host port mapping). So the IP uniquely identifies the instance; the port is irrelevant for "is this node the master?". (The unrelated `checkAndRemoveApplicationInstance` site keeps `socketAddressesMatch` — it compares two real node sockets, not an FDM value.)

## No FDM change / no rollout ordering

This works against the **current, unmodified FDM** (which returns a bare IP) — `ipsMatch` strips the port from `localSocketAddr` and compares IPs. So:

- **No `flux-domain-manager` change is required** (the companion FDM PR was closed).
- **No deployment ordering** — FluxOS can be shipped on its own.
- It can be **validated on the affected node immediately** by deploying only patched FluxOS.
- It's immune to a node's UPnP port changing, and can't be re-broken by FDM behavior.

(An earlier approach had FDM return `ip:port` and FluxOS keep it; this was dropped in favour of the self-contained IP-only fix, since the per-IP invariant makes socket precision unnecessary.)

## Testing

- `tests/unit/socketAddressUtils.test.js` — `ipsMatch` unit tests (bare↔ported same-IP match, same-IP different-port match, cross-IP non-match, null handling).
- `tests/unit/advancedWorkflows.test.js` — full `masterSlaveApps` suite, **9 passing** (7 existing + 2 new), with FDM mocked as a **bare IP** (production format):
  - *does NOT stop its own container when it is the primary on a UPnP (non-default) port* — pins the flap bug.
  - *stops the g: component on a UPnP standby when FDM names a different primary IP* — guards `ipsMatch` doesn't over-match across IPs.
- Both regression tests were verified to **fail under `socketAddressesMatch`** and **pass under `ipsMatch`** with bare-IP FDM input.

## Live validation on the affected node

Deployed this branch to the real Valheim master (`90.228.196.203`, UPnP API port `:16157`) against the **current production FDM** (no FDM change) — switched the node's FluxOS checkout to `fix/masterslave-fdm-port`, truncated logs, and restarted `fluxos.service`.

The node's own debug.log shows it sailing through the exact condition that used to kill it:

```
08:36:02  no primary set -> starting docker component (index 0)     # started once
08:37:03  getMasterIpFromFdm: Got IP 90.228.196.203 from EU FDM     # FDM now names THIS node (bare IP)
08:37:33  Got IP 90.228.196.203
08:38:03  Got IP 90.228.196.203
08:38:33  Got IP 90.228.196.203
08:39:03  Got IP 90.228.196.203                                     # still master, NO stop
```

`Got IP 90.228.196.203` (bare) against a node on `:16157` is precisely what old `socketAddressesMatch` normalized to `:16127` and mismatched, triggering the self-stop. With `ipsMatch` the node recognises itself and stays up.

| metric | before (old code) | after (`ipsMatch`) |
|---|---|---|
| starts / stops | 268 / 268 per day | **1 / 0** |
| container | `exited` (flapping) | **`Up`, RestartCount=0** |
| FDM-master cycles survived | killed ~30s after each | **4+ with zero stops** |

It held running across ~2 minutes of FDM continuously reporting it as master; the old code would have stopped it within 30s of the first such report.

### DIFF
--- a/ZelBack/src/services/appLifecycle/advancedWorkflows.js
+++ b/ZelBack/src/services/appLifecycle/advancedWorkflows.js
@@ -13,7 +13,9 @@ const dockerService = require('../dockerService');
 const verificationHelper = require('../verificationHelper');
 const daemonServiceMiscRpcs = require('../daemonService/daemonServiceMiscRpcs');
 const fluxNetworkHelper = require('../fluxNetworkHelper');
-const { DEFAULT_API_PORT, extractIp, extractPort, socketAddressesMatch } = require('../utils/socketAddressUtils');
+const {
+  DEFAULT_API_PORT, extractIp, extractPort, socketAddressesMatch, ipsMatch,
+} = require('../utils/socketAddressUtils');
 const generalService = require('../generalService');
 // eslint-disable-next-line no-unused-vars
 const upnpService = require('../upnpService');
@@ -180,8 +182,8 @@ function getFdmIndex(appName) {
  * Tries EU, USA, and ASIA FDM servers in order until one succeeds.
  * @param {string} appName - Application name
  * @param {Object} axiosOptions - Axios request options
- * @returns {Promise<{ip: string|null, fdmOk: boolean}>} The master socket address (ip:port,
- *   or bare ip for default-port masters) and success status
+ * @returns {Promise<{ip: string|null, fdmOk: boolean}>} The master IP (FDM returns a bare IP;
+ *   compare it with ipsMatch, which ignores the port) and success status
  */
 async function getMasterIpFromFdm(appName, axiosOptions) {
   const fdmIndex = getFdmIndex(appName);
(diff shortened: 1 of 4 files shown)


Cabecinha84 (approved):
ack

PR #1754, opened 2026-06-09
v8.14.3

Cabecinha84:
## v8.14.3

Release PR merging `development` into `master`. Bundles two fixes already reviewed and merged into `development` (#1752, #1753) plus the version bump.

### Fixes

**Restart containers on any exit + bound the crash-backoff history (#1752)**
Two fixes to the Docker die-event crash recovery, a stopgap ahead of the full reconciler rearchitecture (`feat/container-reconciler`):
- **Restart on clean exits.** Removed the hardcoded `if (exitCode === 0) return`. Before FluxOS took restarts off Docker, containers ran `restart: always`, which restarted clean exits too; clients whose containers exit 0 to free memory relied on that. Deliberate FluxOS stops are still skipped via the `stoppingContainers` check, so only crashes and unintentional self-exits are restarted.
  - *Behaviour change:* a batch/one-shot container that exits 0 to stay stopped will now restart-loop (paced by the backoff ladder). The full reconciler makes this per-component-configurable via the v9 `restartPolicy` field.
- **Bound `restartHistory`.** `recordRestart` pushed a timestamp on every restart and never trimmed, so a perpetually crashing container grew the array forever. Only the count (capped by the ladder) and the last timestamp are read, so trimming to the ladder length is behaviour-preserving.
- **Guard the immediate-restart path.** `handleContainerDie`'s `delay === 0` path now does the same exists/already-running check as the backoff path, so a container removed outside the `appDockerStop` chokepoint no longer triggers a doomed `appDockerStart` and a misleading error log.
- Added a test asserting the backoff history stays bounded to the ladder length after many crashes (would be 9 without the cap), plus coverage for the removed/already-running skip paths.

**Decrypt previous enterprise spec in `updateAppGlobaly` before signature verify (#1753)**
The synchronous `/apps/appupdate` path passed the raw `appInfo` from `globalAppsInformation` as `previousAppSpec`. Enterprise specs are stored there with `compose`/`contacts` stripped to `[]`, and since commit `2c0e419d5` (2026-05-14) `verifyAppMessageUpdateSignature` no longer decrypts `previousAppSpec` itself — callers own the decryption. That commit updated the broadcast/promotion callers but not this API caller, so on secure (Arcane) nodes the `usersToExtend` expire-only comparison compared a decrypted submitted spec against the stripped previous spec and never matched. Result: enterprise subscription renewals signed by a `usersToExtend` address (the apps monitor) were rejected. App-owner updates and non-enterprise renewals were unaffected. The fix decrypts `appInfo` for enterprise apps before passing it as `previousAppSpec`, mirroring `getPreviousAppSpecifications` and aligning the API path with the broadcast path. Validated end-to-end on a live secure node.

### Other
- Version bump `8.14.2` → `8.14.3`.

### Changed files
- `ZelBack/src/services/appLifecycle/advancedWorkflows.js`
- `ZelBack/src/services/appMonitoring/containerCrashRecovery.js`
- `tests/unit/containerCrashRecovery.test.js`
- `package.json`


MorningLightMountain713 (approved):
Ack

PR #1755, opened 2026-06-10
FluxOS container recovery → single level-based reconciler

MorningLightMountain713:
# FluxOS container recovery → single level-based reconciler

> **PR description** for `feat/container-reconciler` → `development` (117 commits, 147 files, +10976/−3238; written against `22366de26`). The design rationale after the line rule below is kept for context.

## What this PR does

Collapses app-container recovery from **five independent imperative actuators** (Docker `die`-watch, hourly resync, `masterSlaveApps` FDM election, the syncthing folder state machines (FSMs), and boot-start) — each of which both *decided* run-state and *called* `appDockerStop/Start` — into **one level-based, per-container `Reconcile`** driven by a single-flight workqueue. Deciders now only **write `controllerDesired`**; `Reconcile` is the sole actuator (start / stop / recreate / uninstall-on-fail) and the sole owner of the durable crash-backoff ladder.

## Why

The old split meant the event stream only saw *live* transitions, never *absence*: a `dockerd` restart under a running FluxOS left exited containers that **nothing** reconciled until the hourly monitor (which itself had a ~2-pass / ~2h worst case). A customer container sat exited for 1h30m. A level-based reconciler reads actual-vs-desired fresh on every trigger, so orphans and missed transitions self-heal on the next sweep.

## Behaviour: `development` → this branch

| Scenario | `development` | This branch |
|---|---|---|
| Container exited while FluxOS was down / dockerd restarted under FluxOS | Invisible to the `die`-watch; recovered only by the hourly monitor, ~2-pass / **up to ~2 h** worst case | Boot reconcile + level sweep recreate/restart it on the **next trigger** (seconds–minutes) |
| Crash-looping app | Backoff state in memory — a FluxOS restart **reset the ladder** and the loop restarted at full rate | `restartHistory` durable in `appsRuntimeState`; the ladder survives FluxOS restarts and reboots |
| Operator `appstop` | Forgotten on reboot / masterSlave election — the app **came back against operator intent** | `operatorStopped` durable; nothing restarts the app until the operator starts it |
| masterSlave election / syncthing FSM transitions | Each FSM called `appDockerStop/Start` directly — five actuators racing, double-actuation possible | FSMs only write `controllerDesired`; one single-flight `Reconcile` per container actuates |
| Invalid g: `containerData` (e.g. roundcube's `/data\|g:/db`, sync flag on a non-primary mount) | `.includes('g:')` substring match **silently adopted** the spec and the app deadlocked forever | Refused at install (volume construction throws); specs already on disk fail **loud** (`invalidSpec`, no actuation) |
| Valid g: shapes (multi-g, sibling index-ref mounts) | Worked by accident of substring matching | Classified canonically and **behaviour preserved** (deferred to masterSlave; suites 42–43 pin this) |
| Syncthing app start on not-yet-synced data | Possible — risked clobbering good replica data | G2 synced-peer guard: never start until a synced peer state is confirmed |
| Enterprise spec not yet decrypted / partial compose state | Recovery paths could act on partial state | Reconcile defers (R3/R4/R5) |
| Node restarts into already-connected peers | `peerThresholdReached` edge could fire before the orchestrator subscribed → ephemeral app-state sync **stalled until the next block** | Level read at subscription; sync starts immediately |
| Docker daemon unreachable | Error-string matching decided behaviour | Probed explicitly; reconcile defers instead of guessing |
| Uninstall-on-failure, tampering checks, port management | — | **Unchanged** (preserved inside Reconcile's recreate-fail path; see *Things that stay put*) |

## Production changes (33 files under `ZelBack/`)

- **The reconciler** — new `appMonitoring/appReconciler.js` (level-based per-container `Reconcile` + single-flight workqueue) and new `appManagement/appsRuntimeState.js` (durable runtime-state collection). `containerCrashRecovery`, `containerHealthMonitor`, and the boot path (`appStartupManager`) trigger it instead of acting directly.
- **Decider inversions** — `masterSlaveApps` (FDM election) and the syncthing FSMs (`syncthingFolderStateMachine`, `syncthingMonitor`) write `controllerDesired` instead of calling docker start/stop; the old monitor restart path and `die`-handler restart logic are removed (direct replacement, no feature flag — see *Cutover* below).
- **Durable backoff + operator-stop** — `restartHistory` and `operatorStopped` persist in `appsRuntimeState`, so crash backoff survives a FluxOS restart and a user `appstop` survives reboot/election.
- **Data-safety guards** — never reconcile on undecrypted enterprise specs or partial compose states (R3/R4/R5); never start a syncthing app on unsynced data (G2, synced-peer guard). On a true cold start (no peer holds the data yet) the r: election seeds exactly one node, chosen deterministically by IP so all nodes agree (no split-brain), and only a genuinely-empty folder may seed — a node holding data waits, never promoting/reverting on an empty global.
- **Canonical g:/r:/s: mount classifier + fail-loud** — parse-based `mountParser` replaces `.includes('g:')` substring matching across `masterSlaveApps`, `appUtilities`, `volumeService`, and the reconciler; unparseable `containerData` is refused loudly (at install via volume construction, and at reconcile for specs already on disk) instead of guessed at.
- **App-sync orchestrator: edge-vs-level fix** — `AppSyncOrchestrator.start()` now reads the peer-threshold *level* after subscribing to the latched `peerThresholdReached` *edge* (new `FluxPeerManager.peerCountIfAboveThreshold()`). Without it, peers connecting before the orchestrator subscribes (fast inbound reconnects on restart) leave ephemeral state sync stalled until the next block. Same level-over-edge principle as the reconciler itself.
- **Observability** — `reconciler:*` events (`desiredChanged` / `actuated` / `swept`) on the FluxEventBus.
- **Config tunables** — decider cadence and port-test delays move to `config/default.js`.

## Test changes

- **Unit** — new suites for `appReconciler`, `appsRuntimeState`, `volumeConstructor`, `volumeService`; existing app-lifecycle suites extended (18 files touched).
- **Integration harness (`test-infra/`)** — 23 new suites (28–49, incl. an enterprise sibling of 28) covering the reconciler end-to-end on real DinD FluxOS fleets: dockerd-restart orphan recovery (plain and encrypted-enterprise), crash/exit-0 restart, durable backoff, operator-stop durability, missing→recreate, masterSlave election, syncthing first-run/stalled/fluxos-restart, boot reconcile, multi-instance spawning, the g: classifier fixtures (install-rejection of `[plain|g]` and self-ref shapes; clean install of valid multi-g and sibling index-ref shapes), and the post-review remediations end-to-end: backup-lease suppression during sweeps, stops landing inside docker-event-stream outages, boot-broadcast ordering (first broadcast complete, never empty), syncthing unreadable-status alert-not-remove, the stall-ladder nudge/removal evidence gates, the promotion pollution gate, and events-consumer outage recovery. New framework pieces to support them: an FDM stub, a syncthing stub with fault-injection controls, reconciler suite helpers, deterministic subnet allocation, poll-based container wait strategies, and per-node evidence dumps on any setup failure.
- **Concurrent gate runner** — `run-all.sh` (per-suite isolation, label-scoped cleanup, atomic subnet claims, per-suite wall-clock bound) and `run-parallel.sh` (memory- and load-gated concurrency, host-wide boot serialisation so fleet boots never contend, ryuk-free cleanup safe for N concurrent test processes). Full 48-suite gate runs in well under an hour vs ~2 h+ sequential.
- **CI** — `.github/workflows/integration-harness.yml`: manual (`workflow_dispatch`) full-gate run on the self-hosted `runonflux` runner, with suite/concurrency inputs, log artifacts, and runner-safe scoped cleanup. Also fixes a long-standing Node CI defect: `test:zelback:unit` chained commands with `;`, so the step's exit status was `docker stop`'s — **unit-test failures never failed CI**. The script now propagates mocha's exit code (cleanup still unconditional); this PR's pushes are the first where the unit step actually gates.

## Validation

- **Unit:** **full canonical suite green** (`npm run test:zelback:unit`, which now actually enforces in CI after the exit-code fix — the CI run on this head reports the exact count), incl. new/updated suites: `appReconciler`, `syncthingFolderStateMachine` (48 — incl. the cold-start election cases), `volumeConstructor` (33), `advancedWorkflows` (51), `appSyncOrchestrator` (57), `fluxPeerManager` (195), `containerHealthMonitor` (12, rewritten for the surviving recreate paths), `peerNotification` (broadcast derivation + hourly reconcile trigger). Single files: `env NODE_CONFIG_DIR=$PWD/tests/unit/globalconfig npx mocha <file> --exit`.
- **Integration:** **full parallel gate green — every suite, 0 failures, at `22366de26`** (incl. the new cold-start suite and the masterSlave g: election suite); the new/rewritten suites were each green solo first.
- **Live nodes — multi-day soak on 8 real nodes** (charlie, squidward, sandwich, chalk, chud, salad, mule, legume), running real customer workloads throughout, bumped through successive branch heads (`8cd361d19` → `84f982984` → `6d99dc6a9` → `22366de26`, current) with `NRestarts=0` at every check. Observations below were captured across the soak against the real app population:
  - **roundcube** (real production spec with the invalid `/data|g:/db` non-primary sync mount, on chud): refused at **both** enforcement points — install (`Error parsing mount at index 1`) and reconcile (`has invalid containerData, not reconciling`) — instead of the old silent adopt-and-deadlock.
  - **simplexftp** (real valid multi-g app, every component g:): boot reconciler classifies it correctly and **defers to masterSlaveApps** — no spurious `invalidSpec`, no double actuation; all four of its syncthing folders **100% synced / idle** on every monitor cycle, with the first-run mount-safety guard correctly gating the first cycle after boot.
  - **Minecraft** (g: app with a domain): masterSlave election functioning — FDM returns a master IP.
  - **Zero `invalidSpec` hits on any healthy app across all 8 nodes** — the classifier separates genuinely-invalid from valid shapes on the live population, not just fixtures.
  - **Zero unexpected actuations fleet-wide**: post-bump logs show no docker stop/kill calls, no `controllerDesired` transitions, and no reconcile actuations anywhere — the level-based actuator is correctly idle on a fleet where actual already equals desired, while every customer app's uptime predates the restart.
  - The orchestrator peer-threshold fix (`aeb185620`) observed live on all 8: `Peer threshold reached` fires on every boot, ephemeral state sync requests and completes.
  - Every bump restarts FluxOS, so the boot-reconcile path has run repeatedly over each node's real container population — the orphan/exited recovery paths themselves are pinned by integration suites 28–33 and 38.

### Pending before merge

- [ ] **Pull soak on 8 live nodes at the current head (`22366de26`) — IN PROGRESS** (deployed from origin, logs truncated at deploy; first recon clean — `NRestarts=0`, zero unwanted actuation, no "poll failed" spam, stall ladder dormant). Production changes since the previously soaked `6d99dc6a9` include the B1 empty-global data-loss fix, the cold-start r:/g: election fix, the syncthing declare-only reset, the events-consumer transport + restart recovery, and the review remediations. Watch: `NRestarts`, the stall-ladder logs, the events consumer, and the cold-start election (exactly one seed; no spurious leader/nudge/removal on a healthy folder).

## Known issues / explicitly out of scope

- **Two PRE-EXISTING g:/masterSlave bugs — out of scope here, resolved by the v9 spec** (byte-identical behavior to the base branch; this PR neither causes nor fixes them):
  - roundcube/rainloop put the sync flag on a non-primary mount (invalid spec); this PR's fail-loud correctly **flags** it instead of looping silently. The v9 spec removes the invalid shape at source.
  - Domainless g: apps (simplexftp/Tor-only) can't be elected a master because FDM/HAProxy only routes apps with domains/ports — addressed by the v9 spec.
- **r: master election is a healthy-network heuristic, not consensus.** The cold-start seed election (data-aware, IP-deterministic, empty-folder-guarded) removes the standoff and split-brain failure modes on a healthy network; a genuinely partition-proof election (a deterministic candidate + a data-aware quorum lease) is **future work in a follow-up PR**. The residual — a peer holding newer data while *disconnected* could be out-seeded — is unchanged from the base branch's leader-seed behaviour.

---

### Problem

App-container recovery is split across **multiple imperative actuators** that each both *decide* and *call* `appDockerStop/Start`:

- `appMonitoring/containerCrashRecovery.js` — Docker `die` event watch, per-container restart + backoff.
- `appMonitoring/containerHealthMonitor.js::monitorAndRecoverApps` — hourly resync; restart stopped, recreate missing, uninstall-on-failure, masterSlave/g:/r: handling, compute broadcast set.
- `appLifecycle/advancedWorkflows.js::masterSlaveApps` — 30s FDM election; starts g: on elected node, stops on standbys.
- `appMonitoring/syncthing*` state machines — stop/start during folder sync, leader election, health.
- `appLifecycle/appStartupManager.js::reconcileAppsOnBoot` — boot-time start of stopped containers.

Consequences: the event stream only sees live transitions (never *absence* / orphans); a dockerd restart under a running FluxOS leaves exited containers nothing reconciles until the hourly monitor; the hourly monitor had a 2-pass debounce (~2h worst case). Observed in production: a customer container exited for 1h30m.

## Model: `desiredState` as the one interface; Reconcile as the one actuator

Kubernetes split — **deciders write desired state; one enforcer drives actual → desired.**

```
  user appstop/start ─┐  writes
  masterSlave election ┼─►  per-component desired state  ──►  Reconcile(id)  ──►  docker start/stop/recreate
  syncthing readiness ─┘                                      (ONLY actuator)     + backoff + uninstall-on-fail
```

### Desired-state shape (two deciders, different authority)

A user `appstop` on a g: app must not be overwritten by the next election cycle, so split authority:

| field | meaning | persistence | written by |
|---|---|---|---|
| `operatorStopped` | user parked it | **durable** (DB) | appController appstop/appkill (set), appstart (clear) |
| `controllerDesired` | election / sync-readiness output (`running`/`stopped`/unset) | **ephemeral** (in-memory, re-derived each boot) | masterSlaveApps, syncthing state machines |
| `restartHistory` | crash backoff ladder timestamps | **durable** (DB) | Reconcile on each restart |
| `lastExitCode`/`lastDiedAt` | diagnostics / tampering | durable (DB) | Reconcile on die |

`controllerDesired` is ephemeral on purpose: after a reboot we must NOT act on a stale election; the live controllers (masterSlaveApps every 30s, syncthing on evaluation) repopulate it. Durable `operatorStopped` + `restartHistory` survive reboot (backoff no longer resets on restart).

### Restart policies (Docker-like, from v9 spec)

The current `if (exitCode === 0) return` is a hardcoded **on-failure** policy — and a regression: before FluxOS took over restarts, Docker ran `restart: always`, which DID restart exit-0 containers (a client relies on this — containers exit 0 to free memory, must come back). Make it a per-component **policy**, defaulting to **always** (restoring prior behavior), configurable via a new v9 component field `restartPolicy`:

| policy | restart a stopped container? |
|---|---|
| `always` (default) | yes, regardless of exit code |
| `unless-stopped` | yes (operatorStopped already separate) |
| `on-failure` | only if exit code ≠ 0 (today's accidental behavior) |
| `no` | only initial start; never after an exit |

Enforced at the **FluxOS level by Reconcile** using the container's *actual* exit code (`docker inspect` `State.ExitCode`, truth even for orphans), NOT pushed to Docker's `RestartPolicy` (which stays `no` — Reconcile is the sole actuator).

**Now (pre-v9):** one resolver `getRestartPolicy(spec)` returns the constant `'always'` — the *only* place policy source lives; no spec-reading anywhere. The enforcement switch `policyAllowsRun(policy, exitCode)` is fully built + unit-tested (only `always` reachable in prod), and Reconcile plumbs the actual exit code. **v9 = change the resolver's one line** to `spec.restartPolicy ?? 'always'` (+ v9 spec validation); Reconcile untouched.

### Effective desired run-state

```
effectiveDesiredRunning(id, spec, actualExitCode):
  if operatorStopped(id):                 return false     // user wins
  if isG(spec) || isR(spec):
     if controllerDesired(id) != running: return false     // not elected / not synced
  return policyAllowsRun(spec.restartPolicy ?? 'always', actualExitCode)

policyAllowsRun(policy, exit):
  always | unless-stopped: true
  on-failure:              exit == null || exit != 0
  no:                      exit == null            // start once, never restart after an exit
```

Reconcile keeps **no** g:/r: run-heuristics (no 30-min grace, no readiness check) — those live in the syncthing controller via `controllerDesired`. The only spec awareness Reconcile keeps: the **start primitive** (g: → `appDockerRestartWithPermissionsFix`, plain → `appDockerStart`) and the **restartPolicy**.

## Reconcile(identifier)

Idempotent, level-based. Inputs read fresh each call: spec (installed? component kind), effective desired, actual docker state, restartHistory.

```
Reconcile(id):
  spec = localSpec(id)
  if !spec: return                         // not installed → nothing to enforce
  desired = effectiveDesired(id, spec)
  actual  = dockerState(id)                // running | stopped | missing

  if desired == STOPPED:
     if actual == running: stop(id) (mark stoppingContainers)   // enforce operator/standby stop
     return

  // desired == RUNNING
  if actual == running: return
  if actual == missing:
     try recreate(id)                       // recreateMissingContainers (reinstall component)
     on fail: recordEvent(container_vanished/recreation_failed/network_pruned); removeAppLocally(name,null,false,true,true)
     return
  // actual == stopped, desired running:
  delay = backoffDelay(id)                  // durable restartHistory ladder [0,30s,5m,15m,30m], reset after 10m stable
  if delay elapsed since last attempt:
     recordRestart(id); start(id, primitiveFor(spec)); startAppMonitoring(id)
  // else leave; a later enqueue re-evaluates
```

Crash recovery's `handleContainerDie` collapses into: record exit → `enqueue(id)`. The restart/backoff logic lives only in Reconcile.

### Triggers (all just enqueue)

- die event → `enqueue(id)`
- hourly timer → `enqueueAll()`
- stream reconnect (orphan sweep) → `enqueueAll()`
- post-install → `enqueue(id)`
- masterSlaveApps / syncthing write controllerDesired → `enqueue(id)`
- boot → `enqueueAll()` (after boot-specific gates, below)

### Workqueue (per-key single-flight)

Reuse `peerNotification`'s idiom, keyed by identifier: if a key is in-flight, mark it dirty and re-enqueue on completion (`setImmediate`). Different keys reconcile in parallel; a given key never overlaps. No lost trigger, no double-actuation.

## Decider inversions

**masterSlaveApps** (`advancedWorkflows.js` ~3914-4051): replace the 5 `appDockerRestartWithPermissionsFix(identifier, appId)` calls with `setControllerDesired(id, RUNNING); enqueue(id)`, and the `appDockerStop(identifier)` (4015) with `setControllerDesired(id, STOPPED); enqueue(id)`. Election, scheduling, lower-index checks, and the syncthing-readiness gate (4019-4044) all stay — masterSlaveApps only writes `running` when it already decided to start. `mastersRunningGSyncthingApps` map unchanged.

**syncthing state machines** (`syncthingMonitor.js`, `syncthingFolderStateMachine.js`, `syncthingHealthMonitor.js`): replace `appDockerStop/Restart` actuation with `controllerDesired` writes (stopped during receive-only sync, running once sendreceive/leader). Side-effects (mount data wipe, folder-mode transitions, permission fixes) stay in the state machines around the write. Ownership between syncthing and masterSlave for the same g: component is composed in the decider — single write requiring both synced AND elected (O2, *Resolved questions*).

## Things that stay put

- **Boot-specific gates** (`manageAppsOnBoot`): location-expiry, node-DOS, daemon/DB timeout, mass-removal. Boot still decides *whether* to keep apps; only the per-container start becomes `enqueueAll()` into Reconcile.
- **Broadcast set** (`peerNotification`): computed as a pure read (installed ∩ running + syncthing-mode apps). Decouple from Reconcile — it no longer returns `masterSlaveAppsInstalled`/`startedApps`.
- **Uninstall-on-failure + tampering**: preserved exactly inside Reconcile's recreate-fail path.

## Cutover — DIRECT REPLACEMENT (O1 resolved)

No coexistence, no feature flag, no shadow mode. Two actuators running side by side (old + Reconcile) would double-actuate and the flags to gate them turn into a jumbled mess — the exact conflict we're removing. Land it as one clean replacement: build Reconcile + workqueue, invert the deciders, and **remove** the old actuation (monitorAndRecoverApps restart path, handleContainerDie restart logic, masterSlave/syncthing direct start/stop) in the same change. Safety net is **thorough testing**, not a runtime flag: unit tests + the integration harness (testcontainers + FluxEventBus SSE) + test-node deployment before fleet.

Foundation: `appsRuntimeState` collection + module, `containerCrashRecovery` reduced to enqueue-only, `onReconnect` orphan sweeps. The desired-state split shipped as designed: durable `operatorStopped` + in-memory `controllerDesired`.

## Boot sequence (the workqueue release gate)

`manageAppsOnBoot` (appStartupManager.js:275) gates everything: first-boot/location-expiry → maybe `removeAllApps`; then `waitForDaemonReady` (5min timeout → removeAllApps), `waitForConfirmationStatus`, DOS check, `waitForDbReady` (timeout → removeAllApps); then `reconcileAppsOnBoot`; then **`globalState.bootContainerStateSettled = true`** + `boot:settled` event.

**The whole reconcile workqueue is gated on `bootContainerStateSettled`** — generalizing crash recovery's existing boot queue to every trigger. During boot, all triggers (die events, masterSlave/syncthing `controllerDesired` writes, hourly) just enqueue keys; the queue holds them and actuates nothing. When `manageAppsOnBoot` opens the gate (after daemon+DB+confirmation+DOS+location all pass), the queue drains with everything ready, so Reconcile reads specs/locations safely. `reconcileAppsOnBoot` reduces to: do its location-validity checks / remove-reassigned, then `enqueueAll`.

## Scope: ONE PR, both correctness domains

Confirmed one PR containing the full architecture: control plane **and** the rewritten masterSlave/syncthing deciders. The `setControllerDesired(id, state, reason)` call is the **internal seam** (separation of concerns), not a PR boundary. Honest logical commits within the one PR.

## Syncthing + masterSlave deciders (correct design)

Deciders compute desired run-state from **live truth** and write `controllerDesired`; Reconcile actuates. `operatorStopped` (durable) always wins; deciders skip operator-stopped components (so the election doesn't churn on a component Reconcile refuses to start).

**Component flags** (`syncthingMonitorHelpers.js:233`): `g:` = masterSlave (run on exactly one elected node, sendreceive); `r:` = receive-only (runs locally once synced); `s:` = sync-only (folder, no container). `requiresSyncing = s|r|g`; container-bearing = `r` or `g`.

**Real signals (replace the heuristics):**
- `getFolderSyncCompletion(id)` → real `{inSyncBytes, globalBytes, syncPercentage, state, isSynced}` — primary readiness signal, replaces `numberOfExecutions >= MAX_SYNC_WAIT_EXECUTIONS` and "force start".
- `checkIfPeersAreSynced(id)` / `getDbCompletion` → peer has 100% — gates safe removal.
- `getPeerSyncDiagnostics()` → `localStatus.state` (syncing/idle/scanning), `peersAreMoreUpdated`, `canSync` — replaces byte-history stall heuristic.
- `isDesignatedLeader(peers, local)` → deterministic leader (broadcastedAt, IP) — replaces `index*3min` stagger.
- FDM `getMasterIpFromFdm(name)` (`advancedWorkflows.js:188`) → authoritative primary IP. **Essential, keep.**

**Desired-state derivation:**
```
g: desired-running = operatorStopped ? false
                   : (FDM says this node is primary) AND (folder data consistent: getFolderSyncCompletion.isSynced)
r: desired-running = operatorStopped ? false : (folder data consistent)
s: never has a container
plain: desired-running = operatorStopped ? false : true
```
The decider writes `controllerDesired` from this each cycle; Reconcile enforces with the durable backoff. Backoff applies to g:/r: too (O4: a crash-looping elected master backs off normally; election keeps it primary, Reconcile paces restarts — no conflict).

**Data-safety sequences that MUST be preserved (stay synchronous in the decider, NOT inverted):**
1. set cache/intent → stop container → delete data (`appDeleteDataInMountPoint`) → set receiveonly. (first run / new app / wipe-resync)
2. fix permissions (`fixAppdataPermissions`) → folder→sendreceive → start. (sync complete)
3. verify mount safety (`verifyFolderMountSafety`) → receiveonly if unsafe. (post-reboot, prevents empty-state propagation)
4. confirm peers synced → remove app. (no data loss)
5. lowest-index/leader confirmed → start. (no split-brain)

These are data-safety operations, not run-state intent: the decider performs them synchronously AND sets `controllerDesired` consistently, so Reconcile (idempotent, level) never races them.

**Essential vs replaceable:** keep — FDM election, `cache.restarted` gate, mount-safety, leader election, lower-index checks, peer-synced-before-remove. Replace — `numberOfExecutions`/`MAX_SYNC_WAIT_EXECUTIONS` time-proxy, `index*3min` stagger, byte-history stall detection (use real sync state), syncthing-restart-on-stall (try rescan/reconnect first).

**FSM durability:** today the FSM state (`receiveOnlySyncthingAppsCache`) is in-memory and lost on FluxOS restart, forcing re-derivation from folder type + mount safety. Correct design derives from live truth each cycle (sync %, FDM, mount), so lost in-memory FSM state is a non-issue — but any genuinely stateful bit (e.g. "wipe already done") must live in `appsRuntimeState`, not an ephemeral Map.

## appController interaction

`appStart`/`appStop`/`appKill` (appController.js:105/239/445) → set/clear `operatorStopped`. Today `appStart` already special-cases g: ("skip start if not currently running" — defers to election); under the new model that becomes: clear `operatorStopped` and `enqueue(id)`, letting the decider+Reconcile decide. Global variants go through `executeAppGlobalCommand`.

## Resolved questions

- **O1 cutover**: direct replacement (above).
- **O2 ownership**: syncthing decider owns `controllerDesired` for the data-readiness dimension; masterSlave owns the election dimension; for g: the effective `running` requires BOTH (synced AND elected) — composed in the decider, single write.
- **O3 stop-before-wipe**: preserved — data-safety stops stay synchronous in the decider, `controllerDesired` set consistently so Reconcile no-ops.
- **O4 backoff vs election**: durable backoff applies uniformly; election keeps the node primary while Reconcile paces restarts. No conflict.


Cabecinha84 (commented):
The architecture is genuinely good — the level-based single actuator, durable operator intent, the docker-unreachable probe, and the boot gating are
  all sound and strictly safer than the old hourly monitor in several scenarios. The agents independently verified a lot: the workqueue has no
  lost-trigger races, the boot gate opens on every exit path (set in a finally), the getPreviousAppSpecifications move is functionally identical with
  no new require cycle, the volumeService extraction is behavior-preserving (with a shell→execFile hardening), the masterSlave election logic survives
  the inversion intact, identifier canonicalization is consistent, the peer-threshold edge-vs-level fix is correct and idempotent, and the
  CI/test-infra portion is safe. The 42-suite harness and 8-node soak are real validation.

  But the review found six issues I'd treat as blockers. I verified the first three directly in the source myself; the others were found independently
  by two or three separate review passes.

  Blockers

  1. Backup/restore guard never matches — reconciler can restart a container mid-backup/restore (data corruption). advancedWorkflows.js:2198/2320/2693
  push the bare app name into backupInProgress/restoreInProgress, but isManagedElsewhere (appReconciler.js:189-190) checks the component identifier
  (component_app). Backup/restore only operates on composed apps, so the guard is dead. The deliberate stop's die event is swallowed via
  stoppingContainers, but an hourly sweep, reconnect sweep, or a pending backoff timer landing inside the multi-minute backup/restore window will see
  "stopped + desired running" and start the container while the tarball is being written or extracted. For g: apps it's worse: controllerDesired is
  still 'running' from before the backup and masterSlave skips the app during backup, so nothing corrects it. Fix is one line: compare the main app
  name.

  2. enqueueAll silently skips all enterprise apps. Enterprise specs are stored locally with compose: [] (appInstaller.js:513-517), and enqueueAll
  (appReconciler.js:376-383) iterates the raw undecrypted list: version >= 4 && Array.isArray([]) is true, so the forEach enqueues nothing and the else
  branch is unreachable. The hourly sweep and the dockerd-reconnect orphan sweep — the PR's headline recovery mechanism — never cover enterprise apps.
  The motivating incident (a customer container exited for 1h30m) would still be unrecoverable if that app were enterprise and the die event was
  missed. The old monitorAndRecoverApps received the decrypted list, so this is a regression. Fix: decrypt leniently in enqueueAll or derive
  identifiers from docker containers.

  3. First post-boot broadcast can erase the node's app locations network-wide. Old boot path awaited every appDockerStart before opening the gate; now
  containers are merely enqueued and the gate opens immediately, so the first apprunning broadcast races the reconciler's actual starts. Containers
  are created with RestartPolicy: 'no', so after a machine reboot everything is stopped, and a durable pre-reboot backoff can legally delay starts up
  to 30 min. If the first snapshot sees zero running apps, checkAndNotifyPeersOfRunningAppsFirstRun bypasses the empty-broadcast guard
  (peerNotification.js:134-137 — verified) and the empty v2 message makes every peer (and the node itself) delete all appsLocations rows for this IP
  (messageStore.js:361-369 — verified) → respawns elsewhere, FDM/DNS drop, corrected only at the next hourly broadcast. Fix: order the first broadcast
  after the boot drain, or drop the firstRun bypass of the empty guard.

  4. appRestart doesn't clear operatorStopped — the reconciler silently reverts an explicit user action. Found independently by three review passes.
  appStart clears the lock, appStop/appKill set it, but appRestart (appController.js:346-465) touches neither, and dockerService.appDockerRestart
  deliberately starts a stopped container. So: user stops app → user restarts it → up to an hour later the reconciler stops it again with no feedback.
  A regression introduced by the new durable lock; one-line fix.

  5. Recreate-fail → removeAppLocally lost its existence re-check, and the missing-classification has a false-positive window. dockerActual
  (appReconciler.js:164-178) discards the probe's own list result; if dockerd comes back between the failed inspect and the probe (or the inspect fails
  transiently while docker is up), an existing container is classified missing → recreate hits a 409 name conflict → removeAppLocally with network
  broadcast. The old g: path had exactly the guard this needs (containerHealthMonitor.js:104-107, containerExistsNow check) and the new path drops it.
  Net-net the new code is still safer than base (whose hourly loop removed apps when docker was down), but the fix is ~5 lines and the path uninstalls
  customer apps.

  6. A leaked stoppingContainers flag permanently wedges the sole actuator for that component. appDockerStop adds the flag and never removes it on
  success — only the die handler consumes it. If the event stream is down at the moment of any deliberate stop (the exact window this PR's reconnect
  sweep exists for), the flag lives forever and every reconcile of that key defers in a silent 5s retry loop; nothing can ever clear it because the
  reconciler's own appDockerStart is unreachable. At base this leak was harmless (the hourly monitor ignored the flag); now the reconciler is the only
  actuator, so it's fatal per-component until FluxOS restarts. Fix: TTL the entries, or let reconcile clear the flag when it observes the container
  already stopped.

  Should be fixed too (same PR or named follow-up)

  - parseMountFlags accepts garbage as flag sets (mountParser.js:38-56): segment.includes(flag) means "logs:/var/log" parses with flags ['g','s'] and
  is classified as a g: app — left stopped at boot, election-managed. Registration validation only checks containerData length, so such specs can exist
  on chain. Since this PR promotes the parser to the single source of truth, tighten the flag segment to /^[rgs]+$/.
  - Stalled-sync give-up shrinks from ~1h to ~10.5min with byte-granularity progress (syncthingFolderStateMachine.js:393-411, 512-555): inSyncBytes
  advances at file-completion granularity, so a single multi-GB file syncing for >5min reads as stalled → local removal → respawn → resync → removed
  again, a churn loop for exactly the large-data workloads g:/r: serve. At minimum let state === 'syncing' veto the stall verdict.
  - The 2.5h unreadable-status removal path (syncthingFolderStateMachine.js:556-575) lacks the checkIfPeersAreSynced guard the stalled path has — a
  last-surviving-copy edge case where local removal destroys the only data.
  - clearControllerDesired is dead code — uninstall never clears the in-memory map, so a reinstalled g: app inherits a stale election verdict, plus a
  slow leak. Wire it into removeAppLocally.
  - No desired-state re-check before appDockerStart (appReconciler.js:304-320): an in-flight reconcile that read controllerDesired='running' can land
  its start after the syncthing stop wrapper's synchronous stop, so the data wipe can run against a running container — the one hole in the
  stop-before-wipe invariant. One-line re-check closes it (and pairs with the dead clearControllerDesired, which is the realistic enabler).
  - No unique index on zelappsruntimestate.identifier — concurrent first upserts can duplicate documents, making the operator lock nondeterministic,
  durably.
  - Backoff ladder never reaches its 30m cap: STABLE_RUN_MS (10m) < the 15m rung, and the "stable run" reset fires while the container sat stopped in
  backoff — effective cycle is 0→30s→5m→15m→reset.
  - Failed appDockerStart is logged and dropped — no scheduleRetry, so recovery waits for the hourly sweep (which, per blocker 2, never comes for
  enterprise apps — compounding).

  Minor / follow-up

  Component-path stop/kill persists the lock after the docker op (whole-app path does it before — flag-first is strictly safer); setOperatorStopped
  swallows DB errors while the API reports success; soft redeploy preserves the lock while hard redeploy wipes it (pick one semantic); boot start
  pacing (old 2s spacing) is gone — N simultaneous decrypts/starts on reboot; the PR itself adds a new loose includes('g:') in peerNotification.js:92
  while preaching the canonical classifier; handleMissingMasterSlaveContainer is now dead code; containerCrashRecovery.subscribe() has a reentrancy gap
  that can double the event stream. CI: scope the runner cleanup to the harness's own flux-e2e-run label rather than org.testcontainers=true, and pass
  the suites dispatch input via env:.

  Bottom line

  This is high-quality work with unusually honest validation, and the direction is right — but blockers 1–3 are confirmed-in-source correctness holes
  in exactly the guarantees the PR claims (data safety during backup/restore, sweep coverage, network presence), and 4–6 are user-visible or
  availability-killing regressions with small, local fixes. I'd request changes citing the six blockers (all fixes are a few lines each except the
  boot-broadcast ordering, which needs a small design decision), and suggest the "should fix" list either ride along or get tracked explicitly. Once
  those land, this is ackable — the soak and harness evidence for the happy paths is strong; what the live validation couldn't exercise is precisely
  the failure windows the blockers live in.

Cabecinha84 (commented):
Re-review complete. I checked out the PR head (6795b8b7cc), read the full fix diff, and ran six independent adversarial
  verification passes against the live code — one per blocker plus a sweep of the secondary list. Tree restored to
  development.

  Bottom line

  The six blockers are genuinely and correctly fixed — verified in source, not just claimed. That part is ackable. But the
  author touched only the six blockers — none of the ~13 "should fix"/minor items you raised were addressed. A few of
  those are correctness/data-safety adjacent. So my recommendation is ACK conditional on tracking the secondary list, with
  B/E/F worth pulling forward — not a clean ack yet.

  The six blockers — all FIXED ✅

  #: 1
  Blocker: Backup/restore guard never matched
  Verdict: Fixed, clean
  Notes: isManagedElsewhere + listRunningApps now compare the bare main app name. The underscore-in-name worry I'd have
  had
    is structurally impossible — APP_NAME_REGEX/APP_NAME_REGEX_LEGACY forbid _ in both app and component names, so   
    split('_')[1] is always correct.
  ────────────────────────────────────────
  #: 2
  Blocker: enqueueAll skipped all enterprise apps
  Verdict: Fixed, clean
  Notes: Now decrypts leniently; undecryptable apps are covered via their existing docker containers (defer-retry,
  converge
    when benchd answers). New decryptEnterpriseSpec shares one in-flight attempt + 60s failure window — no benchd
    hammering. Cache (7-day) keeps steady-state sweeps benchd-free. Residuals all minor/by-design.
  ────────────────────────────────────────
  #: 3
  Blocker: First post-boot empty broadcast wipes locations
  Verdict: Fixed, clean
  Notes: Both halves: empty snapshot is never broadcast now, and the snapshot awaits a new boot-drain AsyncGate (capped 2
    min). Verified the gate is guaranteed to open and there's no new deadlock. Last-app removal still propagates via
    fluxappremoved, so dropping the empty-broadcast doesn't strand rows.
  ────────────────────────────────────────
  #: 4
  Blocker: appRestart didn't clear operatorStopped
  Verdict: Fixed, clean
  Notes: Both component and whole-app branches now clear the lock before the docker op, mirroring appStart. Global variant

    routes back through local restart and clears too.
  ────────────────────────────────────────
  #: 5
  Blocker: Vanished-classification false positive → uninstalls healthy app
  Verdict: Fixed
  Notes: The make-or-break check passes: getAppDockerNameIdentifier returns /flux<name> (slash + prefix) which exactly
    matches docker's Names, so the new indeterminate-defer branch is live, not dead code. Recreate-fail now re-checks
    existence before removal. One narrow residual below.
  ────────────────────────────────────────
  #: 6
  Blocker: Leaked stoppingContainers flag wedges the actuator
  Verdict: Fixed
  Notes: Flag is now operation-scoped (cleared in finally in appDockerStop/Kill), die handler no longer owns it, and a die

    arriving post-clear correctly no-ops (desired already = stopped). Bonus: the stream error/end/close rework correctly
    collapses to one resubscribe (fixes the double-stream + silent close loss).

  Two tiny residuals inside the blocker fixes (not ack-blocking, worth a one-line mention to the author):
  - Blocker 5: the post-recreate-failure getDockerContainerOnly(...).catch(()=>undefined) collapses "docker down" and
  "genuinely gone" into the same undefined → could still uninstall if dockerd dies in that exact few-ms window. Far
  narrower than the original bug.
  - Blocker 6: component-path appStop/appKill still set operatorStopped after the docker op (whole-app sets it before).
  Practically safe (DB write is issued before the die is processed) but asymmetric — moving it before would close the
  theoretical race.

  The secondary list — NONE fixed ⚠️ 

  All 13 of your "should fix"/minor items are still present at HEAD. Ranked by what actually matters:

  Worth pulling into this PR or a fast-follow (correctness/data-safety):
  - B — syncthing stall give-up shrank ~1h → ~10.5min (syncthingFolderStateMachine.js). state === 'syncing' still doesn't
  veto the stall verdict → genuine churn-loop risk (remove→respawn→resync→remove) for exactly the multi-GB g:/r: workloads
  syncthing serves. This is the most operationally concerning one. Live soak didn't hit it (simplexftp's folders were
  small/synced), but a single >5min file will.
  - E — no desired re-check before appDockerStart (appReconciler.js). The one remaining hole in the stop-before-wipe
  invariant — a data wipe can run against a re-started container. One-line fix.
  - F — no unique index on zelappsruntimestate.identifier (appsRuntimeState.js). Concurrent first upserts can duplicate
  the doc → operator lock becomes durably nondeterministic.
  - C — 2.5h unreadable-status removal still lacks the checkIfPeersAreSynced guard (last-copy data loss). They added a
  justification comment arguing it's safe, but not the guard. Worth a deliberate sign-off rather than silently accepting.
  
  Fine as explicitly-tracked follow-ups:
  - G backoff ladder never reaches its 30m cap (STABLE_RUN_MS 10m < the 15m rung) — degraded backoff, not dangerous.
  - D clearControllerDesired still dead → reinstalled g: app inherits stale election + slow map leak. 
  - H failed appDockerStart dropped with no scheduleRetry (hourly sweep recovers it now that enterprise is covered).
  - A parseMountFlags still uses substring includes — logs:/var/log classifies as a g:/s app. Soak showed zero
  misclassification on the real population, but this is now the single source of truth, so /^[rgs]+$/ is cheap insurance.
  - I new loose includes('g:') in peerNotification.js:98 — ironic given the canonical-classifier theme.
  - J handleMissingMasterSlaveContainer dead code; K/L/M lock-ordering / swallowed-DB-error / soft-vs-hard-redeploy lock
  asymmetry.

MorningLightMountain713:
> **Status: all remediations are landed and gated (unit + full integration gate green),
> and the final head is now soak-testing on 8 live nodes — we will not merge until that
> soak is verified.**

Thanks for the depth of this review — we verified every claim against source before
acting, and all six blockers plus all eight should-fixes were confirmed (three of the
should-fixes survived our own attempts to refute them; you were right each time). Here is
what was done, what we'd correct in the review's framing, and one item where we went
further than the review asked after live testing overturned its premises.

## Blockers — fixed tests-first, pushed as `6795b8b7c`

- **B1 (backup lease):** `isManagedElsewhere` and `listRunningApps` now derive the bare
  main name before the lease lookup. Tests pin start/stop/recreate suppression while
  leased and resume-on-release. One correction for the record: this was **not a
  regression** — the base branch's guard compared `stoppedApp === backupItem` (component
  id vs bare name), which was equally dead for composed apps. Pre-existing dead guard,
  faithfully ported, and still right to fix here. End-to-end: new suite 44 holds a real
  backup lease open and fires a dockerd-restart reconnect sweep inside it.
- **B2 (enterprise sweeps):** `enqueueAll` decrypts via a lenient path with a hash-keyed
  success cache; still-encrypted apps get their existing containers enqueued so
  reconciles defer-retry and converge the moment benchd answers. Never skip silently,
  never actuate on encrypted data, never remove. End-to-end: new suite 28-enterprise
  (production storage shape - v8, empty compose, AES-GCM blob) recovers an encrypted app
  through the sweep.
- **B3 (first post-boot broadcast):** three changes — the empty-snapshot bypass is
  deleted; broadcasts wait for the reconciler's boot drain to settle (gated, 2-min cap);
  and a start-nudge refreshes peers immediately when a backoff-delayed component finally
  starts. On the deletion: we know the bypass was added deliberately (`81eb356f5`) for
  the multitoolbox wipe, but the trade does not hold up. An empty v2 snapshot deletes
  every `appsLocations` row for the sender — including on the sender itself — and the
  bypass cannot tell "wiped, runs nothing" from "rebooted, containers not started yet"
  (RestartPolicy `no`; starts are async behind the boot gate). So every ordinary reboot
  races a window where the node erases its own network presence: FDM stops routing to
  it, the spawner respawns its instances elsewhere, and its own boot location check
  then removes healthy local apps as "reassigned". That is a network-wide failure mode
  on the common path, paid for one rare, manual maintenance op whose cleanup the
  network already performs by itself: a clean shutdown broadcasts sigterm and peers
  expire the node's rows in ~7 minutes, and with no shutdown at all the running-TTL
  clears them within ~2 hours. Deliberate removals broadcast a targeted
  `fluxappremoved`; reassignment is caught by the boot-side location check. Every
  legitimate correction already has a bounded, targeted mechanism — the empty snapshot
  is an unbounded network-wide bulk delete that buys minutes on a self-healing case.
  The receive-side empty-v2 handler is KEPT for fleet back-compat (older nodes still
  send them legitimately); a proper in-band path for the wipe flow is queued as a
  follow-up. One framing correction: non-empty v2 messages only upsert
  (`messageStore.js:332-360`) — partial early broadcasts were never destructive; the
  real post-boot hazard is unrefreshed rows dying on the sigterm TTL, which the
  boot-drain gate + start-nudge address. End-to-end: new suite 46 reboots a node and
  asserts the first broadcast is complete, never empty, and that peer rows survive.
- **B4 (apprestart vs operator lock):** `appRestart` now clears the lock with
  `appStart`'s exact semantics (scope follows the name, cleared BEFORE the docker op).
  End-to-end: new suite 32 sibling — the app stays up across the restart-triggered
  reconcile, which fails with a lingering lock.
- **B5 (false "vanished"):** "vanished" now requires docker to confirm absence (probe
  list content; in-list-but-inspect-failed → defer, no tamper entry), and the
  recreate-fail removal re-checks existence at removal time. Removal justified by the
  state of the world at removal time, not classification time.
- **B6 (stoppingContainers leak):** the flag's lifetime is now the stop operation's
  lifetime (`finally`-cleared; the die handler no longer deletes it), and the crash-
  recovery stream handles `error`/`end`/`close` through one resubscribe path. We did
  not take a TTL approach deliberately: v9 specs allow graceful stops up to 6 hours, so
  no deadline is both long enough for a legitimate stop and short enough to bound a
  wedge. End-to-end: new suite 45 lands a stop inside the stream-outage window.

## Should-fixes

- **S1** (`258029f1e`): strict flag parsing (`/^[rgs]+$/` wholesale — `'logs'` no longer
  parses as g+s), spawner moved onto the canonical classifier. All 707 on-chain specs
  checked: none has the broken shape. The `peerNotification` substring match stays
  lenient deliberately: already-accepted apps must keep propagating; strict advertising
  would convert "down but counted" into network-wide respawn churn for the
  roundcube-shape victims without healing anything.
- **S4–S8** (`ead99374e`): uninstall clears `controllerDesired` (S4); the controller
  verdict is re-read immediately before g:/r: starts (S5); `zelappsruntimestate` gets
  boundary normalization, field-wise twin merge, and a unique index (S6); the crash
  ladder resets only on a provable stable run via docker's `FinishedAt` (S7 — your
  finding was confirmed exactly as described: the old reset counted time sat stopped in
  backoff, so the 30m cap was unreachable; thank you); failed starts schedule their own
  retry (S8).
- **Minors** (`3e8cdba93`): lock-before-op on component paths, `setOperatorStopped`
  propagates write failures, soft redeploy runs the hard-removal cleanup seam, dead
  code removed, CI cleanup scoped to the harness label, dispatch inputs via `env:`.

## S2/S3 — superseded: we redesigned follower convergence after live testing

Your S2 (stall window too aggressive) was confirmed — but when we took it to live
syncthing v2 instances (a ~90-minute isolated POC across four fleet nodes, same
binaries as production, distinct networks/NATs) the findings overturned the premises
of the old code, of the planned fix, and of parts of the review, so we rebuilt the
layer on evidence rather than patching the window:

- `idle` does NOT mean done: syncthing's own failed-pull retry backoff parks folders at
  idle with `needBytes > 0` and zero error indication in any polled endpoint.
- `inSyncBytes` is block-granular while pulling — flat bytes genuinely means no blocks.
- `db/completion` reports a stale 100% for DISCONNECTED peers — trusting it is the real
  data-loss hazard; `remoteState === 'valid'` is the connectivity discriminator.
- A genuine inert no-retry state exists; the cure is DEVICE pause/resume (~30s). A
  syncthing process restart is cargo-cult for this purpose: ~38s dead time, node-wide
  collateral, no unique benefit (measured).

The shipped design: evidence-gated stall ladder (ACTIVE states always healthy; idle+flat
with a CONNECTED synced peer → device nudges with doubling backoff; removal only after
repeated failed nudges over a minimum window), peer trust requiring
`remoteState 'valid' + completion 100 + globalBytes > 0`, a promotion gate that
`db/revert`s receive-only pollution instead of broadcasting it cluster-wide, the health
monitor demoted to alert+nudge, and an events consumer that accelerates the same
evaluation the poll runs. **S3's direction is REVERSED, deliberately:** the 2.5h
unreadable-status removal is deleted rather than extended — when the folder status is
unreadable the node can verify nothing, and removal justified by blindness would delete
a possibly-good copy. No removal without positive evidence; alert and wait.

## Found by the new integration suites (first-contact runs)

Three more production bugs surfaced and were fixed tests-first:

- The stall-ladder removal passed the component identifier into `removeAppLocally`,
  which routed into a component-scoped removal and left the app's installed-DB row
  behind — a zombie still broadcast as running (`d52f4f95e`).
- The events consumer's 55s long-poll ran over the service's shared 5s-timeout axios
  instance — every quiet poll aborted client-side (`f004cfc10`).
- The consumer detected syncthing restarts via id regression, but the events API never
  returns events below a stale `since` (verified in `lib/events Since()` and live on an
  isolated v2.0.15: stale `since` → `[]` forever even with fresh low-id events present).
  Restarts are observable only as transport errors, so the consumer now re-anchors and
  publishes exactly one resync on recovery (`d05974bfd`).

## Validation at `2239da80a`

- Unit: 4253 passing / 0 failing / 18 pending (canonical `test:zelback:unit`, which now
  actually gates CI after the exit-code fix).
- Integration: nine new/rewritten suites green solo (37, 44, 45, 46, 47, 48, 49,
  28-enterprise, 32-restart-clears-lock), then the full parallel gate: 48/48 suites
  green — 50 suite files, 231 tests (2026-06-12, at `2239da80a`).
- Fleet soak: the multi-day soak on 8 live nodes stands at `6d99dc6a9` (NRestarts=0
  throughout, real customer workloads); a fresh pull soak on the same 8 nodes at this
  head is underway to cover the post-review production changes, and merge waits on it.

Cabecinha84 (commented):
I checked out the current head (7988d0dd7), diffed it against your last-review head (6795b8b7c), and ran five
  independent adversarial passes plus verified the decisive finding in source myself. Since your last review the author
  landed a large second wave (~627 lines of production code across 14 files): the secondary fixes S1/S4–S8/minors, plus a 
  full syncthing convergence redesign (S2/S3 superseded), plus 3 self-found bug fixes.

  What's genuinely good — and verified intact

  All 6 original blockers still hold at the new head. The second wave did not regress any of them — it strengthened a
  couple (lock-moved-before-op, failed-start retry). Safety net confirmed.

  The secondary list you flagged as untouched is now substantially done, and correctly:
  - S1 strict flag parsing (/^[rgs]+$/) + spawner on the canonical classifier — ✅ fixed
  - S4 clearControllerDesired wired into uninstall, key matches the writers — ✅ fixed (was dead)
  - S6 unique index on zelappsruntimestate.identifier — ✅ fixed, and done properly: dedup-before-index collapses
  pre-existing twins so createIndex won't throw, plus a duplicate-key retry and field-wise twin merge that preserves both
  operatorStopped and restartHistory
  - S8 failed start now schedules its own retry, climbs the backoff ladder, no storm — ✅ fixed
  - S7 backoff ladder reset now keyed on docker FinishedAt so the 30m cap is reachable — ✅ fixed
  - Minors (lock-before-op on component paths, setOperatorStopped propagates write failures, soft-redeploy cleanup seam,
  dead-code removal) — ✅ fixed
  - S3 reversed deliberately and correctly: the 2.5h unreadable-status removal is deleted, not extended — no removal
  without positive evidence. The single FSM removal path is now evidence-gated (remoteState==='valid' && completion===100 
  && globalBytes>0), and d52f4f95e correctly fixed the bare-main-name routing so it no longer leaves a zombie DB row.
  
  The author was rigorous — they confirmed your findings against source, corrected two framing points fairly (B1 was a
  faithfully-ported pre-existing dead guard, not a regression they introduced), and tests-first'd everything.

  The one blocker — newly introduced by the redesign

  db/revert promotion gate can delete the last surviving copy when the cluster global is empty.
  syncthingFolderStateMachine.js:509-516

  - getFolderSyncCompletion (line 169/179) computes isSynced = true whenever globalBytes === 0 (empty global → vacuous
  100%).
  - The new gate if (syncStatus.isSynced && syncStatus.receiveOnlyChangedFiles > 0) then calls
  syncthingService.dbRevert(appId), which discards all locally-changed files in the receiveonly folder.
  - Trace: a non-leader node holding real appdata in a receiveonly folder whose syncthing global index is empty/fresh
  (redeploy where appdata persisted but the folder was re-added; restore-into-receiveonly before index exchange; a lone
  node with no peers broadcasting → isDesignatedLeader returns false). globalBytes=0 ⇒ isSynced=true; the on-disk files
  read as receiveOnlyChangedFiles>0 ⇒ dbRevert deletes them against an empty global. If this was the only copy, it's gone.
  - Why I'm confident it's an oversight, not a trade: the author guarded the removal path against this exact empty-global
  trap 130 lines earlier (line 382, with a comment explaining "a peer that synced empty/wrong data… would falsely satisfy
  'peers are synced' and we would remove the good local copy — data loss"). The revert path has no such guard. Same trap,
  one path protected, the sibling not. And this whole dbRevert gate is new in this PR — it does not exist on development —
  so it's a regression in the fix, in precisely the data-safety guarantee the PR is built to provide.
  - Fix is one line: add && syncStatus.globalBytes > 0 to the gate at 509 (and decide the globalBytes===0 + local data
  case is wait, never revert and never promote-to-sendreceive at 518, which has the same vacuous-isSynced exposure).

  Should-fix (worth pulling in or explicitly tracking)

  - S5 residual race (appReconciler.js:404-410): the verdict re-read closes the big window, but one await 
  appsRuntimeState.recordRestart still sits between the re-read and appDockerStart — a start can still land after a
  stop-wrapper's synchronous stop+wipe. Window shrank from minutes to ~one DB write; consequence is still data corruption.
  One more controllerDesired.get immediately before the start (no await between) closes it.
  - Device pause/resume nudge is device-global, not folder-scoped (syncthingService.js systemPause hits
  /rest/system/pause?device=): a nudge briefly drops every folder shared with that peer, and a thrown systemResume leaves
  the peer paused until a later nudge cycle (this app's last nudge before removal won't retry). The incidental
  systemResume endpoint fix (was hitting /pause) is real and correct.
  - Events-consumer stop/start reentrancy (syncthingEventsConsumer.js:136-151): stop() sets running=false immediately but
  the loop is parked in a 65s long-poll, so a start() before it drains defeats the if (running) return guard and spawns a
  second concurrent loop sharing module-level since. Latent in prod (started once at boot), but reachable from the new
  suites' start/stop cycles.

  Minor / by-design (fine to note and move on)

  peerNotification substring includes('g:') kept lenient deliberately (don't convert "down but counted" into respawn
  churn); getBaseAppName non-injectivity is a pre-existing convention limit now load-bearing for the lock (pathological
  component names only); prepareCollection swallows a createIndex failure (low risk given dedup-before-index). The
  events-consumer claims (long-poll timeout 65s > 55s hold, re-anchor fires exactly one resync, no FSM race via the
  isRunning guard) all check out.

MorningLightMountain713:
Thanks for the second pass — and for verifying the 6 original blockers, the S1/S4–S8/minors, and the S3 reversal all hold at the new head. That matches our own read.

You raised one blocker and two should-fixes. All three are addressed below (the blocker structurally, at the source). Each is quoted first, then our response. After that: the additional items our own validation surfaced — including a cold-start regression that the blocker fix itself exposed, which we've fixed — and the validation summary.

---

## Blocker — empty-global `db/revert` can delete the last surviving copy

> db/revert promotion gate can delete the last surviving copy when the cluster global is empty. (`syncthingFolderStateMachine.js:509-516`) … `globalBytes=0 ⇒ isSynced=true`; the on-disk files read as `receiveOnlyChangedFiles>0 ⇒ dbRevert` deletes them against an empty global. If this was the only copy, it's gone. … Fix is one line: add `&& syncStatus.globalBytes > 0` to the gate at 509 (and decide the `globalBytes===0` + local data case is wait, never revert and never promote-to-sendreceive at 518, which has the same vacuous-isSynced exposure).

**Confirmed and fixed — thank you, this was a real last-copy data-loss.** We reproduced it on real syncthing (a receive-only folder with local files + an empty global; `POST /rest/db/revert` physically deletes them).

Rather than guard the two call sites, we fixed it at the **source** so neither the revert (509) nor the promote (518) — both of which you correctly flagged — can ever treat an empty global as synced:

```js
// getFolderSyncCompletion
isSynced: globalBytes > 0 && syncPercentage === SYNC_COMPLETE_PERCENTAGE,
```

`isSynced` is consumed in exactly those two places, so this single change covers both the data-loss (revert) and the cluster-pollution (empty promote) directions, and an empty/unknown global now falls through to the existing **wait** branch — "no evidence", not "done" — which is the posture you prescribed. The legitimate leader/cold-start seed path is handled separately and is unaffected. Pinned with unit tests and a live-fleet data-safety gate.

## Should-fix — residual start-into-wipe race (`appReconciler.js:404-410`)

> the verdict re-read closes the big window, but one `await appsRuntimeState.recordRestart` still sits between the re-read and `appDockerStart` — a start can still land after a stop-wrapper's synchronous stop+wipe. Window shrank from minutes to ~one DB write; consequence is still data corruption. One more `controllerDesired.get` immediately before the start (no await between) closes it.

**Fixed — structurally, which removes the window entirely rather than shrinking it further.** The root cause is that the wipe ran *outside* the reconciler's per-key single-flight (a second actor mutating the container), so any await before the start re-opens a race. Instead of adding a second synchronous re-read, we made the sync layer **declare-only**: the first-run/new-app reset (stop + appdata wipe) is now a declared desired *data-state* actuated by the reconciler **inside** its single-flight, before the run decision. The reconciler is now the sole caller of the appdata wipe. With the second actor gone, there is no longer a wipe that a start can race — the `recordRestart` window you identified is moot by construction.

## Should-fix — device-global nudge; a thrown resume strands the peer

> Device pause/resume nudge is device-global, not folder-scoped … a nudge briefly drops every folder shared with that peer, and a thrown `systemResume` leaves the peer paused until a later nudge cycle (this app's last nudge before removal won't retry). The incidental `systemResume` endpoint fix (was hitting `/pause`) is real and correct.

**Fixed.** `nudgeFolderDevices` now resumes each device in a `try/finally`, so a throw between pause and resume can no longer strand a peer paused; a failed resume is logged at `error`. You're right that the scope is **device-global** — syncthing has no folder-scoped pause — so we reworded the "surgical/folder-scoped" docstrings to say plainly that the nudge briefly pauses every folder shared with that peer. We did not add a separate resume-sweep before removal: the `finally` already guarantees a resume before the function returns, and the removal gate and the nudge are mutually exclusive per cycle. (Thanks for confirming the `/pause`→`/resume` endpoint fix.)

---

## Additionally found during our own validation

Running our own adversarial passes, a live-syncthing POC, and the multi-node integration gate surfaced these. The first is the most important — it is a regression the blocker fix above *exposed*:

- **Cold-start election regression exposed by the blocker fix — FIXED.** Closing the empty-global=synced hole removed a behaviour that had been silently doubling as the **cold-start bootstrap**: a fresh follower used to promote on its empty global and start. With that (correctly) gone, a brand-new multi-instance r:/g: app placed on several nodes at once — with **no node yet holding the data** — could fail to elect a seed at all. Every holder broadcasts a placement timestamp the moment it is *assigned* the app (placement, not liveness), and the leader election deferred to any such peer, so every node deferred to every other and **nobody seeded** (a cold-start standoff; on the g: path masterSlave then waits forever for a folder that never reaches sendreceive). The fix, in three parts: (1) the election defers only to a peer that **genuinely serves the data**, not one that merely advertises placement; (2) when it must pick a seed it does so by **IP deterministically**, so every node agrees on one seed (no split-brain — the prior re-broadcast-timestamp ordering could differ per node); (3) a node may only seed a **genuinely empty** folder — a node holding data waits, never promoting/reverting on an empty global, preserving the blocker fix above. This makes the election correct for a healthy network; it is not a consensus protocol, and a partition-proof election (a deterministic candidate + a data-aware quorum lease) is tracked as a follow-up. Covered by new unit tests and a new fresh-cold-start integration suite (r: and g:), with the empty-global data-safety suite green throughout.
- **Events-consumer `stop()`/`start()` reentrancy** (latent, not prod-reachable) — rebuilt as a correct idempotent lifecycle so `stop()` actually aborts the in-flight long-poll and a later `start()` spawns exactly one loop. Fixed.
- **Crash-recovery double-stripped the component prefix** — deleted a byte-identical private duplicate of the shared name helper; the boundary now strips once. Behaviour-neutral. Fixed.
- **Stall ladder didn't reset one of its timers on progress** — reset it with its siblings (behaviour-neutral coherence). Fixed.
- **Event-stream gap detector** — we suspected the filtered subscription could false-fire the gap check, but **refuted it against syncthing source**: the event `id` is per-subscription (contiguous), not global, so the check fires only on genuine loss. Left as-is (removing it would silence real losses).
- **`recreateMissing` backoff pacing** — implemented then reverted: routing the missing-container recreate through the crash ladder also gates the recreate→fail→uninstall escape hatch (a vanished-but-previously-healthy container backed off instead of recreating, so the unpullable→uninstall path never fired). The concern is low-reachability and self-limiting (no self-sustaining retry; first vanish recreates immediately; a crashed recreate lands on the paced path; a persistent failure uninstalls), so we kept immediate-recreate.
- **`dockActual` double container-list** — confirmed, deferred to the v9 docker rework (the clean fix is name-based, daemon-authoritative existence across the docker layer; out of scope here).
- **Strict primary-flag regex migration edge** — scanned the entire live global app corpus through the real parser: **zero affected apps**.
- The minor `includes('g:')`/`getBaseAppName`/swallowed-`createIndex` items — verified safe by design (the substring only ever *adds* to the broadcast set; the validator forbids the names that would make `getBaseAppName` ambiguous; the index is self-healing on next boot).

## Validation

- **Unit:** full canonical suite green (now actually enforced in CI after the exit-code fix), including new cases for the empty-global guard and for the cold-start election (seeds exactly one node, defers to a genuinely-serving peer, deterministic by IP, never seeds while holding data).
- **Integration — coverage extended this round, all green:** the full multi-node gate passes with **zero failures** — every reconciler/syncthing suite. New end-to-end gates were added for the work above: a **data-safety gate for the blocker** (a node must never `db/revert` or promote against an empty global when it holds local data), and a **fresh-cold-start gate** (a multi-instance r: and g: app placed with no source must elect exactly one seed — RED before the cold-start fix, GREEN after). Just as importantly, where the existing multi-node suites had implicitly leaned on the old empty-global=synced behaviour to stand up a running peer, we **corrected the harness** so they now validate against the *fixed* production behaviour rather than the old bug — i.e. the suites were strengthened, not relaxed, to keep passing.
- **Soak:** the branch is live-soaking on the test fleet at this head (first recon clean — no unwanted actuation, no spurious leader/nudge/removal). We will not propose merge until that soak is verified over a meaningful window.

Cabecinha84 (commented):
PR #1755 — re-review verdict: ACKABLE (two cheap should-fix items worth pulling in, neither blocking)

  I checked out the current head (22366de26), read the changed code directly, ran four independent adversarial 
  verification passes (S5 race, L1 lifecycle, the new cold-start data-safety core, and a regression sweep of all prior
  fixes), verified the decisive finding in source myself, and ran the unit suites for the changed files (146 passing).
  Tree restored to development.

  Everything you and the prior passes flagged is genuinely fixed

  ┌──────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────┐
  │                        Area                        │                        Status at head                        │
  ├────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  │ 6 original blockers (backup guard, enterprise      │                                                              │
  │ sweep, boot-broadcast wipe, appRestart lock,       │ ✅ all still intact — not regressed by the second wave       │
  │ vanished-classification, leaked stopping flag)     │                                                              │
  ├────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  Status at head: ✅ all fixed (confirmed in source)   
  ────────────────────────────────────────             
  Area: B1 — db/revert against an empty global deletes the only copy (the blocker from your last pass)
  Status at head: ✅ fixed: isSynced = globalBytes > 0 && 100%, so both the dbRevert gate and the promote gate are    
    structurally unreachable on an empty global
  ────────────────────────────────────────
  Area: L1 — events-consumer reentrancy
  Status at head: ✅ fixed: stop() aborts the long-poll and awaits loop exit via FluxController; running stays asserted
    through the whole stop, so a second loop is unreachable. Abort is correctly classified as a clean exit, not a poll
    failure                                            
  ────────────────────────────────────────
  Area: S5 — start-into-wipe data race                 
  Status at head: ✅ closed: the syncthing layer is now declare-only (requestStopAndClearData / setControllerDesired); the

    reconciler is the sole caller of appDeleteDataInMountPoint, actuating stop→wipe→(only then)run inside its per-key
    single-flight. The 3 imperative callbacks are gone with no dangling wiring

  The B1 fix chain is notably careful and self-correcting: fix B1 → it exposed a cold-start standoff → add seed election →
  found a broadcastedAt split-brain → switched to IP-only → realized the election could re-trigger B1 → gated on
  folderIsEmpty. The author confirmed findings against source and tests-first'd everything.

  Two should-fix items (recommend pulling in; not blocking)

  1. Cold-start seed guard doesn't fully match its own stated invariant — syncthingFolderStateMachine.js:489–516, 206–244.
  The seed block flips to sendreceive + starts based purely on isDesignatedLeader, whose "don't seed while holding data /
  while a peer serves" protection is enforced only through the runningPeers proxy (location-DB runningSince rows), not
  through the aPeerHasData/folderIsEmpty syncthing signals it already computed. So if runningPeers is empty while the node
  holds data, the seed can fire — contradicting the comment's claim that it "only seeds when this node holds no data."
  Reachability is low (normal placement and running broadcasts both stamp runningSince, and an empty runningAppList
  returns false), and the worst case only widens the already-documented version-unaware-seed residual — it isn't worse
  than development. Cheap hardening: gate the seed block directly on folderIsEmpty && !aPeerHasData. Worth doing because
  it's the data-safety core and makes the code match the comment.
  2. Clear-data actuation failure has no retry — appReconciler.js:382–397. If appDockerStop/appDeleteDataInMountPoint
  throws, dataDesired stays 'clear', controllerDesired stays 'stopped', no scheduleRetry is armed, and the syncthing
  caller already set its dedupe cache — so the app can sit down until the hourly sweep (~1h). It's fail-safe in the data 
  direction (a failed wipe never proceeds to a start; re-wipe is idempotent). One scheduleRetry(...) on the throw path
  closes it.

MorningLightMountain713:
Thanks for the re-review and the ACK — and for independently confirming at this head that the six original blockers, B1, L1 and S5 all still hold. That matches our own read.

You raised two should-fixes. Both are addressed below — each quoted first, then our response — followed by a short validation note.

---

## Should-fix — clear-data actuation failure has no retry

> Clear-data actuation failure has no retry — `appReconciler.js:382–397`. If `appDockerStop`/`appDeleteDataInMountPoint` throws, `dataDesired` stays 'clear', `controllerDesired` stays 'stopped', no `scheduleRetry` is armed, and the syncthing caller already set its dedupe cache — so the app can sit down until the hourly sweep (~1h). It's fail-safe in the data direction (a failed wipe never proceeds to a start; re-wipe is idempotent). One `scheduleRetry(...)` on the throw path closes it.

**Confirmed and fixed — thank you; this was the one actuation path here that lacked the retry every sibling path has.** The stop→settle→wipe sequence is now wrapped, and a thrown stop or wipe arms its own paced retry (the same near-term cadence the failed-start path uses) instead of dropping to the hourly resync.

We kept the fail-safe direction you noted, deliberately: on the failure path the wipe intent stays pending (the desired data-state is left `clear`) and we do **not** emit the "cleared" signal, so the retried reconcile re-runs the idempotent wipe and — because this block still resolves before the run decision — a start can never proceed onto un-wiped data. Pinned with unit tests: a thrown wipe arms a retry (re-attempts within the retry window rather than ~1h later), and the clear stays pending so a later reconcile re-wipes and never starts the container, even against a contradictory "running" verdict.

## Should-fix — cold-start seed guard doesn't fully match its stated invariant

> Cold-start seed guard doesn't fully match its own stated invariant … The seed block flips to sendreceive + starts based purely on `isDesignatedLeader`, whose "don't seed while holding data / while a peer serves" protection is enforced only through the `runningPeers` proxy …, not through the `aPeerHasData`/`folderIsEmpty` syncthing signals it already computed. So if `runningPeers` is empty while the node holds data, the seed can fire — contradicting the comment's claim that it "only seeds when this node holds no data." … Cheap hardening: gate the seed block directly on `folderIsEmpty && !aPeerHasData`.

**You're right about the defect, and we've corrected the code to tell the truth about it — but we deliberately did not apply the literal one-line gate, because the root cause is architectural, not a local oversight.**

The observation is exactly right: the "holds data → don't seed" protection is enforced only through the running-peer proxy (it fires only when some peer carries a placement marker), so with no running peer a data-holding node can still win the election and seed. The in-code comment overstated that as an absolute guarantee; that was the real bug, and we've reworded it to state precisely what is and isn't enforced.

We did not take `folderIsEmpty && !aPeerHasData` as a gate on the seed block because it regresses a legitimate case into a hard failure: a genuine **sole-source / single-node** deployment — one node holding the only copy, with no peer to sync from — *must* seed, or the app never starts at all (that is the exact cold-start case this work exists to enable). `folderIsEmpty` is also false whenever the folder status is unreadable (we treat unknown as "holds data"), so the literal gate would additionally block that case. In short, the one-liner trades a documented, low-reachability residual for a permanent hang of legitimate single-node apps.

The deeper point — and the reason a local guard isn't the right fix: **this election is a heuristic, not consensus.** It elects by gossip (the `apprunning` view) and lowest-IP, so it can guarantee neither a single master under partition (the split-brain this path already had to be made IP-only to avoid) nor that whoever seeds actually holds the newest data. The residual the code flags — a fresh seed winning over a disconnected peer's newer copy — is inherent to electing by reachability instead of by data, and no purely-local guard can close it (it either reintroduces the cold-start standoff or still can't see the offline newer copy).

Properly closing it is a redesign we plan to land in a future PR: a **consensus-grounded election** — a deterministic candidate over the confirmed on-chain node set, plus a **data-aware quorum lease**. The candidate only serves once a **strict majority of the app's assigned host set** confirms it, with the data-version check folded into that quorum (if any member reports a newer version it syncs up first and promotes only once caught up); a minority partition can't form a majority, so it serves nothing rather than risk a second master (the CP trade), and a time-bounded lease + epoch fencing keep an old and new master from ever overlapping and fence out a returning stale ex-master. The same mechanism unifies r: and g: and demotes FDM from election authority to a routing consumer. It closes both the split-brain and the stale-seed residual coherently — which a one-line local patch can't — so for this PR we've made the code honest about the limitation and deferred the fix to that redesign. Reachability of the current residual is low (every running node broadcasts a runningSince marker, so an empty running-peer set means the node is effectively alone), and the worst case is the same version-blind overwrite already documented — no worse than the prior behaviour.

---

## Validation

- **Unit:** the changed files' suites are green, including the new clear-data retry cases (retry armed on a thrown wipe; the clear stays pending and never starts on un-wiped data). The seed-guard change is comment-only and behaviour-neutral — the empty-global data-safety and cold-start election suites are unchanged and remain green.
- The branch continues to soak on the test fleet; we won't propose merge until that's verified over a meaningful window.

Cabecinha84 (approved):
ack

PR #1756, opened 2026-06-14
feat(info): expose connected MongoDB server version on /flux/info

MorningLightMountain713:
## What

Adds the connected MongoDB server version to `GET /flux/info` as `data.flux.mongoDbVersion`, alongside the existing Docker and Syncthing versions. This was the one backing-service version FluxOS never surfaced.

## How

- `dbHelper.js`: reads the server version once per connection via a `buildInfo` command and caches it in a new `getMongoDbVersion()`. The read is **informational** — it swallows its own errors and returns `null`, so it can never turn `/flux/info` into an error response. The cache is warmed on the initial connect (`initiateDB`) and cleared on close so a reconnect re-reads it.
- `fluxService.getFluxInfo()`: surfaces `info.flux.mongoDbVersion`, grouped with the other backing-service versions.

`buildInfo` is the lowest-privilege, most universally available source for the human-readable version (the driver handshake only exposes the wire-protocol version). It works on MongoDB 3.6+, so it succeeds on every node the driver can already talk to.

No standalone endpoint was added: like Docker and Syncthing, MongoDB is a backing service, so it belongs only in the consolidated `/flux/info` rather than minting a new per-field route.

## Testing

- New unit tests in `tests/unit/dbHelper.test.js` cover `getMongoDbVersion()` (returns `null` when unconnected; semver string when connected).
- Existing `dbHelper` (34) and `getFluxInfo` (14) unit suites pass with no new lint errors.
- Validated end-to-end against a live node running MongoDB: `/flux/info` reports `flux.mongoDbVersion` correctly (e.g. `8.0.20`).

### DIFF
--- a/ZelBack/src/services/dbHelper.js
+++ b/ZelBack/src/services/dbHelper.js
@@ -18,6 +18,12 @@ const mongoUrl = `mongodb://${config.database.url}:${config.database.port}/`;
 let openDBConnection = null;
 
 /**
+ * Cached MongoDB server version, populated once per connection.
+ * @type {string | null}
+ */
+let mongoDbVersion = null;
+
+/**
  * Returns MongoDB connection, if it was initiated before, otherwise returns null.
  *
  * @returns {mongodb.MongoClient | null}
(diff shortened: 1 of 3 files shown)


Cabecinha84 (approved):
ack

PR #1757, opened 2026-06-15
refactor: drop all zelAppSpecifications legacy references (supersedes #1717)

MorningLightMountain713:
## What

Removes the `zelAppSpecifications` legacy field entirely. The field was renamed to `appSpecifications` in Feb 2021, but read paths still carried an `appSpecifications || zelAppSpecifications` fallback and `explorerService` still created the legacy `zelAppSpecifications.*` indexes.

This **supersedes #1717**, which was branched before the `appMessaging` refactor (`messageStore` / `messageVerifier` / `appHashSyncService` split, `getPreviousAppSpecifications` moving to `registryManager`). Rebasing it produced conflicts against code that had since been rewritten, so the change was rebuilt fresh on `development`.

## Changes

- **Hard cutoff** — removed every `appSpecifications || zelAppSpecifications` read fallback (`registryManager`, `appUninstaller`, `appHashSyncService`, `messageStore`, `messageVerifier`, `appQueryService`, `fluxCommunicationMessagesSender`), including the now-dead legacy-key scan block in `getPreviousAppSpecifications` and the legacy `$or` branch in `appQueryService`.
- **Legacy indexes** — dropped the three `zelAppSpecifications.*` `createIndex` calls.
- **One-time migration** — `$rename` of `zelAppSpecifications` → `appSpecifications` plus a legacy-index drop, run once per process in `initiateBlockProcessor`, gated by a module flag and wrapped in `try/catch` so a malformed or fresh-sync DB can't stall block-processor init. Idempotent; safe to remove along with the flag once the fleet has run it.

## Why this is safe

- The message hash/signature is computed over the spec object's **value**, not the wrapper key — so the field name is freely renameable without invalidating any hash or signature.
- Inbound messages are already normalised to `appSpecifications` at ingestion (`storeAppTemporaryMessage`), and the permanent-message store already requires `appSpecifications`.
- Verified on the live network that stored permanent messages no longer carry the legacy field; the migration covers any node still holding pre-2021 records at rest.

## Addresses review feedback from #1717

1. **try/catch around the migration** — `col.indexes()` can throw on a never-created collection; the call is wrapped and logs on failure.
2. **Peer wire compatibility** — confirmed (see "Why this is safe"): no path still emits the legacy field, and the value-based hash makes the rename transparent.
3. **Cleanup follow-up** — the migration function and gating flag are commented as safe to delete once the fleet has fully upgraded.

## Testing

- Unit suites green for every touched module: `explorerService`, `registryManager`, `appUninstaller`, `appHashSyncService`, `messageStore`, `messageVerifier`, `appQueryService`, `fluxCommunicationMessagesSender`.
- Lint clean on all changed files (no new findings).

_PR description written against commit `ee594bdf0`._

### DIFF
--- a/ZelBack/src/services/appDatabase/registryManager.js
+++ b/ZelBack/src/services/appDatabase/registryManager.js
@@ -1998,7 +1998,7 @@ async function rescanGlobalAppsInformation(height = 0, removeLastInformation = f
 
     // eslint-disable-next-line no-restricted-syntax
     for (const message of results) {
-      const updateForSpecifications = message.appSpecifications || message.zelAppSpecifications;
+      const updateForSpecifications = message.appSpecifications;
       updateForSpecifications.hash = message.hash;
       updateForSpecifications.height = message.height;
       // eslint-disable-next-line no-await-in-loop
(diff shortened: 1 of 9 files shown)


Cabecinha84 (approved):
ack

PR #1761, opened 2026-07-03
fix: FluxOS-owned app volume mounting, syncthing mount-safety hardening, and enterprise app lifecycle fixes

MorningLightMountain713:
This PR remediates the root causes of a production data-loss incident involving a two-instance syncthing-replicated (`g:`) application, plus additional defects found while building integration coverage for the fixes.

**The problems being solved**

1. **After a reboot, nothing reliably re-mounted an app's data volume.** Remounting was delegated to a root `@reboot` crontab entry that FluxOS itself could delete and that nothing ever re-asserted — a node could come back broadcasting an app as "running" while its data volume was never mounted.
2. **An unmounted app directory silently accepted writes.** The bare mountpoint was an ordinary writable directory, so container binds and syncthing pulls landed on the host filesystem (bypassing the app's quota), and a syncthing folder left `sendreceive` with a stale index could broadcast the "missing" files as deletions — wiping the healthy replica.
3. **A single transient UPnP port-mapping failure force-removed the entire app** and broadcast the removal network-wide, triggering a re-placement that came up empty.
4. **Enterprise apps were invisible to the startup crontab/mount inventory** (their components live in the encrypted blob), so every FluxOS restart deleted their remount entries; separately, their local-install path lost `height`/`hash`, which the expiry sweep eventually punishes with a forced, broadcast removal.
5. **Volume deletion at uninstall depended on parsing the crontab entry** — when the entry was missing, the multi-GB volume image was orphaned on disk.

**Volume mounting (the core change)**
- FluxOS now owns app data-volume mounting end to end: volumes are mounted from the installed-apps database and deterministic image discovery at startup, by the reconciler before any actuation, and by the syncthing monitor's repair path. The legacy `@reboot` crontab entries are no longer created; existing entries are removed only once the FluxOS-owned mount demonstrably covers them (an entry whose volume cannot mount is kept as the remaining safety net).
- Volume deletion at uninstall derives the image path deterministically instead of parsing the crontab, eliminating orphaned volume images when entries are missing.
- An unmounted app directory is now physically inert: the bare mountpoint is set immutable (`chattr +i`), so writes while the volume is unmounted fail loudly instead of silently landing on the host filesystem (and bypassing the app's quota).

**Syncthing mount-safety**
- The mount-safety guard gates on the mountpoint, not directory content: a `sendreceive` folder over an unmounted directory is demoted to `receiveonly` and its container held, regardless of content. Content on a bare directory means data already leaked to the host filesystem — it must never buy a pass (in the incident, a stale `sendreceive` folder broadcast deletions that gutted the healthy replica).
- A stale ("phantom") index claiming data over an empty mounted volume is demoted before it can broadcast the missing files as deletions.
- Promotion to `sendreceive` — including the cold-start leader seed — requires that same verification to pass *before* the flip, so a folder whose index claims data the disk does not hold (or whose directory is not a mountpoint) is never promoted at all; the per-cycle check above covers folders that degrade while already `sendreceive`. A legitimately empty cold-start seed (an empty index over an empty disk) still seeds.
- The demotion also runs from the monitor's skip path (previously unreachable when the mount was unrepairable), and the `.stfolder` marker can only ever be created inside a mounted volume.
- The reconciler now honors a pending stop even while a volume is unavailable — previously the volume-unavailable defer blocked the very stop the mount-safety hold requested, leaving the container running over a missing volume.

**Enterprise app lifecycle**
- Enterprise apps store `compose: []` locally (components live in the encrypted blob). The crontab/mount inventory never decrypted, so every FluxOS restart classified enterprise apps as uninstalled and deleted their remount entries. The inventory now decrypts like every other consumer, with a decryption-free fallback that derives component ids from the volume-image filenames; a failed database enumeration aborts the sweep instead of reading as "nothing installed".
- The local-install path preserved neither `height` nor `hash` through the enterprise decrypt+format step; since the expiry sweep force-removes any installed app without a height, locally-installed enterprise apps were eventually removed (with broadcast). Both fields now survive, mirroring the existing pattern in `registryManager`.

**UPnP**
- `restoreAppsPortsSupport` no longer force-removes (and broadcasts removal of) an app on a single UPnP mapping failure; removal requires sustained failure across multiple cycles and a minimum monotonic duration.

**Tests**
- Unit coverage across the touched services (crontab/mount cleanup, reconciler, installer, port manager, syncthing state machine), with the key guards mutation-verified.
- Two new integration suites: `60-volume-mount-ownership` (install/reboot/uninstall/enterprise volume lifecycle, immutable-mountpoint enforcement, legacy-entry handling) and `61-syncthing-mount-safety` (demotion of unsafe `sendreceive` folders, phantom-index handling, cold-start non-demotion, marker discipline).
- Harness hardening: hermetic image registry (no live-internet dependencies; enforced at spec construction), deterministic seeded-sync state (a folder pinned synced holds matching data on disk), stall-tolerant confirmation-staleness defaults with the fast windows scoped to the suites that test those flows, and teardown that cannot leak volumes.
- The **full integration harness — all 54 suites — has been run against this branch and passes in its entirety** (including the two new suites).

Written against `57dc362dd`.

### DIFF
--- a/ZelBack/src/services/appLifecycle/advancedWorkflows.js
+++ b/ZelBack/src/services/appLifecycle/advancedWorkflows.js
@@ -1,9 +1,9 @@
 const config = require('config');
 const util = require('util');
 const df = require('node-df');
+const fs = require('node:fs');
 const path = require('node:path');
 const nodecmd = require('node-cmd');
-const systemcrontab = require('crontab');
 const axios = require('axios');
 const dbHelper = require('../dbHelper');
 const log = require('../../lib/log');
@@ -25,6 +25,8 @@ const {
   globalAppsMessages,
   globalAppsLocations,
   appsFolder,
+  appVolumesPath,
+  legacyAppVolumesPath,
 } = require('../utils/appConstants');
 const { specificationFormatter } = require('../utils/appSpecHelpers');
 const { checkAndDecryptAppSpecs } = require('../utils/enterpriseHelper');
@@ -48,8 +50,19 @@ const timeTostartNewMasterApp = new Map();
 
 // Promisified functions
 const cmdAsync = util.promisify(nodecmd.run);
-const crontabLoad = util.promisify(systemcrontab.load);
-const fluxDirPath = process.env.FLUXOS_PATH || path.join(process.env.HOME, 'zelflux');
+
+/**
+ * Runs a command as root via execFile (no shell, args passed as params) and
+ * throws on failure, preserving the throw-to-cleanup flow of volume
+ * construction.
+ * @param {string} cmd Binary to run.
+ * @param {string[]} params Arguments.
+ * @returns {Promise<void>}
+ */
+async function execAsRoot(cmd, params) {
+  const result = await serviceHelper.runCommand(cmd, { runAsRoot: true, params });
+  if (result.error) throw result.error;
+}
 
 // We need to avoid circular dependency, so we'll implement getInstalledAppsForDocker locally
 // eslint-disable-next-line no-unused-vars
@@ -493,14 +506,15 @@ async function createAppVolume(appSpecifications, appName, isComponent, res) {
       if (res.flush) res.flush();
     }
 
-    let execDD = `sudo fallocate -l ${appSpecifications.hdd}G ${useThisVolume.mount}/${appId}FLUXFSVOL`; // eg /mnt/sthMounted
+    // volume image path: at the root of the chosen host volume, or in the
+    // appvolumes directory when the root filesystem hosts it
+    let volumeFile = path.join(useThisVolume.mount, `${appId}FLUXFSVOL`);
     if (useThisVolume.mount === '/') {
-      const execMkdir = `sudo mkdir -p ${fluxDirPath}appvolumes`;
-      await cmdAsync(execMkdir);
-      execDD = `sudo fallocate -l ${appSpecifications.hdd}G ${fluxDirPath}appvolumes/${appId}FLUXFSVOL`; // if root mount then temp file is /flu/appvolumes
+      await execAsRoot('mkdir', ['-p', appVolumesPath]);
+      volumeFile = path.join(appVolumesPath, `${appId}FLUXFSVOL`);
     }
 
-    await cmdAsync(execDD);
+    await execAsRoot('fallocate', ['-l', `${appSpecifications.hdd}G`, volumeFile]);
     const allocateSpace2 = {
       status: 'Space allocated',
     };
@@ -518,11 +532,7 @@ async function createAppVolume(appSpecifications, appName, isComponent, res) {
       res.write(serviceHelper.ensureString(makeFilesystem));
       if (res.flush) res.flush();
     }
-    let execFS = `sudo mke2fs -t ext4 ${useThisVolume.mount}/${appId}FLUXFSVOL`;
-    if (useThisVolume.mount === '/') {
-      execFS = `sudo mke2fs -t ext4 ${fluxDirPath}appvolumes/${appId}FLUXFSVOL`;
-    }
-    await cmdAsync(execFS);
+    await execAsRoot('mke2fs', ['-t', 'ext4', volumeFile]);
     const makeFilesystem2 = {
       status: 'Filesystem created',
     };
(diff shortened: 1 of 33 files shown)


Cabecinha84 (approved):
I found no correctness bugs. The shell→execFile/runCommand conversion is also a genuine security improvement (removes shell-string-interpolation injection surface in the volume paths), and path.join fixes the old string-concat that glued appvolumes onto the flux dir.

  Non-blocking observations

  None of these should hold up the merge; worth raising as comments or follow-ups:

  1. Broadened reconciler scope (the one to be aware of). reconcile() now defers all actuation for every app until its volume is mounted — previously this coupling effectively only mattered for g:/r: apps. It's well-guarded (paced 30s retry, honors a pending stop, one-shot tampering event), and discovery
  mirrors creation, so an app should only stay inert when its backing filesystem genuinely isn't up (the intended behavior). The residual risk is any historical volume-placement layout not covered by df(/dev,non-loop,non-boot) + appVolumesPath + legacyAppVolumesPath. Worth a one-line confirmation from the
  author that no legacy layout is missed.
  2. portManager serial retry delay. On a wide UPnP outage, restoreAppsPortsSupport adds a serial delay(30s) per failing app within a single cycle (N apps → up to N×30s). It only slows an already-degraded, non-data-loss path, but could be parallelized or capped.
  3. Cosmetic dead branch in getVolumeFilePath/getComponentAppIdsFromVolumeFiles: the (filesystem.includes('loop') && mount === '/') eligibility clause is always excluded by the following mount !== '/' guard (root is handled by the appVolumesPath fallback). Harmless, slightly confusing.

  Verdict: ACK / approve ✅

  Rationale: it fixes a confirmed production data-loss incident at the root cause; the wiring is verified correct with no bugs found; it reduces attack surface; CI is green; and it ships heavy test coverage (mutation-verified unit tests + two new integration suites, with the author reporting the full 54-suite
  gate passing). The concerns above are awareness/optional items, not defects.

MorningLightMountain713:
Thanks for the review — taking the three observations in order:

> **1. Broadened reconciler scope.** [...] The residual risk is any historical volume-placement layout not covered by df(/dev,non-loop,non-boot) + appVolumesPath + legacyAppVolumesPath. Worth a one-line confirmation from the author that no legacy layout is missed.

Confirmed — no layout is missed. The volume-eligibility filter has been byte-identical since FLUXFSVOL placement was introduced, and images have only ever been written to (a) the root of a chosen non-root eligible mount, covered by the same df sweep, or (b) the appvolumes directory when the root filesystem was chosen — covered in both its spellings, since `appVolumesPath` and the string-glued `legacyAppVolumesPath` are both explicit candidates computed from the same inputs. A bare `/` placement never existed; root was redirected into appvolumes from the first version. If df itself fails, discovery degrades to the appvolumes candidates and the reconciler defers with paced retries — deferral, not loss. Also verified empirically on live nodes (one Arcane, one legacy): every FLUXFSVOL image present on disk falls inside the discovery candidate set, and the df-eligible mounts match the placement filter exactly.

> **2. portManager serial retry delay.** On a wide UPnP outage, restoreAppsPortsSupport adds a serial delay(30s) per failing app within a single cycle (N apps → up to N×30s). It only slows an already-degraded, non-data-loss path, but could be parallelized or capped.

Agreed. Router operations need to stay serial (consumer IGDs mishandle concurrent SOAP), so rather than parallelizing, 60e7e4da8 caps it: the 30s retry pause is now paid at most once per cycle, shared across apps — every failing mapping still gets its retry attempt, and strike/removal hysteresis is unchanged. A wide outage now costs ~30s plus per-app timeouts instead of N×30s, which also keeps the cycle inside the 10-minute restore interval. This is a quick mitigation: the proper treatment — distinguishing gateway-wide failure from per-app mapping refusal, with a node-level response — belongs to the upcoming UPnP/nat-upnp rework and will be handled there.

> **3. Cosmetic dead branch in getVolumeFilePath/getComponentAppIdsFromVolumeFiles:** the (filesystem.includes('loop') && mount === '/') eligibility clause is always excluded by the following mount !== '/' guard (root is handled by the appVolumesPath fallback). Harmless, slightly confusing.

Agreed, removed in 60e7e4da8 — collapsed to a single conjunction. Behavior is identical (root-hosted images are found via the explicit appvolumes candidates), verified against live df output on an Arcane node and a legacy node: old and new candidate sets are byte-identical and every on-disk FLUXFSVOL image remains discoverable.

PR #1762, opened 2026-07-03
v8.16.0

Cabecinha84:
This PR remediates the root causes of a production data-loss incident involving a two-instance syncthing-replicated (g:) application, plus additional defects found while building integration coverage for the fixes.

The problems being solved

After a reboot, nothing reliably re-mounted an app's data volume. Remounting was delegated to a root @reboot crontab entry that FluxOS itself could delete and that nothing ever re-asserted — a node could come back broadcasting an app as "running" while its data volume was never mounted.
An unmounted app directory silently accepted writes. The bare mountpoint was an ordinary writable directory, so container binds and syncthing pulls landed on the host filesystem (bypassing the app's quota), and a syncthing folder left sendreceive with a stale index could broadcast the "missing" files as deletions — wiping the healthy replica.
A single transient UPnP port-mapping failure force-removed the entire app and broadcast the removal network-wide, triggering a re-placement that came up empty.
Enterprise apps were invisible to the startup crontab/mount inventory (their components live in the encrypted blob), so every FluxOS restart deleted their remount entries; separately, their local-install path lost height/hash, which the expiry sweep eventually punishes with a forced, broadcast removal.
Volume deletion at uninstall depended on parsing the crontab entry — when the entry was missing, the multi-GB volume image was orphaned on disk.
Volume mounting (the core change)

FluxOS now owns app data-volume mounting end to end: volumes are mounted from the installed-apps database and deterministic image discovery at startup, by the reconciler before any actuation, and by the syncthing monitor's repair path. The legacy @reboot crontab entries are no longer created; existing entries are removed only once the FluxOS-owned mount demonstrably covers them (an entry whose volume cannot mount is kept as the remaining safety net).
Volume deletion at uninstall derives the image path deterministically instead of parsing the crontab, eliminating orphaned volume images when entries are missing.
An unmounted app directory is now physically inert: the bare mountpoint is set immutable (chattr +i), so writes while the volume is unmounted fail loudly instead of silently landing on the host filesystem (and bypassing the app's quota).
Syncthing mount-safety

The mount-safety guard gates on the mountpoint, not directory content: a sendreceive folder over an unmounted directory is demoted to receiveonly and its container held, regardless of content. Content on a bare directory means data already leaked to the host filesystem — it must never buy a pass (in the incident, a stale sendreceive folder broadcast deletions that gutted the healthy replica).
A stale ("phantom") index claiming data over an empty mounted volume is demoted before it can broadcast the missing files as deletions.
Promotion to sendreceive — including the cold-start leader seed — requires that same verification to pass before the flip, so a folder whose index claims data the disk does not hold (or whose directory is not a mountpoint) is never promoted at all; the per-cycle check above covers folders that degrade while already sendreceive. A legitimately empty cold-start seed (an empty index over an empty disk) still seeds.
The demotion also runs from the monitor's skip path (previously unreachable when the mount was unrepairable), and the .stfolder marker can only ever be created inside a mounted volume.
The reconciler now honors a pending stop even while a volume is unavailable — previously the volume-unavailable defer blocked the very stop the mount-safety hold requested, leaving the container running over a missing volume.
Enterprise app lifecycle

Enterprise apps store compose: [] locally (components live in the encrypted blob). The crontab/mount inventory never decrypted, so every FluxOS restart classified enterprise apps as uninstalled and deleted their remount entries. The inventory now decrypts like every other consumer, with a decryption-free fallback that derives component ids from the volume-image filenames; a failed database enumeration aborts the sweep instead of reading as "nothing installed".
The local-install path preserved neither height nor hash through the enterprise decrypt+format step; since the expiry sweep force-removes any installed app without a height, locally-installed enterprise apps were eventually removed (with broadcast). Both fields now survive, mirroring the existing pattern in registryManager.
UPnP

restoreAppsPortsSupport no longer force-removes (and broadcasts removal of) an app on a single UPnP mapping failure; removal requires sustained failure across multiple cycles and a minimum monotonic duration.
Tests

Unit coverage across the touched services (crontab/mount cleanup, reconciler, installer, port manager, syncthing state machine), with the key guards mutation-verified.
Two new integration suites: 60-volume-mount-ownership (install/reboot/uninstall/enterprise volume lifecycle, immutable-mountpoint enforcement, legacy-entry handling) and 61-syncthing-mount-safety (demotion of unsafe sendreceive folders, phantom-index handling, cold-start non-demotion, marker discipline).
Harness hardening: hermetic image registry (no live-internet dependencies; enforced at spec construction), deterministic seeded-sync state (a folder pinned synced holds matching data on disk), stall-tolerant confirmation-staleness defaults with the fast windows scoped to the suites that test those flows, and teardown that cannot leak volumes.
The full integration harness — all 54 suites — has been run against this branch and passes in its entirety (including the two new suites).


MorningLightMountain713 (approved):
Ack

PR #1763, opened 2026-07-03
fix: syncthing monitor verifies at decision points, never as steady-state churn

MorningLightMountain713:
## The problem being solved

On any node running an actively-writing `g:` application, the syncthing monitor degenerated into permanent maximum-rate churn:

- **Process-spawn churn**: every monitoring pass forked `mountpoint(1)` per app folder (some folders twice) and re-ran `sudo mkdir -p .../.stfolder` per syncthing folder — observed live at **3–4 shell commands per second, sustained indefinitely** (880 mountpoint forks and 206 mkdirs in ten minutes on a node with six apps).
- **Event-storm degeneration**: the monitor's edge accelerator re-armed on *any* folder event. A large install sync — or simply a busy app writing into its own synced folder — emits events continuously, so the "debounced" accelerator drove the full monitoring pass every ~2 seconds forever, instead of the intended 30-second level cadence.
- **Log spam**: un-rate-limited per-pass observations ("mounted but has no content … Potential data loss risk") wrote the same warning line every pass — 110 times in ten minutes for one legitimately-empty folder — burying real signals.
- **Architectural mismatch**: mount safety was re-verified for every folder on every pass, even though syncthing already has a native storage-loss signal: the `.stfolder` marker lives *inside* the mounted volume, so a vanished mount takes the marker with it and syncthing halts the folder and raises `FolderErrors` — no deletion broadcast. (The historical failure of that mechanism was FluxOS re-creating the marker on the bare directory each pass; that was fixed in #1761's mount-gated marker creation.)

## The design

Mount safety is verified **at decision points and in reaction to syncthing's own signals — never as a steady-state sweep**:

- **Startup** (first monitor pass after process start — the reboot case): full sweep with repair, unchanged.
- **Promotion flips** (receiveonly → sendreceive): the pre-flip safety gate, unchanged.
- **`FolderErrors` reaction**: the events consumer now accumulates flagged folder ids; the next pass mount-verifies exactly those folders (with repair, demote-to-receiveonly and container hold on failure). A plain app's bad mount no longer gates syncthing processing of unrelated folders.
- **The edge accelerator** (extracted into `syncthingMonitorAccelerator.js`) only reacts to folders the state machine is actively transitioning, plus `FolderErrors` and resync requests; early runs keep a minimum gap from the last completed pass (both knobs tunable via `config.syncthing`); a request landing mid-pass re-arms when the pass ends instead of being dropped. Steady-state folder activity never accelerates anything.
- **`isPathMounted` reads `/proc/self/mountinfo`** — one silent file read instead of a process fork, with octal-escape decoding and a `mountpoint(1)` fallback if the read fails.
- **`.stfolder` is created once** (mount-gated exactly as before); an existing marker is never re-asserted.
- **Observations log on change**: a persistent condition writes one line when first seen, a re-log at most every 5 minutes while it lasts, and one recovery line when it clears.

One deliberate contract change: a phantom index arising **mid-run** is caught at the next decision point (startup, promotion, flagged verify) rather than by per-cycle re-verification. The phantom hazard windows are precisely those decision points; steady-state paranoia was the churn.

## Results

- **Unit**: 168 tests across the six affected suites, including new coverage for the accelerator (scoping, min-gap, burst coalescing, mid-pass re-arm), the FolderErrors drain, marker create-once, observation gating, and a steady-state-no-sweep guard. The five key behaviors are mutation-verified (each fix reverted individually makes its guard test fail).
- **Live verification: this branch is deployed and running on a production node with real apps** — a three-component WordPress (wp + mysql + operator, its data folder syncthing-synced and actively written to by the running site: the exact workload that drove the old code to maximum-rate churn), Folding@home (`g:`, legitimately-empty folder — the observation-spam case), and a blockbook instance. Measured since restart: `mountpoint` forks **880/10min → 0**, marker re-assertions **206/10min → 0**, repeated observation warns **110/10min → 1** (a single change-gated line at boot for the empty folder). All app containers ran untouched across the deploy.
- **The mountinfo reader** was validated against live node mountinfo (loop mounts, trailing slashes, octal escapes; ~0.8 ms per probe, zero forks).
- **Extended soak: ~18 hours in production, spanning 11 FluxOS process restarts** (routine watchdog update-check restarts — the watchdog restarts FluxOS each cycle while a node runs a version ahead of the published release, as this pre-release deployment does; not crashes or instability) — zero `mountpoint` forks, zero marker re-assertions, and the empty-folder observation logged exactly once per process start (1:1 with firstRun sweeps). The change-gated logging contract holds across process lifecycles, not just at deploy time.
- **Same-hour A/B against an unpatched node**: over an identical idle hour on the same version base, an unpatched node (4 apps, all healthy) logged **123 forked `mountpoint` probes and 120 identical "data loss risk" warnings**; the patched node logged **zero of each** — the only `.stfolder` creations on either node were genuine new-install decision points. The unpatched base cadence is steady churn even without the event-storm degeneration described above.
- **Enterprise masterSlave app installed live on the patched node** (v8 encrypted-spec app, two `g:` components): accelerator early-runs fired at the 10s min-gap *only* while the folders were in active transition, interleaved with the 30s level passes; held-until-synced promotion ran through the pre-flip gate (112.6MB synced, then flipped to sendreceive); the standby containers were correctly held. A FluxOS restart then showed the firstRun decision-point sweep verifying both folders and steady state returning to silence. Every new mechanism in this PR observed on a workload class the original deployment didn't include.
- **Integration**: suite 61 gains a steady-state-quiet test (zero probes/re-assertions/warns over four-plus monitor cycles on healthy nodes) and its demote tests now inject the `FolderErrors` event real syncthing raises natively (the stub has no disk watcher). Suite 60 is unchanged: its self-heal path rides the reconciler, and its crontab/reboot tests use the intact startup sweeps.

## Merge gate

- [x] Unit suites green (168/168), key behaviors mutation-verified
- [x] Targeted syncthing/mount-family integration subset (15 suites) — all green (one new harness test needed a plumbing fix, `96a8dafbc`; all product-behavior tests passed first run)
- [x] **Full integration harness — 52/52 green** (full 52-suite parallel gate, run at `96a8dafbc`)

Written against `7022b2287`; harness-test fix `96a8dafbc`.

### DIFF
--- a/ZelBack/src/services/appMonitoring/syncthingEventsConsumer.js
+++ b/ZelBack/src/services/appMonitoring/syncthingEventsConsumer.js
@@ -50,6 +50,11 @@ const controller = new FluxController();
 // folderId -> { time, errors } from the last FolderErrors event
 const folderErrorsByFolder = new Map();
 
+// folder ids seen in FolderErrors since the last drain - the monitor pass
+// drains this to mount-verify exactly the flagged folders (targeted reaction,
+// never a steady-state sweep)
+const erroredFolderIdsSinceLastDrain = new Set();
+
 /**
  * One long-poll iteration: fetch events after `since`, detect lost-events
  * conditions, surface folder activity to the monitor.
@@ -100,6 +105,7 @@ async function pollOnce() {
     }
     if (event.type === 'FolderErrors') {
       folderErrorsByFolder.set(folder, { time: event.time, errors: event.data.errors || [] });
+      erroredFolderIdsSinceLastDrain.add(folder);
       fluxEventBus.publish('syncthing:folderErrors', { folder, time: event.time, errors: event.data.errors || [] });
     }
     if (callbacks.onFolderActivity) callbacks.onFolderActivity(folder, event.type);
@@ -177,9 +183,21 @@ function getFolderErrors(folderId) {
   return folderErrorsByFolder.get(folderId);
 }
 
+/**
+ * Folder ids flagged by FolderErrors since the last drain; draining clears
+ * the accumulator. Consumed by the monitor pass for targeted mount verifies.
+ * @returns {string[]} Folder ids
+ */
+function drainErroredFolderIds() {
+  const ids = [...erroredFolderIdsSinceLastDrain];
+  erroredFolderIdsSinceLastDrain.clear();
+  return ids;
+}
+
 module.exports = {
   start,
   stop,
   isRunning,
   getFolderErrors,
+  drainErroredFolderIds,
 };
--- a/ZelBack/src/services/appMonitoring/syncthingFolderStateMachine.js
+++ b/ZelBack/src/services/appMonitoring/syncthingFolderStateMachine.js
@@ -24,6 +24,46 @@ const {
 
 const { isPathMounted } = volumeService;
 
+const monotonicMs = () => Number(process.hrtime.bigint() / 1000000n);
+
+// Per-folder mount-safety observation log gate: a persistent condition writes
+// one line when first seen (re-logged at most every OBSERVATION_RELOG_MS while
+// it lasts) and one recovery line when it clears - never a line per monitor
+// pass. Process-lifetime state, bounded by the folder ids ever observed; it
+// only dedupes log lines, so a stale entry for a removed app costs nothing.
+// appId -> { observation, lastLoggedMs }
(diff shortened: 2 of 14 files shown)


Cabecinha84 (approved):
ack

PR #1764, opened 2026-07-04
v8.16.0

Cabecinha84:
## v8.16.0

Release from `development` into `master`.

### Highlights

**Volume mounting & ownership**
- FluxOS now owns app volume mounting; unmounted app dirs are inert (#1761)
- Verify folder safety before any flip to `sendreceive`
- Crontab cleanup and boot mount sweep now see enterprise apps
- Enterprise local install keeps the spec height through the formatter
- Honor a pending stop when the data volume is unavailable

**Syncthing monitor**
- Monitor probes at decision points, never as steady-state churn (#1763)
- Mount-safety guard gates on the mountpoint, not content
- Demote unsafe `sendreceive` folders even when the monitor cycle is skipped

**UPnP**
- Require sustained UPnP mapping failure before app force-removal
- Bound the UPnP retry pause to once per restore cycle

**Tests (infra)**
- New volume mount ownership + syncthing mount-safety suites (suites 60/61)
- Enterprise-blind inventory regressions (unit + suite 60)
- Various hermeticity and robustness improvements to the infra tests


MorningLightMountain713 (approved):
Ack

PR #1766, opened 2026-07-07
fix(reconciler): heal containers running detached from their docker network

Cabecinha84:
## Summary

Heals containers that are **running but attached to no docker network** — a state a plain `docker start` can never repair. The reconciler previously trusted any running container as "already where we want it", so such a container stayed broken indefinitely.

## The bug, from a real incident

A shared-db WordPress app whose **operator** component (`runonflux/shared-db`) could not reach its **mysql** component by name.

Reconstructed from the node's FluxOS logs (debug + error) and live `docker`/`journalctl`:

- **Healthy** until `2026-06-25T03:55:40`.
- At that instant a `docker start` of the operator failed:
  ```
  failed to set up container networking: driver failed programming external connectivity
  on endpoint fluxoperator_wordpress...: failed to bind host port 0.0.0.0:36622/tcp: address already in use
  ```
  (The `768e…` in that message is the **endpoint** id, not the container — container `a5c71e…` was created `2026-06-07` and never recreated.)
- From that minute the operator ran **detached** from `fluxDockerNetwork_<app>`: `State=running`, `NetworkMode=fluxDockerNetwork_<app>`, but `NetworkSettings.Networks = {}` and **no published ports**. No IP, no embedded DNS (so it couldn't resolve `mysql`), API port `36904` dead.
- `monitorSharedDBApps` logged `operatorStatus error: AggregateError` **every 5 min for ~11 days** (1 isolated error before, 2728+ continuous after). Live daemon logs confirmed the embedded resolver forwarding the operator's own name externally (absent from the network) and timing out.
- The node was rebooting **uncleanly every ~4h** (`cleanShutdown=false`), producing the transient host-port conflict at a restart; each later boot did a plain `docker start` → came up detached again, silently.

**Root cause of the persistence:** the reconciler's actuation had `if (actual.running) return; // already where we want it` — it never verified network attachment.

## Design

Invariant checked on **every reconcile of a running container** (post-start *and* the periodic sweep, so it fixes freshly-broken *and* long-broken containers):

> a **running** container whose `HostConfig.NetworkMode` is `fluxDockerNetwork_*` must appear in `NetworkSettings.Networks[NetworkMode]` with an `IPAddress`.

Repair is a **recreate**, not a reconnect: `docker network connect` on a live container does not restore published host ports (e.g. `36904`), so only recreating (fresh endpoint via the tested soft-install path) fully heals it.

The remove step is destructive, so it is guarded — see "Safety" below.

## Changes

- **`dockerService.js`**
  - `classifyContainerNetworkAttachment(info)` — pure classifier over a docker inspect object → `{ managed, running, networkMode, attached }`.
  - `isContainerDetachedFromNetwork(attachment)` — `managed && running && !attached`. Host/none/bridge are never "detached".
  - `dockerNetworkState(networkName)` → `'exists' | 'absent' | 'unknown'`. Distinguishes a stale endpoint from a pruned network. Absence is never inferred from an error: on a failed inspect it probes docker's network list and uses its ANSWER (the same pattern `dockerActual` uses); if docker can't answer at all, `'unknown'` and the caller defers.
- **`containerHealthMonitor.js`** — `recreateMissingContainers(id, { softOnly })`. Without it, the recreate silently falls back to `installApplicationHard` → `createAppVolume` → `fallocate` + `mke2fs`, i.e. it **reformats the app's data volume**, on any `verifyAppVolumeMount` failure. That is tolerable for a container that is gone anyway; it is catastrophic for the heal, which deliberately removes a *live* container whose data is intact. `softOnly` makes the recreate throw (and be retried) instead.
- **`appReconciler.js`**
  - `dockerActual` classifies attachment from the inspect it already does and carries it in `actual` (no second inspect / no TOCTOU).
  - Running branch: verify `actual.attachment`; if detached, confirm, check the preconditions, then recreate — paced, and **without** the uninstall escalation.
  - `!exists` branch: a container **we** removed for a heal is recreated on the no-uninstall path, gated on a **durable** flag rather than in-memory state.
  - Every successful start/recreate arms a short verification pass (a container born detached at boot would otherwise wait for the hourly sweep).
  - Stops the container's stats monitor before force-remove (matching the uninstaller) so its interval can't leak/error-spam against a gone container — and restores it if the remove fails.
- **`appsRuntimeState.js`** — the durable heal state: `networkHealRemoval` ("I removed this container on purpose") plus `networkHealHistory`, the heal's own backoff ladder. Both merged by `prepareCollection`.
- **`appTamperingDetectionService.js`** — document the new `network_detached` event type (audit row only — no scoring, no auto-uninstall).

## Safety

The heal removes a container *before* it recreates it, so every guard below exists to make sure that trade is never a bad one.

- **Never destroys what it cannot rebuild.** The remove is gated on the recreate's own preconditions, checked first: the app must have a `compose` spec (`recreateMissingContainers` throws unconditionally without one — a v1-3 app's container would have been destroyed permanently, since the heal by design never escalates to an uninstall that would re-place the app), and the data volume must verify as mounted (the recreate is `softOnly`, so an unverifiable volume means it would fail *after* the container was gone). If either fails, the container is left running and the reason is logged loudly.
- **Never reformats the app's data.** The recreate is `softOnly` (see Changes): a transient volume-check failure retries instead of running `mke2fs` over the user's volume.
- **Never uninstalls the app.** `recreateMissing`'s failure path runs `removeAppLocally` (the whole app); a transient recreate failure must not uninstall a degraded-but-serving app. The heal recreates directly and re-arms a paced retry; no path from the heal reaches an uninstall.
- **Survives a FluxOS restart mid-heal.** The container is legitimately absent between the force-remove and the recreate. That fact is persisted in `appsRuntimeState.networkHealRemoval` **before** the remove, and cleared as soon as the container exists again. Without it, a restart in that window would make the next process read the absence as tampering (a false `container_vanished`) and, on a failed recreate, uninstall the app. The read of that flag **fails closed**: a DB error throws and the reconciler defers, because guessing "not a heal removal" is the destructive guess.
- **Never actuates on a container it no longer owns.** `isManagedElsewhere` is sampled at reconcile entry, but the heal spends seconds confirming — so it is re-checked at actuation time, right before the force-remove. A redeploy or backup that starts during the confirmation must not have its container removed underneath it.
- **Confirmed, and persisted.** The detached state must survive an in-pass settle + re-inspect **and** have persisted for 60s of wall-clock since first sighting. (Wall-clock, not a count of reconcile passes: two passes can land in the same instant, since the workqueue re-runs a dirty id immediately.) A dockerd restart with live-restore can answer an inspect before libnetwork has restored endpoint IPs — and the reconnect sweep enqueues every component at exactly that moment — so a short confirm alone would rebuild the node's whole workload.
- **Refuses a node-wide rebuild.** If several containers look detached at once, that is a daemon-level fault, not N stale endpoints: the heal says so loudly and destroys nothing.
- **Paced, not capped.** Heal attempts walk their own durable ladder (same shape as the crash backoff, decaying to a 30m cap, but a *separate* history — sharing one would let a heal's attempts hold down the very container it just recreated, and would block the heal of a crash-looping container). A hard attempt cap would park the heal in a terminal state until the FluxOS process restarted — not level-based, and on an unattended node "leave it for manual intervention" means "leave it broken". Retrying forever is convergent at the network level: while the container is broken or absent this node stops advertising it in `apprunning`, its location record expires on TTL and the app is re-placed elsewhere — without destroying this node's bind-mounted data, while the paced retry keeps working the local problem.
- **Only heals a stale endpoint.** If the network itself was pruned, a recreate would fail on a missing `NetworkMode`, so the container is left in place and a `network_pruned` event is recorded. If docker cannot say whether the network exists, nothing is destroyed.
- **Fails loud, not silent.** The pre-remove writes (tamper event, ladder, flag) abort the remove and re-arm a retry if they throw, rather than escaping `reconcile` and leaving the container broken until the hourly sweep. A heal whose recreate keeps failing records `recreation_failed` / `network_pruned` — a looping heal must not be an invisible outage.
- **Bounded node-tampering-event footprint.** `network_detached` is recorded once per episode (not per attempt) into the tampering-events collection that `appTamperingBlocklistService` counts. That count only DOSes a node that is *also* on the manually-curated blocklist and over the threshold, so a normal node is never DOSed by it; recording once keeps the contribution minimal for what is an operational, not malicious, fault.
- Only `fluxDockerNetwork_*` containers are eligible; stopped / awaiting-controller containers are excluded; an inspect error is treated as "not detached".

## Testing

Run with `NODE_CONFIG_DIR=$PWD/tests/unit/globalconfig` (fully `proxyquire`-mocked — no DB/docker needed). **128 passing**, 0 failing:

- **`tests/unit/appReconciler.test.js`: 84** — the `network-detach heal` block covers: attached → no-op + heal state cleared; the state cleared for a container that merely *exists*, even stopped (the flag must not leak); a detached read that clears on re-inspect destroys nothing; a detach that has not persisted destroys nothing; a persisted detach sets the durable flag **before** the remove, stops monitoring, recreates with `{ softOnly: true }`, never uninstalls; **refuses to remove a v1-3 container it could never recreate**; **refuses to remove a container whose volume cannot be verified**; aborts when another operation takes the container over during the confirmation; **refuses a node-wide rebuild** when many containers look detached at once; network confirmed gone → no destroy + `network_pruned`; network *unreadable* → no destroy and no false `network_pruned`; paced on its own ladder (and provably not on the crash one); no remove if the heal cannot be recorded durably; a failed force-remove restores monitoring; a failed recreate retries, records `recreation_failed`, never uninstalls; a persistently broken heal keeps retrying (no terminal give-up) with one tamper event per episode; a **FluxOS restart mid-heal** still recreates while a genuine vanish still takes the vanished path; an **unreadable heal flag defers** rather than guessing. Plus post-start attachment verification.
- **`tests/unit/appsRuntimeState.test.js`: 32** — the durable flag (persists across a fresh module instance, dropped on uninstall, **throws rather than answering false** on a read failure) and the heal ladder (paces, does not touch `restartHistory`, resets once healthy).
- **`tests/unit/containerHealthMonitor.test.js`: 12** — `softOnly` throws instead of hard-installing (component and whole-app paths), still soft-installs normally, and leaves the default vanished-container path free to hard-install.
- **`tests/unit/dockerService.test.js`: 11** for the pure classifier + `dockerNetworkState` (notably: `unknown`, never `absent`, when docker can't answer).

## Scope / follow-up

Defense-in-depth (auto-heal). The environmental trigger — a node rebooting uncleanly every ~4h — is a separate operational issue, out of scope here.

### DIFF
--- a/ZelBack/src/services/appManagement/appsRuntimeState.js
+++ b/ZelBack/src/services/appManagement/appsRuntimeState.js
@@ -191,29 +191,80 @@ async function setNetworkHealRemoval(identifier, removed) {
 /**
  * Whether the reconciler removed this container itself for a network heal.
  *
+ * Deliberately does NOT reuse getState: getState swallows a read failure and
+ * returns null, which here would read as "not a heal removal" - the DESTRUCTIVE
+ * direction (the caller would treat its own removal as a vanished container,
+ * record a false tampering event and, on a failed recreate, uninstall the app).
+ * A read failure must be a failure, so the caller can defer instead of guess.
+ *
  * @param {string} identifier
- * @returns {Promise<boolean>}
+ * @returns {Promise<boolean>} - throws if the state cannot be read
  */
-async function isNetworkHealRemoval(identifier) {
-  const state = await getState(identifier);
+async function isNetworkHealRemoval(rawIdentifier) {
+  const identifier = canonical(rawIdentifier);
+  const database = collection();
+  const state = await dbHelper.findOneInDatabase(database, appsRuntimeState, { identifier }, { projection: { _id: 0 } });
   return state?.networkHealRemoval === true;
 }
 
 /**
- * Clears the heal-removal flag. Reads first so the healthy steady state (every
- * reconcile of every attached container) costs a lookup, not a write.
+ * Appends a network-heal attempt to its OWN durable ladder. Kept separate from
+ * restartHistory: sharing it would cross-contaminate in both directions - a heal
+ * that recreates a g: component (created, not started) would then have the very
+ * container it just fixed held down by the crash ladder its own attempts grew,
+ * and a crash-looping container could not be healed for up to the 30m cap.
  *
  * @param {string} identifier
  */
-async function clearNetworkHealRemoval(identifier) {
+async function recordNetworkHealAttempt(identifier) {
+  // No catch: pacing that silently fails to persist means the next pass sees an
+  // empty ladder, waits 0, and hammers the destructive heal. The caller defers.
+  const state = await getState(identifier);
+  const history = (state && state.networkHealHistory) || [];
+  history.push(Date.now());
+  if (history.length > MAX_HISTORY) {
+    history.splice(0, history.length - MAX_HISTORY);
+  }
+  await setFields(identifier, { networkHealHistory: history });
+}
+
+/**
+ * How long (ms) before the next network-heal attempt is allowed - 0 means now.
+ * Same ladder shape as the crash backoff (immediate, 30s, 5m, 15m, 30m cap), so
+ * a container that keeps coming back detached is retried forever but at a
+ * decaying rate. The ladder is reset by clearNetworkHeal once the container is
+ * seen healthy, so a later, unrelated episode starts from the bottom again.
+ *
+ * @param {string} identifier
+ * @returns {Promise<number>}
+ */
+async function networkHealWaitMs(identifier) {
+  const state = await getState(identifier);
+  const history = (state && state.networkHealHistory) || [];
+  if (history.length === 0) return 0;
+  const last = history[history.length - 1];
+  const index = Math.min(history.length, BACKOFF_DELAYS_MS.length - 1);
+  return Math.max(0, BACKOFF_DELAYS_MS[index] - (Date.now() - last));
+}
+
+/**
+ * Drops all network-heal state (the removal flag and the heal ladder) once the
+ * container is healthy again. Reads first so the healthy steady state - every
+ * reconcile of every attached container - costs a lookup, not a write.
+ *
+ * @param {string} identifier
+ */
+async function clearNetworkHeal(identifier) {
   try {
-    if (!(await isNetworkHealRemoval(identifier))) return;
-    await setFields(identifier, { networkHealRemoval: false });
+    const state = await getState(identifier);
+    if (!state) return;
+    if (state.networkHealRemoval !== true && !(state.networkHealHistory || []).length) return;
+    await setFields(identifier, { networkHealRemoval: false, networkHealHistory: [] });
   } catch (err) {
-    // A stale `true` only ever makes the reconciler MORE conservative (it keeps
+    // A stale flag only ever makes the reconciler MORE conservative (it keeps
     // recreating instead of escalating to an uninstall), so a failed clear is safe
     // to log and move on from - unlike a failed set, which must abort the remove.
-    log.error(`appsRuntimeState - failed to clear network heal removal for ${identifier}: ${err.message}`);
+    log.error(`appsRuntimeState - failed to clear network heal state for ${identifier}: ${err.message}`);
   }
 }
 
@@ -278,6 +329,7 @@ async function prepareCollection() {
           // likewise, a heal-removal anywhere means a heal may be mid-flight: keep
           // the reconciler on the recreate path rather than the uninstall path
           networkHealRemoval: twins.some((t) => t.networkHealRemoval === true),
+          networkHealHistory: [...new Set(twins.flatMap((t) => t.networkHealHistory || []))].sort((a, b) => a - b).slice(-MAX_HISTORY),
           restartHistory: [...new Set(twins.flatMap((t) => t.restartHistory || []))].sort((a, b) => a - b).slice(-MAX_HISTORY),
           updatedAt: Math.max(...twins.map((t) => t.updatedAt || 0)),
         };
@@ -308,7 +360,9 @@ module.exports = {
   restartWaitMs,
   setNetworkHealRemoval,
   isNetworkHealRemoval,
-  clearNetworkHealRemoval,
+  recordNetworkHealAttempt,
+  networkHealWaitMs,
+  clearNetworkHeal,
   recordExit,
   remove,
   BACKOFF_DELAYS_MS,
--- a/ZelBack/src/services/appManagement/appsRuntimeState.js
+++ b/ZelBack/src/services/appManagement/appsRuntimeState.js
@@ -170,6 +170,54 @@ async function restartWaitMs(identifier, lastFinishedAtMs = null) {
 }
 
 /**
+ * The network-detach heal force-removes a container in order to recreate it with
+ * a fresh endpoint. Between those two steps the container is legitimately absent,
+ * and that fact MUST be durable: if FluxOS restarts in that window, the next
+ * reconcile would otherwise see a missing container with no heal in flight, call
+ * it vanished (a false tampering signal) and, on a failed recreate, uninstall the
+ * whole app. This flag is the reconciler's memory of "I removed this on purpose";
+ * it is set before the remove and cleared once the container is seen running AND
+ * attached again (see appReconciler).
+ *
+ * @param {string} identifier
+ * @param {boolean} removed
+ */
+async function setNetworkHealRemoval(identifier, removed) {
+  // No catch: the flag is what keeps a failed heal from escalating to an app
+  // uninstall. If it cannot be persisted, the caller must not remove the container.
+  await setFields(identifier, { networkHealRemoval: removed });
+}
+
+/**
+ * Whether the reconciler removed this container itself for a network heal.
+ *
+ * @param {string} identifier
+ * @returns {Promise<boolean>}
+ */
+async function isNetworkHealRemoval(identifier) {
+  const state = await getState(identifier);
+  return state?.networkHealRemoval === true;
+}
+
+/**
+ * Clears the heal-removal flag. Reads first so the healthy steady state (every
+ * reconcile of every attached container) costs a lookup, not a write.
+ *
+ * @param {string} identifier
+ */
+async function clearNetworkHealRemoval(identifier) {
+  try {
+    if (!(await isNetworkHealRemoval(identifier))) return;
+    await setFields(identifier, { networkHealRemoval: false });
+  } catch (err) {
+    // A stale `true` only ever makes the reconciler MORE conservative (it keeps
+    // recreating instead of escalating to an uninstall), so a failed clear is safe
+    // to log and move on from - unlike a failed set, which must abort the remove.
+    log.error(`appsRuntimeState - failed to clear network heal removal for ${identifier}: ${err.message}`);
+  }
+}
+
+/**
  * Records the last observed exit for diagnostics / tampering signals.
  *
  * @param {string} identifier
@@ -227,6 +275,9 @@ async function prepareCollection() {
           identifier,
           // a lock anywhere is a lock: never auto-start a deliberately stopped app
           operatorStopped: twins.some((t) => t.operatorStopped === true),
+          // likewise, a heal-removal anywhere means a heal may be mid-flight: keep
+          // the reconciler on the recreate path rather than the uninstall path
+          networkHealRemoval: twins.some((t) => t.networkHealRemoval === true),
           restartHistory: [...new Set(twins.flatMap((t) => t.restartHistory || []))].sort((a, b) => a - b).slice(-MAX_HISTORY),
           updatedAt: Math.max(...twins.map((t) => t.updatedAt || 0)),
         };
@@ -255,6 +306,9 @@ module.exports = {
   isOperatorStopped,
   recordRestart,
   restartWaitMs,
+  setNetworkHealRemoval,
+  isNetworkHealRemoval,
+  clearNetworkHealRemoval,
   recordExit,
   remove,
   BACKOFF_DELAYS_MS,
--- a/ZelBack/src/services/appMonitoring/appReconciler.js
+++ b/ZelBack/src/services/appMonitoring/appReconciler.js
@@ -374,13 +374,13 @@ async function recreateForNetworkHeal(identifier) {
 }
 
 /**
- * One heal cycle for a running-but-detached container: record the signal,
- * force-remove the broken container (keeping bind-mounted data), then recreate
- * it with a fresh endpoint. Force-remove is required because the recreate path
- * (appDockerCreate) 409s on an existing container name.
+ * One heal cycle for a running-but-detached container: force-remove the broken
+ * container (keeping bind-mounted data), then recreate it with a fresh endpoint.
+ * Force-remove is required because the recreate path (appDockerCreate) 409s on an
+ * existing container name. The durable network_detached signal is recorded once
+ * per episode by the caller, not here (this runs up to MAX times per episode).
  */
-async function attemptNetworkHeal(identifier, mainAppName) {
-  await appTamperingDetectionService.recordEvent(mainAppName, 'network_detached', `Container ${identifier} running with no network endpoint on its own network`);
+async function attemptNetworkHeal(identifier) {
   fluxEventBus.publish('reconciler:actuated', { identifier, action: 'networkDetached' });
   // Stop the per-minute stats monitor before removing the container (mirrors the
   // uninstaller). Otherwise its interval runs against a gone container - and, if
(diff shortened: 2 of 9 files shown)


alihm (approved):
Ack

MorningLightMountain713 (changes requested):
The approach is right — the level-based check and recreate-as-repair should stay as-is, and none of the findings below touch them. They're all in the guard mechanics around the heal, and most trace back to the in-memory `networkHealState` machine (`obs`/`tries`/`gaveUp`), so after the individual findings there's a restructure suggestion that removes most of them structurally rather than patching each one.

### 1. A transient error can strand a container removed forever (blocking)

Two things combine:

- `dockerNetworkExists()` treats **any** inspect failure as "network doesn't exist" — a momentary docker error becomes a wrong conclusion.
- `gaveUp` means two different things: in the network-missing branch it's "already logged, don't spam"; in the `!exists` branch it's "stop retrying forever".

The sequence: a transient error in the network check → `gaveUp = true` plus a false `network_pruned` event → the check passes on a later pass, the heal proceeds and force-removes the container → the recreate fails once (port still held) → the next pass hits `if (st.gaveUp) return` in the `!exists` branch → the container stays removed until FluxOS restarts.

Fix: give `dockerNetworkExists` three outcomes — exists / confirmed absent / couldn't tell — using the same probe-and-use-the-answer pattern `dockerActual` already uses, and defer (scheduleRetry) on "couldn't tell". Track the log-once behavior separately (the `volumeMissingNoted` Set idiom fits) instead of overloading `gaveUp`.

### 2. "Never uninstalls the app" doesn't survive a FluxOS restart mid-heal

`networkHealState` is in-memory. If FluxOS restarts between the force-remove and the recreate, the new process sees a missing container with no heal in flight and takes the vanished path — a false `container_vanished` tamper event, and the uninstall escalation if that recreate fails. The one fact the heal genuinely needs to remember — *"I removed this container on purpose"* — must survive a restart. `appsRuntimeState` is the existing home for durable reconciler inputs; one flag set before the force-remove, cleared once the container is next seen running **and attached**, closes this completely.

### 3. Detection is slow after boot — up to an hour

Nothing triggers a reconcile after a successful start, so a container that comes up detached at boot isn't looked at again until the hourly sweep. This pathology is *born at start time*, so anchor detection to the start we just performed: `scheduleRetry(identifier, ~30s)` after every successful start/recreate, as a verification pass. Detached-at-boot then heals in under a minute, with the sweep remaining a backstop. (Docker's network events can't cover this: a container born detached never emits a disconnect event.)

### 4. The debounce counts passes, not time

Two reconciles can queue back-to-back (die event + sweep), so both "consecutive observations" can come from effectively the same instant. Suggest confirming inside a single pass instead: see detached → settle a few seconds → re-inspect → act only if still detached. That guarantees the time separation the debounce is meant to provide, removes `obs` entirely, and has in-file precedent (`DATA_CLEAR_SETTLE_MS`).

### 5. One give-up path leaves the stats monitor off

If the force-remove itself keeps failing, we give up with the container still present and running — but monitoring was stopped and never restarted. On remove failure, `startAppMonitoring` again.

### Suggested restructure

`tries` + `gaveUp` implement a hard cap that parks the heal in a dead state until process restart. That's un-level-based: a reconciler should keep trying at a bounded *rate*, not stop. `appsRuntimeState` already carries the durable crash-backoff ladder (decaying to a 30m cap) — pacing heal attempts on that instead of counting them removes the terminal state entirely: no `gaveUp` flag (finding 1's conflation can't exist), no stranding, and steady-state retries cost the same as the sweep.

"Retry forever, never uninstall" is also convergent at the network level: if the container stays broken or absent, this node stops advertising it in `apprunning`, the location record expires on TTL, and the app gets re-placed elsewhere — the network self-heals without destroying this node's bind-mount data, while the paced retry keeps working the local problem. Strictly better than parking it for manual intervention on an unattended node.

Net effect: `networkHealState` disappears, replaced by four existing idioms — an in-pass confirm, a post-start verification retry, one durable flag in `appsRuntimeState`, and the backoff ladder plus a record-once Set. The core of the PR (classifier, running-branch invariant, recreate via the tested soft-install path, never-uninstall policy, network-missing carve-out) is unchanged, and the guard code gets smaller, not bigger.

### Naming

- `gaveUp` — needs splitting regardless of the restructure; it names two different concepts, and that's the root of finding 1.
- `dockerNetworkExists` — overpromises as implemented: it actually means "an inspect succeeded just now". Honest once error is distinguished from absence.
- `parseContainerNetworkAttachment` — nothing is parsed; `classifyContainerNetworkAttachment` (or `get…`) is truer. Cosmetic.
- `st` / `obs` — this file otherwise uses full words (`bootDraining`, `volumeMissingNoted`, `controllerDesired`); `healState` / `observations` would match the house style.
- `MAX_NETWORK_HEAL_ATTEMPTS`, `DETACHED_CONFIRM_OBSERVATIONS`, `recreateForNetworkHeal`, `isContainerDetachedFromNetwork` — all good, say exactly what they do.

Cabecinha84:
> The approach is right — the level-based check and recreate-as-repair should stay as-is, and none of the findings below touch them. They're all in the guard mechanics around the heal, and most trace back to the in-memory `networkHealState` machine (`obs`/`tries`/`gaveUp`), so after the individual findings there's a restructure suggestion that removes most of them structurally rather than patching each one.
> 
> ### 1. A transient error can strand a container removed forever (blocking)
> Two things combine:
> 
> * `dockerNetworkExists()` treats **any** inspect failure as "network doesn't exist" — a momentary docker error becomes a wrong conclusion.
> * `gaveUp` means two different things: in the network-missing branch it's "already logged, don't spam"; in the `!exists` branch it's "stop retrying forever".
> 
> The sequence: a transient error in the network check → `gaveUp = true` plus a false `network_pruned` event → the check passes on a later pass, the heal proceeds and force-removes the container → the recreate fails once (port still held) → the next pass hits `if (st.gaveUp) return` in the `!exists` branch → the container stays removed until FluxOS restarts.
> 
> Fix: give `dockerNetworkExists` three outcomes — exists / confirmed absent / couldn't tell — using the same probe-and-use-the-answer pattern `dockerActual` already uses, and defer (scheduleRetry) on "couldn't tell". Track the log-once behavior separately (the `volumeMissingNoted` Set idiom fits) instead of overloading `gaveUp`.
> 
> ### 2. "Never uninstalls the app" doesn't survive a FluxOS restart mid-heal
> `networkHealState` is in-memory. If FluxOS restarts between the force-remove and the recreate, the new process sees a missing container with no heal in flight and takes the vanished path — a false `container_vanished` tamper event, and the uninstall escalation if that recreate fails. The one fact the heal genuinely needs to remember — _"I removed this container on purpose"_ — must survive a restart. `appsRuntimeState` is the existing home for durable reconciler inputs; one flag set before the force-remove, cleared once the container is next seen running **and attached**, closes this completely.
> 
> ### 3. Detection is slow after boot — up to an hour
> Nothing triggers a reconcile after a successful start, so a container that comes up detached at boot isn't looked at again until the hourly sweep. This pathology is _born at start time_, so anchor detection to the start we just performed: `scheduleRetry(identifier, ~30s)` after every successful start/recreate, as a verification pass. Detached-at-boot then heals in under a minute, with the sweep remaining a backstop. (Docker's network events can't cover this: a container born detached never emits a disconnect event.)
> 
> ### 4. The debounce counts passes, not time
> Two reconciles can queue back-to-back (die event + sweep), so both "consecutive observations" can come from effectively the same instant. Suggest confirming inside a single pass instead: see detached → settle a few seconds → re-inspect → act only if still detached. That guarantees the time separation the debounce is meant to provide, removes `obs` entirely, and has in-file precedent (`DATA_CLEAR_SETTLE_MS`).
> 
> ### 5. One give-up path leaves the stats monitor off
> If the force-remove itself keeps failing, we give up with the container still present and running — but monitoring was stopped and never restarted. On remove failure, `startAppMonitoring` again.
> 
> ### Suggested restructure
> `tries` + `gaveUp` implement a hard cap that parks the heal in a dead state until process restart. That's un-level-based: a reconciler should keep trying at a bounded _rate_, not stop. `appsRuntimeState` already carries the durable crash-backoff ladder (decaying to a 30m cap) — pacing heal attempts on that instead of counting them removes the terminal state entirely: no `gaveUp` flag (finding 1's conflation can't exist), no stranding, and steady-state retries cost the same as the sweep.
> 
> "Retry forever, never uninstall" is also convergent at the network level: if the container stays broken or absent, this node stops advertising it in `apprunning`, the location record expires on TTL, and the app gets re-placed elsewhere — the network self-heals without destroying this node's bind-mount data, while the paced retry keeps working the local problem. Strictly better than parking it for manual intervention on an unattended node.
> 
> Net effect: `networkHealState` disappears, replaced by four existing idioms — an in-pass confirm, a post-start verification retry, one durable flag in `appsRuntimeState`, and the backoff ladder plus a record-once Set. The core of the PR (classifier, running-branch invariant, recreate via the tested soft-install path, never-uninstall policy, network-missing carve-out) is unchanged, and the guard code gets smaller, not bigger.
> 
> ### Naming
> * `gaveUp` — needs splitting regardless of the restructure; it names two different concepts, and that's the root of finding 1.
> * `dockerNetworkExists` — overpromises as implemented: it actually means "an inspect succeeded just now". Honest once error is distinguished from absence.
> * `parseContainerNetworkAttachment` — nothing is parsed; `classifyContainerNetworkAttachment` (or `get…`) is truer. Cosmetic.
> * `st` / `obs` — this file otherwise uses full words (`bootDraining`, `volumeMissingNoted`, `controllerDesired`); `healState` / `observations` would match the house style.
> * `MAX_NETWORK_HEAL_ATTEMPTS`, `DETACHED_CONFIRM_OBSERVATIONS`, `recreateForNetworkHeal`, `isContainerDetachedFromNetwork` — all good, say exactly what they do.

Thanks — the review was right on all five, and the restructure is in (adef6cb).
  `networkHealState` (obs/tries/gaveUp) is gone, replaced by: in-pass confirmation,
  a durable `networkHealRemoval` flag written before the force-remove, a post-start
  verification pass, and ladder-paced retries with no terminal state.

  One deliberate deviation from your suggestion, and one thing that turned up
  afterwards which matters more than anything in the original review.

  ### Deviation: the heal gets its own ladder, not `restartHistory`

  You suggested pacing heal attempts on the existing crash-backoff ladder. I gave
  the heal a separate durable history (`networkHealHistory`, same shape, same 30m
  cap) because sharing `restartHistory` cross-contaminates in both directions. The
  concrete case: `recreateMissingContainers` *creates but does not start* a g:
  component, so the very next reconcile takes the stopped-container path and calls
  `restartWaitMs` — which now reads the ladder the heal itself just grew. A
  correctly recreated, perfectly healthy container gets held stopped for 5–30
  minutes. Symmetrically, a container that happens to be crash-looping cannot be
  healed for up to the cap. Everything else about your reasoning holds: paced, no
  attempt cap, no terminal state, retry-forever is convergent via `apprunning` TTL
  re-placement.

  ### What a follow-up review found: the heal could reformat the app's data volume

  This is in the part of the design that predates your review and that neither of
  us flagged, so I want it on the record rather than buried in a diff.

  The heal force-removes a **live** container and rebuilds it with
  `recreateMissingContainers`. That function picks soft vs hard install from
  `verifyAppVolumeMount` inside a `try/catch` that swallows *every* error into
  `volumeMounted = false` (containerHealthMonitor.js:47-57) — and
  `installApplicationHard` → `createAppVolume` runs `fallocate` + **`mke2fs`** on
  the app's volume file with no existence check (advancedWorkflows.js:517,535). So
  a transient volume-check failure, in a window the heal itself opens by destroying
  a healthy container, reformats the user's data. My own comment in the first draft
  ("v=false: Flux data lives on bind mounts, so nothing is lost; the recreate reuses
  it via a soft install") asserted a soft path that was never guaranteed.

  Same root cause, second symptom: `recreateMissingContainers` throws
  unconditionally on a spec with no `compose` array (v1-3 apps, which the reconciler
  explicitly supports). The heal would force-remove such a container, fail every
  recreate, and — by design — never escalate. Container destroyed permanently, app
  never re-placed. The "never uninstall" policy had quietly become "never recover".

  Fixed in dbfa061 by one rule: **never destroy what you cannot prove you can
  rebuild.** The recreate is now `softOnly` (throws and retries instead of
  reformatting), and the preconditions are checked *before* the remove.

  The same review also produced, and I've fixed:

  - `isManagedElsewhere` re-checked at actuation time — it was sampled at reconcile
    entry, seconds before the force-remove, so a redeploy/backup starting during the
    confirmation had its container removed underneath it.
  - `isNetworkHealRemoval` now fails **closed**. It went through `getState`, which
    swallows read errors and returns null — read as "not a heal removal", the
    destructive guess: a DB blip could make the reconciler treat its own removal as
    a vanished container and uninstall the app. It throws now; the caller defers.
  - The durable flag is cleared as soon as the container **exists**, not only when
    seen running+attached. It leaked on every path that leaves a container stopped
    (awaiting-controller g:, operator stop), and a stale flag permanently diverted a
    *genuine* disappearance away from the vanished path (no `container_vanished`, no
    escalation — tampering detection blind to a real `docker rm`).
  - The detach must persist 60s of wall-clock (timestamp, not a pass count — your
    point about back-to-back passes still stands), plus a circuit breaker: if
    several containers look detached at once, that's the daemon, not N stale
    endpoints, and nothing is destroyed. A dockerd restart with live-restore can
    answer inspects before libnetwork restores endpoint IPs, and the reconnect sweep
    enqueues everything at exactly that moment — the 3s confirm alone would have
    rebuilt the node's entire workload.
  - Pre-remove writes abort the remove and re-arm a retry if they throw, instead of
    escaping `reconcile` and leaving the container broken until the hourly sweep.
  - A failing heal recreate now records `recreation_failed` / `network_pruned` —
    still no uninstall, but no longer an invisible outage.

MorningLightMountain713 (approved):
Ack

PR #1768, opened 2026-07-16
feat(apps): honest, attributable app-tampering incident model

MorningLightMountain713:
## What this does

Rebuilds app-tampering detection into an **honest, attributable, deduplicated incident model**, and makes its tamper score safe to reason about. All behaviour is centralized in `appTamperingDetectionService.js`; call sites still pass plain `(appName, eventType, details)`.

- **Reboot detection by kernel `boot_id`.** `/proc/sys/kernel/random/boot_id` (a per-boot UUID, stable across process restarts) is compared to a stored marker. A genuine reboot appends to a rolling `bootHistory` (capped at 200) in `nodestartuptracker`, recording both the uptime-derived machine boot moment (`bootedAt`) and when FluxOS first saw it (`at`). A same-`boot_id` process restart records nothing.
- **Incident model, not an event log.** One document per `(appName, eventType, incidentKey)` where `incidentKey = boot_id + hourly bucket`. Repeats `$inc` a `count` and advance `lastSeen` instead of inserting a new row, so a retry storm collapses to a single deduplicated incident while distinct incidents in later hours stay separate.
- **Attribution stamped at write time.** Every incident carries node identity (`nodeTxid`/`nodeOutidx`/`nodeIp`), operator identity (`pubkey`/`paymentAddress`), and app attribution (`ownerZelid`/`appHash`). Identity is best-effort with daemon backoff; a one-shot backfill enriches anything written before `fluxd`'s RPC was up.
- **Weighted score over distinct incidents.** `computeTamperScore` sums a per-incident `severity` (stamped at write time) over `schemaVersion >= 1` documents: `container_vanished` = 3 (a container positively confirmed missing is the strongest local interference signal — containers persist as `exited` across a clean reboot); mount/volume/network types = 1; `recreation_failed` = 0 (operational: registry/image/disk).
- **Startup purge of pre-schema rows.** On boot the service removes any document without a `schemaVersion`, so an upgraded node presents an honest collection from its first boot rather than serving a month of un-attributed, un-deduplicated rows until they TTL.
- **Capped public API.** `GET /apps/tamperingevents/:appname?` returns incidents most-recent first, default 500 / max 1000 (`?limit=`), with `_id` as a stable paging key. The `appname` filter is string-coerced so a bracket-notation query can't inject a Mongo operator.

## Why

The existing feature collects data too noisy and too unattributable to act on, which is why its enforcement path is dormant behind a manually-curated, currently-empty blocklist. Two problems dominate:

1. **The score can't tell tampering from operational churn.** Scoring on `countDocuments({})` counts every row of every type, undifferentiated and un-deduplicated, so a single flapping app or one routine reboot can dominate the count.
2. **Reboot detection measures the wrong thing.** A wall-clock "restart gap" heuristic is blind to a slow reboot cadence, false-positives on crash loops, and is evadable by clock skew.

This branch fixes both: `boot_id` gives a reboot signal that can't be faked by clock skew or confused by process restarts, and the dedup + severity + identity schema turns the raw log into per-incident, per-operator data a fleet-level system can trust.

## Fleet evidence

The scoring design was validated against two full-network scans (tooling and raw results in `dev/tampering/`).

**Noise is concentrated, not fleet-wide (full-network scan, 7,538 nodes).** 96% of all `mount_vanished` events came from just 30 nodes running structural FoldingAtHome churn; a single node emitted **21,724** events in 46 hours (the sweep re-recording every ~30s for the same missing directory). This is exactly the row-per-observation noise the old `countDocuments` score could not distinguish from tampering — the incident rollup collapses each such fault to **one document plus a count**.

**Benign boot-clustered noise is negligible (full-network scan, 6,561 nodes).** Correlating every event timestamp against each node's boot time (scan time − `/flux/systemuptime`), of **2,901 nodes rebooted within the 30-day retention window** (2,599 ArcaneOS / 302 legacy), only **6 (0.21%)** showed a genuine benign mount/volume event in the first 30 minutes after boot — each affecting **1–2 apps**, one event apiece, all self-healing, none recurring. Fleet-wide, `container_vanished` had **1** in-window event out of 14,906. This is why boot proximity needs no special handling in scoring: a benign post-boot blip costs an honest node at most ~2 points for 30 days (dedup), while persistent damage keeps re-recording at full weight in every hourly bucket — the score naturally separates the two.

## Safety

Nothing in this PR can DOS a node. Enforcement (`appTamperingBlocklistService`) is **double-gated**: a node self-limits only if its txhash is on the manually-curated blocklist (currently `[]`) **AND** its score exceeds the threshold, and enforcement is skipped entirely on ArcaneOS. The scoring/schema changes here do not touch that gate; they only make the underlying data trustworthy enough to eventually drive it.

## Scope

Data-quality and attribution only — an honest local event log and an incident schema a fleet-level system can consume. The network-wide per-operator aggregate and any enforcement redesign are **not** in this PR, and enforcement stays double-gated behind the empty blocklist.

## Follow-up

A small number of emitter-side refinements are intentionally out of scope here: their call sites are in the reconciler, which is under active restructuring on the v9 branch. They'll land in a follow-up PR once that work settles, to avoid conflicting with it.

## Testing

- Unit suites for both services green (86 tests), lint clean.
- Adjacent call-site suites (appReconciler, crontab/mounts cleanup, syncthing state machine) green.
- Deployed and exercised on live nodes: real tampering (container/network/mount/volume removal) recorded and healed; cold-migration from the current release verified (pre-schema purge, marker upgrade, index build, `bootHistory` seeding); reboot path recording confirmed.
- Enforcement-safety paths (empty blocklist, benchmark unavailable, daemon not synced, txhash undetermined, ArcaneOS) covered by the blocklist suite.

### DIFF
--- a/ZelBack/src/services/appTamperingBlocklistService.js
+++ b/ZelBack/src/services/appTamperingBlocklistService.js
@@ -6,30 +6,13 @@ const fluxNetworkHelper = require('./fluxNetworkHelper');
 const generalService = require('./generalService');
 const daemonServiceMiscRpcs = require('./daemonService/daemonServiceMiscRpcs');
 const benchmarkService = require('./benchmarkService');
+const appTamperingDetectionService = require('./appTamperingDetectionService');
 
 const BLOCKLIST_URL = `${config.github.rawBaseUrl}/helpers/tamperingblockednodes.json`;
 const CHECK_INTERVAL_MS = 12 * 60 * 60 * 1000; // 12 hours
 const SYNC_POLL_MS = 60 * 1000; // 60s while waiting for daemon sync
 const TAMPER_SCORE_THRESHOLD = 10;
 const DOS_MESSAGE_PREFIX = 'Node flagged via tampering blocklist';
-
-// Per-type weights for the tamper score. Not every event is tamper evidence:
-// recreation_failed is an operational fault (registry, image pull, disk) and
-// node_reboot is an observation feeding operator-level analysis — both weigh 0
-// via omission, because punishing them would penalize honest nodes for infra
-// noise. container_vanished weighs highest: containers persist as `exited`
-// across a clean reboot, so a positively-confirmed missing container is the
-// strongest local indicator of host-side interference this data has. The
-// mount/volume/network types are real signals but dominated by late-data-disk
-// and network-not-ready boot races, so they weigh low until the phase-1
-// schema can flag boot-storm context explicitly.
-const TAMPER_EVENT_WEIGHTS = {
-  container_vanished: 3,
-  network_pruned: 1,
-  network_detached: 1,
-  mount_vanished: 1,
-  volume_missing: 1,
-};
 
 const tamperingEventsCollection = config.database.local.collections.appTamperingEvents;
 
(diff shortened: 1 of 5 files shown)


Cabecinha84 (approved):
ack

PR #1769, opened 2026-07-21
fix(apps): recreate a pruned per-app docker network in the reconciler heal path

MorningLightMountain713:
## Summary

App containers can get stuck in docker `created` state, looping indefinitely on `failed to set up container networking: network fluxDockerNetwork_<app> not found`, when the app's per-app docker network goes missing (e.g. `docker network prune`, a docker daemon restart). The per-app network was only ever created at install time (`registerAppLocally`); nothing in the reconciler/heal loop recreated it, so `recreateMissingContainers` and every container start failed and retried forever. On the vanished-container path the repeated failure eventually escalates to `removeAppLocally`.

## Fix

- **`containerHealthMonitor.recreateMissingContainers` now ensures the app network exists before recreating any container.** This is the core fix — both the vanished-container path and the network-detach heal path funnel through `recreateMissingContainers`, so a pruned network is recreated before `appDockerCreate` instead of looping on "network not found".
- **`ensureAppDockerNetwork` extracted from `registerAppLocally`** — the single creator for install and heal:
  - *Idempotent* — returns early when the network exists, doing no allocation and no `DOCKER-USER` iptables rebuild (an intact interface is already in the chain), so a heal recreate on an existing network is cheap.
  - *Deterministic but collision-safe allocation* — lowest-free `172.23.x.0/24`, scanning **all** docker networks (docker enforces subnet uniqueness globally), and advancing past any octet a create loses to a concurrent allocation. It gives up only on genuine octet exhaustion — never a fixed retry count, because a premature throw on the vanished path would escalate to an app uninstall while subnets were still free.
  - *Reserves the legacy-pinned octets* in the free-octet scan so a non-legacy app cannot squat an octet a legacy gateway-pinned app must heal onto.
- **`dockerService.getFreeFluxAppNetworkOctet(excludeOctets)`** — lowest-free octet across all networks, with a caller exclude set for the collision retry.
- **`softRegisterAppLocally` (soft-install / soft-redeploy) now delegates to `ensureAppDockerNetwork`** instead of a near-duplicate allocator. The duplicate still used `Math.random()*256` (which could pick octet `0`, colliding with the base network) and a fixed 21-try cap that could throw while subnets were still free and escalate a soft-register to an uninstall — the exact fragility this PR argues against. One creator now serves install, soft-install, and heal; the install stream messages are unchanged.
- Legacy gateway-pinned apps (`appsThatMightBeUsingOldGatewayIpAssignment`) keep their name-derived octet, unchanged. `appReconciler.js` is unchanged.

## Testing

- **Verified end-to-end on a live node.** With the fix deployed, a running app's container was removed and its per-app docker network pruned, then FluxOS was restarted. The reconciler detected the vanished container, recreated the missing network via `ensureAppDockerNetwork`, and restarted the container within ~5s — the app returned to `running`. The healed (non-legacy) app was re-assigned a fresh lowest-free octet, confirming the deterministic-allocator path; this is octet-independent for correctness (host port mappings and the `172.23.0.0/16` firewall coverage do not depend on the octet).
- Unit coverage across `appInstaller`, `containerHealthMonitor`, `dockerService`: network-ensured-before-recreate; deterministic lowest-free allocation; non-flux networks counted; exclude-set advancement; exhaustion → clear error; idempotent exists path does no create and no firewall rebuild; `unknown` docker-network state falls through to an idempotent create; **legacy-octet reservation; exists-path status stream**. The exhaustion test is **mutation-verified** — reintroducing a fixed retry cap makes it fail.
- All affected unit suites green with no regression: appInstaller (incl. 2 new cases), containerHealthMonitor, dockerService, appReconciler, advancedWorkflows (51 — confirms the soft-register consolidation is a behaviour-preserving delegation), appSpawner, peerNotification.
- Hardened under two adversarial review passes plus maintainer review; the allocator's concurrency behaviour (no false failure / no spurious uninstall during a mass-prune recovery), the all-networks scan, and the no-firewall-on-exists path were added in response.

## Deferred follow-ups (not in this PR)

- The per-app subnet allocator is a hand-rolled reimplementation of docker's native IPAM, kept only for legacy gateway pinning and the `172.23.0.0/16` firewall coupling. Replacing it with docker `default-address-pools` allocation is a separate cross-repo change (node daemon.json + careful rollout ordering).
- The `DOCKER-USER` rebuild runs once per created network during a mass-prune recovery (N healing apps → N rebuilds, each a brief protection drop). Inherent to the per-app network design and not worsened here; a batching opportunity that folds naturally into the IPAM migration above.

Written against commits `5f24a94` + `4db9b3f`.

### DIFF
--- a/ZelBack/src/services/appLifecycle/advancedWorkflows.js
+++ b/ZelBack/src/services/appLifecycle/advancedWorkflows.js
@@ -41,9 +41,6 @@ const appNetworkLinker = require('./appNetworkLinker');
 
 const isArcane = Boolean(process.env.FLUXOS_PATH);
 
-// Legacy apps that use old gateway IP assignment method
-const appsThatMightBeUsingOldGatewayIpAssignment = ['HNSDoH', 'dane', 'fdm', 'Jetpack2', 'fdmdedicated', 'isokosse', 'ChainBraryDApp', 'health', 'ethercalc'];
-
 // Master/slave app tracking
 const mastersRunningGSyncthingApps = new Map();
 const timeTostartNewMasterApp = new Map();
@@ -921,47 +918,14 @@ async function softRegisterAppLocally(appSpecs, componentSpecs, res) {
     await appNetworkLinker.checkAppNetworkRequirements(appSpecifications);
 
     if (!isComponent) {
-      let dockerNetworkAddrValue = Math.floor(Math.random() * 256);
-      if (appsThatMightBeUsingOldGatewayIpAssignment.includes(appName)) {
-        dockerNetworkAddrValue = appName.charCodeAt(appName.length - 1);
-      }
-      const fluxNetworkStatus = {
-        status: `Checking Flux App network of ${appName}...`,
-      };
-      log.info(fluxNetworkStatus);
-      if (res) {
-        res.write(serviceHelper.ensureString(fluxNetworkStatus));
-        if (res.flush) res.flush();
-      }
-      let fluxNet = null;
-      for (let i = 0; i <= 20; i += 1) {
-        // eslint-disable-next-line no-await-in-loop
-        fluxNet = await dockerService.createFluxAppDockerNetwork(appName, dockerNetworkAddrValue).catch((error) => log.error(error));
-        if (fluxNet || appsThatMightBeUsingOldGatewayIpAssignment.includes(appName)) {
-          break;
-        }
-        dockerNetworkAddrValue = Math.floor(Math.random() * 256);
-      }
-      if (!fluxNet) {
-        throw new Error(`Flux App network of ${appName} failed to initiate. Not possible to create docker application network.`);
-      }
-      log.info(serviceHelper.ensureString(fluxNet));
-      const fluxNetworkInterfaces = await dockerService.getFluxDockerNetworkPhysicalInterfaceNames();
-      const accessRemoved = await fluxNetworkHelper.removeDockerContainerAccessToNonRoutable(fluxNetworkInterfaces);
-      const accessRemovedRes = {
-        status: accessRemoved ? `Private network access removed for ${appName}` : `Error removing private network access for ${appName}`,
-      };
-      if (res) {
-        res.write(serviceHelper.ensureString(accessRemovedRes));
-        if (res.flush) res.flush();
-      }
-      const fluxNetResponse = {
-        status: `Docker network of ${appName} initiated.`,
-      };
-      if (res) {
-        res.write(serviceHelper.ensureString(fluxNetResponse));
-        if (res.flush) res.flush();
-      }
+      // One creator for install and heal: idempotent, collision-safe, and
+      // exhaustion-bounded. Replaces the old random-octet + fixed-21-retry loop
+      // that could pick octet 0 (colliding with the base network) and throw while
+      // subnets were still free, escalating a soft-register to an uninstall.
+      // Dynamic require because appInstaller <-> advancedWorkflows is a circular
+      // dependency (dynamically required throughout this file); to be untangled in v9.
+      const appInstaller = require('./appInstaller');
+      await appInstaller.ensureAppDockerNetwork(appName, res);
     }
 
     const appInstallation = {
--- a/ZelBack/src/services/appLifecycle/appInstaller.js
+++ b/ZelBack/src/services/appLifecycle/appInstaller.js
@@ -44,6 +44,11 @@ const volumeService = require('../utils/volumeService');
 // Legacy apps that use old gateway IP assignment method
 const appsThatMightBeUsingOldGatewayIpAssignment = ['HNSDoH', 'dane', 'fdm', 'Jetpack2', 'fdmdedicated', 'isokosse', 'ChainBraryDApp', 'health', 'ethercalc'];
 
+// The octets those legacy apps pin by name (charCodeAt of the last character).
+// Reserve them in the free-octet scan so a non-legacy app can't take one before
+// the legacy app heals onto its fixed octet.
+const legacyPinnedOctets = appsThatMightBeUsingOldGatewayIpAssignment.map((name) => name.charCodeAt(name.length - 1));
+
 // Helper functions and constants for installApplicationHard
 const util = require('util');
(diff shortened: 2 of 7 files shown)


Cabecinha84 (approved):
Verdict: Solid, well-scoped, and correct. No blocking defects. The fix addresses a real bug I confirmed against the source, the allocator logic is sound, and the safety reasoning holds up. Below are the verification results plus a few opportunities.

  The bug is real (verified)

  The claim checks out end-to-end:
  - The heal paths call recreateMissingContainers → installApplicationSoft/Hard, and those functions call dockerService.appDockerCreate but never create the per-app network (appInstaller.js:819, 895). The network was only ever created in registerAppLocally.
  - So a pruned fluxDockerNetwork_<app> makes every recreate fail on "network not found."
  - On the vanished-container path this is genuinely dangerous: appReconciler.recreateMissing escalates a persistent failure to appUninstaller.removeAppLocally (appReconciler.js:396). A pruned network could therefore uninstall a healthy app. That elevates this from "annoying loop" to "data-loss-adjacent,"
  which justifies the care taken.

  Both heal entry points (recreateMissing line 369, and recreateForNetworkHeal with softOnly line 416) funnel through recreateMissingContainers, so the single ensureAppDockerNetwork call at the top covers both. Correct placement.

  The allocator is correct

  - Termination is bounded. Each lost create adds its octet to tried, and getFreeFluxAppNetworkOctet(tried) excludes those, so the loop advances monotonically and terminates in ≤255 iterations or at null. It cannot spin forever.
  - The removed fixed retry-cap is the right call. A premature throw would escalate to uninstall while subnets were still free — the JSDoc explains exactly this, and the mutation-verified exhaustion test locks it in.
  - Global scan is correct. Docker enforces subnet uniqueness across all networks, so scanning non-flux networks too (not just getFluxDockerNetworkSubnets) is necessary and well-documented.
  - Docker-down doesn't produce a misleading error, because getFreeFluxAppNetworkOctet calls docker.listNetworks() uncaught — a daemon outage rejects there and propagates a real docker error, rather than looping 255 times to a false "no free subnet."

  Deterministic re-assignment is safe

  Moving from random octet to lowest-free means an app can get a different subnet after a prune. I verified nothing depends on the specific octet:
  - Firewall rules operate on the whole 172.23.0.0/16 block (fluxNetworkHelper.js:2047, 2153), not per-app octets.
  - Inter-component / networkWith links resolve by docker DNS container name, not IP (appNetworkLinker.connectComponentToLinkedApps).
  - The idempotent exists-path skipping the DOCKER-USER rebuild is justified — the interface is already in the chain; rebuilding would briefly drop RFC1918 protection for every flux container for no gain.

  Tests

  containerHealthMonitor suite passes fully here (12/12), including the new "ensures the app docker network before recreating any container" test with a calledBefore ordering assertion. The appInstaller/dockerService suites I couldn't run to completion in this checkout — but only for environmental reasons
  (missing local ZelBack/config/userconfig.js and a slow/hanging integration-style proxyquire setup), not PR defects. The new test code itself reads correctly and the coverage matrix (collision advance, exhaustion, legacy pin, unknown state, non-flux networks, exclude set) is genuinely thorough.

  Opportunities / minor observations

  1. The duplicated allocator in advancedWorkflows.softRegisterAppLocally (lines 936–947) still has both flaws this PR fixed — the Math.random()*256 octet (which can even pick octet 0, colliding with the base network) and the fixed 21-iteration cap. And its failure also escalates to removal (REMOVAL REASON: 
  Soft registration failure, line 1090). The PR explicitly defers consolidating this, which is a reasonable scope boundary — but it's the highest-value follow-up, because the same "fixed cap → throw → uninstall" fragility the PR argues against still lives on that path. Worth a tracking issue so it isn't
  forgotten.
  2. Legacy pinned octets aren't reserved by the free-octet scan. On a densely-packed node, a non-legacy app could claim a legacy app's charCodeAt-derived octet (typically 97–122) before that legacy app heals, causing the legacy heal to throw. In practice this only bites past ~96 apps on a node, and it's not a
  regression (random allocation had the same latent collision) — but excluding the legacy octet set inside getFreeFluxAppNetworkOctet would close it cheaply.
  3. Firewall rebuild runs once per created network during a mass-prune recovery. N healing apps → N full DOCKER-USER rebuilds, each a brief protection drop. Inherent to the per-app design and not made worse here; just noting it as a batching opportunity for a future pass.
  4. Cosmetic: on the exists-early-return, the streamed "Checking Flux App network…" status is emitted but the closing "Docker network initiated." isn't — a negligible install-stream asymmetry.

MorningLightMountain713:
Thanks for the thorough review, @Cabecinha84 — especially independently verifying the uninstall-escalation and the octet-independence. Addressed the actionable items in `4db9b3f`:

**1. Duplicate allocator in `softRegisterAppLocally`** — consolidated. It now delegates to `ensureAppDockerNetwork`, so the `Math.random()*256` (octet-0 collision) + fixed-21-cap → throw → uninstall path is gone; install, soft-install, and heal share the one idempotent, collision-safe, exhaustion-bounded creator. The now-dead legacy list in `advancedWorkflows.js` is removed as well.

**2. Legacy octets not reserved by the free-octet scan** — fixed. `ensureAppDockerNetwork` seeds the exclude set with the legacy `charCodeAt` octets, so a non-legacy app can't claim a legacy app's pinned octet before it heals onto it. New unit test covers it.

**3. N firewall rebuilds during a mass-prune recovery** — agreed it's inherent to the per-app design and not worsened here, so I left it as a documented deferred follow-up. It folds into the docker-native IPAM migration, where the whole allocator (and this rebuild) go away.

**4. Cosmetic exists-path stream asymmetry** — fixed; the exists early-return now emits an "already exists" status. New unit test.

New coverage is the two `ensureAppDockerNetwork` cases (legacy-octet reservation, exists-path status); `advancedWorkflows` (51) stays green, confirming the soft-register consolidation is a behaviour-preserving delegation.

Also validated the heal end-to-end on a live node: pruned the per-app network + removed the container, restarted FluxOS, and the reconciler recreated the network via `ensureAppDockerNetwork` and restarted the container in ~5s.

PR #1771, opened 2026-07-25
fix(apps): unwedge the container reconciler, close the terminal crash paths, and make masterSlave elections resilient

Cabecinha84:
## What happened in production

On 2026-07-25 a customer's g:-synced app was down for ~8 hours across two nodes. The root cause was the container reconciler wedging: a recreate whose docker pull never settled held the per-app single-flight guard forever, silently — no retry, no log, no recovery. Two other real defects rode along (a terminal-exec crash and a masterSlave election loop that died on its first throw), but the wedge was the outage.

While fixing the wedge we found that the failure paths *around* it were worse than the wedge itself: four separate code paths could **delete an established app and its data** in response to what is usually a transient, node-local failure. This PR closes all of them.

## The core contract this PR establishes

**A failure never destroys an established app's data.** A failed recreate, redeploy, or update converges to *kept-down-with-data*: the app stays in the local DB with its volume intact, and the reconciler retries it on a backoff ladder. A down instance stops broadcasting `apprunning`, so the network replaces it elsewhere through the normal spawner machinery — removal is never needed for availability. Only an app that never ran on a node (nothing to preserve) may be cleaned up.

### Container reconciler
- A recreate's provision runs under a hard ceiling, so a hung pull can no longer wedge the per-app single-flight forever; a superseded pass can never write state over its replacement.
- Docker pulls are judged by **silence, not elapsed time** — a slow-but-progressing pull of a large image is healthy; only a stalled stream trips the ceiling.
- The deletion gate **fails closed**: a runtime-state read error, a provision timeout, or a docker-runtime timeout is never grounds for removal. An app that has ever started on this node is kept on failure, always.
- The appdata wipe requires a positive answer that nothing is running (unreachable/indeterminate container state aborts the wipe), and a detached provision can no longer race it.
- Short docker calls (lookup, inspect, list, image-size) are bounded so a wedged daemon fails a pass quickly instead of hanging it. Container/network creates are deliberately **not** bounded — a timeout there is not a refusal, and acting as if it were caused duplicate-resource bugs; the comments carry the reasoning.

### Image verification and the update path
- `ImageVerifier` classifies its own failures (transient network/rate-limit/server errors vs. permanent refusals), and a request that received no response at all is a *connectivity* answer, not a verdict. A recreate rides out a registry outage on the locally present image instead of failing the app.
- The mutable-tag image updater now **verifies the new image before anything is torn down** (compliance plus per-component size/arch checks). Previously a digest change triggered the teardown first and verified after — a bad push or a registry error at the wrong moment removed the app and its data. A registry rate-limit aborts the whole cycle rather than burning the budget across apps.

### Soft redeploy — the keep contract
- A failed soft redeploy **never removes the app**. All three failure sites (the redeploy driver, the registration step, the block-driven spec-update path) keep the app and its data for the reconciler.
- The redeploy's own teardown deletes the app's local DB row and the registration step re-inserts it minutes later; a failure inside that window used to hit "no row → remove with force". The row is now **restored** in that window (the spec being applied is re-inserted), because a rowless app is invisible to every recovery sweep. Only an app that was never locally installed falls back to cleanup.
- Concurrency collisions (another install or removal holding the operation flag when the registration runs) are **typed** and treated as deferrals, not failures: the spec-update path no longer force-removes an app over a timing collision, and a collision can never clear another operation's flag.
- The kept app's synced components are re-marked so the sync layer's first-encounter handling cannot clear their data; a mark is never written for an app the node doesn't hold.
- Every busy-guard early return across the four redeploy entry points now closes its streamed API response (previously the client hung to a gateway timeout).

### Syncthing data-safety guards
- The phantom-index discriminator (which demotes a sendreceive folder whose index claims data over an empty disk — the guard that stops a wiped node broadcasting deletions to healthy replicas) is now **files-aware**: an index claiming files over a disk holding only a directory skeleton is correctly caught, while a dirs-only payload (the 2026-07-04 false positive) stays healthy. The sync-status plumbing carries `globalFiles` through so the discriminator actually sees it.
- The disk walk's exclusions are **root-anchored**, matching the `.stignore` FluxOS writes (`/backup` is anchored): files under a *nested* directory named `backup` are synced payload and now count, so such an app is no longer falsely held down as a phantom.
- Creating a fresh volume **invalidates any surviving synced-mark** for that component — at the point of no return, not at entry, so an aborted pre-flight (e.g. out of space) preserves the mark protecting the intact existing data.

### masterSlave scheduler
- Election passes are abandoned per-pass: a stuck pass can no longer kill the scheduler loop or have its stale writes land over a newer pass.
- A heartbeat watchdog judges passes by **silence**, on monotonic time; a legitimately long pass that keeps making progress is left alone.
- The permissions-fix chain is single-flight per app with a hard time budget, and its result is actually read.

### Terminal (socket.io exec)
All four crash paths closed: EPIPE on a flooded pty stream, the hijacked stream's `error` event, a client disconnecting before exec setup completes, and deferred callbacks firing after teardown.

### Test-event stream
Event ids are seeded from the monotonic clock, so a FluxOS restart mints ids above anything previously served and last-event-id consumers don't silently discard every post-restart event.

## Verification

- **441 unit tests** across the touched suites, every behavioral fix proven to fail with the fix reverted (both directions where a placement was in question). Tests that route through catches carry explicit reached-the-path assertions so they cannot pass vacuously.
- **Full 54-suite integration gate: 54/54 pass, clean run**, including new system-level cases pinning the registry-verdict keep, the slow-registry election budget, the live-session terminal crash, the phantom-index guard, and volume-mount ownership across reboots.
- The branch went through repeated independent adversarial reviews to convergence; each round's findings were verified against the code before being fixed, and each fix landed with its own pinning test.

## Notes for reviewers

- **v9 relationship:** `feat/fluxos-v9` rebases on top of this branch after merge. Most of this PR's redeploy hardening maps onto structure v9 already has (per-app operation leases, no rowless window, native keep-on-failure); the two genuine carry-overs (the fresh-volume mark invalidation and the root-anchored walk) are flagged in their commit messages.
- **Documented, deliberately not built here:** the image-update pre-flight cannot cover the install's node-side gates (hardware, network requirements, port mapping — the mapping *is* the action). With keep-on-failure, missing them costs a torn-down app waiting on the reconciler's ladder, never data. The unbounded create under the fresh-install path (`registerAppLocally`) pins the installation flag until process restart and predates this branch; it is owned by the operation-registry redesign.

*Description written against `01a7dec89`.*

### DIFF
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -359,6 +359,10 @@ module.exports = {
     masterSlaveStaggerMs: 180000, // per-index delay before a standby may claim an unclaimed primary
     masterSlaveProbeTimeoutMs: 10000, // probing the previous primary's /apps/listrunningapps
     masterSlaveFdmTimeoutMs: 10000, // FDM /appips lookup
+    // cap on a reconciler recreate's provisioning (registry verify + image pull): a
+    // black-holed registry must fail the recreate, not wedge the component's
+    // reconcile single-flight for the TCP stack's worst case
+    recreateProvisionCapMs: 180000,
   },
   lockedSystemResources: {
     cpu: 10, // 1 cpu core
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -355,7 +355,11 @@ module.exports = {
     imageUpdateDelayAfterRedeployMs: 120000,
     imageUpdateDelayBetweenComponentsMs: 1000,
     masterSlaveIntervalMs: 30000, // masterSlave (g:) FDM election cycle
-    masterSlaveMaxPassMs: 120000, // abandon a masterSlave pass wedged this long, so the next cycle still runs
+    masterSlaveMaxPassMs: 120000, // abandon a masterSlave pass SILENT this long (heartbeat-based), so the next cycle still runs
+    // The primary-promotion chmod -R over the whole appdata tree: generous (a cold
+    // walk of a huge tree is legitimate work) but finite, so a wedged walk surfaces
+    // as a failed fix instead of running forever.
+    permissionsFixTimeoutMs: 3600000,
     masterSlaveStaggerMs: 180000, // per-index delay before a standby may claim an unclaimed primary
     masterSlaveProbeTimeoutMs: 10000, // probing the previous primary's /apps/listrunningapps
     masterSlaveFdmTimeoutMs: 10000, // FDM /appips lookup
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -355,6 +355,10 @@ module.exports = {
     imageUpdateDelayAfterRedeployMs: 120000,
     imageUpdateDelayBetweenComponentsMs: 1000,
     masterSlaveIntervalMs: 30000, // masterSlave (g:) FDM election cycle
+    masterSlaveMaxPassMs: 120000, // abandon a masterSlave pass wedged this long, so the next cycle still runs
+    masterSlaveStaggerMs: 180000, // per-index delay before a standby may claim an unclaimed primary
+    masterSlaveProbeTimeoutMs: 10000, // probing the previous primary's /apps/listrunningapps
+    masterSlaveFdmTimeoutMs: 10000, // FDM /appips lookup
   },
   lockedSystemResources: {
     cpu: 10, // 1 cpu core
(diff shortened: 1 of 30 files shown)


MorningLightMountain713:
Superseeded

PR #1772, opened 2026-07-27
fix(apps): stop a terminal crashing FluxOS, and let a stopped g: app come back

MorningLightMountain713:
> **Status: full parallel gate PASSED against `8ecfe4529` — all 54 suites green.** Description updated against `1bc6cf644`, which includes the review fixes.

## What happened

Four customer `g:` (masterSlave) apps — four different owners — went down or silently lost redundancy in a single morning. Two were offline for 12 and 25 hours. All four traced to the same mechanism, and none of them recovered on their own.

## Root cause

An `appstop` arrived without a matching `appstart`. That is a legitimate API call and FluxOS did exactly what it was told: it wrote the durable `operatorStopped` lock and the election loop began skipping that component.

Everything after that was FluxOS being unable to recover from a normal operator action:

**The exclusion is silent.** A `g:` component can sit unelected indefinitely while the election loop emits nothing about it. In the logs that is indistinguishable from a loop that has died. Two separate investigations, with full node access and complete logs, read that silence as a wedged scheduler.

**The last primary cannot re-elect itself.** The loop remembers the primary it last saw; when that is *this node*, it is disqualified from the no-history start (which requires no remembered primary) **and** from the previous-primary branch (which requires the remembered primary to be a different node). Nothing evicts that record while the app stays installed, so only a process restart cleared it.

**With every instance in that state the app cannot come back at all.** That is what kept two apps down for 12 and 25 hours; recovery needed a hand-run FluxOS restart on a customer node.

**Opening a terminal on a stopped container killed the FluxOS process.** This is the one defect here that takes down a *node* rather than an app, and it is the one that defeated a customer's own recovery — so it is worth separating from the rest. Of the six stop events traced across the four apps, exactly one had a crash beside it (verified from systemd on each node); the other five ran continuously through the stop. But that one mattered disproportionately: `container.exec` hands back a null exec when the daemon rejects the create — which it does whenever the container is not running — and `exec.start` on null threw an unhandled TypeError out of a socket.io callback. In one incident this landed between a customer's stop and their retry: the node was down for the 31 seconds systemd took to restart it, their restart click failed at the proxy, and because the stop had already set the durable lock, nothing brought the app back.

**And a latent one, found by review rather than incident:** the index-0 fast path started without checking whether a peer was already running. Registration in the routing layer lags a node actually starting by up to ~110 s (measured), and throughout that window it reports no primary while an instance is live — so a node restarting inside another node's promotion window could start a second writer on a shared syncthing volume.

## What this PR changes

Three production files, plus CI and the harness's own config.

| fix | file |
|---|---|
| Four terminal crash paths guarded — null exec, null stream, writes to a destroyed stream, unhandled stream `error` | `dockerTerminalHandler.js` |
| A terminal session is never recorded as opened when the client leaves mid-setup | `dockerTerminalHandler.js` |
| A failed container lookup records its cause, so an unreachable daemon is distinguishable from an absent container | `dockerTerminalHandler.js` |
| Operator-stop exclusion announced, latched so it is affordable at election cadence | `advancedWorkflows.js` |
| Stale self-naming primary record evicted, so a stopped last-primary can be elected again | `advancedWorkflows.js` |
| Index-0 start gated on an all-peer probe, run concurrently | `advancedWorkflows.js` |
| Peer containers matched by whole name — a substring test also matches a longer app whose name merely begins the same way | `advancedWorkflows.js` |
| One settled `g:` app no longer costs every later one its cycle (`return` where `continue` was meant) | `advancedWorkflows.js` |
| Peer-probe cancel timers cleared once the request settles | `advancedWorkflows.js` |
| A freshly created volume invalidates any surviving synced-mark | `advancedWorkflows.js` |
| Event ids seeded from the monotonic clock so they survive a restart; ring wraparound keyed on a real write count | `fluxEventBus.js` |
| Runner dependencies installed with `npm install` — `package-lock.json` is gitignored, so `npm ci` had nothing to read | `integration-harness.yml` |
| Harness runs no longer report suite activity to the live analytics service | `test-infra/config/shared.js` |

The eviction and the peer check ship together deliberately: the eviction puts nodes back on the fast path, and the fast path is the unguarded one. Either alone is worse than both.

The synced-mark fix is placement-critical. It drops the mark immediately before the allocation, **not** at function entry — the pre-flight above it can abort with the existing volume and its data untouched, and stripping the mark there hands intact data to the skip/second-encounter chain, which clears it. Entry placement turns a survivable failure into data loss. Both directions are pinned by tests.

The `return`/`continue` fix is pre-existing on `development` and not introduced by this branch. It is fixed here because it is the same silent-skip failure the rest of this PR exists to remove: one app whose previous master is still running abandoned the whole pass, skipping every remaining `g:` app for that cycle.

## Verification

- Unit tests across the touched suites, **each fix prove-failed** — reverted individually and confirmed to fail for the right reason. The synced-mark pair discriminates in both directions: remove the fix and the invalidation test fails; move it to entry and the preservation test fails.
- Two new harness suites, both green:
  - terminal crash safety — container-not-running and container-absent, asserting a socket error reaches the client *and* the process survives
  - operator-stop recovery — an operator-stopped instance stays down, and both-stopped-then-started elects exactly one again (the multi-hour outage, reproduced)
- eslint clean on every touched file.
- **Full parallel gate against `8ecfe4529`: PASSED, all 54 suites green.** (56 files run as 54 suites; `28-*` and `32-*` each own two files.)

Two suites were red on `development` before this branch and are green here, both for reasons worth stating:

**`60-volume-mount-ownership`** was failing its post-restart case under full-gate load while passing in isolation — which looks exactly like contention and is not. The reconciler was emitting `volumeUnavailable` on schedule every time; the *consumer* could not see it, because event ids restarted at 1 and every post-restart event fell below the id the test had last seen. The busier the run, the higher that id, hence the load correlation. Fixed by the event-id seed above; this is a real defect for any SSE client using `Last-Event-ID`, not a test artefact.

**`61-syncthing-mount-safety`**'s phantom-index case is now **pending, not failing**. It documents a genuine product gap — `checkDirectoryHasSyncScopedContent` walks with `countDirs: true`, so an emptied-but-present `appdata/` counts as content and a stale index over an empty volume is never detected. That behaviour was itself a deliberate 2026-07-04 fix for the opposite false positive, so this is a trade-off rather than a regression, and it is unrelated to this PR (verified by running the suite against unmodified `development`, which fails identically). Marked pending with the reasoning inline rather than deleted, because deleting it erases the only record of the gap.

**Already resolved on v9**, which splits the check in two: `checkDirectoryHasSyncScopedContent` keeps `countDirs: true`, and a sibling `checkDirectoryHasSyncScopedFiles` counts files only — the deletion-broadcast hazard is per-file, so a surviving directory skeleton protects nothing. The pending test is the natural proof of that split once v9 lands, which is why it is kept rather than removed.

## Deliberately not in this PR

The real architectural problem is that the routing layer is being used as a lock service. It is eventually consistent, its "no primary" answer is ambiguous between *nobody should run this* and *nobody has reported in yet*, and it forgets a stopped app entirely — which is why each node keeps a private in-memory guess, and why that guess became a deadlock.

Fixing that properly means an owner for "who runs this component" that lives inside FluxOS. That work belongs in v9, where this whole subsystem has already been restructured, and it carries an unsolved problem of its own. It is not appropriate to bolt onto a release as a hotfix, and none of the four incidents were caused by a bad election *decision* — they were caused by an unpaired stop, a crash, and a node that could not re-elect itself. All three are fixed here, locally and cheaply.

A separate design note records what v9 needs after the rebase, including the fact that these fixes live in a file v9 deleted and will be **silently dropped** by a naive rebase.

---

*Description written against `1bc6cf644`.*

### DIFF
--- a/ZelBack/src/lib/socketIoHandlers/dockerTerminalHandler.js
+++ b/ZelBack/src/lib/socketIoHandlers/dockerTerminalHandler.js
@@ -8,74 +8,153 @@ const log = require('../log');
 async function dockerTerminalHandler(socket) {
   const clientIp = socket.handshake.headers['x-forwarded-for']?.split(',')[0]?.trim() || socket.handshake.address;
 
+  // Anything that throws in a socket.io listener is unhandled and takes the whole
+  // FluxOS process down, so every failure here has to leave through
+  // socket.emit('error', ...) instead.
+  //
+  // The try/catch below only covers this listener's own body - the dockerode and
+  // socket callbacks registered inside it run LATER, after the try has exited, so
+  // they are each wrapped in `guard` rather than relying on it.
   socket.on('exec', async (zelidauth, nameOrId, dockerCmd, dockerEnv, dockerUser) => {
-    const auth = {
-      zelidauth,
+    // wrap a callback that runs outside this listener's try/catch
+    const guard = (label, fn) => (...args) => {
+      try {
+        return fn(...args);
+      } catch (error) {
+        log.error(`dockerTerminalHandler: ${label} for ${nameOrId}: ${error.message}`);
+        socket.emit('error', 'Terminal session error.');
+        return undefined;
+      }
     };
-    const container = await dockerService.getDockerContainerByIdOrName(nameOrId);
-    const mainAppName = nameOrId.split('_')[1] || nameOrId;
-    const authorized = await verificationHelperUtils.verifyAppOwnerOrHigherSession(auth, mainAppName);
-    if (authorized !== true) {
-      socket.emit('error', 'Not authorized.');
-      return;
-    }
-    if (!container) {
-      socket.emit('error', 'Container not found.');
-      return;
-    }
-
-    const parts = nameOrId.split('_');
-    const component = parts.length > 1 ? parts[0].replace(/^(zel|flux)/, '') || null : null;
-    const analyticsAppName = parts.length > 1 ? mainAppName : mainAppName.replace(/^(zel|flux)/, '');
-    trackTerminalSession(zelidauth, analyticsAppName, 'open', clientIp, component);
+    // Registered BEFORE the awaits below: a disconnect can land while auth and
+    // the docker lookups are still in flight (deterministically, for a client
+    // that emits 'exec' and closes), and a listener added after socket.io's
+    // single 'disconnect' emit never fires - leaking the hijacked stream and
+    // its exec for as long as the container lives.
+    let execStream = null;
+    let clientGone = false;
     socket.on('disconnect', () => {
-      trackTerminalSession(zelidauth, analyticsAppName, 'close', clientIp, component);
+      clientGone = true;
+      if (execStream) execStream.destroy();
     });
-
-    const cmd = {
-      AttachStdout: true,
-      AttachStderr: true,
-      AttachStdin: true,
-      Tty: true,
-      Cmd: serviceHelper.commandStringToArray(dockerCmd),
-      Env: serviceHelper.commandStringToArray(dockerEnv),
-      User: dockerUser,
-    };
-    container.exec(cmd, (err, exec) => {
-      const options = {
-        Tty: true,
-        stream: true,
-        stdin: true,
-        stdout: true,
-        stderr: true,
-        hijack: true,
+    try {
+      const auth = {
+        zelidauth,
       };
-      socket.on('resize', (data) => {
-        const { rows, cols } = data;
-        exec.resize({ h: rows, w: cols }, () => {
-        });
+      const mainAppName = nameOrId.split('_')[1] || nameOrId;
+      // Authorise BEFORE touching Docker: the lookup below is a remote-controlled
+      // operation on an attacker-supplied name, and must not be reachable by an
+      // unauthenticated caller.
+      const authorized = await verificationHelperUtils.verifyAppOwnerOrHigherSession(auth, mainAppName);
+      if (authorized !== true) {
+        socket.emit('error', 'Not authorized.');
+        return;
+      }
+      // getDockerContainerByIdOrName reads .Id off an undefined lookup result when
+      // the container is absent, so this rejects rather than returning null.
+      const container = await dockerService.getDockerContainerByIdOrName(nameOrId).catch(() => null);
+      if (!container) {
+        socket.emit('error', 'Container not found.');
+        return;
+      }
+
+      const parts = nameOrId.split('_');
+      const component = parts.length > 1 ? parts[0].replace(/^(zel|flux)/, '') || null : null;
+      const analyticsAppName = parts.length > 1 ? mainAppName : mainAppName.replace(/^(zel|flux)/, '');
+      trackTerminalSession(zelidauth, analyticsAppName, 'open', clientIp, component);
+      socket.on('disconnect', () => {
+        trackTerminalSession(zelidauth, analyticsAppName, 'close', clientIp, component);
       });
-      /* eslint-disable no-shadow */
-      exec.start(options, (err, stream) => {
-        if (err) {
-          log.error(err);
-          // socket.emit('error', 'Error executing the command.');
+
+      // the client may have gone away during the awaits above - do not create
+      // an exec nobody is attached to
+      if (clientGone || !socket.connected) return;
+
+      const cmd = {
+        AttachStdout: true,
+        AttachStderr: true,
+        AttachStdin: true,
+        Tty: true,
+        Cmd: serviceHelper.commandStringToArray(dockerCmd),
+        Env: serviceHelper.commandStringToArray(dockerEnv),
+        User: dockerUser,
+      };
+      container.exec(cmd, guard('exec create', (err, exec) => {
+        // dockerode passes back a null exec when the daemon rejects the exec
+        // create (most commonly the container is not running - state created or
+        // exited). Without this guard the code below dereferences null
+        // (exec.start / exec.resize) and the resulting TypeError is thrown from
+        // inside this callback, which is unhandled and crashes the whole FluxOS
+        // process. Fail the terminal session cleanly instead.
+        if (err || !exec) {
+          log.error(`dockerTerminalHandler: exec create failed for ${nameOrId}: ${err ? err.message : 'no exec instance (is the container running?)'}`);
+          socket.emit('error', 'Error opening a terminal. Is the container running?');
           return;
         }
-        stream.on('data', (chunk) => {
-          socket.emit('show', chunk.toString());
-        });
-
-        socket.on('cmd', (data) => {
-          if (typeof data !== 'object') {
-            stream.write(data);
+        const options = {
+          Tty: true,
+          stream: true,
+          stdin: true,
+          stdout: true,
+          stderr: true,
+          hijack: true,
+        };
+        socket.on('resize', guard('resize', (data) => {
+          const { rows, cols } = data;
+          exec.resize({ h: rows, w: cols }, () => {
+          });
+        }));
+        /* eslint-disable no-shadow */
+        exec.start(options, guard('exec start', (err, stream) => {
+          // Same defensive check as above: on failure stream can be null, and the
+          // stream.on(...) below would throw an unhandled TypeError out of this
+          // callback (crashing the process). Bail out cleanly instead.
+          if (err || !stream) {
+            log.error(`dockerTerminalHandler: exec start failed for ${nameOrId}: ${err ? err.message : 'no stream'}`);
+            socket.emit('error', 'Error executing the command.');
+            return;
           }
+          stream.on('data', guard('stream data', (chunk) => {
+            socket.emit('show', chunk.toString());
+          }));
+
+          // The hijacked stream is the raw upgraded docker socket, and node
+          // detaches its own error handler at upgrade - so without this listener
+          // a write racing the exec teardown (EPIPE on a keystroke as the shell
+          // exits) is an uncaught exception that exits the whole process.
+          stream.on('error', guard('stream error', (error) => {
+            log.error(`dockerTerminalHandler: stream error for ${nameOrId}: ${error.message}`);
+            socket.emit('error', 'Terminal session error.');
+          }));
+
+          // Hand the stream to the disconnect teardown registered before the
+          // awaits, and close the race the other way: if the client vanished
+          // while exec.start was in flight, its disconnect has already fired
+          // and nothing else would ever destroy this stream.
+          execStream = stream;
+          if (clientGone || !socket.connected) {
+            stream.destroy();
+            return;
+          }
+
+          // A keystroke racing the container's teardown fails ASYNCHRONOUSLY via
+          // the stream's 'error' event (handled above) - a destroyed socket's
+          // write() just returns false. The type filter is what stops a
+          // non-string payload throwing synchronously out of write().
+          socket.on('cmd', guard('cmd', (data) => {
+            if (typeof data !== 'object') {
+              stream.write(data);
+            }
+          }));
+        }));
+        socket.on('end', () => {
+          log.info('--------end---------');
         });
-      });
-      socket.on('end', () => {
-        log.info('--------end---------');
-      });
-    });
+      }));
+    } catch (error) {
+      log.error(`dockerTerminalHandler: ${nameOrId}: ${error.message}`);
+      socket.emit('error', 'Error opening a terminal.');
+    }
   });
 }
 
--- a/ZelBack/src/lib/socketIoHandlers/dockerTerminalHandler.js
+++ b/ZelBack/src/lib/socketIoHandlers/dockerTerminalHandler.js
@@ -62,8 +62,13 @@ async function dockerTerminalHandler(socket) {
         return;
       }
       // getDockerContainerByIdOrName reads .Id off an undefined lookup result when
-      // the container is absent, so this rejects rather than returning null.
-      const container = await dockerService.getDockerContainerByIdOrName(nameOrId).catch(() => null);
+      // the container is absent, so this rejects rather than returning null. An
+      // unreachable daemon rejects here too and reaches the client as the same
+      // "not found" - log the cause so the two are distinguishable.
+      const container = await dockerService.getDockerContainerByIdOrName(nameOrId).catch((error) => {
+        log.error(`dockerTerminalHandler: container lookup failed for ${nameOrId}: ${error.message}`);
+        return null;
+      });
       if (!container) {
         socket.emit('error', 'Container not found.');
         return;
--- a/ZelBack/src/lib/socketIoHandlers/dockerTerminalHandler.js
+++ b/ZelBack/src/lib/socketIoHandlers/dockerTerminalHandler.js
@@ -33,15 +33,26 @@ async function dockerTerminalHandler(socket) {
     // its exec for as long as the container lives.
     let execStream = null;
     let clientGone = false;
+    let analyticsOpened = false;
+    // Derived synchronously from nameOrId so the disconnect listener below can
+    // own BOTH halves of the analytics pair. A close listener registered later
+    // (after the awaits) can never fire when the client left during them -
+    // socket.io emits 'disconnect' exactly once - which is how an 'open' ends up
+    // recorded with no matching 'close'.
+    const mainAppName = nameOrId.split('_')[1] || nameOrId;
+    const parts = nameOrId.split('_');
+    const component = parts.length > 1 ? parts[0].replace(/^(zel|flux)/, '') || null : null;
+    const analyticsAppName = parts.length > 1 ? mainAppName : mainAppName.replace(/^(zel|flux)/, '');
     socket.on('disconnect', () => {
       clientGone = true;
       if (execStream) execStream.destroy();
+      // pairs whatever was opened, regardless of where in the flow we were
+      if (analyticsOpened) trackTerminalSession(zelidauth, analyticsAppName, 'close', clientIp, component);
     });
     try {
       const auth = {
         zelidauth,
       };
-      const mainAppName = nameOrId.split('_')[1] || nameOrId;
       // Authorise BEFORE touching Docker: the lookup below is a remote-controlled
       // operation on an attacker-supplied name, and must not be reachable by an
       // unauthenticated caller.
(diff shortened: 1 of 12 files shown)


ZelBack/src/services/appLifecycle/advancedWorkflows.js
```diff
@@ -3749,15 +3783,47 @@ async function masterSlaveApps(globalStateParam, installedApps, listRunningApps,
...
+                const checkPeersRunning = async (scope) => {
+                  const limit = scope === 'all' ? runningAppList.length : index;
+                  if (limit <= 0) return false; // not found, or no lower nodes to check
 
                   const { CancelToken } = axios;
                   const timeout = 10 * 1000;
 
-                  // Check all nodes with lower index
-                  for (let i = 0; i < index; i += 1) {
+                  for (let i = 0; i < limit; i += 1) {
```
alihm, line 3847: For an index-0 node, this loop now probes all app locations sequentially with a 10-second timeout each. If an app has many unreachable locations, primary recovery, and processing of every subsequent g: app in this
    single election pass is delayed by up to 10s * peer count, after which the next pass is scheduled. Probe concurrently or impose a pass-wide deadline so a partial network outage cannot stall election for minutes.

ZelBack/src/lib/socketIoHandlers/dockerTerminalHandler.js
```diff
@@ -8,74 +8,153 @@ const log = require('../log');
...
-      Env: serviceHelper.commandStringToArray(dockerEnv),
-      User: dockerUser,
-    };
-    container.exec(cmd, (err, exec) => {
-      const options = {
+      const parts = nameOrId.split('_');
+      const component = parts.length > 1 ? parts[0].replace(/^(zel|flux)/, '') || null : null;
+      const analyticsAppName = parts.length > 1 ? mainAppName : mainAppName.replace(/^(zel|flux)/, '');
+      trackTerminalSession(zelidauth, analyticsAppName, 'open', clientIp, component);
+      socket.on('disconnect', () => {
```
alihm: When the client disconnects while authorization or the Docker lookup is awaiting, the early listener sets clientGone, but execution later records an analytics open before checking that flag. The corresponding close
    listener is registered after Socket.IO's one-time disconnect event, so analytics receives a phantom terminal session that never closes even though no exec was created.

Cabecinha84 (commented):
Issues to raise

  Should be fixed before merge (both already flagged by alihm)

  1. The index-0 probe is sequential with a 10s-per-peer ceiling — advancedWorkflows.js:3825.

  checkPeersRunning('all') walks every entry in runningAppList one at a time. The whole masterSlaveApps pass is a
  sequential loop over every g: app on the node, so one app with unreachable peers delays every g: app behind it. Rough
  worst case on a node with 5 g: apps × 5 locations during a partition: ~200s per cycle — and the cycle re-arms 30s
  after the pass finishes (finally at the end of the function), so it stretches rather than overlaps. That means the
  probe is slowest exactly when recovery matters most.

  alihm's suggestion is right and cheap: fire the probes concurrently and take the first affirmative answer (Promise.all
  + .some, or an any-style race), or impose a pass-wide deadline. Since the probe is already best-effort, concurrency
  costs nothing semantically. I'd rather this land now than as a follow-up, given the whole point of the PR is recovery
  latency.

  2. Phantom analytics session on disconnect-during-await — dockerTerminalHandler.js:64-71.

  trackTerminalSession(zelidauth, analyticsAppName, 'open', clientIp, component);
  socket.on('disconnect', () => { trackTerminalSession(..., 'close', ...); });  // never fires if already disconnected
  if (clientGone || !socket.connected) return;                                  // ← too late

  If the client goes away during auth or the Docker lookup, socket.io's single disconnect has already fired, so the
  close listener never runs — analytics records an open session that never closes. Fix is to move the if (clientGone ||
  !socket.connected) return; above the analytics block. Analytics-only, no crash, but it's two lines.

  Non-blocking — worth noting in the thread

  3. since()'s totalWritten is now meaningless — fluxEventBus.js:57. const totalWritten = this.#nextId - 1; assumed the
  counter started at 1. With the hrtime seed it's ~10¹²–10¹³, so count is always RING_BUFFER_SIZE and startIdx is always
  #writeIndex. I traced the partially-filled case: empty slots are undefined and skipped by the if (entry && ...), and
  iterating 1024 slots from writeIndex still yields oldest→newest, so there is no correctness bug — just a full-buffer
  scan instead of a partial one (negligible) and a variable whose name now lies. A separate #writeCount would remove the
  trap for the next person.

  4. The event-bus unit test is mildly timing-dependent. It asserts a second bus seeds above the first bus's second id,
  i.e. that ≥2µs elapse between two constructions. Two new + two publish + a 1024-slot since() makes that near-certain,
  but it is a wall-clock assumption in a unit test.

  5. test-infra/runner has no lockfile. package-lock.json is gitignored repo-wide (.gitignore:42), but
  integration-harness.yml:76 runs npm ci in test-infra/runner — which cannot work without one. Pre-existing, not caused
  by this PR, but adding socket.io-client is the first time it bites anyone running the harness fresh. Version (~4.8.1)
  correctly matches the server's socket.io in the root package.json.

  6. Redundant condition — advancedWorkflows.js:3797. The && !runningAppsNames.includes(identifier) in the eviction
  guard is already guaranteed by the enclosing if (!runningAppsNames.includes(identifier)) at line 3723. Harmless and
  arguably self-documenting; just noting it isn't load-bearing.

  7. Repeated exec on one socket accumulates listeners. Each exec now registers one extra disconnect listener on top of
  the pre-existing resize/cmd/end ones, so ~10 execs on a single socket would trip MaxListenersExceededWarning.
  Pre-existing pattern, marginally worsened, no crash. Only matters if the UI reuses a socket across sessions.

  Adjacent finding — not this PR, but same function and same failure story

  advancedWorkflows.js:3910 uses return; where it means continue;:

  if (previousMasterStillRunning) {
    return;   // aborts the ENTIRE pass, skipping every remaining g: app
  }

  I confirmed this is pre-existing (origin/development:3834), so it is not a regression here. But the effect is that
  whenever one app's previous master is still running, every g: app after it in the list is silently skipped for that
  cycle — which is squarely in the family of "the election loop looks dead but isn't" that this PR is trying to make
  visible. Worth a follow-up issue, and worth mentioning while the author has the file paged in.

MorningLightMountain713:
Thanks both — all eight points were worth raising. Six are fixed, two I'm pushing back on with reasoning below. A full parallel gate is running against `8ecfe4529`; I'll post the tally when it lands.

## Fixed

**1. Serial peer probing — `19e720b77`**

Correct, and worse than described: the probe is gated on the component not already running locally, so it re-runs on *every* election pass for the whole duration of the permissions fix rather than once. Probes now run concurrently.

I deliberately did **not** shorten the 10s per-probe timeout. The check fails open, so a shorter timeout makes a slow-but-alive primary read as absent and starts a second writer on a shared volume — precisely the outcome the probe exists to prevent. Concurrency gives the entire speed win at no safety cost. The test asserts all N requests are issued before any settles; against the serial version it fails with `expected 1 to equal 3`.

**2. Phantom analytics session — `583b00181`**

Confirmed. Rather than move the guard above the analytics block, both halves of the pair now belong to the single `disconnect` listener registered before the awaits, gated on a flag. That removes the ordering dependency rather than correcting one instance of it. The test asserts neither `open` nor `close` is recorded when the client leaves mid-await.

**3. `totalWritten` is meaningless — `2db384f39`**

Agreed on both counts: a real trap, and no correctness bug. I traced the partially-filled case — empty slots are skipped by the `entry &&` test and the walk still yields oldest→newest. Replaced with a real `#writeCount`, plus a test for the exactly-at-capacity boundary, which had no coverage.

**4. Timing-dependent event-bus test — `d5b0f9b8c`**

Fair, and the `#writeCount` change above made it marginally worse: `since()` now scans 2 slots instead of 1024, so *less* time elapses between the two constructions. It now waits for the clock to pass the ids already served, which makes the assertion hold by construction and is a fairer stand-in for the seconds a restart takes. 20/20 clean, and it still fails against the unseeded counter.

**5. No lockfile but CI runs `npm ci` — `ebcb9fd8d`**

Real — I hit this independently switching branches before reading the review. CI now uses `npm install`.

I'd initially argued for committing the lockfile instead, but that carves a ~108 KB generated file out of a repo-wide convention to benefit a workflow with zero recorded runs, and the gate resolves its dependencies with `npm install` regardless.

One correction: this isn't newly bitten by `socket.io-client`. With no lockfile at all, `npm ci` fails on a fresh checkout whatever the dependency list is. The new dependency only matters to someone carrying a stale lockfile locally.

**6. Adjacent finding — `return` where `continue` was meant — `97f0903e0`**

Confirmed and fixed; thanks for flagging it. It sits directly in the per-app loop with no callback boundary, so one app whose previous master is still running abandons the entire pass and silently skips every remaining `g:` app until the next cycle. The trigger is the routing layer reporting no primary for a component that is in fact running, which its registration lag makes routine.

**This is pre-existing on `development`** (`origin/development:3834`) and not introduced by this branch — but it is the same silent-skip failure the rest of this PR exists to remove, so it is fixed here rather than deferred.

Worth recording why no suite caught it: the election loop is per-app and takes only the first `g:` component of each app, so `return` and `continue` are indistinguishable unless two `g:` apps are installed on one node — which no suite does. That loop body has run exactly once in every suite that has ever exercised it.

## Not changed

**7. Redundant condition**

Confirmed redundant — it is guaranteed by the enclosing `if` 74 lines above.

Keeping it deliberately. This code is due to be lifted into v9's restructured lifecycle, where that enclosing condition will not travel with it. A guard that reads correctly in isolation survives the move; one that silently depends on distant context does not. Happy to drop it if you'd rather.

**8. Listener accumulation**

Checked against the frontend rather than reasoned about. The UI opens a **new socket per terminal tab** and emits `exec` exactly once on it, with `removeAllListeners()` + `disconnect()` on close and cleanup on unmount. There is no retry path, no reconnect re-emit, and no re-exec on container change — every failure path tears the socket down instead. socket.io v4 also forces a fresh manager when the same namespace is requested again, so ten terminals means ten independent sockets, not ten `exec`s on one.

One correction to the severity: the warning threshold is unchanged by this PR. `resize`, `cmd` and `end` already registered one listener per `exec`, and the limit is per event name — so this adds a fourth name that would warn at the same point, rather than making it happen sooner. Not reachable from the UI, so leaving it alone.

## One extra, found while chasing #2

Harness nodes resolved `analytics.url` to the application default — the live endpoint — because the per-node config is layered *over* `ZelBack/config/default.js` rather than replacing it, and the fleet network permits egress. Test runs were therefore reporting generated app names and fixture identities as real traffic, silently, because analytics is fire-and-forget with warning-level failures. Fixed in `8ecfe4529`.

ZelBack/src/lib/socketIoHandlers/dockerTerminalHandler.js
MorningLightMountain713: resolved

ZelBack/src/services/appLifecycle/advancedWorkflows.js
MorningLightMountain713: resolved

Cabecinha84 (commented):
Every one of the eight points raised by you and alihm has been genuinely addressed in code (six fixed, two pushed back
  on with reasoning I find sound). I re-read the resulting diff line by line and could not find a correctness defect in
  what landed. The reason to hold the ack is this:

  ▎ "A full parallel gate is running against 8ecfe4529; I'll post the tally when it lands."

  That tally has not been posted. The PR body still claims 54/54 PASSED, but that run was against 874ae6dc1 — six
  commits ago. Four of the six commits since then touch production code:

  ┌───────────┬──────────────────────────────────────────────────────────────────┐
  │  commit   │              production change after the green gate              │
  ├───────────┼──────────────────────────────────────────────────────────────────┤
  │ 583b00181 │ restructured the terminal handler's disconnect/analytics pairing │
  ├───────────┼──────────────────────────────────────────────────────────────────┤
  │ 19e720b77 │ rewrote the peer probe from sequential to concurrent             │
  ├───────────┼──────────────────────────────────────────────────────────────────┤
  │ 97f0903e0 │ return → continue in the election loop                           │
  ├───────────┼──────────────────────────────────────────────────────────────────┤
  │ 2db384f39 │ since() now keyed on #writeCount instead of #nextId              │
  └───────────┴──────────────────────────────────────────────────────────────────┘

  97f0903e0 in particular is a control-flow change in the per-app election loop — exactly the kind of change unit tests
  cannot fully cover and the harness can. Ack once the gate tally for 8ecfe4529 is posted; no further code changes
  needed on my read.

  Verification I did

  Unit suites on the PR head (NODE_CONFIG_DIR=$PWD/tests/unit/globalconfig):

  - advancedWorkflows.test.js — 60 passing
  - dockerTerminalHandler.test.js — 2 passing
  - fluxEventBus.test.js — 14 passing

  Point-by-point on the review responses:

  1. Concurrent probe — confirmed. Promise.all over a pre-built peer list, .some(Boolean). Self-exclusion via if (i ===
  index) continue, and 'lower' reduces to the exact old behaviour (limit = index, limit <= 0 → false). His refusal to
  shorten the 10s timeout is right: the probe fails open, so a shorter budget makes a slow-but-live primary read as
  absent and starts a second writer on a shared volume.
  2. Phantom analytics — confirmed, and the fix is better than what we asked for. Both halves now live in the single
  pre-await disconnect listener gated on analyticsOpened, which removes the ordering dependency rather than correcting
  one instance of it.
  3. totalWritten — replaced with a real #writeCount. The > RING_BUFFER_SIZE boundary is correct (at exactly full,
  startIdx = 0 = #writeIndex), and he added the boundary test that was missing.
  4. Timing-dependent test — now waits for the clock to pass served ids. Assertion holds by construction.
  5. npm ci — now npm install --no-audit --no-fund. His correction is right: with no lockfile at all, npm ci fails
  regardless of socket.io-client.
  6. return → continue — I verified the brace nesting independently: the innermost enclosing loop is for (const
  installedApp of appsInstalled.data) at :3651, with only if blocks in between and no callback boundary. continue binds
  where he says it does.

  Two extras I checked that nobody raised:

  - globalState.receiveOnlySyncthingAppsCache and appCaches.receiveOnlySyncthingAppsCache are the same Map instance, so
  the new delete in createAppVolume cannot miss the cache masterSlaveApps reads.
  - analytics: { url: '' } really does disable analytics (analyticsService.js:257 + the early if (!analyticsUrlCached)
  return). His self-reported harness-leaking-to-live-analytics finding is real and correctly fixed.

  Push-backs — both acceptable

  - #7 redundant condition. Confirmed redundant; his "guard that reads correctly in isolation survives the v9 lift"
  argument is legitimate. Either way is fine — not worth a round trip.
  - #8 listener accumulation. His correction is technically right and I'd retract my framing:
  MaxListenersExceededWarning is per event name, and resize/cmd/end already registered one each per exec. Adding a
  fourth name doesn't lower the threshold. Combined with one-socket-per-tab in the UI, it isn't reachable.

  Non-blocking notes for the thread

  1. Substring match on the index-0 path is a wider surface now. appsRunning.find((app) =>
  app.Names[0].includes(identifier)) is a substring test, so component x of app myapp matches a container for myapp2.
  Pre-existing and copied, not introduced — but it now sits on the index-0 start decision, where a false positive means
  "never start while FDM reports no primary" rather than just skipping a staggered branch. A ===-on-parsed-name or
  boundary-anchored match would close it.
  2. getDockerContainerByIdOrName(...).catch(() => null) collapses every failure — including an unreachable docker
  daemon — into "Container not found." Correct for crash safety, weaker for diagnosis; a log.error in the catch is one
  line.
  3. Probe timers are never cleared. Each concurrent probe arms a 10s setTimeout guarded only by isResolved, so N
  un-unref'd timers now linger per pass instead of one at a time. Pre-existing pattern, harmless, but it is the kind of
  thing that makes a shutdown hang for 10s.
  4. The PR description is stale. It says "written against 874ae6dc1" and its change table omits the return→continue
  fix, the analytics pairing fix, and the harness analytics config. Since this description is explicitly meant to be the
  record for the v9 rebase note (the fixes live in a file v9 deleted), it should be refreshed before merge.
  5. 61-syncthing-mount-safety phantom-index → it.skip. I'd accept it. The reasoning is inline, the gap is documented
  rather than erased, and he verified it fails identically on unmodified development. Worth an issue so the files-aware
  discriminator doesn't get lost.

MorningLightMountain713:
## Gate result, and the follow-up notes

**Full parallel gate against `8ecfe4529`: all 54 suites green.**

A further gate is in progress covering the review notes below, and I'll post that tally too.

### The non-blocking notes

**1. Substring match — fixed, `8ca91b60a`.** This one was worth more than "non-blocking": it is on the index-0 probe added by this PR, and it fails toward *never start*. `simplexsmp` against `simplexsmp1` are both live on the network today, and for those compose apps the identifier `smp_simplexsmp` is contained in `/fluxsmp_simplexsmp1` — so a node holding `simplexsmp` at index 0 would be blocked by any peer running `simplexsmp1`. `getAppIdentifier` already yields exactly the container name, so both probes now compare whole names. Applied to the previous-master check as well, which shares the defect.

It also turned up a test problem worth stating: the election fixture stubbed an identifier `getAppIdentifier` never produces (`zel_<app>`) while the peers it stubbed reported `flux<app>` containers — a pair that only matches under a substring test. So the "don't start when a peer runs it" assertions were passing for the wrong reason and would have kept passing against a broken exact-match implementation. The fixture now derives the identifier the way production does.

**2. Opaque lookup failure — fixed, `1bc6cf644`.** The client still gets "Container not found." (right thing to tell a user), but the underlying error is logged so an unreachable daemon is distinguishable from an absent container.

**3. Probe timers never cleared — fixed, `ed3316bf0`.** Cleared when the request settles, which makes the `isResolved` flag redundant, so it is gone. Applied to both probes.

**4. Stale description — done.** Rewritten against `1bc6cf644`; the change table now lists every fix, and the `return`/`continue` change is marked pre-existing on `development`.

**5. Issue for the skipped `61` case — not needed.** v9 already resolves it: it splits the check into `checkDirectoryHasSyncScopedContent` (`countDirs: true`) and a sibling `checkDirectoryHasSyncScopedFiles` (`countDirs: false`), on the grounds that the deletion-broadcast hazard is per-file and a surviving directory skeleton protects nothing. That is exactly the files-aware discriminator, so the pending test becomes its proof when v9 lands. Noted in the description.

Thanks for the review — notes 1 and 3 were both real, and 1 in particular was a defect I introduced.

Cabecinha84 (approved):
Ack

PR #1773, opened 2026-07-27
Release 8.17.1: terminal exec crash safety + g: election recovery

Cabecinha84:
## Release 8.17.1

`master` is at **8.17.0**. This PR carries the terminal crash fixes and g: election recovery accumulated since that release.

---

### Terminal (`dockerTerminalHandler.js`)

Each of these took down the entire FluxOS process — an exception thrown in a socket.io listener has nobody to catch it.

- **A terminal on a stopped container killed FluxOS.** dockerode hands back a null `exec` when the daemon rejects the exec create (most commonly a container in `created`/`exited`), and the code dereferenced it. Same for `stream` in `exec.start`. Both now fail cleanly via `socket.emit('error', ...)`.
- **No `error` handler on the hijacked stream.** Node detaches its own handler at upgrade, so a write racing the exec teardown (EPIPE on a keystroke as the shell exits) was an uncaught exception.
- **Leaked exec/stream when the client disconnects during the awaits.** The `disconnect` listener is now registered *before* authorisation and the docker lookups — a listener added afterwards never fires, because socket.io emits `disconnect` exactly once.
- **Analytics `open` with no matching `close`.** Same root cause; the pair is now owned entirely by the listener registered up front, and a session that never opened is never recorded.
- **Authorisation now precedes the Docker access.** The lookup is a remote operation on a caller-supplied name and must not be reachable unauthenticated.
- A failed lookup now logs its cause: an unreachable daemon and a missing container both reached the client as the same "not found".

### Master/slave election (`advancedWorkflows.js`)

- **A g: app stopped by the operator could not come back.** When the remembered primary was this node but the component was not running here, the stale memory disqualified it on both paths (the no-history start requires no remembered primary; the previous-primary path requires a *different* node). With every instance in that state, the app only returned after a FluxOS restart.
- **One settled g: app cost every later one its cycle.** A `return` where a `continue` belonged abandoned the whole pass whenever the previous primary still held the app — and FDM's registration lag makes that routine, not exotic.
- **Container matching by substring.** `myapp` matched `myapp2`, `simplexsmp` matched `simplexsmp1`. A false positive here means the component never starts at all. Now compares whole names.
- **An index-0 start was blind.** With no lower-index nodes, the existing probe always answered "nobody". Given FDM's registration lag (~110s measured in production), that meant starting a second writer on a shared volume. It now probes every peer.
- **Probes now run concurrently.** Each probe is bounded at 10s and an unreachable peer burns the whole budget; sequentially that cost 10s × peers, repeated every 30s pass and ahead of every later g: app. The per-probe timeout is deliberately unchanged — the check fails open, so shortening it would make a slow peer read as absent.
- **Cancel timers are now cleared.** Every probe left a live 10s timer behind, per peer, on every pass.
- **A fresh volume invalidates any surviving synced-mark.** Without this, a fresh install skipped the new-install receiveonly protection and read as instantly ready to become primary. The mark is dropped at the point of no return (the allocation), not at entry — the pre-flight can abort with the volume and its data intact.
- **Operator-stop exclusion is now announced.** An operator stop is durable in the DB, so the app sat unelected indefinitely with the loop emitting nothing at all — indistinguishable in the logs from a loop that has died.

### Event bus (`fluxEventBus.js`)

- **Event ids did not survive a FluxOS restart.** They restarted at 1, making the whole post-restart stream invisible to a consumer filtering on `Last-Event-ID` until the counter climbed past the id it last saw — which on a busy node is never. They are now seeded from the monotonic clock.
- `since()` now counts its own writes instead of inferring from the id, since the id no longer carries count information.

### Tests and infra

- New integration harness suites 52 (operator-stop recovery) and 53 (terminal exec crash safety).
- New unit tests for `dockerTerminalHandler` and `fluxEventBus`; broader coverage in `advancedWorkflows`.
- The harness no longer reports suite activity to the live analytics service — generated app names, fixture identities and 198.18.x addresses were going out as real traffic.
- CI installs runner dependencies with `npm install` (`npm ci` was failing).

---

## Testing

- Unit tests: `npm test`.
- Integration harness: suites 52, 53 and 61.


MorningLightMountain713 (approved):
Ack

PR #1774, opened 2026-07-31
Placement diversity: computed placement over a published IP table, network policy in its own repository, and a hardened synced-app election

MorningLightMountain713:
## What this delivers

- **Deploy forms can finally tell the truth.** Two new endpoints for the frontend answer *"where can this app run, and on how many nodes?"* before the user pays: `GET /apps/placementlocations` serves live node, provider and tier counts per continent and country, and `POST /apps/placementfeasibility` returns a per-spec verdict computed by the same code the network will enforce, to a signed-in Flux ID.
- **Synced apps now reach their full instance count.** The old `/16` rule could strand an app below its count forever — a customer's 3-instance WordPress sat at 1 of 3 with the site returning 503, permanently. Placement now computes what is actually possible and converges to it.
- **Diversity is real, not octet theatre.** Fault domains are provider organisations from a published IP table: Hetzner's 12 Finnish `/16`s count as one domain, and a `/16` that spans 10 countries stops cross-blocking nodes in different countries.
- **Network policy moves out of the application repository.** Everything the network enforces — blocked and vetted repositories, enterprise nodes, the tampering blocklist, and the IP location table — now lives in `RunOnFlux/fluxos-network-policy`, with its own writer set and CI that validates every publication before it can be served. Merging application code no longer carries the power to change network policy, and signed policy releases become a small isolated step instead of a cross-cutting one.
- **Nobody can buy an impossible deployment.** A provably unsatisfiable spec is refused at registration with a clear error before payment; a deliverable-but-fragile one is accepted with a warning carrying the numbers.
- **A dead node no longer takes a synced app down for two hours.** The election drops a provably-gone holder at the pick, so failover completes in under a minute instead of waiting out a 125-minute broadcast expiry.
- **A `g:` app's ordinary day is correct.** Nine primary-election defects fixed — second writers, ghost claims, permanently-unelectable nodes — and every ranking the fleet must agree on is now deterministic on every node.
- **Operators can see why.** Placement and election decisions log the evidence they decided from, and syncthing's errors finally reach the node's logs instead of dying unread in the daemon.

## How it fits together

Synced (`g:`/`r:`/`s:`) apps could be made permanently unplaceable by a placement rule that never checked whether it was satisfiable — and the same app class ran on a primary election that had never been tested in the arrangement production actually produces. This PR fixes both halves of a synced app's life, plus the infrastructure both needed:

[...]

PR #1775, opened 2026-08-03
Memory: bound the sync path, take heavy dependencies off the boot path, and make node-list readiness askable

MorningLightMountain713:
> **Second in the stack: `development ← #1774 ← this`.** Based on `feature/placement-diversity` and targets it, so the diff here is only the 18 memory commits. It cannot go to `development` until #1774 merges, at which point this retargets.

## What this does

FluxOS grew into memory it never gave back, in two distinct ways.

A sync response was verified and stored whole, so the events, their verification copies and the driver's encoding of every location write were all held at once — one response cost **676MB of RSS in two seconds**, measured second-by-second on a live node.

Separately and more durably, the signature-verification pool started `os.cpus().length - 1` worker isolates at boot and never stopped them. On a 16-core node that is **15 full V8 heaps that grow for the life of the process**, to serve 18 batches in 10.5 hours, all inside a 4-second window at boot.

### Measured: one node, both commits, matched uptime

`master` `fe5ef7239` against this branch on the same node, same 8 apps, same `node_modules` (`master`, the base `4af5e4382` and this branch declare byte-identical dependencies, so nothing was reinstalled between arms). Arms run sequentially, each on a single process — a restart mid-arm would have voided it, and neither had one.

| uptime | this branch | `master` | delta |
|---|---|---|---|
| 45 min | 343.1MB | 654.7MB | **−311.6MB** |
| 1h00 | 355.8MB | 655.1MB | **−299.3MB** |
| 1h15 | 356.2MB | 655.6MB | **−299.4MB** |
| 1h30 | 356.9MB | 655.8MB | **−298.9MB** |
| **boot peak** | **479.1MB** | **889.4MB** (at 79s) | **−410.3MB** |
| **threads** | **12** | **26** | **−14** |

**~300MB lighter and ~410MB lower at peak, at 1.5h uptime with 8 apps.** The delta held within 13MB across 45 minutes of matched samples while `master` sat flat within 1.1MB. Measured boot-to-boot variance for a single configuration on these nodes is ~50MB, so this is a signal roughly six times the noise.

The cost also **grows with uptime**, because worker isolate heaps grow. On a node at 7.5 days, `master` held **1,023.6MB RSS with 15 `WorkerThread`s**, of which **~490MB was worker-isolate heap** — 690.7MB of the process sat in 256KB V8 heap pages while the main isolate's `heapTotal` was 156.8MB, against a ~43MB baseline gap measured with one worker.

**Earlier figures in this branch are withdrawn.** "~250MB settled", "736MB original" and "boot peak 1011MB → ~388MB" are not carried forward: they were taken at short or unstated uptimes, without a control arm, and the peak pair was never measured under matching conditions. Every figure above names the node, the app count and the uptime it came from — the absence of exactly that context is what allowed two different "settled" figures to circulate.

## The changes

**Elastic verify pool** — `utils/verifyPool.js`

`cpus-1` isolates started once at boot and never stopped: 15 on a 16-core node, holding **~490MB at 7.5 days**, to serve 18 batches inside a 4-second window. Now one resident worker, scaling on outstanding chunks to `cpus-1`, reaped after 60s idle. Chunk size 256 measured best — chunking cut retention 167MB → 34MB, and smaller chunks are worse because more chunks means more workers.

What this does NOT do is fan a single sync response across workers. A slice is 250 events and a chunk is 256, so one response is one chunk on one worker, and verification within it is serial where it previously spread over `cpus-1`. That is the trade the slice size is chosen for — raising it above the chunk size buys back a fraction of a second on a path that runs 18 times in 10.5 hours, and costs worker isolates at boot, which is the whole point of the change. The pool scales across concurrent peers, not within one response.

**Throwaway workers for heavy dependencies** — `utils/workerRunner.js`

Cloud registry SDKs cost **+38MB / 427 modules** at every boot via eager requires, and openpgp **+19MB** for one boot-time check. A worker per exchange, terminated after, is the only way the pages actually return.

**Sync responses carried in slices** — `serviceHelper.processInSlices`

Bounds the apprunning events, their verification copies and the write encoding to one slice at a time. Two things are deliberately NOT bounded by it: evictions and non-apprunning state events accumulate across the whole response, because both are applied only after every slice has been processed - an eviction clears a node's locations outright, so applying one inside a slice would let a later slice re-add what it removed. The bound is therefore over the dominant traffic, not over the response. Slice size 250 sits **deliberately** below the 256 chunk size: 1024 gives 12 worker isolates and nearly doubles boot peak.

**Bounded location writes**

~2000 broadcasts × ~20 apps = ~40,000 database operations, each carrying its own merge program, all accumulated before a single bulkWrite.

**openpgp off the boot path**

A node that already has a keypair now does no PGP work at boot. Re-verifying an unchanged key cost ~75ms of worker spawn and library load to perform 2.4ms of work, on every boot of every node.

**Node-list readiness is a property, not a value**

`networkStateService` answered "not started" and "genuinely empty" with the same value, so discovery read an unknown list as a network with no nodes, sized both deterministic loops to zero, and connected to nobody. `isReady()` and `onReady()` make it askable and observable.

**A config read that lands mid-write is rejected**

`config/userconfig.js` is rewritten whole by all seven of its writers, and by configd from *another process*. `require()` resolves a truncated file to an object rather than throwing, so the catch never fired and the process was left holding a config with no `initial` — where every read of it throws until the next write settles. A node that already holds a config now keeps it, rather than publishing defaults that carry no zelid and no keypair.

## Correctness

The peering fix is a production defect, not a test artifact. Before, on a real node:

```
13:00:18   My node was found on index: -1 of 0 nodes
13:01:33   My node was found on index: 1991 of 6082 nodes
```

Every node lost its **first ~75 seconds of peering** on every boot, because discovery autostarts ahead of the node-list fetch and read the empty list as final. After:

```
19:40:28   My node was found on index: 1990 of 6103 nodes    ← first pass
```

Also fixed here, found by adversarial review of this branch's own work: the verify pool could wedge permanently and silently stop verifying all signed gossip; evictions were systematically undone by later slices; the listening socket rejected clients whose deflate window was smaller than ours; a worker failure carrying no message resolved as success, which on the PGP path reads as a corrupt identity and overwrites the node's keypair; and location pruning took its input from the raw response rather than the verified broadcasts, so an unsigned event could drive row deletions.

## Verification

- Integration gate **64/64** at `3ea1ed0e5`, this branch's pre-rebase head, and each change gated separately so any failure was attributable to one change rather than needing a bisect. The rebase onto the updated #1774 replayed all 18 commits with no conflict and changed nothing in them — the only delta underneath is test-infra — the 82-suite gate at the top of the stack (#1780, `b55b510c0`) re-covers this branch's content in place: **82 of 82 green**.
- Unit suite **4,777 → 4,813** (18 pending), measured on this branch's rebased head and on its base #1774; the four tests added with the config fix were each proven to fail against the unfixed code, and fail for the right reason — the unfixed loader publishes `{}` over a good config and leaves the process with no `initial` at all.
- New harness suite covering eviction, pruning and forged events in a sync response — proven red against the vulnerable build before being trusted.
- Every memory figure measured on a live node with the inspector attached to a **running** process, never a restarted one, and paired with `/proc/<pid>/smaps` accounting. `heapTotal` reports only the main isolate — worker isolates are invisible to it and fully visible in RSS, which is why this cost went unmeasured for three sessions.

## Settled by measurement — please don't re-litigate

- glibc trim/mmap tunables are **not needed** (worth 49MB before the sync slicing, ~0 after).
- jemalloc is **~22MB worse** than stock at 15 workers.
- `--max-old-space-size` moves settled RSS by 8MB; it is not the lever.
- **`--max-semi-space-size` is not the lever either.** Three boots on one node: the two *default* boots differed by **41.8MB**, and the flag's boot landed inside that spread. It does cut boot peak (324.9MB against defaults of 386.7 and 443.0) and volatility, but costs ~15% more CPU, and it is not deliverable from this repo — V8 fixes the size at heap configuration.
- Transferables/SharedArrayBuffer are **slower** than V8's structured clone here.
- The 250-vs-256 mismatch between slice and chunk size is deliberate and commented.
- **The module graph is not a lever.** 1,217 modules load, only 166 of them ours; the main isolate's used heap is 63.2MB spread so thin that its largest single class is `object:Object` at 5.37MB over 79,474 instances. The `~97MB main isolate heap` figure in older notes is stale.

## Accounted for, not fixed here

Quantified so they are not rediscovered later as regressions:

| item | size |
|---|---|
| bson scratch buffer | 17MB |
| ssh2 poly1305 WASM | 16MB |
| dockerode's grpc transport | 65 modules + protobufjs's 39 |
| appmonitor `statsStore` | **73MB at 10 apps, 365MB at 50** |

**bson** allocates `1024*1024*17` at module load and holds it for process life, regardless of document size. Dependency behaviour, not ours to change.

**ssh2** exports `init` from `crypto.js` as an IIFE evaluated at module load, so `require('ssh2')` instantiates the WASM with no SSH connection involved. It arrives via `dockerode → docker-modem/lib/ssh.js`, whose only use is under `protocol === 'ssh'` — and FluxOS talks to the local Docker socket. The fix is a one-line lazy require upstream.

**dockerode's grpc transport** is the same shape: `dockerode/lib/session.js` requires `@grpc/grpc-js` eagerly. Same upstream ask.

**appmonitor** keeps 7 days of 3-minute samples at a measured **7.3MB per app**, observed on two apps that had reached the full 3,360-sample window. It takes 7 days to arrive, which is why a freshly restarted node reads it as negligible. Being fixed separately.

`config/userconfig.js` writes are still not atomic — only the read is hardened here. The root cause is that FluxOS writes node runtime state into the operator's file at all, which v9 has already dissolved by moving the keypair and last-known IP into the local database.


Cabecinha84 (commented):
PR #1775 — Review

  Status of the PR itself

  There is no prior review history at all. Zero review submissions, zero issue comments, zero inline comments. The
  activity you're seeing is the author's own: 18 commits and 6 force-pushes, all rebases onto the moving
  feature/placement-diversity base. So there is nothing to "check whether it was addressed" — this is a first review.

  Also note it targets feature/placement-diversity (#1774), not development. It cannot merge until #1774 lands.

  What I verified

  - Merge-base is exactly #1774's head (839b3ddb4), so the diff is purely the 18 commits — the PR's claim there is
  accurate.
  - Read the post-change state of every substantive file, not just the diff.
  - Ran the unit suite: 4803 passing, 5 pending, 23 failing. All 23 failures are dockerService tests needing a live
  Docker daemon (Cannot read properties of undefined (reading 'Id')) — environmental, unrelated to this PR.
  - Spot-checked the claims in the description against the code: eviction-applied-after-all-slices,
  prune-from-verified-broadcasts-only, per-operation (not per-broadcast) batch cap, worker failure as an explicit ok
  flag rather than a truthy error string. All four check out.
  - The compression change is now correct: socketServer.js leaves both window bits negotiable (true), and only the
  outbound peer socket in fluxCommunication.js:1061 pins 13. Old peers negotiate fine in both directions, and browsers
  on /ws/id|sign|payment are no longer at risk of a 400.

  The measurement discipline in the description is genuinely unusual — matched-uptime control arm, withdrawn earlier
  figures, explicit "settled, don't re-litigate" list. I have no quarrel with the memory work itself.

  Findings

  1. Removing the blocking waitStarted() is broader than the problem it fixes — this is my main ask

  fluxCommunicationUtils.js lost await networkStateService.waitStarted() from all four accessors (deterministicFluxList,
  getNodeCount, getFluxnodeFromFluxList, socketAddressInFluxList).

  The stated justification is sound for two callers: self-re-arming loops would be retired for the process lifetime by
  an await that may never return. But the change removed the wait for every caller, and only fluxDiscovery got a
  replacement guard (fluxCommunication.js:1341). Ten other call sites now silently read "unknown" as "empty network".

  The one that destroys state is ipLocationSync.refreshNodeLocations() (ipLocationSync.js:48). With an empty list,
  ipLocationStore.js:639 computes:

  const departed = [...nodeView.keys()].filter((ip) => !listed.has(ip));  // → everything

  and then issues deleteMany({ _id: { $in: departed } }) — wiping the entire nodeLocations collection and the in-memory
  view. Nothing refills it in that pass (missing is derived from the same empty list), and the next refresh is
  REFRESH_INTERVAL_MS = 24h away, with the retry backoff reset because status().ready is still true. The only guard
  there is if (!status.ready) return, which is about the IP-location baseline, not the node list.

  To be fair to the author: I could not construct a live boot path that reaches it. startSync() is gated on
  waitForDbReady(); dbReady is only ever set in AppSyncOrchestrator.#rebuildDb(); that runs downstream of await
  this.#waitForNetworkState() (appSyncOrchestrator.js:150). So dbReady ⇒ network state populated, and
  networkStateService.stop() is test-only. The wipe is currently prevented — by a three-hop ordering coincidence that
  nothing documents, no test pins, and that this PR just removed the structural guarantee for.

  The fix is cheap and the codebase already contains the exact precedent, in code from the base branch:

  // placementFeasibility.js:175
  // an empty node list is missing data, not an empty network - reporting
  // zero candidates would read as proven impossibility to the callers

  Apply the same reasoning in ipLocationStore.refreshNodeLocations (bail on an empty nodeList), or keep the wait on the
  one-shot paths and drop it only from the re-arming loops.

  Secondary sites with the same shape, worth the author's eye but all self-healing: nodeStatusMonitor.js:93 (every
  location fails the list check and falls into the HTTP-probe path), fluxNetworkHelper.js:329 isFluxAvailable (returns
  false → the asking peer takes DOS increments during our first ~75s), appHashSyncService.js:501,
  enterpriseNodesService.js:15, placementFeasibility.js:681.

  2. verifyPool: a requeue from inside dispatch() has no guaranteed re-entry — confirmed by repro

  verifyPool.js:87 — when postMessage throws, the job is unshifted back onto the queue and the loop continues. If no
  free slot remains in that pass, nothing schedules another dispatch(): the re-entry points are all worker events
  (message/exit/timeout), and no worker is holding a job. I reproduced it:

  Verify worker could not be given a batch: () => {} could not be cloned.
  RESULT: STILL PENDING after 8s
  stats: {"workers":1,"busy":0,"queued":1,"maxWorkers":19}

  One idle worker, one queued job, verify() never settles. MAX_JOB_ATTEMPTS never gets to fire because nothing
  re-enters. In production it self-clears once two further verify() calls burn the remaining attempts, so it stalls one
  sync response rather than wedging permanently — and reachability is low, since chunks are now JSON-derived broadcasts
  that clone fine. But this is precisely the failure class commit 5a58b59c4 claims to have closed, and it's a two-line
  fix (re-dispatch after a requeue, or count the failed handover against attempts before continuing).

  3. New lint errors introduced

  fluxCommunication.js:249-258 — the for (const event of [...evictions, ...stateEvents]) body kept its old indentation
  after the loop header changed. 10 fresh indent errors under the repo's own rules. (The base file already carries 14
  pre-existing ones elsewhere, so this isn't a new category — but these ten are this PR's.) npm run lint --fix on that
  hunk.

  4. Minor: prune input diverges from the old prune input

  fluxCommunication.js builds newestByIp without the validTill < Date.now() expiry filter that messageStore.js:669
  applies when building its own v2AppsByIp. So the sync path can now prune from an expired broadcast where it previously
  wouldn't. Impact is bounded by the broadcastedAt: { $lte: ... } clause — only rows already at least as stale can be
  deleted — so I'd call it a comment-or-align item, not a defect.

  5. Note on the stated bound

  evictions and stateEvents accumulate across all slices, so the slicing bound really only applies to apprunning events.
  Correct given the traffic mix, but the description reads as a stronger guarantee than the code gives.

  Verdict

  Don't ACK yet — request changes, but small ones.

  The engineering is strong and the claims hold up under checking. Nothing here is a design objection. What I'd ask for
  before ACK:

  1. A guard on the empty node list in ipLocationStore.refreshNodeLocations (or restore the wait on non-loop callers) —
  the one that matters.
  2. Re-dispatch after requeue in verifyPool.
  3. Fix the ten indent errors.

  (1) and (2) are a handful of lines each. Once those land I'd ACK without further review — I'd suggest saying exactly
  that in the comment so it doesn't cost another full round trip on an 18-commit branch.

MorningLightMountain713:
All five answered, in five commits on top of a rebase onto #1774's current tip. Unit suite 4838 passing / 18 pending / 0 failing. Full integration gate on the branch tip `3c8a04172`: **`RESULT  suites_pass=67 suites_fail=0 FAILED:[ ]`**, 60m26s.

Two of the five landed differently from what you asked for, and one of your findings turned out to be better than you pitched it.

## 1. The removed `waitStarted()` — fixed, and wider than the ask

You're right about the wipe, and right that nothing pins the ordering that currently prevents it. I traced the chain myself and it holds — `startSync` ← `waitForDbReady` ← `dbReady` ← `#rebuildDb` ← `#waitForNetworkState`, with both `#rebuildDb` call sites downstream and `serviceManager` supplying the optional wait. So: unreachable today, on a coincidence.

**But I didn't take the guard-in-`refreshNodeLocations` fix, because the problem is bigger than that one caller.** I went through every consumer of the four accessors: **fourteen call sites, of which two were given the replacement.** `fluxDiscovery` (`onReady` plus `isReady`) and `placementComputation` (503). The other twelve read an unknown list as an empty network — and the filtered accessors return `[]` for an unknown state too, so the `filter` variants are in the same position.

Beyond the one you found, two more are worth naming. `placementLocations` answers all-zero counts while its sibling *in the same file* refuses with 503 — the two endpoints an owner cross-checks against each other, disagreeing about the same network. And `checkDeterministicNodesCollisions` — the loop the removal was justified by — reads an unknown list, concludes this node is absent from the confirmed list, **logs that false conclusion as the reason**, and returns before `checkMyFluxAvailability()`, which is the call that sets `dosState = 0`.

That last one is what decided the shape. **The hazard had a local fix, and that function already demonstrates it two lines above the list read**: it checks `isDaemonSynced`, reschedules, and returns. The node list needed the same three lines. Instead the contract of four shared accessors changed for every caller, to avoid a hazard in two of them.

So the default is inverted: the unsafe reading is free, and correctness has to be opted into. Two of fourteen opted in.

The accessors wait again. The three re-arming loops — `checkDeterministicNodesCollisions`, `monitorNodeStatus`, and `fluxDiscovery`, which already did — check and re-arm instead, so they never reach a blocking call and the deadlock you were protecting against cannot occur. `monitorNodeStatus` checks at the point of use rather than at entry, because everything above its list read is app removal that needs no list. Request handlers answer 503, since a client is holding a connection and can wait for neither.

`isReady()`/`onReady()` stay. They're the right primitives — they were just the only thing standing between an unknown list and a `deleteMany`.

Worth recording: **no test asserted that an accessor returns `[]` when the state is unready.** The contract change had no coverage at all, while `fluxCommunicationUtils.test.js:81` still stubbed the removed `waitStarted`. That absence is why twelve callers could be degraded silently.

## 2. `verifyPool` requeue — confirmed and fixed

Exactly as you describe. `requeue` is only `queue.unshift(job)`; the re-entry points are `verify()` and three worker events, and after a failed handover no worker holds anything to raise one. Dispatch re-enters itself now, bounded because the attempt is counted before the post — a chunk that cannot be cloned is abandoned after `MAX_JOB_ATTEMPTS` and its caller rejected. Your repro is now a regression test, and it fails without the fix on the "STILL PENDING" path you saw.

## 3. The indent errors — your count was exact, the reason wasn't

Ten errors, `fluxCommunication.js:249-258`, all this branch's. Confirmed, fixed.

**But not "under the repo's own rules", and there are no 14 pre-existing ones.** `.eslintrc.js` extends `eslint:recommended` only and configures **no `indent` rule anywhere** — no root, ZelBack or override config adds one. With the rule switched on, those ten are the *only* indent errors in the whole of `ZelBack` and `tests`.

The reason nothing enforced it is worse than a missing rule. **`npm run lint` cannot run at all, for anyone**: `babel.config.js` requires `@vue/cli-plugin-babel/preset`, which is not in `package.json` as any kind of dependency, so `npm install` can never supply it and `@babel/eslint-parser` dies before reading a file. The Vue frontend it was written for is not in this repo. And neither CI workflow runs lint, so nothing would have caught it regardless.

Fixed together: the babel config is deleted, the `import` plugin is declared (it was in `rules` but not `plugins`, so both `import/*` rules answered "rule not found"), and `indent` is on. `test-infra` is ignored rather than linted — it's ESM, where an import must carry its `.js` extension and this config forbids exactly that, so linting it produces ~1,300 errors that are the rules pointed at the wrong module system. That leaves 52 pre-existing errors across `ZelBack` and `tests`, none of them this branch's, left for their own change rather than swept into one about memory.

## 4. The prune input — this is better than "comment-or-align"

Aligned, and it's worth saying what it actually is. On the base, `messageStore` built this map itself, with the expiry filter in the loop. The move out of `messageStore` dropped **two** things at once: the input became the raw response, and the filter went missing. The raw-input half was caught and fixed on this branch. This is the other half of the same regression, not a pre-existing asymmetry.

You're right that it was bounded — the prune only deletes rows at or below the announcement's own timestamp, so a lapsed one could reach nothing that wasn't already lapsed. That bound lives in a database filter in another file, which is a poor place to keep the reason something is safe.

## 5. The stated bound — description corrected, code left alone

Correct: `evictions` and `stateEvents` are declared outside the slice loop and pushed into from within, so the slicing bounds the apprunning path only.

The code is right, though, and I checked the base to be sure rather than trusting the comment. Before slicing existed, the handler looped the entire response, collected evictions, and applied them first — whole-response scope, evictions ahead of the rest. Applying them per-slice would interleave them with apprunning storage, and since an eviction clears a node's locations outright, a later slice would re-add exactly what it removed. Accumulating preserves the original semantics under slicing.

So the claim was too broad, not the behaviour. The description now says which two things are exempt and why.

Cabecinha84 (approved):
Verdict: safe to ACK. All five findings from the previous round are genuinely fixed, I verified each against the code
  rather than the reply, and the extra sweep turned up nothing that should gate a merge.

  Head 3c8a04172, merge-base 68f593fdc = exactly the current tip of feature/placement-diversity, so the diff is
  precisely this PR's 23 commits / 47 files / +1917 −282. It still targets #1774 and cannot merge until that lands.

  The five findings, re-checked

  1. waitStarted() — fixed, and the fix is structurally better than what I asked for. All four accessors in
  fluxCommunicationUtils.js wait again. The three re-arming loops check and re-arm instead of blocking:
  checkDeterministicNodesCollisions (fluxNetworkHelper.js:1408), monitorNodeStatus (nodeStatusMonitor.js:85),
  fluxDiscovery (fluxCommunication.js:1346), and the two request handlers answer 503. I traced the wipe I was worried
  about: ipLocationSync.js:48 → ipLocationStore.js:642 can no longer see an empty list, because the accessor blocks. I
  also checked the case my original finding didn't cover — a later fetch returning empty after readiness:
  networkStateManager.js:382 only assigns #state under if (state.length), so a populated state is never regressed to
  empty. The deleteMany hazard is closed on both paths.

  Worth noting this is not a new blocking risk for HTTP callers: the base branch's waitStarted also blocked once start()
  had been called, so addOutgoingPeer, isFluxAvailable and getEnterpriseList behave exactly as they did before. The
  only changed case is "before start()", which is the hole being closed. And serviceManager.js:416 calls start() with no
  waitTimeoutMs, so it cannot reject and leave started pending forever.

  2. verifyPool requeue — fixed and bounded. dispatch() now re-enters via a do/while on handoverFailed. Termination is
  sound: job.attempts += 1 happens before the postMessage, so an unclonable chunk is abandoned after MAX_JOB_ATTEMPTS
  and its caller resolves all-false. My repro is now tests/unit/verifyPool.test.js:41, and the surrounding suite
  (elastic sizing, cap at spare cores, order across chunk boundaries, reap, crash-abandon, stop-while-pending) is
  genuinely good coverage.

  3. Lint — fixed, and his correction of my reasoning is right. I ran it: npx eslint --ext .js ./ now works, 0 indent
  errors repo-wide, 52 pre-existing errors (30 prefer-destructuring, 19 no-unused-vars, 2 vue/max-len, 1
  no-constant-condition), none in any file this PR touches. babel.config.js was indeed the blocker and nothing else
  references it; eslint-plugin-import and @babel/eslint-parser are both already devDependencies, and requireConfigFile:
  false was already set, so the deletion is safe. Neither CI workflow runs lint, confirmed.

  4. Prune input — aligned. The RUNNING_EXPIRY_MS filter is now applied where newestByIp is built
  (fluxCommunication.js:227), matching messageStore's own loop.

  5. Description corrected, code left alone. Correct call.

  What I ran

  - Full unit suite: 4828 passing, 5 pending, 23 failing. All 23 are dockerService tests needing a live Docker daemon
  (Cannot read properties of undefined (reading 'Id')) — identical to the pre-existing set, unrelated to this PR.
  - The ten PR-relevant suites in isolation: 485 passing, 0 failing.
  - eslint, as above.

  Additional checks that came back clean

  - Deflate compatibility. socketServer.js window bits are unchanged (true/true, comments only); only memLevel moved
  9→8. origin/master's server also uses true for both, so the outbound client pinning 13 negotiates fine against the
  entire deployed fleet. No handshake risk.
  - batchVerifyBroadcasts posting items directly — I expected this to enlarge the structured clone, but items is already
  {index, messageToVerify, pubKey, signature}. The old .map() was pure waste. The comment is accurate.
  - decryptMessage(msg, key = null) — both callers (dockerService.js:829, registryCredentialHelper.js:49) pass one
  argument, so behaviour is preserved and self-key verification runs for exactly the right cases.
  - Worker paths resolve (ZelBack/src/services/workers/, utils/verifyWorker.js), runInWorker's settle() is single-shot
  and terminates the worker on every path including a postMessage throw.

  Non-blocking nits — worth saying in the ACK, not worth another round trip

  1. Dead prune path in messageStore.storeBatchAppRunningMessages. It now has exactly one caller
  (fluxCommunication.js:235), always with prune: false. The function still builds v2AppsByIp on every slice and never
  uses it, and the if (!prune) branch is unreachable. Small allocation in precisely the path this PR is about — the
  option and the map can go, since pruneAppRunningLocations owns this now.
  2. The 503 in placementComputation also hits an internal caller, appSpawner.js:537. Unreachable today, because
  trySpawningGlobalApplication returns early on !globalState.dbReady and dbReady ⇒ network state ready — which is the
  same undocumented ordering coincidence this PR spent a commit arguing against relying on. If it ever were reachable,
  the throw lands in the outer catch and that app is skipped for the pass. Either a comment naming the internal caller,
  or an isReady() check in the spawner, would settle it.
  3. ipLocationStore.refreshNodeLocations still has no guard of its own. Defensible — the accessor contract is the right
  place, and I verified it holds — but two lines bailing on an empty nodeList would make the deleteMany safe
  independently of a contract in another file.
  4. configManager.reloadConfig records the hash of the truncated file before loadConfig(true) reads it. The config is
  correctly retained now, which is the fix; the residual is that if the watcher does not fire again after the writer
  finishes, the node keeps the old config until the next write. A short retry on the throw path would close it.
  5. config.fluxapps.nodeMonitorCheckIntervalMs is not defined in ZelBack/config/default.js (falls back to 120000),
  while its sibling nodeMonitorDosRecoveryDelayMs is. Add the key.
  6. scheduleReap() is only called from the worker message handler, so a burst ending in an exit or an abandon leaves
  the raised workers resident until the next successful verification.
  7. npm run lint still exits non-zero on the 52 pre-existing errors. His reasoning for not sweeping them into a memory
  PR is right, and CI doesn't run lint — just be aware the commit title "the linter runs" ships with a red script.

MorningLightMountain713:
Thanks for the ACK, and for re-checking each fix against the code rather than the reply — the `networkStateManager.js:382` check in particular covers the case my own finding didn't.

All seven nits are addressed below. Two did not survive verification and are answered rather than fixed; one turned out to be worse than either of us scoped it.

Rebased onto the new `feature/placement-diversity` tip, which has moved since your review. Unit suite: 4,865 passing, 18 pending, 0 failing.

Three changes here did not come from your nits, and are called out at the end so they are not mistaken for them.

---

**1. Dead prune path — removed.** Confirmed exactly as you describe: `v2AppsByIp` is consumed after the `if (!prune) return`, and the single caller passes `prune: false`, so the map was assembled on every slice and discarded. The option, the map and the unreachable tail are gone, and the caller no longer passes the argument. Removing locations a node no longer reports stays with `pruneAppRunningLocations`, which works from the newest broadcast per node rather than from whichever announcements shared a batch — which is why it was turned off here.

**2. The 503 reaching `appSpawner.js:537` — fixed, and it was worse than "that app is skipped for the pass".** The spawner's catch does not merely skip: it treats any throw there as a pre-install error and sets the app's hash in `trySpawningGlobalAppCache` with a six-hour TTL. So a location table that is seconds from ready would have cost that app six hours of spawn attempts on that node. It is now caught where it is raised and deferred to the next pass instead. Your reading of the reachability is right — it depends on `dbReady` implying network-state readiness, which is exactly the undocumented ordering this branch has been removing reliance on, so it is handled rather than commented.

**3. `refreshNodeLocations` guard — added.** Worth saying this was the most valuable of the seven. The function derives "departed" from absence, so an empty list means every held address has left and the whole collection is deleted in one call. You verified the accessor contract holds, and it does — but the cost of it ever not holding is the entire location view, rebuilt only over following passes and giving wrong placement answers meanwhile. Empty and absent lists are both refused now, at the cost of one skipped cleanup pass in a case that should not arise. Two tests cover it, and two existing tests that used `[]` as a shortcut into the drop and view-load paths now name a list that still departs the held address.

**4. `reloadConfig` — fixed, though not the way you suggested.** The finding is right: the hash was stamped before `loadConfig` read the file, so a read landing mid-write recorded the half-written file as the version this node holds. The suggested retry has nowhere to attach — `loadConfig` catches its own failures, so the caller never saw a throw. It now reports whether it read anything, and the hash is recorded only when it did.

**5. `nodeMonitorCheckIntervalMs` — added** to `ZelBack/config/default.js` at 120000, matching the fallback and its declared siblings.

**6. `scheduleReap()` — this one does not reproduce, and I think the call graph is misleading here.** The only workers that can still be resident when a burst ends are the ones that replied, and replying is what arms the reap: a worker that dies splices its own slot out of the pool in its exit handler, and a batch abandoned as unpostable never held a slot. Measured rather than argued — a burst in which every worker crashes ends holding exactly one worker at every pool size tried, and a mixed burst ends with its raised workers held by the replies that armed the reap for them. I could not construct a case that leaks. Rather than change a hot path on the strength of a scenario neither of us can produce, the reasoning is now recorded above `scheduleReap` so the next reader does not re-derive it from the call graph.

**7. `npm run lint` — no change here, and you are right that it should not be swept into this PR.** For the record the debt has drifted since your review: 47 errors across 19 files, not 52, and 25 `prefer-destructuring` rather than 30. None are in files this branch touches. The single `no-constant-condition` is a deliberate `while (true)` walk-up in `pathSecurity.js` rather than anything suspicious. Our intent is to clear it in its own PR on top of this stack once the rest of the chain is rebased and reviewed — hand-fixed rather than a repo-wide `--fix`, so the diff stays readable — which gets the script green without putting nineteen unrelated files into a memory PR.

---

## Not from your review

**A privilege check read a global nobody guaranteed.** `idService` and `verificationHelperUtils` read `globalThis.userconfig` directly, which `ConfigManager`'s constructor populates — so they relied on some other module having imported the manager first. A node answers HTTP before that is guaranteed, and a check reached in that window threw a `TypeError` that arrived at the caller as a server error on an ordinary restart. They read through the manager now; importing it is what guarantees the config has been read. Null is still answered for, because a failed load installs defaults with no zelid: the three session checks deny, and the two granting sites refuse the login rather than resolving the caller to `user`, which would hand an operator a session with their own rights stripped and no indication why.

**The empty-list guard says why it refused.** The `refreshNodeLocations` guard returned `{refreshed: 0, dropped: 0}` silently while its caller only logs a pass that changed something, so refusing to act on a node list that could not be obtained looked identical to a pass with nothing to do.

**Suite 76 was asserting two different claims as one.** It proved both that a single fault domain is permitted to hold all N instances — the Bahrain regression it exists for — and that it is capped at N. The first is arithmetic every node computes identically. The second depends on the admitting nodes having heard each other's claims: a node counts the claimants ahead of it in a gossiped list, and two nodes whose lists are each missing the other both count themselves inside the remaining share and both install. The surplus is not removed inside the window either, as `checkAndRemoveApplicationInstance` is paced at 44 blocks. It reproduced at roughly one run in fourteen, worse under a loaded gate. The live test now proves the regression it was written for; the ceiling is its own skipped test, naming deterministic placement as what makes it true by construction. No product change — the admission race is real and belongs to the spawner rebuild, not to this PR.

---

On the figures you gave that I could not reproduce: the unit suite here runs 4,865 passing with 0 failing rather than your 23 `dockerService` failures, because this environment has a live Docker daemon. Worth knowing when comparing runs.

Cabecinha84 (commented):
PR #1775 — Review of the current state

  Head 8a5e9aa35 · 35 commits · 58 files · +2200 / −327 · base = development @ 438874b01 (exact merge-base — #1774 has
  landed, this is now a direct-to-development PR, no stacking left).

  Prior history exists (one review round, an ACK on 08-10, seven nits answered on 08-11), but I re-derived everything
  below from the code as it stands now rather than from the replies.

  Verdict

  Safe to ACK. One minor defect worth fixing (F1 below), small enough to take as a follow-up rather than block on.
  Nothing I found threatens correctness of the sync path, the auth path, or boot.

  What I verified myself

  The risky mechanisms, walked rather than trusted:

  1. The waitStarted() semantics flip — the biggest blast radius in the PR. It went from "return immediately if not
  started" to "block until the node list is populated". I enumerated all 18 accessor call sites. The three
  self-re-arming loops that blocking would have retired for the life of the process — fluxDiscovery,
  checkDeterministicNodesCollisions (fluxNetworkHelper.js:1408), monitorNodeStatus (nodeStatusMonitor.js:81) — each
  check isReady() and re-arm on a timer instead of awaiting. HTTP handlers either ride out a short boot window
  (addOutgoingPeer) or refuse with 503. No boot deadlock: networkStateService.start() is called unconditionally at
  serviceManager.js:416, ahead of startDbDependentServices, and nothing inside the start path re-enters the accessors.

  2. The new 503 does not break app-message validation. checkPlacementFeasibility (placementFeasibility.js:378-386)
  wraps the call in try/catch and returns null, so registration/update/gossip validation degrades to "no feasibility
  check" instead of rejecting a valid app message. appSpawner.js:540 catches statusCode === 503 specifically and returns
  shortDelayTime — it does not fall into the pre-install catch that would have parked the app for six hours. Both
  correct.

  3. isReady() vs populated ordering. networkStateManager.fetchNetworkState emits populated and then sets #started =
  true (lines 407-409), so resolveStarted() fires while isReady() is still false. But onReady/waitStarted resume on a
  microtask, after the synchronous emit returns — by which time #started is true. No window.

  4. verifyPool failure handling — all four paths: post throws → job requeued, slot stays free, the do/while re-drives
  and terminates (attempts increment before the handover); worker exits holding a job → slot spliced, job requeued,
  replacement raised; worker times out → slot retired/spliced/terminated, job requeued, the later exit event is a no-op;
  stop() → every in-flight and queued job abandoned. abandonJob resolves new Array(chunk.length).fill(false), so index
  alignment with items[] in batchVerifyBroadcasts holds and an unverifiable batch is dropped, never accepted. Timers are
  unref()'d. Result order survives chunking and concurrent verify() calls.

  5. Eviction / prune ordering — the thing most likely to silently corrupt state. Evictions and other state events
  accumulate across all slices and are applied after the last one in the original [evictions, stateEvents] order.
  pruneAppRunningLocations runs from newestByIp built across the whole response from verified broadcasts only, and is
  skipped entirely if any location bulkWrite failed. The expiry filter at fluxCommunication.js:222 matches the one
  messageStore applies. storeBatchAppRunningMessages is now upsert-only and has exactly one caller, so no other path
  silently lost the prune it used to get for free.

  6. Smaller ones: queryFind in storeAppInstallingMessage is {name, ip} with no broadcastedAt, so spreading the
  withdrawal guard adds rather than overwrites. promotedFolderIds is always a Set by syncthingMonitor.js:772 (built at
  566, no early return between), and every newFoldersConfiguration entry comes from createSyncthingFolderConfig, which
  always sets type — the else delete branch can't fire on undefined. configManager defaults set zelid: null, so
  nodeAdminZelid() genuinely returns null and both idService sites refuse instead of quietly downgrading the operator to
  'user'. socketServer.js is memLevel 9→8 plus comments only — clientMaxWindowBits/serverMaxWindowBits: true are
  unchanged, so the handshake-rejection defect really is reverted on the listening side; the 13-bit window is set only
  on the outgoing peer socket.

  Test evidence: full unit suite 4856 passing / 5 pending / 23 failing — every failure is dockerService needing the
  runonflux/website container that npm run test:zelback:unit starts and I didn't. Targeted run of the eight suites this
  PR touches: 287 passing, 0 failing. CI build check green. I did not run the 82-suite integration harness (needs the
  docker fleet).

  Findings

  F1 — ensureIdentityVerified can loop on identity regeneration (minor, worth fixing)

  pgpService.js — on a public/private mismatch it calls createIdentity() but never records that it did.
  adjustPGPidentity writes the file only; globalThis.userconfig refreshes solely when configManager's fs watcher fires
  (apiServer.js:212). So every decryptMessage() inside that window re-derives from the same stale private key, sees the
  same mismatch, and spawns another openpgp keygen worker plus another full rewrite of config/userconfig.js.

  Callers are registryCredentialHelper.js:49 and dockerService.js:829 — both app-install paths, so a multi-component
  install can burst several calls before the watcher lands. The old code did this check exactly once per boot inside
  generateIdentity().

  Reachable only with a corrupt or hand-edited keypair, and it self-resolves once the reload lands. But the fix is one
  line: set verifiedPrivateKey = privateKey (or a separate repairAttempted marker) around the createIdentity() call so a
  persistent mismatch cannot repeat.

  Same fix covers the related point that createIdentity() is now reachable at runtime at all — the PR's own
  serviceManager.js:397 comment argues that rewriting userconfig.js isn't atomic and must be sequenced at boot, and this
  path reintroduces that write at an unsequenced moment.

  F2 — The elastic pool barely elasticises in production (informational; the description overstates it)

  SYNC_EVENTS_PER_SLICE = 250 sits below CHUNK_SIZE = 256, and all four batchVerifyBroadcasts call sites now receive a
  ≤250-item slice. So verify() always builds exactly one chunk, and ensureWorkers() can only exceed one worker when
  several sync handlers are in flight from different peers at once — realistically 1–3, never cpus-1. verifyPool.start()
  has no production caller at all.

  The practical consequence isn't in the "Accounted for" table: signature verification within a single sync response is
  now fully serial on one worker where it previously fanned out over cpus-1. Given the PR's own figure of 18 batches in
  10.5 hours that is almost certainly the right trade — but the write-up reads as though elastic scaling is a live
  property, and it mostly isn't. Worth stating plainly, or dropping the machinery.

  F3 — npm run lint still exits 1 (process note)

  The linter genuinely runs now, which is a real improvement over the broken state on development. But the repo is not
  clean: 47 errors, 10 in ZelBack/src (IOUtils, analyticsService, explorerService, fluxService, geolocationService,
  upnpService, utils/globalState, utils/pathSecurity, utils/volumeConstructor) and the rest in tests/unit. None in files
  this PR touches, and CI has no lint step (nodejs.yml), so nothing breaks — just don't read "the linter runs" as "lint
  is a gate". Also worth knowing: /test-infra/ is now entirely unlinted by the new .eslintignore entry.

  F4 — Harness coverage narrowed (a call to make consciously, not a defect)

  waitForInstanceCount gained exact: false, and test 76's ceiling assertion was split out into an it.skip. The skip
  documents a real over-installation race — two nodes each missing the other's claim both install, ~1 run in 14 — which
  now has no gate at all. It's pre-existing behaviour, not introduced here, and the reasoning (a flaky red trains people
  to skim past the gate) is sound. But merging means the ceiling stays ungated until the deterministic-placement
  spawner rebuild lands. If you want it tracked, ask for an issue linked from the skip comment.

MorningLightMountain713:
Thanks — F1 is a real defect and it is fixed, though not the way you proposed. F2's write-up point is fair and the body now says it; the characterisation of the machinery is not, and I have set out why. F3 and F4 are accurate observations that need no change here.

## F1 — fixed in `669b577dd`, and the window was worse than "self-resolves"

You are right about the mechanism: the mismatch path calls `createIdentity()` and records nothing, `adjustPGPidentity` only writes the file, and the in-process copy refreshes solely when the watcher fires. Two things make it sharper than a burst of wasted keygens:

**Other writers put the corrupt pair back.** `fluxService.js:1398/1453/1509/1559/1623` and `fluxNetworkHelper.js:1108` each rebuild `userconfig.js` whole from `userconfig.initial.*`, `pgpPublicKey` included. One firing inside the window writes the *old* keypair over the freshly generated one, and `verifiedPrivateKey` is still null — so the next decrypt regenerates again. That is a cycle, not a bounded burst.

**Concurrent installs mean overlapping whole-file rewrites.** There was no single-flight, so N concurrent `decryptMessage` calls give N overlapping rewrites of the file `configManager.loadConfig` already defends against partial reads of, and which `serviceManager.js:397` awaits at boot precisely because *"that write is not atomic — a reload landing inside it leaves the process with no `userconfig.initial` at all"*. The PR moved that write onto an app-install path where nothing sequences it.

**Why not the marker you suggested.** `verifiedPrivateKey = privateKey` records the key we have just proved corrupt as verified. Worse than the semantics: `adjustPGPidentity` wrapped everything in `try/catch` and logged, so `createIdentity` resolved successfully when nothing had been written. With the marker set, a node whose write failed would never attempt the repair again — the one node that needs it.

What landed instead:

- `adjustPGPidentity` calls `configManager.reloadConfig()` after the write, so the next reader sees the identity that is in force and the cross-writer revert cannot happen. It is already public and synchronous.
- Its `try/catch` is gone, so a failed write propagates and the repair stays armed.
- `repairInFlight` shares one repair between concurrent callers.
- `createIdentity` records the key it generated as verified — its halves were made together, so the check that saves has one answer. No marker, and no claim about a key we proved corrupt.
- A second warn line says what the repair cannot do: whatever was sealed to the previous public key was never readable here and has to be published again.

Three tests, two of them red against the unfixed code: three concurrent callers produce one keygen and one write; a second decrypt after the identity is replaced neither derives nor regenerates; and a failed write leaves the repair armed rather than disabled. That third one is the guard against the marker route.

## F2 — the write-up point is right, the characterisation is not

**Right, and fixed.** The body implied fan-out within a sync response. It now states plainly that a slice is one chunk on one worker, that verification within a response is serial where it previously spread over `cpus-1`, and that the pool scales across concurrent peers rather than within one response.

**Not right: "the machinery mostly isn't a live property."** Two of its three parts are the entire memory result and are load-bearing — one resident worker instead of `cpus-1` started at boot and never stopped, and idle reaping after 60s. That is what removes the 15 isolates holding ~490MB at 7.5 days. The elasticity engages too; it engages across concurrent peers, which is the shape the traffic has. What is unreachable is the `cpus-1` ceiling, not the mechanism.

**The mismatch you found is deliberate and measured**, and is stated in the change list and again under "settled by measurement": slice size sits below chunk size because at boot a node syncs from many peers at once and each concurrent response raises workers — 1024 was measured at 12 isolates and nearly double the boot peak. I tried 512 anyway and reverted it: raising the slice trades boot-peak worker isolates, which is what this PR exists to remove, for a fraction of a second on a path that runs 18 times in 10.5 hours. Measured here at 0.116 ms per verification, so a 250-event slice is ~29 ms of worker time and a full 2,500-event response is ~0.6 s serial.

`verifyPool.start()` and `stop()` having no production caller is correct — `verify()` raises workers itself. They stay as the test surface.

## F3 — accurate, and none of it is this PR's

47 problems, and I intersected the error list against this PR's changed files: the overlap is empty. #1782 takes the tree to zero, which I verified in a throwaway worktree at that branch. One trap if you reproduce it: without `config/userconfig.js` present, `import/extensions` reports 8 phantom "missing file extension" errors for requires of it, so a clean checkout looks red when it is not.

On `/test-infra/` being unlinted: it suppresses **1,311** problems, of which 736 `indent` and 532 `import/extensions` are the ESM-versus-CommonJS mismatch the ignore comment names, and 4 more are `db`/`print` in `fixtures/mongo-init.js`, which are mongosh globals rather than Node ones. Those numbers were being reported before this PR too — the entry made the output usable rather than hiding working coverage. The residue worth having is the 32 `no-unused-vars` hits, and getting them back means giving `test-infra` its own eslint setup: `sourceType: module`, `import/extensions` configured for ESM, and a mongosh env for `fixtures/`. That is a follow-up rather than something to smuggle into a memory PR.

## F4 — the skip stands, and now says so on its first line

The condition was already in the comment, three paragraphs down. A reader deciding whether a skipped test is abandoned or waiting reads the first line, so it says there that the spawner redesign is what makes the ceiling true by construction and what un-skips it (`5b25add9d`).

No issue link: this repository has one open issue in total, so an issue is not where anything here is tracked, and adding the stack's only one would be ceremony rather than tracking.

Cabecinha84 (approved):
ack

PR #1777, opened 2026-08-04
Monitoring: fix the dead endpoints, retire the on/off and pause controls, hold one store of extracted samples

MorningLightMountain713:
> **Base `development`** — #1774, #1775 and #1778 have merged; this carries its own 54 commits.

## Merge order

| | repo | PR |
|---|---|---|
| **1st** | fluxos-frontend (FluxCloud) | [#140](https://github.com/RunOnFlux/fluxos-frontend/pull/140) |
| **2nd** | flux-domain-manager (FDM) — merged **and deployed to all FDM servers** | [#154](https://github.com/RunOnFlux/flux-domain-manager/pull/154) |
| **3rd** | **this PR** → cut the release | |
| after rollout | palworld-server-website | [#35](https://github.com/RunOnFlux/palworld-server-website/pull/35) |
| after rollout | minecraft-server-website | [#23](https://github.com/RunOnFlux/minecraft-server-website/pull/23) |
| after rollout | deploy-with-git-frontend | [#29](https://github.com/RunOnFlux/deploy-with-git-frontend/pull/29) |

#140 goes first so the buttons are gone before the endpoints stop answering. #154 keeps an operator-stopped `g:` master named as the load-balancer primary; it must be **live on the FDM fleet before this release retires pause**, or stopping a `g:` server drops it from the load balancer and the primary can move to another node's copy of the data on the next start. #154 is a no-op until nodes serve `/apps/heldcomponents`, so it can be deployed ahead of this release. The three dashboards call routes this PR retires.

**The three dashboards deploy after the release has rolled out — not when this merges.** Merging puts nothing on a node; the deprecation notice on a going-away button costs nothing if it lands late, whereas a dashboard that tells customers to stop a `g:` server before #154 is on the fleet hands them a hold that can silently lose their edit.

Monitoring endpoints that resolve their own state, one store of extracted samples, and two controls retired.

## Breaking: `/apps/appstats` and `/apps/appmonitor` narrow

Both used to return dockerode's reading verbatim. Both now rebuild it from the stored extract, so most fields are gone.

Our own UI is unaffected — checked against `fluxos-frontend`. These are public node APIs, so **third-party tooling scraping the dropped fields is affected**; `/apps/appmonitor` is the charting endpoint and the likelier to be scraped directly.

<details>
<summary><b>Full list of dropped fields</b></summary>

| kept | dropped |
|---|---|
| `cpu_stats.cpu_usage.total_usage`, `system_cpu_usage`, `online_cpus` | `percpu_usage`, `usage_in_kernelmode`, `usage_in_usermode`, `throttling_data` |
| `precpu_stats.cpu_usage.total_usage`, `system_cpu_usage` | `precpu_stats.online_cpus` |
| `memory_stats.usage`, `limit`, `stats.inactive_file` | the rest of `memory_stats.stats`, `max_usage`, `failcnt` |
| `blkio_stats.io_service_bytes_recursive` — **normalised**: exactly two summed entries, `op` lowercased (`read`/`write`); cgroup-v1's per-device entries and capitalised ops are gone | every other blkio series |
| every interface's `rx_bytes`, `tx_bytes` | per-interface packet, error and dropped counters |
| `nanoCpus`, `disk_stats` | `read`/`preread`, `pids_stats`, `num_procs`, `storage_stats`, container id and name |

`stats.inactive_file` is kept because `docker stats` subtracts it from `usage` to report a container's real memory; without it no consumer can.

</details>

Holding a week of samples per component is what the narrowing buys.

### App monitoring on/off — `/apps/startmonitoring`, `/apps/stopmonitoring`

Monitoring is not a setting. FluxOS starts it for every app on install, at boot, and whenever a container starts. There is no state in which an app is installed and not monitored, so the endpoints offered to turn on something that was already running.

Turning it *off* was worse than a no-op. The samples the monitoring loop collects are the same samples the CPU throttler reads to decide whether an app is pinned at its allocation. Stopping monitoring stopped those samples, and nothing periodic restarts them — so an app whose container did not happen to restart was never throttled again, silently and indefinitely. A per-app switch that disables a node-wide protection is not a control we want an app owner holding.

Node-wide monitoring control is gone: `startAppMonitoringAPI` and `stopAppMonitoringAPI` answer with a `Deprecated` error unconditionally, with no privilege check, so the operator escape hatch is retired rather than restricted.

### Pause / unpause — `/apps/apppause`, `/apps/appunpause`

FluxOS models **desired state**. A paused app is a transient condition, not a desired state an operator would ask for, and modelling it properly means teaching every component that reasons about app state to distinguish "paused" from "stopped" and from "running".

That complexity buys nothing, because the customer use case is virtually non-existent — an owner who wants an app to stop can stop it.

It also does not currently work safely. Docker reports a paused container as `running`, so the reconciler and the load balancer both keep treating it as healthy and routing traffic to it while it is frozen and answering nothing.

### End state of the retired routes

The routes still exist and answer; they are not silently dropped.

**Read the `code` column as a body field, not a wire status.** All five answer **HTTP 200** carrying `{ status: 'error', data: { code: 410, name: 'Deprecated', message } }`; `/apps/appmonitorstream`'s message names the polling replacements (`/apps/appstats` for a live reading, `/apps/appmonitor` for the series).

| Route | Wire | Body |
|---|---|---|
| `/apps/startmonitoring`, `/apps/stopmonitoring` | `200` | `code: 410`, `Deprecated` — "Application monitoring is managed by the node and runs for every app. This endpoint no longer has any effect and will be removed." |
| `/apps/apppause`, `/apps/appunpause` | `200` | `code: 410`, `Deprecated` — "Pausing applications is no longer supported. Use appstop to stop an application." Ownership is still verified first, so an unauthorised caller gets `unauthorized` rather than the deprecation notice. |
| `/apps/appmonitorstream` | `200` | `code: 410`, `Deprecated` — "The stats stream has been removed. Poll /apps/appstats for a live reading or /apps/appmonitor for the collected series." The streaming behaviour itself is gone (see below). |

### Why in-band

FluxOS answers a request that reached a handler in the body, at HTTP 200, and reserves wire statuses for refusals that happen **before** one runs. Roughly 315 handlers answer in-band against 27 that set a status, and the 27 are middlewares (`requireHttps` 403, `routeGuards` 503/400), request-shape rejections (`paymentService` 413) and resource lookups (`fluxEventBus` 404) — not application results. `operationsController` writes `res.status(200).json(messageHelper.createErrorMessage(...))` explicitly, which is the rule stated in code.

`messageHelper.dataOrThrow` documents the reason: the in-band shape exists *"so a transport failure converted to an in-band error shape can never impersonate an empty result"*. A wire `410` that also carried an in-band error would blur exactly the line that distinction is drawn on.

These four handlers run, check ownership, and report that an operation is no longer supported. That is an answer, not a transport refusal, so it goes in the body.

An error rather than a success matters here: a caller must not be told a container stopped when it has not.

## Three monitoring endpoints shipped dead

The router called them with `(req, res)` and dropped the trailing arguments the handlers dereferenced. Every call threw and fell into the catch block, which returned an error response.

No test caught this, because the tests asserted *that a response happened* — which the catch block satisfies. The handlers now resolve their own state, and `tests/unit/routes.test.js` asserts what each path resolves to rather than that something came back.

## `/apps/appmonitorstream` no longer streams

It piped docker's stats stream straight to the response, and the stream's `onProgress` handler wrote the full raw reading to the node's log on every event — roughly one ~2.4 KB entry per second, per open connection, unbounded in duration, with no cap on concurrent streams.

It was gated on `appownerabove`, but **app owner is not a trusted role with respect to the node**: anyone can register an app and thereby hold owner privilege on it. A customer could fill a node's disk with logs.

It had zero frontend references and no other caller. `dockerContainerStatsStream` goes with it.

## One store, holding what is actually read

`dockerContainerStats` returned dockerode's output untrimmed and the whole thing went into the store verbatim, retained for seven days. Between them, the charts and the throttler read about thirteen values out of it.

Everything else was carried for seven days and never read: `percpu_usage` (one entry per core, twice), the ~40 kernel counters in `memory_stats.stats`, four unused `blkio_stats` arrays, `pids_stats`, `storage_stats`, and every interface's packet, error and dropped counters. (Every interface's byte counters are kept — a spec's `networkWith` attaches containers to further networks, and that traffic belongs in the reading.)

`extractSample` now reduces a reading to the values consumers read; `expandSample` puts it back into the docker shape on the way out, so the shape callers parse is preserved even though the field set is narrower (see above). The store is written every sample interval instead of every third one, so the series is **finer than before while being far smaller**.

`lastHourstatsStore` is gone. It never held an hour: the throttler emptied it on every pass and runs every `cpuCheckIntervalMs` (900,000 ms in production), so it held at most ~15 samples. The writer maintained a sliding hour; the reader consumed a 15-minute tumbling window and discarded the rest.

### Both clocks on every sample

Samples carry `timestamp` (wall clock — what charts plot, what a requested range means) and `elapsed` (monotonic, via `process.hrtime.bigint()`).

Retention, the CPU decision window and the hour bound all read `elapsed`, so an NTP step cannot expire the store early or re-admit samples a decision has already counted.

### The throttler reads the shared store through a watermark

`cpuDecisionWindow` returns samples newer than `max(lastCpuDecisionAt, now − 1h)`. Nothing is destroyed to make a decision, which is what allows the store to be shared with the charts at all.

Two properties that are load-bearing:

- **The watermark does not advance when a decision is skipped.** When docker cannot inspect the container, or fewer than five samples have arrived, the samples must carry into the next decision.
- **The window keeps an upper bound.** Previously the one-hour filter capped what accumulated across skipped passes. In a shared seven-day store there is no such cap, so without `max(…, now − 1h)` a decision following a long run of skips would compute the 80% rule over days of samples.

### Live view no longer duplicates the sampler

`/apps/appstats` made the same three docker calls live on every request, while a sample at most sixty seconds old sat unread in memory. Live mode polls it every 5 s alongside `appinspect` and `apptop`, so an open browser tab drove five docker calls per app every five seconds.

`appstats` now serves the freshest held reading if it is under `statsFreshnessMs` (5 s), otherwise takes one and holds it. Work is bounded by **time rather than by request count**: ten viewers cost what one viewer costs. Three mechanisms deliver that together: a landed reading serves every caller for the next five seconds, callers arriving while one is in flight share it (`statsInFlight`), and the 60 s storage cache bounds what each miss pays for the `du` walk. A viewer who is alone gains nothing — their 5 s poll lands just after the window expires — which is fine: the cost being bounded is the ten-viewer case.

### Other changes

- A disk reading the sizer did not stand behind (`partial` or `error`) is not stored; the last good reading is carried forward in its place, labelled `status: 'stale'` — the figure still charts, the label does not claim it is fresh.
- A container restart still resets the retained series: `startAppMonitoring` replaces the whole monitoring record on every start, as it always has. The seven days of retention is an upper bound, not a guarantee across restarts.
- Thinning past a day compares **timestamps rather than array positions**, so spacing holds whatever the sampler's cadence, and the newest sample is always included. The old `index % 20` was tuned to a 3-minute cadence and would have degraded silently when the cadence changed.
- Containers that are not running are no longer sampled. `getDockerContainerOnly` already returns the list-API summary carrying `State`, so the check is free.
- `getAppFolderSize` removed from both modules that exported it — neither had a caller and both shelled out to `sudo du`.
- The size log that stringified the whole store and re-encoded it to bytes every three minutes per app is removed.
- App disk usage is read from the filesystem once a minute rather than by shelling out — and on a dedicated volume it is now a different measurement, not just a cheaper one. A mount on its own device is read with `statfs` and reports that filesystem's allocated blocks; a mount sharing the node's filesystem still walks the tree with `du -sb` as before.

  A mount that cannot be measured no longer contributes a silent zero to a total reported as `status: 'success'`. `du` prints a running total and exits non-zero when it cannot read an entry — a temp file vanishing mid-walk on a busy app — so that number is kept; only a run producing no usable total leaves a mount unmeasured, and then the reading is `status: 'partial'` naming it, and is not cached.

  Expect the figure to move on apps that changed nothing. A dedicated volume now includes the filesystem's own overhead — journal, inode tables, reserved metadata — so a nearly empty one reports a few hundred MB where it used to report close to zero. In the other direction, `du -sb` counted apparent size, so an app using sparse files reported the holes as data and now reports only what is allocated.
- The sample interval moves to `config.fluxapps.statsSampleIntervalMs` (60,000 default), matching `cpuCheckIntervalMs` immediately beside it, which was already config-driven.

## Retiring pause changes what happens to a `g:` master

A paused container reports `State: 'paused'` but is still returned by `docker ps`. So it read as **running** to `listrunningapps`, to FDM's health check, and to the masterSlave election — which is how pausing held a `g:` primary in place while its owner worked on it. Retiring pause takes that away, and `appstop` is what is left.

`appstop` writes a durable operator-stop lock, and the node holding it correctly refuses to restart the component and excludes itself from the election. That intent never left the node: a peer saw no container and no in-memory commitment, concluded the component was free, and elected a new primary over an owner who had stopped theirs. Whether it happened turned on whether that node's FluxOS had restarted since it was elected — `controllerDesired` is in-memory and is written only at the moment a node wins an election, never re-asserted while it goes on being the primary.

`/apps/heldcomponents` now reports the durable lock alongside running containers and in-memory commitments. The answer is about **ownership**, not about what is up. `52-masterslave-operator-stop-recovery` gains the case end to end: stop the master, restart FluxOS to empty the in-memory state, clear FDM so the election opens at once, then assert the restarted node still reports the component held and the standby records a refusal.

`appDockerPause` and `appDockerUnpause` are deleted rather than left as dead exports, so no `.pause()` on a container remains anywhere in ZelBack. The retirement is structural rather than a handler that declines.

Two further fixes to the same election, found while covering it:

- A peer that **has** `/apps/heldcomponents` and cannot serve it answers in band, which is shape-identical to a peer too old to have the route — and the old-peer path falls back to the container list, which cannot see the lock. That now reads as `UNKNOWN`, so a node that cannot answer is never mistaken for one with nothing to hold.
- A staggered start whose turn comes due at index 0 checked only the nodes **ahead** of it, of which there are none, and started without asking any peer. It now escalates to asking every peer when the lower-only scope is empty. The per-place wait moves to `config.fluxapps.masterSlaveStaggerMs` so the path is reachable by a test at all.

## Release notes — what an operator sees

**`/apps/appmonitorstream` no longer streams.** Like the other four retired routes it answers HTTP 200 with `code: 410` in the body and a sentence naming the replacements. Nothing in our own UI called the stream on any branch, so the loss of the streaming behaviour reaches third-party tooling only — and such a caller is told where to go rather than left with a silent failure.

The replacement for a live view is `/apps/appstats`, which returns the most recent reading, or `/apps/appmonitor` for the retained series. Both narrow as described above.

## Testing

| Suite | Result |
|---|---|
| `npm run test:zelback:unit` | **5,323 passing / 18 pending / 0 failing** |
| `62-app-monitoring-endpoints` (new) | not re-run since the rebase |
| `63-cpu-throttling` (new) | not re-run since the rebase |

The integration harness is `workflow_dispatch` only and has no recorded runs, so the two new suites are stated as un-run rather than quoted from a local result. They are scheduled to run on this tip before merge, along with `52-masterslave-operator-stop-recovery`, which gains a case here.

**The CPU throttler had no test coverage at all**, and it physically re-allocates container CPU on every node. Eight characterisation tests went in **first, as their own commit, passing against the unmodified code**: throttle down at ≥80% of the window, no action below it, restore when the applied allocation is under spec, no decision on ≤4 samples, no sample counted twice, the burst-active skip, the docker-inspect-failed skip, and per-component decisions on a composed app.

Two mutations confirmed they were load-bearing — weakening `stats.length * 0.8` to `* 0.5` killed exactly the 80% test; removing the post-decision clear killed exactly the reuse test. Nothing else moved either time.

After the rework **all eight pass with identical assertions**; only the fixture's shape changed.

One test changed meaning deliberately: the burst-active case asserted the store was *emptied*. That was asserting the old mechanism — with a watermark the samples correctly remain for the charts. It now asserts what matters, which is that a later pass does not re-count them.

### The two new harness suites

`62-app-monitoring-endpoints` is the first coverage these endpoints have ever had. The unit tests prove `extract → expand` against a fixture, which is to say against assumptions about what docker populates and how it spells it. Only a real reading proves the extract picks up fields that are actually there — `blkio_stats` op casing and `networks.eth0` are exactly what a stub gets wrong and a live container does not.

`63-cpu-throttling` asserts the chain unit tests cannot reach, because they hand the throttler its window directly. The sampler and the throttler are separate functions on separate timers; if the sampler stops filling the store, every unit test still passes, no app is ever throttled again, and nothing says so. It also asserts the allocation comes **back**, without which the suite would pass against a throttler that only ratchets down.

## Notes for review

- **No migration surface.** `appsMonitored` is a plain in-memory object — not persisted, not broadcast. A restarted node starts empty. The wire contract is only the `/apps/appmonitor` and `/apps/appstats` response bodies; both narrow as described above. Any change to the response *shape* belongs in the API versioning work, not here.
- `disk_stats.used` is still computed and read by nothing. Left deliberately — it is part of a response shape, and removing it is a wire change.
- **Why the retired routes still answer, rather than being dropped.** Ordering the frontends first does not close the window on its own, because bundles are cached: a browser still holding an older one calls the deprecated routes after the new frontend deploys. Answering with a reason makes that call fail visibly, in the body the UI already parses, instead of appearing to succeed.

---


Cabecinha84 (commented):
PR #1777 — deep review

  refactor/monitoring-single-store → feat/volume-file-operations · 33 files, +2104/−1882, 23 commits · no prior human
  review · mergeable · Codecov patch 84.7%

  What I verified myself (not taken from the description)

  ┌─────────────────────────────────────────────────────┬───────────────────────────────────────────────────────────┐
  │                       Check                       │                           Result                            │
  ├───────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────┤
  │ Affected unit suites (appInspector,              │                                                              │
  │ monitoringOrchestrator, appController,           │ 288 passing / 0 failing locally                              │
  │ appUtilities, routes, appReconciler, 3           │ (NODE_CONFIG_DIR=tests/unit/globalconfig)                    │
  │ middleware, nodeConfirmationService)             │                                                              │
  ├──────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  │ dockerContainer.State !== 'running' skip — the   │ Correct. getDockerContainerOnly returns a                    │
  │ highest-risk line in the PR, since a wrong shape │ dockerListContainers entry (dockerService.js:183), whose     │
  │  silently kills all sampling and the throttler   │ State is the string "running"                                │
  ├──────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  │ cpuThrottlingRuns >= stats.length * 0.8 with an  │ Guarded. if (inspect && stats && stats.length > 4) is still  │
  │ empty window (0 >= 0 → throttle everything)      │ in place on both the v≤3 and per-component paths             │
  ├──────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  │ deprecatedPauseResponse dropped its try/catch    │ Not reachable. verifyPrivilege has its own catch and returns │
  │ around verifyPrivilege — would hang the request  │  false (verificationHelper.js:44)                            │
  │ on rejection                                     │                                                              │
  ├──────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  │ appConstants.appsFolderPath (flat require in     │                                                              │
  │ appUtilities, vs the old destructured {          │ Resolves. appConstants exports it flat                       │
  │ appConstants } in appInspector)                  │                                                              │
  ├──────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  │                                                  │ No. getContainerStorage is consumed only by the two          │
  │ Does the new disk accounting feed app removal?   │ monitoring paths; checkStorageSpaceForApps uses              │
  │                                                  │ dockerGetUsage().Containers[].SizeRootFs. No eviction risk   │
  ├──────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  │ Can dedicatedVolumeUsage wander onto an          │ No. Every mount Source comes from                            │
  │ unrelated host filesystem?                       │ volumeConstructor.constructVolumes, always under the apps    │
  │                                                  │ folder                                                       │
  ├──────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  │ Dangling refs to getAppFolderSize,               │                                                              │
  │ appMonitorStream, dockerContainerStatsStream,    │ None                                                         │
  │ lastHourstatsStore, appsMonitoredTemplate,       │                                                              │
  │ appDeleteDataInMountPoint                        │                                                              │
  ├──────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  │ fs.statfs availability                           │ Node v22 — fine                                              │
  ├──────────────────────────────────────────────────┼──────────────────────────────────────────────────────────────┤
  │                                                  │ Checked fluxosuidevelopment@ec8bda0: it reads exactly the    │
  │ Frontend compatibility with expandSample         │ fields emitted, and does i.op.toLowerCase() === "read", so   │
  │                                                  │ the lowercase op reconstruction matches                      │
  └──────────────────────────────────────────────────┴──────────────────────────────────────────────────────────────┘

  The engineering is genuinely good. The bug it fixes is real (three endpoints threw on every call), the CPU-throttler
  characterisation tests landing first as their own commit against unmodified code is the right way to do this, and the
  watermark design is sound — max(lastCpuDecisionAt, now − 1h) correctly preserves both invariants it claims.

  I did not find a correctness bug that would break a node. But there are two things I would not ack without, and three
  small code fixes.

  ---
  Blocking-ish

  1. "Response bodies are unchanged" is not true

  The PR states this twice. It isn't. /apps/appstats used to return dockerode's reading verbatim; it now returns a
  reconstruction from 13 stored values. Dropped from the public node API with no mention:

  read / preread · pids_stats · num_procs · storage_stats · memory_stats.stats (the ~40 kernel counters) ·
  memory_stats.max_usage / failcnt · cpu_stats.cpu_usage.percpu_usage / usage_in_kernelmode / usage_in_usermode ·
  cpu_stats.throttling_data · precpu_stats.online_cpus · every blkio array except io_service_bytes_recursive · every
  network interface except eth0 · container id / name.

  Our UI is fine — I checked. But this is a node API that app owners, dashboards and third-party tooling call directly,
  and throttling_data / memory_stats.stats are the two most commonly scraped of that set. The change may well be the
  right call; presenting it as "no wire change" is what I'd push back on. Either state the narrowing plainly in the PR
  body and release notes, or keep the extract for the store and pass the live reading through untouched on
  /apps/appstats (it already has the full object in hand on a miss — though that costs response-shape determinism, so
  I'd lean to just documenting it).

  2. Merge order is coupled to the frontend, and the PR doesn't say so

  fluxosuidevelopment@ec8bda0 (current development) still calls all four retired endpoints:

  src/services/AppsService.js:63,72        /apps/apppause, /apps/appunpause
  src/services/AppsService.js:383–405      /apps/startmonitoring, /apps/stopmonitoring
  src/@core/components/AppControl.vue      the buttons, single-node and multi-node paths

  If this lands before fluxos-frontend#140, every one of those buttons returns Deprecated to the user. And node
  operators pinned to an older UI bundle get that regardless of merge order. Needs an explicit statement of the ordering
  constraint (and ideally a note that older UI bundles will show errors on those buttons — which is arguably fine, but
  should be a decision, not a surprise).

  ---
  Worth fixing before merge (small)

  3. getContainerStorage can silently report zero disk usage

  appUtilities.js — dedicatedVolumeUsage stats appConstants.appsFolderPath inside the per-mount try. That stat is a
  per-call, not per-mount, condition. If it ever fails (FLUX_APPS_FOLDER misconfigured, folder missing, permissions),
  every mount hits the catch and gets continued, and the function still returns:

  { bind: 0, volume: 0, rootfs: <n>, used: <n>, status: 'success' }

  The UI only blanks its values on status === 'error', so it would display 0 % disk use indefinitely — cached, for every
  app. Hoist the apps-folder stat out of the loop, do it once, and let its failure either mark the result status:
  'error' or fall back to walkedUsage.

  4. The watermark can drop samples the old code counted

  appInspector.js — stats = cpuDecisionWindow(...) snapshots via .filter(), then two awaits run (dockerContainerInspect,
  cpuBurstHelper.isBurstActive) before lastCpuDecisionAt = monotonicMs(). Samples pushed during those awaits are
  excluded from this decision and then excluded from the next one too — they're lost. The old code held a live array
  reference, so they were counted.

  Tiny window in practice, but it's the exact class of thing the watermark exists to prevent. One-liner:

  const decisionAt = monotonicMs();
  stats = cpuDecisionWindow(appsMonitored[app.name]);
  // ...
  appsMonitored[app.name].lastCpuDecisionAt = decisionAt;

  5. Disk usage now means two different things

  dedicatedVolumeUsage returns (blocks - bfree) * bsize — allocated blocks, including ext4 journal, inode tables and
  metadata. walkedUsage still returns du -sb, which is apparent size (sparse files at full logical size). Same field,
  two definitions, picked by whether the volume happens to be dedicated.

  User-visible effect: every app on a dedicated volume will report a few hundred MB used at rest instead of ~0, and
  sparse-file apps (databases, preallocated game servers) will report less than before. It's arguably the more honest
  number, and I confirmed it can't trigger a storage violation — but the PR sells it as a pure perf change ("read from
  the filesystem once a minute rather than by shelling out"). Support will get tickets. Call it out, and make sure the

  I did not find a correctness bug that would break a node. But there are two things I would not ack without, and three
  small code fixes.

  ---
  Blocking-ish

  1. "Response bodies are unchanged" is not true

  The PR states this twice. It isn't. /apps/appstats used to return dockerode's reading verbatim; it now returns a
  reconstruction from 13 stored values. Dropped from the public node API with no mention:

  read / preread · pids_stats · num_procs · storage_stats · memory_stats.stats (the ~40 kernel counters) ·
  memory_stats.max_usage / failcnt · cpu_stats.cpu_usage.percpu_usage / usage_in_kernelmode / usage_in_usermode ·
  cpu_stats.throttling_data · precpu_stats.online_cpus · every blkio array except io_service_bytes_recursive · every
  network interface except eth0 · container id / name.

  Our UI is fine — I checked. But this is a node API that app owners, dashboards and third-party tooling call directly,
  and throttling_data / memory_stats.stats are the two most commonly scraped of that set. The change may well be the
  right call; presenting it as "no wire change" is what I'd push back on. Either state the narrowing plainly in the PR
  body and release notes, or keep the extract for the store and pass the live reading through untouched on
  /apps/appstats (it already has the full object in hand on a miss — though that costs response-shape determinism, so
  I'd lean to just documenting it).

  2. Merge order is coupled to the frontend, and the PR doesn't say so

  fluxosuidevelopment@ec8bda0 (current development) still calls all four retired endpoints:

  src/services/AppsService.js:63,72        /apps/apppause, /apps/appunpause
  src/services/AppsService.js:383–405      /apps/startmonitoring, /apps/stopmonitoring
  src/@core/components/AppControl.vue      the buttons, single-node and multi-node paths

  If this lands before fluxos-frontend#140, every one of those buttons returns Deprecated to the user. And node
  operators pinned to an older UI bundle get that regardless of merge order. Needs an explicit statement of the ordering
  constraint (and ideally a note that older UI bundles will show errors on those buttons — which is arguably fine, but
  should be a decision, not a surprise).

  ---
  Worth fixing before merge (small)

  3. getContainerStorage can silently report zero disk usage

  appUtilities.js — dedicatedVolumeUsage stats appConstants.appsFolderPath inside the per-mount try. That stat is a
  per-call, not per-mount, condition. If it ever fails (FLUX_APPS_FOLDER misconfigured, folder missing, permissions),
  every mount hits the catch and gets continued, and the function still returns:

  { bind: 0, volume: 0, rootfs: <n>, used: <n>, status: 'success' }

  The UI only blanks its values on status === 'error', so it would display 0 % disk use indefinitely — cached, for every
  app. Hoist the apps-folder stat out of the loop, do it once, and let its failure either mark the result status:
  'error' or fall back to walkedUsage.

  4. The watermark can drop samples the old code counted

  appInspector.js — stats = cpuDecisionWindow(...) snapshots via .filter(), then two awaits run (dockerContainerInspect,
  cpuBurstHelper.isBurstActive) before lastCpuDecisionAt = monotonicMs(). Samples pushed during those awaits are
  excluded from this decision and then excluded from the next one too — they're lost. The old code held a live array
  reference, so they were counted.

  Tiny window in practice, but it's the exact class of thing the watermark exists to prevent. One-liner:

  const decisionAt = monotonicMs();
  stats = cpuDecisionWindow(appsMonitored[app.name]);
  // ...
  appsMonitored[app.name].lastCpuDecisionAt = decisionAt;

  5. Disk usage now means two different things

  dedicatedVolumeUsage returns (blocks - bfree) * bsize — allocated blocks, including ext4 journal, inode tables and
  metadata. walkedUsage still returns du -sb, which is apparent size (sparse files at full logical size). Same field,
  two definitions, picked by whether the volume happens to be dedicated.

  User-visible effect: every app on a dedicated volume will report a few hundred MB used at rest instead of ~0, and
  sparse-file apps (databases, preallocated game servers) will report less than before. It's arguably the more honest
  number, and I confirmed it can't trigger a storage violation — but the PR sells it as a pure perf change ("read from
  the filesystem once a minute rather than by shelling out"). Support will get tickets. Call it out, and make sure the
  frontend PR knows.

  ---
  Nits

  - PR body is stale on one point. It says "Node-wide monitoring control is retained but reserved for the Flux team" —
  that was commit 71ac97e2d, but the final state has startAppMonitoringAPI/stopAppMonitoringAPI returning 410
  unconditionally, with no auth check at all. No control is retained. Fix the body, or confirm losing the operator
  escape hatch is intended.
  - No single-flight in latestStats. "Ten viewers cost what one viewer costs" holds for staggered polls, but ten
  simultaneous requests landing just past the 5 s window all see a stale sample and all issue docker calls. A promise
  cache would make the claim literally true.
  - dockerService.appDockerPause / appDockerUnpause are now dead — unreachable except from their own unit tests. Remove
  with the feature or leave a note.
  - /apps/appmonitorstream removal is well-justified (unbounded ~2.4 KB/s/connection log write, gated on a role anyone
  can obtain by registering an app), but it's an endpoint deletion with no deprecation window while the other four got
  410s. Deserves an explicit API-changelog line.
  - nanoCpus is refreshed on run % 3 === 1. If that first inspect throws, samples carry nanoCpus: null until run 4, and
  the frontend's rawCpu / (nanoCpus / cores / 1e9) goes to Infinity for up to 3 minutes. Old code pushed no sample at
  all in that case.
  - Scope creep: three nodeConfirmationService test commits, the routes.js middleware extraction, and
  minimumInstancesV8Block in test-infra/config/shared.js have nothing to do with monitoring. All harmless, all
  reviewable, but they widen the blast radius of a revert.
  - appMonitor with no range serializes the whole 7-day store (10,080 samples/component). Still far smaller than before,
  and the UI always passes a range — but nothing clamps it.

  ---
  Verdict

  Ack with conditions — don't ack as-is.

  The code is sound and I could not break it. What I'd ask for before approving:

  1. Correct the "response bodies are unchanged" claim and document the /apps/appstats field narrowing (required — it's
  a public API change being described as not one).
  2. State the merge-order constraint against fluxos-frontend#140 (required).
  3. Fix the silent-zero path in getContainerStorage (#3) and the watermark timing (#4) — both one-liners.
  4. Document the disk-usage semantic change (#5).

  None of that is architectural. If David turns those around, it's a clean ack.

MorningLightMountain713:
Thanks — all five addressed, two by code and three by correcting the description. Branch is now `a2eb0a094`.

**1. "Response bodies are unchanged" — you were right, and it appeared three times, not two.** The body now names the `/apps/appstats` narrowing field by field: `read`/`preread`, `pids_stats`, `num_procs`, `storage_stats`, `memory_stats.stats`, `max_usage`/`failcnt`, `percpu_usage`, the kernel/user split, `throttling_data`, `precpu_stats.online_cpus`, every blkio series except `io_service_bytes_recursive`, every interface except `eth0`, and the container id and name.

I checked our own UI independently of your check, on a different branch (`fluxos-frontend` at `beb1aa1`): none of the dropped fields are referenced, it reads `networks.eth0.rx_bytes`/`tx_bytes` only, and it matches blkio entries with `i.op.toLowerCase()`, so the lowercase reconstruction lines up.

I did **not** take the option of passing the live reading through on a cache miss. It would return 13 fields when warm and 40-plus when cold, so a consumer scraping `throttling_data` would get it intermittently depending on timing — worse than a consistent narrowing.

**2. Merge-order coupling — stated.** The body now carries the constraint against fluxos-frontend#140 and the note that operators on an older bundle see errors on those buttons regardless of merge order.

**3. `getContainerStorage` silent zero — fixed** in `5fce81347`, though not quite as scoped. The apps-folder stat is one fact about the call, so it is resolved once, where the classifying happens, and a failure to resolve it fails the call rather than sizing every mount at zero and caching `status: 'success'`.

Falling back to walking was tempting and wrong: `statfs` charges a dedicated volume its whole filesystem (which is why sibling mounts are deduped by device) while `du -sb` reports apparent size, so substituting one for the other keeps the field populated while changing what it means. Refusing to answer is the honest failure.

Two things your scoping missed, both caught by the existing tests: a container with **no mounts** has nothing to classify and must not need the apps folder at all, and resolving it before the docker inspect swallows the inspect's error. It now sits inside the mounts branch.

**4. Watermark — fixed** in `a2eb0a094`, one line as you said, and it covers all four stamp sites (both paths, each stamping on the burst-active skip and on the decision). There is a test that interleaves a sample arriving mid-decision; it fails against the old ordering.

**5. Disk-usage semantics — documented.** Allocated blocks versus apparent size, and that apps on dedicated volumes will report a few hundred MB at rest where they used to report ~0.

**Nits.** The stale "node-wide monitoring control is retained" line is corrected — both endpoints return 410 unconditionally with no privilege check, so the escape hatch is retired rather than restricted. The single-flight, `appDockerPause`/`appDockerUnpause` deadness, the `/apps/appmonitorstream` changelog line, the `nanoCpus` refresh gap and the unclamped `appMonitor` range are all fair; none are changed here and I would rather they were their own change than widen this one further.

Verified at the top of the stack: **85 of 85 suites green** at `f6c6bc399`.

Cabecinha84 (commented):
PR #1777 — re-review of the current state (efdbfb832)

  Status: refactor/monitoring-single-store → feat/volume-file-operations · 33 files, +2167/−1882, 25 commits · MERGEABLE
  / CLEAN · CI build passing · Codecov patch 84.7%. Head has been force-pushed twice since my Aug 8 review (a2eb0a094 →
  f6c6bc399 → efdbfb832).

  What I verified myself on the current head

  ┌───────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────┐
  │                               Check                               │                   Result                    │
  ├───────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────┤
  │ Affected unit suites (appInspector, appUtilities,                 │                                             │
  │ monitoringOrchestrator, appController, routes, appReconciler,    │ 254 passing / 0 failing                      │
  │ alwaysRespond, isLocal, requireHttps)                            │                                              │
  ├──────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────┤
  │                                                                  │ 23 failing — identical 23 failures on the    │
  │ dockerService.test.js                                          │  branch 3b4fe64e1. Pre-existing/environmental, │
  │                                                                │  not this PR                                   │
  ├────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────┤
  │ npx eslint on every changed source file +                      │ Clean                                          │
  │ ZelBack/src/middlewares/                                       │                                                │
  ├────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────┤
  │ Dangling refs: getAppFolderSize, appMonitorStream,        │                                                     │
  │ dockerContainerStatsStream, lastHourstatsStore,           │ None                                                │
  ├─────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────┤
  │ Every startAppMonitoring/stopAppMonitoring call site    │                                                       │
  │ migrated to the new signature (appUninstaller,          │ All 30, none passing the dropped appsMonitored arg    │
  │ advancedWorkflows, appInstaller, appController,         │                                                       │
  │ appReconciler, orchestrator)                            │                                                       │
  ├─────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────┤
  │ Circular-dependency risk from the new top-level         │                                                       │
  │ requires (monitoringOrchestrator → appQueryService,     │ Loads clean; globalState is a leaf                    │
  │ appInspector → globalState)                             │                                                       │
  ├─────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────┤
  │                                                         │ Executed appMonitor() against a real store entry —    │
  │ extract → expand round trip                             │ shape is correct, blkio op lowercase, networks.eth0   │
  │                                                         │ present                                               │
  ├─────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────┤
  │                                                         │ No. getContainerStorage has exactly 2 call sites,     │
  │ Does the new disk accounting feed eviction?             │ both monitoring. checkStorageSpaceForApps still uses  │
  │                                                         │ dockerGetUsage().SizeRootFs                           │
  ├─────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────┤
  │                                                         │ No. Volume images are per-component                   │
  │ Can statfs double-charge components of a composed app?  │ (flux<component>_<app>FLUXFSVOL), so each container's │
  │                                                         │  statfs reads its own filesystem. The in-call         │
  │                                                         │ countedDevices dedup covers sibling mounts            │
  ├─────────────────────────────────────────────────────────┼───────────────────────────────────────────────────────┤
  │ Harness suites 62/63 registered                         │ Yes — run-all.sh glob-discovers tests/*.js            │
  └─────────────────────────────────────────────────────────┴───────────────────────────────────────────────────────┘

  The two code fixes: both landed, both correct

  #3 getContainerStorage silent zero — fixed, and better scoped than I proposed. The apps-folder stat is resolved once,
  inside the Mounts.length branch, and its failure propagates to the outer catch → status: 'error', not cached. Putting
  it inside the mounts branch (rather than at the top of the function, as I'd said) is right: a container with no mounts
  has nothing to classify, and hoisting it above the docker inspect would have swallowed the inspect's error. Declining
  to fall back to walkedUsage is also the correct call — substituting apparent size for allocated blocks keeps the
  field populated while silently changing its meaning.

  #4 Watermark timing — fixed. decisionAt = monotonicMs() is taken before cpuDecisionWindow() and stamped at all four
  sites (v≤3 and per-component, each on the burst-active skip and on the decision). The regression test genuinely
  interleaves a sample during the dockerContainerInspect await and asserts lastCpuDecisionAt < midFlight.elapsed — it
  fails against the old ordering. I traced the boundary arithmetic: no sample is counted twice and none is stranded.

  Three documentation items are not in the PR body

  This is where I'd stop short of an ack. The reply says all five were addressed; the body currently carries only one of
  the three doc items.

  1. /apps/appmonitor narrows exactly like /apps/appstats did — and the body says twice that it doesn't. This is new,
  and it's the same defect that #1 was about.

  The old store held the full reading: statsStore.push({ timestamp: now, data: statsNow }) where statsNow was
  dockerContainerStats() plus disk_stats/nanoCpus. appMonitor returned that array verbatim. It now returns { timestamp,
  data: expandSample(stats) } — the identical 13-field reconstruction. Every field listed in the /apps/appstats
  paragraph is dropped from /apps/appmonitor too.

  The body states the opposite in two places:

  - line 9: "/apps/appmonitor is unchanged."
  - line 119: "/apps/appmonitor is unchanged and /apps/appstats narrows as described above."

  Our own UI is fine (you and I both checked, on different frontend commits). But /apps/appmonitor is the charting
  endpoint — it is more likely to be scraped by third-party tooling than appstats, and the body currently tells a reader
  it is safe.

  2. Merge-order coupling is still absent. The body has one line — "The frontend counterpart is fluxos-frontend#140 — it
  removes the buttons these endpoints backed." No ordering constraint, no note about older UI bundles. grep for
  "order", "older", "bundle" in the body returns nothing. fluxos-frontend#140 is still OPEN (base development, last
  touched Aug 4), so the coupling is live.

  3. Disk-usage semantics still undocumented. The body still reads "App disk usage is read from the filesystem once a
  minute rather than by shelling out" — sold as a pure perf change. No "apparent", no "allocated", no "statfs", no "at
  rest". Apps on dedicated volumes will report a few hundred MB of ext4 metadata where they used to report ~0, and
  sparse-file apps will report less.

  Possibly these were written and lost in the body edit at 16:02 (after the 15:14 reply). Either way they aren't there
  now.

  New, minor

  4. statsSampleIntervalMs is missing from the unit-test config. It was added to ZelBack/config/default.js (60000) and
  test-infra/config/shared.js (2000), but tests/unit/globalconfig/default.js — which is a full fork of the production
  config, and what npm run test:zelback:unit points NODE_CONFIG_DIR at — still has cpuCheckIntervalMs with no
  statsSampleIntervalMs beside it. I confirmed the value resolves to undefined under the test config, which makes
  setInterval(cb, undefined) a 1 ms timer. Nothing spins today because no unit test calls startAppMonitoring for real
  (appInspector.test.js:1648 only asserts it's a function), but it's a live trap for the next test that does, and it
  means the configured cadence is unexercised at unit level. One line.

  5. monitoringOrchestrator.stopMonitoringOfApps now has zero callers — its only users were the two API bodies that
  became 410s. Dead export.

  On the deferred nits

  nanoCpus refresh gap and the unclamped appMonitor range were all acknowledged and explicitly deferred. I confirmed
  each is still present and I agree they're separate changes. One correction to my own earlier nanoCpus nit: if the
  run-1 inspect throws, the catch skips the whole sample, so nothing is pushed carrying nanoCpus: null — the null window
  only opens if inspect resolves without HostConfig. Narrower than I said.

  I'd still ask for the appmonitorstream changelog line, since it's the one route deleted outright while four others got
  410s.

  Verdict

  Don't ack yet — but nothing architectural is left.

  The engineering is in good shape and I could not break it: the two code fixes are correct, the watermark arithmetic
  holds, the disk accounting can't reach eviction, the volume-per-component layout makes statfs safe, and the
  PR-relevant suites are green with lint clean.

  What I'd ask for is one more pass on the body plus a one-line config fix:

  1. Correct "/apps/appmonitor is unchanged" in both places and fold it into the narrowing paragraph — it drops the same
  fields (required; this is the same public-API misstatement as #1, on the endpoint more likely to be scraped).
  2. Add the merge-order constraint against fluxos-frontend#140 and the older-bundle note (required — it was reported as
  done and isn't there; #140 is still open).
  3. Document the disk-usage semantic change (reported as done, isn't there).
  4. Add statsSampleIntervalMs to tests/unit/globalconfig/default.js.

MorningLightMountain713:
Thanks — all four asks are done, plus the two you raised as minor and the changelog line. Two of your findings turned out to have a stronger argument behind them than the body was making, and one of them narrows a risk rather than widening it. Head is now `ae199f640`; the branch was also rebased onto the current `feat/volume-file-operations` today, so the commit ids below are post-rebase.

## The three documentation items

**1. `/apps/appmonitor` narrows, and the body said the opposite.** You were right, and it was wrong in both places — lines 9 and 119. `appMonitor` maps every sample through `expandSample`, and `/apps/appstats` reaches the same function via `latestStats`, so both drop exactly the same fields. The two endpoints are now described together in one paragraph, with the field list stated once, and the note that `/apps/appmonitor` is the charting endpoint and so the likelier of the two to be scraped directly.

**2. The merge-order constraint is in, and it comes with a correction that narrows this.** fluxos-frontend#140 is still open, so the coupling is live, and the body now says merge it first and explains why ordering alone does not close the window — bundles are cached, so a browser holding an older one calls the deprecated routes after the new frontend deploys. That is the reason those four routes answer `410` with a sentence rather than being dropped: the call fails visibly instead of appearing to succeed.

The correction: I checked `development` for what the frontend actually calls. `startmonitoring` and `apppause` appear in two files each — **`appmonitorstream` appears in zero**. So the one route removed *outright* is the one nothing of ours has ever called, and the entire coupling is the four routes that still answer. The blast radius of the deletion is third-party tooling only.

**3. The disk-usage semantics are documented, and they change in both directions.** You were right that the body sold a measurement change as a performance change. What actually happens is a split:

| mount | before | after |
|---|---|---|
| on its own device | `sudo du -sb` | `fs.statfs` → `(blocks - bfree) * bsize` |
| sharing the node's filesystem | `sudo du -sb` | unchanged |

`dedicatedVolumeUsage` compares the mount's `st_dev` against the apps folder's; same device falls through to `walkedUsage`, which is still the shell-out. So a dedicated volume now reports that filesystem's allocated blocks including ext4's own overhead — journal, inode tables, reserved metadata — where a nearly empty one used to report close to zero. And in the other direction, `du -sb` counted apparent size, so an app using sparse files reported holes as data and now reports only what is allocated. The body says both, because an operator watching that figure across an upgrade sees it move on apps that changed nothing.

## The two minor ones

**4. `statsSampleIntervalMs` is in the unit config** (`c19ebc8e6`). Confirmed your reading by resolving it: it read `undefined` under `tests/unit/globalconfig`, and its only consumer passes it straight to `setInterval`, which coerces `undefined` to `1`. Nothing spins today because no unit test arms it for real — `appInspector`'s test asserts only that `startAppMonitoring` is a function, and `appController` and `appReconciler` stub it — so it was a trap for the first test that does, and the reason the configured cadence went unexercised at unit level.

**5. `stopMonitoringOfApps` is deleted** (`ae199f640`), along with its `describe` block. Its four call sites were the handler bodies that became `410`s, so this PR is what killed it. `startMonitoringOfApps` stays — `serviceManager.js:476` calls it at boot — and `resolveAppSpecs` keeps its remaining caller.

Worth saying why it was not kept for a shutdown path, since that was the obvious alternative. There *is* a real graceful shutdown — `apiServer.js` handles `SIGTERM`, broadcasts `fluxnodesigterm`, shortens location TTLs, writes the shutdown reason and stops every app container with a 9s force-kill fallback. Wiring a monitoring stop into it would mean deliberately taking away the CPU throttler's input seconds before `process.exit(0)`, to suppress a possible log line. The samplers already self-heal on that path: a tick that finds its container gone calls `stopAppMonitoring` itself.

**6. The `/apps/appmonitorstream` changelog line is in**, as a short release-notes section. The asymmetry is the point and it is now stated plainly: four routes answer `410` with a reason, and the fifth is simply gone and answers `404` — indistinguishable from a typo or a broken node to a caller that does not know it was removed.

## One thing your review made clearer than the body did

Your `stopMonitoringOfApps` finding is what prompted this. "Stop monitoring" is not only a charting concern — an `appsMonitored` entry holds `statsStore`, `lastCpuDecisionAt`, `nanoCpus` and the sampler timer, and `checkApplicationsCpuUSage` reads that series to decide CPU limits. So stopping monitoring for an app removed the throttler's only input and CPU enforcement for that app stopped happening.

Which means the retired control was reachable by an app owner, and an app owner is not a trusted role with respect to the node — the same argument the body already makes about `/apps/appmonitorstream`. That is a better reason for retiring the on/off control than "there is nothing for a caller to turn on or off", and it is the reason the deletion above is the right call rather than a tidy-up.

## Testing

FluxOS unit suite **5,197 passing / 18 pending / 0 failing**. The count is six lower than before this round; those six are the `stopMonitoringOfApps` tests that went with the function. `npx eslint` clean on every changed file.

Your two deferred nits — the `nanoCpus` refresh gap and the unclamped `appMonitor` range — are still deferred, and I agree they are separate changes. Thanks for the correction on the first: you are right that a run-1 inspect that throws skips the whole sample, so the null window only opens if the inspect resolves without `HostConfig`.

Cabecinha84 (commented):
Findings

  1. getContainerStorage still caches a silent zero as status: 'success' — just not the one you fixed

  Round 1 fixed the fs.stat(appsFolderPath) resolution so a failure there fails the whole call. But per-mount failures
  still continue with size = 0, and the total is then cached for 60 s with status: 'success':

  } catch (error) {
    log.warn(`Failed to get size for ${mountType} mount ${source}: ${error.message}`);
    continue;   // contributes 0
  }
  ...
  cache.set(appName, storage);   // status: 'success'

  Old code did the same thing minus the caching, so it isn't a regression — but the cache now makes a transient du/stat
  failure stick for a minute and present as a successful reading of almost no disk. Given the reasoning the author
  already accepted for the shared-device stat ("refusing to answer is the honest failure"), the same argument applies
  one level down. Minor, but consistent to ask for.

  2. latestStats has no in-flight dedup — the description and a test name overclaim it

  The body says "ten viewers cost what one viewer costs" and the test is called "should take one reading however many
  callers ask at once". The test awaits three calls sequentially:

  await appInspector.appStats(req, res);
  await appInspector.appStats(req, { json: sinon.stub() });
  await appInspector.appStats(req, { json: sinon.stub() });

  Genuinely concurrent callers each miss the freshness check and each issue their own dockerContainerStats +
  dockerContainerInspect. Impact is small in practice (viewers poll unsynchronised, and getContainerStorage is now
  cached separately), but the claim isn't what the code does and the test doesn't prove what its name says. Either add a
  shared in-flight promise, or rename the test and soften the sentence in the body. Minor — same claim-vs-code class as
  the three items already corrected in earlier rounds.

  3. The sampler loop itself has no unit coverage, and its only coverage is a manual workflow

  startAppMonitoring's interval body is where the State !== 'running' skip, run % 3 === 1, the seven-day retention
  filter on elapsed, and the statsSampleIntervalMs wiring all live. No unit test enters it — the only assertion is
  expect(appInspector.startAppMonitoring).to.be.a('function'). That matches Codecov's 18 uncovered lines in
  appInspector.js.

  The PR's answer is harness suite 63-cpu-throttling, which does assert the full chain and even asserts the allocation
  comes back — genuinely good. But .github/workflows/integration-harness.yml is workflow_dispatch only, and gh run list
  shows no recorded runs at all. So the claimed 5/5 and 3/3 are unverified by anything I can see.

  This is the gap I'd actually press on: a fake-timer test over one interval tick would cover the four behaviours above
  cheaply, or dispatch the harness on this branch and link the run.

  4. Smaller things

  - startMonitoringOfApps try/catch is per-app, not per-component. The comment says "one app that cannot be monitored
  must not leave the rest of them unthrottled" — but for a v4 app, a component that throws skips every later component
  of that same app. One-line fix (move the try inside the compose loop).
  - minimumInstancesV8Block: 2000000 in test-infra/config/shared.js lowers a minimum for every harness suite, not
  just 63. Harness-only, comment explains why, but it could quietly mask a regression in the v8 minimum-instances rule
  elsewhere.
  - appDockerPause / appDockerUnpause are now dead code in dockerService.js — nothing calls them. Worth removing with
  the controls they backed.
  - The "410" is a body field, not an HTTP status. messageHelper.createErrorMessage(..., 'Deprecated', 410) goes out
  over res.json as HTTP 200. That's the FluxOS convention and consistent with the rest of the codebase, but the body and
  the table read as if the wire status were 410. Doc nit.
  - alwaysRespond on appstart/appstop/apprestart/appremove/redeploy/redeploycomponent is scope beyond monitoring, but
  it's correct, self-contained, and tested. No objection.

MorningLightMountain713:
Thanks for round 3 — all of it is on the current head (`190076e61`), and there's a set of independent fixes since your pass worth flagging too.

**Your round-3 findings**

1. **`getContainerStorage` caching a silent zero** — fixed in `14eef3698`. A per-mount `du`/`stat` failure now fails the whole reading instead of contributing 0 and caching a near-empty total as `success`, matching the shared-device-stat argument you pointed at.
2. **`latestStats` dedup** — rather than soften the claim, the dedup is now real (`fbfe88a7e`): concurrent callers share one in-flight reading, so "ten viewers cost what one viewer costs" is what the code does. The test asserts the shared call, not a log line (`2f5d6d029`).
3. **Sampler loop coverage** — the interval body now has direct unit coverage (`f7a134d05`): the `State !== 'running'` skip, the `run % 3` cadence, the seven-day retention filter and the interval wiring are asserted by a fake-timer test. Suite 63 still asserts the full chain end-to-end and passes in the gate.
4. **Smaller** — per-app→per-component try/catch fixed (`fbf4eeb19`); dead `appDockerPause`/`appDockerUnpause` removed (`6d2c7464f`); the 410 body-vs-wire distinction is now spelled out in the API table (`e3523ad8a`); `minimumInstancesV8Block` and `alwaysRespond` left as-is per your note.

**Independent fixes since round 3** — an independent pass over the diff plus follow-on work; the first two are defects this branch introduced earlier and are worth your eyes:

- **`clearAppVolumeData` deleted as root but enumerated as the unprivileged user** (`190076e61`). A component whose image hardens its data dir (postgres does `chmod 700 $PGDATA`) makes the `readdir` fail `EACCES`; the reconciler then holds `dataDesired` at `clear` and retries every 5s forever, so the component never starts — a permanent availability bug for exactly the database images `g:` mode serves. Now a single root `find … -exec rm -rf`, nothing enumerated host-side.
- **`clearAppVolumeData` never inspected `runCommand`'s `.error`** (`6259a6460`). `runCommand` resolves `{ error }` rather than rejecting, so a failed wipe fell through to `log.info('Deleted data')`, and the reconciler's catch that guards against starting on un-wiped data was unreachable. Each `rm`'s error is now inspected and a failure rejects.
- **A paused container read as healthy** (`dfddea94e`). Docker reports `Running: true` for a paused container; `dockerActual` reported it up, so the reconciler no-op'd, the sampler skipped it, and it disagreed with `appStartupManager` every boot. It now normalises the state (stop → exited).
- **`deprecatedPause/Unpause` could hang the socket** (`7f5d1b4b5`). `appname.split('_')` ran on `req.query.appname` before auth, and `?appname=a&appname=b` arrives as an array with no `.split`; the rejection was dropped and the socket stayed open. Order is now validate → authenticate → act. (This is the `split` shape round 1 raised and closed as not reachable — it was.)
- **Event-bus discipline** (`0b57a9110`) plus harness/timing hardening: the `g:` election loop now emits facts-as-events and cadence-as-a-counter, so suites wait on the loop's own passes rather than a production-derived clock.
- **A wrong in-band handler count and middleware-order rationale** are corrected (`8863e289e`).

**Dependency:** the FDM side this relies on — keeping an operator-stopped `g:` master named as the load-balancer primary — is now `RunOnFlux/flux-domain-manager#154`, tested end-to-end (operator-stop, and FDM restart while stopped). The merge order in the description is unchanged: FDM deployed first, then this to the fleet.

Cabecinha84 (commented):
Findings

  1. Undocumented wire change: blkio op casing and entry count (doc-only, but it's a public API)

  expandSample always emits exactly two entries, op: 'read' and op: 'write', lowercase. Docker on cgroup v1 reports
  Read/Write, and reports one entry per device in the stack. The extract matches case-insensitively and sums — correct,
  and there's a good test for it (matches docker's capitalised blkio ops) — but the output is normalised. The
  breaking-change table lists blkio_stats.io_service_bytes_recursive as kept, unqualified. A third-party scraper
  filtering op === 'Read', or reading a per-device entry, now gets nothing rather than a narrowed value. Our UI
  lowercases, so we're unaffected. One line in the table would close it.

  2. A carried-forward disk reading is indistinguishable from a fresh one appInspector.js:632

  if (statsNow.disk_stats?.status && statsNow.disk_stats.status !== 'success' && lastKnownDisk) {
    statsNow.disk_stats = lastKnownDisk;
  }

  lastKnownDisk carries its own status: 'success'. The reasoning for not storing null is right (dashboards read used ||
  0, so absent charts as a drop to the floor). But under a persistent failure — a volume unmounted, du permanently
  refused — the series shows a flat, stale line labelled success indefinitely, with nothing in the sample saying it was
  carried. { ...lastKnownDisk, status: 'stale' }, or a carried: true flag, would keep the number and stop the label
  lying. Chart-only; nothing enforcement-related reads disk_stats (I checked — checkStorageSpaceForApps uses
  dockerGetUsage().Containers[].SizeRootFs).

  3. A paused container docker stop refuses has no operator escape hatch appReconciler.js:~780

  The paused branch stops the container and, on failure, scheduleRetry(MANAGED_RETRY_MS) — forever. appDockerUnpause is
  deleted, so if any Docker version or edge case declines stop on a paused container, FluxOS has no primitive left to
  release it. In practice modern dockerd handles it (SIGTERM, then SIGKILL at timeout), and the only containers that can
  be paused are pre-rollout ones, so this is a tail risk rather than a defect. Worth naming the manual remedy (docker
  unpause) in that log.error so an operator reading it knows what to do.

  4. statsSampleIntervalMs doesn't get the protection its own comment claims appInspector.js:~657

  The comment says "A missing key coerces to NaN and anything below 1 is treated as ONE MILLISECOND — that has happened
  to this exact key before", and then guards with ?? 60 * 1000. ?? covers the missing-key incident it names, and nothing
  else: a key present as 0, "60000" or a typo'd string goes straight into setInterval. Given the incident history,
  Number.isFinite(x) && x >= 1000 ? x : 60000 is cheap and actually delivers what the comment promises.

  5. The freshness window does less than the body says (doc nit)

  statsFreshnessMs is hardcoded at 5 s while the sampler is config-driven at 60 s. A single viewer polling
  /apps/appstats every 5 s misses the window on essentially every poll, so it still takes a live dockerContainerStats +
  dockerContainerInspect each time. What actually delivers "ten viewers cost what one viewer costs" is the new
  statsInFlight single-flight map (real, and well tested — four cases including the rejected-reading one), plus the 60 s
  containerStorageCache that kills the du walk. The body attributes it to the freshness window.

  6. getContainerStorage hard-fails for mounts that never needed the apps folder (nit)

  fs.stat(appConstants.appsFolderPath) resolving once inside the mounts branch was the right fix for round 1. But a
  container whose mounts all live outside appsFolder — a docker-managed anonymous volume under the data root, which the
  code comment itself identifies as reachable — still can't be sized at all if that stat fails, where before it would
  just have been walked. Extremely narrow (the apps folder exists whenever an app is installed).

  ---

  Context worth knowing, not asks

  - A container restart discards the retained week. startAppMonitoring replaces the whole appsMonitored[name] object,
    and the reconciler and appStart call it on every start/restart/recreate. Pre-existing behaviour, unchanged — but now
    that "a week of samples" is the headline, it's worth knowing that any restart resets the chart history to empty.
  - Memory strictly improves. Old store: ~3360 full readings/week/component. New: 10080 extracts. Even at a pessimistic
    600 B/sample the new store is several times smaller than the old one, so the finer series is genuinely free.
  - The body says "38 commits"; the branch carries 51 against the current merge-base. Cosmetic.
  - Codecov's 94% patch figure is 34 commits behind head — stale, ignore it either way.

  ---

  The part I'd call out as good

  tests/unit/routes.test.js — the test that reads routes.js, maps each const x = require(...), and asserts every x.fn(
  the file names is actually exported. That is the real fix for the class of bug that motivated the whole PR (three
  monitoring endpoints shipped wired to names their service no longer exported, answering 500 on the first user request,
  with a green suite). It's resolved statically rather than by invoking ~200 production handlers, which is the right
  call. That test will keep paying out long after this PR.

  Similarly, the eight characterisation tests landed first, as their own commit, against unmodified throttler code, with
  two mutations to prove they were load-bearing, and all eight passing afterwards with identical assertions — that's
  the correct way to rework code that physically re-allocates CPU on every node.

  ---

MorningLightMountain713:
Thanks for round 4 — all six are addressed on the current head (`e41cef725`); four in code, two in the PR body, which is also brought up to date with the commits it was missing.

1. **blkio wire change** — the breaking-change table now states the normalisation: exactly two summed entries, ops lowercased; cgroup-v1's per-device, capitalised entries are gone. Updating the table also corrected its networks rows — the extract keeps every interface's byte counters, and the table now says so.
2. **Carried disk reading** — the carried copy is stored as `{ ...lastKnownDisk, status: 'stale' }`: the figure still charts (the dashboard blanks only on `error`), the label no longer claims freshness. This also keeps the live path honest when a poll lands inside the freshness window right after a sampler tick.
3. **Paused stop with no escape hatch** — the error log now names the remedy with the real container name: `manual remedy on the node: docker unpause <name>`. Behaviour unchanged: a transient failure self-heals through the retry; the log line serves the persistent case, which is exactly where a by-hand unpause works.
4. **`statsSampleIntervalMs`** — the guard refuses any value that is not a finite number ≥ 1000, falls back to 60s, and warns naming the rejected value. Pinned by tests for the missing key, a present-but-wrong value, and a configured non-default value actually being used.
5. **Freshness window** — the body now credits the three mechanisms together: a landed reading serves every caller for the next five seconds, callers arriving mid-read share the in-flight one, and the 60s storage cache bounds what each miss pays for the `du` walk. One clarification in return: the window is load-bearing for the ten-viewer claim — once a reading lands it serves all staggered viewers until it ages out; the single-flight map only covers arrivals during the read itself. Where the window does nothing is a viewer who is alone, which is your arithmetic — both are true.
6. **`getContainerStorage`** — the shared device is resolved on first need, per mount. A container whose mounts never sit under the apps folder is walked even when the folder cannot be statted; a mount under the apps folder that cannot be classified still fails the whole reading — the honest answer, and the existing test pins it.

On the context notes: the restart-resets-the-series point is now stated in the body, and the commit count is corrected.

Since round 4, a further blind pass over the whole diff surfaced two real defects, both fixed in `0e5a8ed1b` and worth your eyes:

- **`clearAppVolumeData` classified "nothing to clear" by matching `find`'s English error text.** `find` renders strerror in the node's locale and sudo keeps `LANG`/`LC_*`, so on a non-English node a missing appdata directory read as a failed wipe and the reconciler retried it every 5s forever. The classification is now a root `test -d` after the wipe — exit status, no prose — and the post-wipe order means an ambiguous race falls toward a retry, never toward falsely-clean.
- **A paused container with an unavailable volume and a pending stop was never released.** The volume-unavailable branch honoured a pending stop only for a running container and returns before the paused normalisation; paused reports not-running, so the stop was skipped and the container stayed frozen. The branch now honours the stop for running or paused.

The same pass drove a test-hardening commit (`e41cef725`): the mongo projection and each source of the `/apps/heldcomponents` union, the live-stats memory, the sampler's configured cadence, the export surfaces and the deprecation codes all now have tests that go red when the code they pin breaks — each proven by mutation. `/apps/appmonitorstream` also now answers the same in-band 410 as the other retired routes, naming the polling replacements, instead of a bare 404; the body's table reflects it.

Cabecinha84 (approved):
ack

PR #1779, opened 2026-08-07
Restore: acquire before destroying, and stop one dead volume from silencing a node

MorningLightMountain713:
## Overview

On 2026-08-04 a customer's Palworld world was destroyed by a restore. This PR fixes the path that destroyed it — and four further ways data could still be lost, which only became visible once these tests were run against real nodes for the first time.

**What went wrong.** One node had a single app whose backing volume had been deleted. That one unrepairable volume jammed a node-wide gate in the syncthing monitor, so *no* syncthing folder was ever configured for *any* app on that node, and no `g:` primary election ever ran there. A customer's Palworld instance was stranded on it — never synced, holding 373 bytes. Someone backed that instance up and restored it the next day. The restore deleted appdata before fetching anything, then fanned out a forced hard redeploy to every other instance, which `rm -rf`'d the 35 GB volume on the one node that actually held the world.

**What this delivers:**

- **5 product fixes** — the node-wide gate, the backup gate, the restore rewrite, the mount-safety dedup, and a syncthing restart that could take the daemon down permanently.
- **6 new harness suites** (34, 89–93), covering both ends of the incident for the first time — including one that boots a real syncthing daemon.
- **Three further review rounds serviced (2026-08-11, 2026-08-24 ×2): 17 further commits** — restore/monitor safety fixes in the same incident family, the HTTP status carried structurally instead of regexed out of error prose, streaming-runner hardenings, a harness supply-chain fix, an atomic claim primitive, and a dead `g:` election guard brought back to life. Details in "Review rounds two to four" below.
- **Verification:** 5,515 unit tests passing, 18 pending, 0 failing on the current head (`d2a79fabe`), and the full harness gate green on this branch — 85 suites.

Four of the five product fixes were found *by running* the new suites, not by writing them.

**First of four in the stack: `development ← this ← #1780 ← #1781 ← #1782`.** Everything that was below has merged; this targets `development` directly.

---

## What this changes

**The gate.** `syncthingAppsCore` abandoned the whole cycle on any unmounted folder, and set `syncthingInitializedSuccessfully` only after that point — so `syncthingAppsFirstRun` latched forever, and that flag gates the `g:` primary election node-wide. One deleted volume stopped every masterSlave app on the node from ever electing. An unsafe mount is now an app-level fault: that app is demoted and held out of the pass, and the rest of the pass proceeds.

[...]

PR #1778, opened 2026-08-05
App volume file operations: move, copy, compress, extract — and every write to a tenant volume moves into a container

MorningLightMountain713:
## What this delivers

- **Move, copy, compress and extract on an app's volume.** Reorganising a media library no longer means downloading it and re-uploading it; installing a 500-file mod pack is one archive upload and an extract, not 500 uploads. Four new endpoints, each answering `202` with a job to poll.
- **Every operation that writes to an app's volume now runs in a throwaway container** with that volume as the only thing mounted — including the ones that already existed (`createfolder`, `renameobject`, `removeobject`) and, in this PR, upload.
- **Nothing is ever half-written at its destination.** Work goes to a staging directory and is swapped into place in one atomic step, so a failed, cancelled or power-cut operation leaves the caller's data exactly as it was - not because a recovery path puts it back, but because there is no in-between state to be caught in.
- **A remotely-triggerable root compromise is closed.** The boot sweep passed the contents of a tenant-writable file to a root `mv` as its destination — `sudo mv -T <volume>/.flux-old-pwn /etc/cron.d/pwn` on the next restart. Restarts are routine.
- **An app owner can no longer take a node down by opening their own file manager.** Folder measurement followed symlinks with unbounded recursion: `ln -s / escape` measured the host, `ln -s .. loop` never finished. It ran as the FluxOS process, before any container existed, from two entry points.
- **Downloads stop leaving world-writable copies.** Both download endpoints ran `chmod 777` on the file before sending it. FluxOS is root, so it granted nothing — it just left every downloaded file permanently world-writable and ran as root against a path the application can swap.
- **A node that cannot reach ghcr.io gets the executor image from a peer.** The fleet already holds the image; asking a peer costs the network nothing it does not already have, and it keeps the file browser working on a node behind a restrictive firewall or during a registry outage.

## Why the work runs in a container

The operations act on paths inside a directory a tenant fully controls, while FluxOS holds `/var/run/docker.sock`. Getting a path check wrong is not a containment bug, it is a host compromise. So containment comes from the container having **nowhere to escape to** rather than from a sequence of checks being correct:

```
the app's volume and nothing else   a path that escapes it lands nowhere
ReadonlyRootfs                      writes outside the volume fail
NetworkMode none                    a hostile archive cannot phone home
no-new-privileges + nosuid mount    a setuid payload is inert twice over
CapDrop ALL + CHOWN,FOWNER,DAC_OVERRIDE
pids and memory limits              bound a runaway archive
```

[...]

PR #1780, opened 2026-08-07
Reconciler: the only actuator for every operator command, and a ladder that paces faults not operators

MorningLightMountain713:
## Overview

Two things, both about who decides a container's run state.

**The crash-recovery ladder paces faults, not every stop.** It used to pace all of them, so an operator restarting their own app walked the same escalating waits as a container that could not stay up. A customer restarting their Palworld server six times in twenty-one minutes reached the fifteen-minute rung, which from the outside is indistinguishable from the app refusing to come back. That is what the support ticket was about.

**Every operator command is desired state the reconciler converges to.** `appStart`, `appStop`, `appRestart` and `appKill` record the operator's intent and let the single actuator converge on it; no route in `appController` drives a container. Seventeen `appDockerStart`/`Stop`/`Restart`/`Kill` calls remain outside the reconciler — install, uninstall, redeploy and backup/restore, mount recovery, and the two-hourly sweep that stops containers FluxOS does not own — and those paths are not what this changes.

**First of three in the stack: `development ← this ← #1781 ← #1782`.** #1779 below has merged (`63327e97f`), so this targets `development` directly. #1781 and #1782 are based on this branch and merge in stack order.

## Why the exit code cannot decide this

The exit code was already persisted — `recordExit` writes `lastExitCode` beside `lastDiedAt` — and nothing read it. It is now an input, but only in the direction it is sound.

**For a container that went down on its own: non-zero, or `OOMKilled`, proves a fault, and zero proves nothing.** An image whose entrypoint is a wrapper script ending in `exit 0` reports a clean stop for a segfault. Palworld's official image does exactly this:

```sh
"${STARTCOMMAND[@]}"     # the game — segfaults here, $? = 139
LogAction "Ending Server"
exit 0                   # unconditional; the game's status is discarded
```

Its runtime state on a live node records `lastExitCode: 0` for a death confirmed as a SIGSEGV. No init we wrap around the container recovers this: the loss happens inside the image's own script, a level below anything we can inject, so `--init`/tini would faithfully propagate the zero that was already laundered.

Pacing on the code alone would therefore leave that class of container restarting without limit. That is what the ceiling is for, and why it consults nothing but the rate.

The qualifier at the top of this section is load-bearing, and the case it excludes is not yet handled. FluxOS stops containers deliberately — an election moving a `g:` primary, a backup taking the app down — and those go out as SIGTERM, so an image that does not trap it exits 143. This lineage reads that as a fault and books a rung, which paces a healthy app for the node's own tidy shutdown. It is an improvement on what it replaces, which paced every restart whatever the cause, but it is not the whole rule: the exit code is an unreliable narrator in both directions, and only one of them is addressed here. Teaching the ladder to recognise the node's own drains needs desired state to decide it rather than the exit code, and that is a follow-up PR rather than a widening of this one.

## The ceiling, and what it reaches

Five automatic restarts inside five minutes is itself evidence of a fault, whatever Docker reported. It disposes into the same ladder rather than into a failed state a human has to clear — nobody is watching a customer's node at 3am.

It is judged before the restart is recorded, against five entries spanning five gaps, so its reach is `window / count`. Measured against the real module:

```
every  59s   paced
every  60s   paced
every  61s   never paced, however long it continues
```

A container exiting in milliseconds trips it in about a second. Anything slower than one restart a minute is deliberately left alone.

## The ladder is measured against running, not against waiting

`restartWaitMs` clears the ladder once a component has genuinely run for `STABLE_RUN_MS`, and measures that from its last rung. A component holding rungs therefore earns one for every restart — otherwise the gap being measured is the wait the component just served, and any rung longer than ten minutes clears the ladder every time it fires.

The exit code decides what puts a component on the ladder; it does not decide whether it stays. `restartWaitMs` takes no `crashed` argument for that reason — whether to wait depends only on whether rungs stand.

Measured against the real module, driving the reconciler's own loop order for two hours with a container dying two seconds after every start:

| container | starts in 2h | rungs reached |
|---|---|---|
| reports exit 0 for a segfault | 12 | 30s, 5m, 15m, 30m cap |
| exits non-zero | 7 | 30s, 5m, 15m, 30m cap |
| healthy, runs for an hour | 2 | never paced |
| operator presses restart | — | never paced, ladder cleared |

An operator restart is recognised by the path it takes, not by its exit code: `appStart` and `appRestart` clear the operator lock, and `setOperatorStopped(id, false)` drops the rungs with it. Recognising it by exit code instead hands the same exemption to a laundered segfault, which also exits 0.

## Operator commands are desired state

`appStart` probed docker for "is this container running" and used the answer as a proxy for "should this node be running it". Those diverge both ways: a masterSlave primary whose container is stopped was refused a start though the election names it the writer, and a standby whose container happened to be up was started. It classified `g:` with `containerData.includes('g:')`, which `mountParser`'s own comment calls wrong — `'/data|g:/db'` matches while the flag sits in an invalid non-primary position.

The election owns that decision and the reconciler consults it on every pass, so the handler clears the lock and lets it decide. A component the election is holding is reported as pending, naming the election, rather than as skipped.

`appRestart` raises a durable restart generation; the reconciler bounces a running container once the generation passes the one it last actuated, and a start of a stopped container satisfies the same request. `appKill` records a durable force mode which the reconciler honours with `appDockerKill` where it stops the container.

Every one of the four reports what happened rather than what was asked for: `Application X stopped`, or `will be stopped: docker is not reachable`.

A command that addresses nothing is refused rather than reported done. The components are resolved before anything is written, and an app that is installed but whose parts this node cannot work out — a spec that will not decrypt, falling back to a container listing that is empty or that failed outright — yields none. Acting on that empty list wrote no intent and settled vacuously against nothing, so all four answered `Application X stopped` having touched nothing and recorded nothing for a later pass to pick up. They now say the app was not changed and why, worded for the operator: HomeUI shows the message verbatim.

## Monitoring follows the container

`startAppMonitoring` resets `statsStore`, and `appStart` called it on every request — so asking an app to start discarded the series the charts read. `appStop` stopped the sampler before knowing whether the stop had happened, so an unreachable docker left a running container unmonitored.

The reconciler owns both ends. It stops the sampler on each of the four paths where it stops a container, and `ensureAppMonitoring` puts one back for a running container that has none without resetting a live one.

## The kill endpoint

`appKill` was implemented, exported and covered by unit tests calling it directly, and no route registered it — on this lineage or on v9. Every route is a string literal, there is no dispatch table, and `executeAppGlobalCommand` builds a URL against the same route table, so a peer could not reach it either. `routeWiring.test.js` reads the real route table, so its absence now fails a test.

`appKill` and `appRemove` both ask for `appownerorfluxteam` rather than `appownerabove`, which takes the node operator out of both. Hosting an app is not owning it: an operator who can remove one can script that against every install and keep a customer's app off their node indefinitely, which the customer experiences as an app that will not stay deployed and cannot diagnose.

There is no operator need on the other side of that. An app cannot exceed what was bought — `Memory`, `MemorySwap`, `NanoCpus` and `StorageOpt` are set on the container from the spec, so an app inside its allocation is spending cycles the operator sold, and an app outside it is a containment defect to be fixed in the limits rather than papered over on one node. What an operator does keep is the control that is honest about its own cost: stop FluxOS, or drop the node, both of which give up the payment along with the obligation.

That reasoning does not stop at these two, and the rest is already decided. `appStart`, `appStop` and `appRestart` still ask for `appownerabove`, and a follow-up PR takes the remaining app control away from the node operator entirely, with the frontend controls narrowed alongside it — the same way this one lands with #150. It is separate only so that each gate ships with the control that exposes it.

`appownerorfluxteam` admits exactly `{owner, fluxTeam, fluxSupport}`, which is the same set the vetted-app branch below it admitted. That branch existed to keep the operator away from vetted apps, so with the operator out of the gate above it can no longer refuse anyone who reaches it. It goes, and with it two database reads and a vetted lookup on every uninstall.

The same reasoning closes the other end. `installapplocally` and `testappinstall` asked for `adminandfluxteam` on their by-name branch, which resolves an app from the marketplace, the global registry or a permanent message — so an operator reaching it chose *which* customer's app ran on hardware they control, and for a `g:`/`r:` app the new instance syncs that customer's data down to it. Both move to `fluxteam`. Leaving install open while closing remove would have left the operator able to place an app on their node and unable to take it off.

The temporary-message branch above them is deliberately untouched and stays open to any logged-in user: that is how an app is tested before it is registered, it is addressed by hash rather than by name, and it expires on its own via `temporaryAppAllowance`.

This is user-visible on a reachable route, so the frontend control that offered it is narrowed to fluxteam in RunOnFlux/fluxos-frontend#150. **The two land together.** `local.vue` renders Uninstall to `admin`, so this merging alone leaves a node operator pressing a button that now refuses them. The same PR fixes the panel refresh behind finding 5: `AppControl.vue` read `response.data.status` once and refetched once, so an accepted command showed a green toast beside a status that had not moved yet. It now refreshes on a spreading schedule (0/2/5/10/20s).

## Config fallbacks

`RESTART_BURST_WINDOW_MS` falls back to the value config ships. Every `config.fluxapps.x ?? literal` in the service layer is a second copy of a shipped value — forty keys, none of them a default the module owns — and `tests/unit/configFallbacks.test.js` compares each against `ZelBack/config/default.js`. An expression it cannot evaluate fails rather than being skipped.

`config.get()` removes the second copy entirely and throws by name on a missing key. That is 40 call sites here and 108 on the v9 lineage, so it is its own change.

## The boot-lock harness commits

Three commits touch only `test-infra/runner/framework/boot-lock.js` and its call site, and are named here because they are otherwise unexplained in a reconciler PR: *the boot lock says how long it queued and how long it held*, *a boot's duration is unreadable without the shape of the fleet*, and *the boot lock admits a bounded number of boots, not one*. They are named by subject rather than by hash: this branch is rebased whenever the one below it moves, and every hash in it changes when that happens.

Fleet boot is serialised host-wide by a semaphore, and nothing recorded what that cost. The first two make acquire report how long a build queued and how many were ahead of it, release report how long it held, and add the shape of the fleet being built — without which a hold duration mixes one-node and ten-node builds indistinguishably.

The third acts on what the first two measured: the gate was **89% boot-lock-held wall clock**, 4,439s of 4,978s across 130 boots, so its duration was very nearly the serial sum of the boots it serialised. The lock now admits `E2E_BOOT_LOCK_WIDTH` boots, default 2. Arrival order is untouched — the queue still drains in the order it formed, which keeps starvation structural rather than statistical.

Measured across two gates of the same 82 suites at `MAXN=6`, the same 130 boots and a near-identical fleet-size distribution:

| | width 1 | width 2 |
|---|---|---|
| gate wall clock | 4,978s | **4,192s** |
| queue wait, p50 | 87.8s | **24.1s** |
| queue wait, total | 11,697s | **3,747s** |
| lock held, p50 | 32.6s | 40.2s |
| lock held, total | 4,439s | 5,746s |

Boots contend and each costs more, but they overlap, and the overlap wins by 786 seconds. `E2E_BOOT_LOCK_WIDTH=1` restores the old behaviour without reverting the commit.

All three emit TAP comments (`# boot-lock ...`), so none can alter a result: `run-all.sh` tallies with `grep -c "^ok "`, anchored to line start.

## An address change no longer stops at the first composed app

`adjustExternalIP` restarted the apps that survive an address change by calling `appController.appDockerRestart(app.name)`. That resolves one container by name and a composed app has none under its bare app name — its containers are `<component>_<app>` — so the first one threw a `TypeError` that nothing caught, taking with it every app behind it in the loop, the `fluxipchanged` broadcast, the confirmation transaction and the geolocation update. One log line was the trace.

The surviving apps are now handed to `appReconciler.requestRestartOf` as durable per-component restart requests, through a seam (`setOnAddressChanged`, wired in `serviceManager`) rather than a require — `appUninstaller` requires `fluxNetworkHelper` and `appReconciler` requires `appUninstaller`, so reaching upward from the network layer closes a cycle, and twenty-odd app-layer modules import it. `componentIdsOf` is shared with `enqueueAll` so the rule for what components an app has, including the docker fallback for specs that will not decrypt, has one home.

## An address change no longer uninstalls the node's own apps

Before restarting an app, `adjustExternalIP` asks whether it is already running at the address being moved to, and force-removes it if so — on the sound principle that one instance per IP is what the host port mapping allows, so an instance already there means the ports are gone. It treated any answer as another node's. The node's own registration is one of those answers: it stores its own running-app row locally, at the address benchmark reports, so the row found at that address is usually itself. The app was removed — forced, and broadcast to the network — for being exactly where it belonged.

The address is still compared at IP granularity — that is the question being asked, and a UPnP sibling on another port holds the ports just as surely. Own-ness is the full socket address, which is what separates the two, using the same `socketAddressesMatch` test the collision check in that file already uses.

## A stop is honoured while a data volume is unavailable

The reconciler defers every actuation for a component whose data volume will not mount — a start there writes to the host filesystem instead of the volume. A stop is the exception, because it takes nothing from the app dir, and leaving a container running over a missing volume is the state the mount-safety hold exists to end.

That exception covers the operator's stop as well as the controller's. The operator's outranks the controller's everywhere else in a pass, and an unmountable volume is the state support reaches for a stop *in*, so a stop that waits for the mount is a stop that never arrives when it is wanted. The force flag is carried through it, so an `appkill` there is a kill rather than a graceful stop.

## An operator command addresses what the app is made of

The components a command acts on come from `appReconciler.componentIdsOf`, which decrypts the spec to enumerate them and falls back to the running containers when it cannot.

Reading `compose` off a stored spec answers neither case reliably. An enterprise app keeps its component names inside an encrypted blob, so its stored compose is empty — a whole-app command derived from it addresses nothing, and every one of zero components reaches its desired state, so it reports success. A component name taken verbatim addresses nothing either when it is not one of the app's, and still writes a durable operator lock under it that nothing clears.

`componentIdsOf` returns one spelling whichever source it used. Docker holds the namespaced name (`flux<component>_<app>`) where a decrypted compose gives the bare one, so the list carried either depending on whether the spec read. Every consumer until now canonicalised on ingest and so could not tell them apart; the membership check above compares the list against a name an operator typed, and refused every component-level command against an app whose spec would not decrypt. Stripped at the source, so one spelling reaches all three consumers.

`containersReachedStopped` distinguishes a container that is absent from one that is stopped. `dockerActual` reports `exists` separately from `running`, and reading the pair as one answered "stopped" for a container that was never there — the answer `appStart` already gave correctly.

## The restart generation is raised by the database

`requestRestart` issues an atomic increment rather than reading the current generation and writing one more than it. `getState` answers null both for "no record yet" and "the read failed", and treating the second as zero writes a generation below the one already actuated — which the reconciler reads as nothing pending, so the restart is reported and never happens. An increment has no such answer to misread, creates the field at 1 when there is no record, and counts two concurrent requests as two rather than losing one.

The upsert retry that protects a first write from a concurrent one is shared with `setFields` rather than copied, since without it the loser of that race is dropped silently.

## Changes

| file | change |
|---|---|
| `appController.js` | all four operator handlers record intent and probe the outcome; `app:operatorIntent` published from inside the reconciler's slot; `appDockerRestart` deleted |
| `appReconciler.js` | honours the restart generation and the force mode; owns monitoring on its stop and steady-state paths; `requestRestartOf`, and `componentIdsOf` shared with `enqueueAll` |
| `fluxNetworkHelper.js` | `setOnAddressChanged`; the surviving apps are handed over once; the duplicate check excludes the node's own registration |
| `serviceManager.js` | wires that seam to `appReconciler.requestRestartOf` |
| `appUninstaller.js` | `appownerorfluxteam` on the uninstall route; the vetted branch it made redundant removed |
| `appsRuntimeState.js` | `restartGeneration` / `operatorStopForce`; a component holding rungs earns one per restart; twin-merge carries all three |
| `appInspector.js` | `ensureAppMonitoring`; a cleared interval no longer looks live |
| `routes.js` | `/apps/appkill/:appname?` |
| `verificationHelper*.js` | `appownerorfluxteam` |
| `ZelBack/config/default.js` | `restartBurstCount: 5`, `restartBurstWindowMs: 300000` |

`globalState.stoppingContainers` stays. Its remaining consumers are the uninstaller, redeploy, backup/restore and mount recovery, none of which this touches. The network-change restart is no longer one of them.

## Testing

Unit suite **5,635 passing, 0 failing** (18 pending) on the current head. Both fixtures up — without mongo the run skips 682 tests and still exits 0, so a bare pass is not a result.

Every guarantee added here was mutated to prove the test would notice: the force mode ignored, every stop forced, the restart request never actuated, the bounce never recorded, a start not satisfying a pending request, a pending outcome reported as success, the handler driving docker again, the intent event published outside the slot, the route removed, the kill privilege widened, the uninstall privilege widened back to the one that admits the node operator, the component identifier left namespaced by one of its two sources, an empty component list reporting success, the ladder cleared by its own wait, and a fallback diverging from config.

Twenty-nine mutations run, twenty-eight caught. The one that was not is worth stating: it reverted a line measuring stability from the last restart in either array rather than from the last rung. Nothing failed, because a component holding rungs earns one for every restart — so the two are always the same value and the line could not change an outcome. It was removed rather than shipped.

One test was removed for being unable to fail, and one rewritten: it asserted that an operator restart is not paced but expressed that through the exit code, which is the exemption a laundered segfault also receives.

### On a fleet

The unit tests stub docker, so they prove a handler writes the right intent and nothing further. What a customer sees is a container's run state and a response string, and until now nothing drove either. Six tests were added to four existing suites, and one new suite covers the address-change path:

| suite | test |
|---|---|
| `32-…-restart-clears-lock` | apprestart bounces a container that is **already running**, and bounces it once. The suite's existing test restarts one that was down, which is a start — either implementation satisfies it, and the restart generation only decides anything when the container is up. |
| `32-…-operator-stop-durable` | appkill stops it, asserted on the `forced` flag — "the container stopped" passes against a graceful stop too. And the node operator is refused a kill while still allowed a stop, driven from the node's own identity, both halves. |
| `52-masterslave-…` | appstart on an instance the election is not electing reports the election rather than claiming a start. Runs first in that suite, before anything there stops an instance — after that, a non-running holder could be one an operator stopped, which is a different answer reaching the same assertion. |
| `54-…-burst-ceiling` | the ladder is not cleared by the wait it just served, reaching the rung that outlasts a stable run. |
| `35-…-masterslave-election` | a backup's stop costs the elected primary no rung: the hold is released and it converges back to running inside a cycle. The same assertion fails on the branch below, where every restart earns a rung and the component is on the five-minute one by then — so it discriminates against the behaviour this PR changes, proven by running it on both. |
| `56-address-change-restarts-apps` | **new.** Three nodes, one COMPOSED app. Both components restart on their own identifiers after the node's address moves, and a peer accepts the `fluxipchanged` broadcast — which is what distinguishes a handler that ran to the end from one that stopped at the first app. |

Suite 54 previously ran against production's ladder, so the only rung reachable inside a sane timeout was the first — everything past it, including the rung this PR's defect lived at, needed a twenty-one minute wait and so was never driven. Its ladder and stable-run window are now compressed together, preserving the relationship that matters (a rung outlasting the window that clears the ladder: 30s against 20s here, 15m against 10m in production) and asserting that relationship holds, because compressing one without the other leaves the suite green having stopped testing what it exists for.

The backoff actuation now carries the rung. `waitMs` is what *remains* of a rung and the worker re-enqueues during a wait, so one rung reports several times with a falling number and two backoffs cannot be compared — which is precisely the question a ladder that resets itself turns on. It reaches the log line too, so support can read how far a component has escalated rather than only how long is left.

Suite 56 changes a node's address for real rather than simulating one: the new address goes onto its interface and the old one comes off, and benchmark's public-IP probe reports the move while the deterministic list still says where it was. Its peers then fail to reach it at the old address because it genuinely is not there, which is what makes it ask benchmark whether it moved — a node that is still reachable never asks. The list being untouched is what lets those same peers still recognise it as the sender of the broadcast that follows.

Simulating either half breaks the other, which is why it is done properly: blocking the node's packets makes a peer's answer arrive after the asking node has given up, so it reads as "I could not ask" rather than "you are unreachable" and never reaches benchmark; hiding the node from peers' lists makes them answer promptly and then reject its broadcast as coming from a stranger.

Two harness assumptions had to go with it, both of them things an address change necessarily violates. The runner reached a node at a fixed address, so a node that moved went blind to its own client — `nodeClient.followTo` re-points it and reconnects. And the daemon stub identified a node by the address its requests arrive from, so a node that moved lost its identity, read its own address as the not-found fallback, and skipped its availability check on every cycle; it is now recognised at its reported address too.

Run against images rebuilt from the tree under test and verified from inside the image, not from build output. Suite 56 on the current head: 2 tests, 0 failures, twice consecutively at 3m35s and 2m32s. Suite 35: 4 tests, 0 failures.

### DIFF
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -144,6 +144,8 @@ module.exports = {
     // count. A container that fails to START never ran, so it is never a fault
     // and never walks the ladder directly - it reaches it only by filling this
     // window, and a window narrower than that retries forever.
+    // Counted as restarts ALREADY RECORDED, so at 5 the sixth restart is the one
+    // that earns a rung and the seventh is the first one held back.
     restartBurstCount: 5,
     restartBurstWindowMs: 300000,
     // How long a finished operation stays readable at /apps/operations/:jobId,
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -132,11 +132,20 @@ module.exports = {
     // and the run length that counts as stable (resets the ladder)
     crashBackoffDelaysMs: [0, 30000, 300000, 900000, 1800000],
     crashBackoffStableRunMs: 600000,
-    // cause-blind backstop for images whose entrypoint discards the payload's
-    // exit status: this many automatic restarts inside the window is treated as
-    // a crash and enters the ladder above
+    // Backstop for images whose entrypoint discards the payload's exit status.
+    // A clean exit proves nothing, so for those images restart RATE is the only
+    // fault evidence left and this is the only thing that ever paces them.
+    // This many automatic restarts inside the window is treated as a crash and
+    // enters the ladder above, which reaches anything restarting closer together
+    // than window/count - 60s apart at these values. Slower is deliberately left
+    // alone: Palworld's segfault-restart cycle is ~77s at its worst, and coming
+    // straight back is better for the customer than being paced.
+    // Keep the window wider than the reconciler's retry interval times this
+    // count. A container that fails to START never ran, so it is never a fault
+    // and never walks the ladder directly - it reaches it only by filling this
+    // window, and a window narrower than that retries forever.
     restartBurstCount: 5,
-    restartBurstWindowMs: 60000,
+    restartBurstWindowMs: 300000,
     // How long a finished operation stays readable at /apps/operations/:jobId,
     // and how long a client is told to wait between polls while one runs. A
     // RUNNING job never expires - only terminal ones are retained on a clock.
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -132,6 +132,11 @@ module.exports = {
     // and the run length that counts as stable (resets the ladder)
     crashBackoffDelaysMs: [0, 30000, 300000, 900000, 1800000],
     crashBackoffStableRunMs: 600000,
+    // cause-blind backstop for images whose entrypoint discards the payload's
+    // exit status: this many automatic restarts inside the window is treated as
+    // a crash and enters the ladder above
+    restartBurstCount: 5,
+    restartBurstWindowMs: 60000,
     // How long a finished operation stays readable at /apps/operations/:jobId,
     // and how long a client is told to wait between polls while one runs. A
     // RUNNING job never expires - only terminal ones are retained on a clock.
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -410,7 +410,7 @@ module.exports = {
     daemonPONFork: 2020000, // block height where PON (Proof of Node) fork activates - chain works 4x faster after this block
     blocksAllowanceInterval: 1000, // ap differences can be in 1000s - more than 1 day
     removeBlocksAllowanceIntervalBlock: 1625000, // after this block we can start having app updates without extending subscription - block expected in April 19th 2024
-    ownerAppAllowance: 1000, // in case of node owner installing some app, the app will run for this amount of blocks
+    ownerAppAllowance: 1000, // a by-name local install (fluxteam only) runs for this amount of blocks before the expiry sweep removes it
     temporaryAppAllowance: 200, // in case of any user installing some temporary app message for testing purposes, the app will run for this many blocks
     expireFluxAppsPeriod: 100, // every 100 blocks we run a check that deletes apps specifications and stops/removes the application from existence if it has been lastly updated more than 22k blocks ago
     updateFluxAppsPeriod: 9, // every 9 blocks we check for reinstalling of old application versions
--- a/ZelBack/src/routes.js
+++ b/ZelBack/src/routes.js
@@ -1229,6 +1229,12 @@ module.exports = (app) => {
   app.get('/apps/apprestart/:appname?/:global?', alwaysRespond, requireBootSettled, (req, res) => {
     appController.appRestart(req, res);
   });
+  // No :global - a kill is deliberately per-node. Its privilege is narrower than
+  // its siblings' too: appKill checks appownerorfluxteam, so the node operator
+  // cannot order a hard kill of an app they only host.
+  app.get('/apps/appkill/:appname?', alwaysRespond, requireBootSettled, (req, res) => {
+    appController.appKill(req, res);
+  });
   app.get('/apps/apppause/:appname?/:global?', alwaysRespond, requireBootSettled, (req, res) => {
     appController.appPause(req, res);
   });
--- a/ZelBack/src/services/appLifecycle/appInstaller.js
+++ b/ZelBack/src/services/appLifecycle/appInstaller.js
@@ -970,8 +970,19 @@ async function installAppLocally(req, res) {
         blockAllowance = config.fluxapps.temporaryAppAllowance;
       }
       if (!appSpecifications) {
-        // only owner can deploy permanent message or existing app
-        const ownerAuthorized = await verificationHelper.verifyPrivilege('adminandfluxteam', req);
+        // Placing a registered app on a node is not the node operator's call, for
+        // the same reason removing one is not: hosting an app is not owning it.
+        // This branch resolves an app BY NAME from the marketplace, the global
+        // registry or a permanent message, so an operator reaching it is choosing
+        // which customer's app runs on hardware they control - and for a g:/r: app
+        // the new instance syncs that customer's data down to it. The spawner
+        // decides placement; the owner decides everything else.
+        //
+        // The temporary-message branch above is untouched and stays open to any
+        // logged-in user: that is how an app is tested before it is registered,
+        // it is addressed by hash rather than by name, and it expires on its own
+        // (temporaryAppAllowance).
+        const ownerAuthorized = await verificationHelper.verifyPrivilege('fluxteam', req);
         if (!ownerAuthorized) {
           const errMessage = messageHelper.errUnauthorizedMessage();
           res.json(errMessage);
@@ -1134,8 +1145,19 @@ async function testAppInstall(req, res) {
       }
 
       if (!appSpecifications) {
-        // only owner can deploy permanent message or existing app
-        const ownerAuthorized = await verificationHelper.verifyPrivilege('adminandfluxteam', req);
+        // Placing a registered app on a node is not the node operator's call, for
+        // the same reason removing one is not: hosting an app is not owning it.
+        // This branch resolves an app BY NAME from the marketplace, the global
+        // registry or a permanent message, so an operator reaching it is choosing
+        // which customer's app runs on hardware they control - and for a g:/r: app
+        // the new instance syncs that customer's data down to it. The spawner
+        // decides placement; the owner decides everything else.
+        //
+        // The temporary-message branch above is untouched and stays open to any
+        // logged-in user: that is how an app is tested before it is registered,
+        // it is addressed by hash rather than by name, and it expires on its own
+        // (temporaryAppAllowance).
+        const ownerAuthorized = await verificationHelper.verifyPrivilege('fluxteam', req);
         if (!ownerAuthorized) {
           const errMessage = messageHelper.errUnauthorizedMessage();
           res.json(errMessage);
--- a/ZelBack/src/services/appLifecycle/appUninstaller.js
+++ b/ZelBack/src/services/appLifecycle/appUninstaller.js
@@ -21,7 +21,6 @@ const { specificationFormatter } = require('../utils/appSpecHelpers');
 const { stopAppMonitoring } = require('../appManagement/appInspector');
 const appsRuntimeState = require('../appManagement/appsRuntimeState');
 const volumeService = require('../utils/volumeService');
-const imageManager = require('../appSecurity/imageManager');
 const fluxEventBus = require('../utils/fluxEventBus');
 
 const fluxDirPath = process.env.FLUXOS_PATH || path.join(process.env.HOME, 'zelflux');
@@ -1202,37 +1201,20 @@ async function removeAppLocallyApi(req, res) {
       throw new Error('No Flux App specified');
     }
 
-    const authorized = await verificationHelper.verifyPrivilege('appownerabove', req, appname);
+    // The node operator is deliberately NOT here. Hosting an app is not owning it:
+    // an operator who can remove one can script the removal against every install
+    // and keep a customer's app off their node indefinitely, which the customer
+    // experiences as an app that will not stay deployed and cannot diagnose.
+    // Ending an app is the owner's call, or the team's on their behalf.
+    //
+    // One gate, and vetted status does not narrow it further: appownerorfluxteam
+    // admits exactly {owner, fluxTeam, fluxSupport}, which is who may uninstall a
+    // vetted app too. A second check against the same set can only ever agree, and
+    // asking it costs two database reads and a vetted lookup per uninstall.
+    const authorized = await verificationHelper.verifyPrivilege('appownerorfluxteam', req, appname);
     if (!authorized) {
       const errMessage = messageHelper.errUnauthorizedMessage();
       return res.json(errMessage);
-    }
-
-    // For vetted apps, only app owner or Flux Team can uninstall
-    // First, get app specifications to check if vetted
-    const dbopen = dbHelper.databaseConnection();
-    const appsDatabase = dbopen.db(config.database.appslocal.database);
-    const database = dbopen.db(config.database.appsglobal.database);
-    const appsQuery = { name: appname };
-    const appsProjection = {};
-
-    let appSpecsForVettedCheck = await dbHelper.findOneInDatabase(appsDatabase, localAppsInformation, appsQuery, appsProjection);
-    if (!appSpecsForVettedCheck) {
-      appSpecsForVettedCheck = await dbHelper.findOneInDatabase(database, globalAppsInformation, appsQuery, appsProjection);
-    }
-
-    if (appSpecsForVettedCheck) {
-      const appIsVetted = await imageManager.isAppVetted(appSpecsForVettedCheck);
-      if (appIsVetted) {
-        // Check if user is specifically the app owner or Flux Team
-        const isAppOwner = await verificationHelper.verifyPrivilege('appowner', req, appname);
-        const isFluxTeam = await verificationHelper.verifyPrivilege('fluxteam', req);
-
-        if (!isAppOwner && !isFluxTeam) {
-          const errMessage = messageHelper.createErrorMessage('This is a vetted application. Only the app owner or InFlux Support Team are allowed to uninstall it.');
-          return res.json(errMessage);
-        }
-      }
     }
 
     if (global) {
--- a/ZelBack/src/services/appManagement/appController.js
+++ b/ZelBack/src/services/appManagement/appController.js
@@ -164,10 +164,11 @@ async function executeAppGlobalCommand(appname, command, zelidauth, paramA, bypa
  * @param {object|null} appSpecs full app spec (null for a component command)
  * @param {boolean} stopped
  */
-async function setAppOperatorStopped(appname, appSpecs, stopped) {
+async function setAppOperatorStopped(appname, appSpecs, stopped, { awaitPass = false } = {}) {
   const ids = (!appname.includes('_') && appSpecs && appSpecs.version > 3)
     ? appSpecs.compose.map((c) => `${c.name}_${appSpecs.name}`)
     : [appname];
+  let allActuated = true;
   // eslint-disable-next-line no-restricted-syntax
   for (const id of ids) {
     // Written through the reconciler's per-key slot rather than straight to the
@@ -178,9 +179,10 @@ async function setAppOperatorStopped(appname, appSpecs, stopped) {
     // the write lands, and enqueues on release - so the two cannot interleave,
     // and the next pass reads what was just written.
     // eslint-disable-next-line no-await-in-loop
-    await appReconciler.applyIntent(id, async () => {
+    const actuated = await appReconciler.applyIntent(id, async () => {
       await appsRuntimeState.setOperatorStopped(id, stopped);
-    });
+    }, { awaitPass });
+    if (!actuated) allActuated = false;
     // A stop retracts the controller's desire as well as taking the lock. The
     // lock only suppresses the reconciler while it is held; a desire left
     // standing is reconciled against the stopped container the moment the lock
@@ -196,6 +198,29 @@ async function setAppOperatorStopped(appname, appSpecs, stopped) {
     // the sync layer marks a component processed before it asks.
     if (stopped) appReconciler.clearControllerDesired(id);
   }
+  return { ids, actuated: allActuated };
+}
+
+/**
+ * What the containers are actually doing, once the reconciler has had its pass.
+ *
+ * `actuated` says a pass ran, not that it achieved anything: a pass that finds
+ * docker unreachable completes by deferring. So the answer to "is it stopped"
+ * comes from probing, and dockerActual is the probe that can tell a container
+ * being gone from docker being unreachable - which is the difference between
+ * reporting done and reporting pending.
+ * @param {string[]} ids Component identifiers.
+ * @returns {Promise<{settled: boolean, reason: string|null}>}
+ */
+async function containersReachedStopped(ids) {
+  // eslint-disable-next-line no-restricted-syntax
+  for (const id of ids) {
+    // eslint-disable-next-line no-await-in-loop
+    const actual = await appReconciler.dockerActual(id);
+    if (!actual.reachable) return { settled: false, reason: 'docker is not reachable' };
+    if (actual.running) return { settled: false, reason: 'the reconciler has not stopped it yet' };
+  }
+  return { settled: true, reason: null };
 }
 
 async function appStart(req, res) {
@@ -366,43 +391,58 @@ async function appStop(req, res) {
       return res ? res.json(appResponse) : appResponse;
     }
 
+    // THE RECONCILER STOPS IT, NOT THIS HANDLER.
+    //
+    // Two things drove the container before this: the handler called
+    // appDockerStop directly while the reconciler actuated off its own per-key
+    // queue. Two writers to one container is what made an operator stop
+    // interleave with a pass - the pass read the lock, the stop landed, the pass
+    // started the container it had already decided to start. Writing the intent
+    // and letting the single actuator converge removes the second writer rather
+    // than narrowing the window between them.
+    //
+    // The contract is unchanged: awaitPass holds this handler until the pass has
+    // run, so a success still means the container is stopped, in the same
+    // wall-clock the direct call took.
     const isComponent = appname.includes('_'); // it is a component stop
-    let appRes;
+    let ids;
+    let actuated;
+    let stoppedName;
 
     if (isComponent) {
-      // lock BEFORE the docker op (matching the whole-app path): a crash between
-      // the stop and the lock write would leave a stopped container the
-      // reconciler restarts against the operator's intent
-      await setAppOperatorStopped(appname, null, true);
+      ({ ids, actuated } = await setAppOperatorStopped(appname, null, true, { awaitPass: true }));
       appInspector.stopAppMonitoring(appname, false);
-      appRes = await dockerService.appDockerStop(appname);
+      stoppedName = appname;
     } else {
       // Check if app exists before stopping
       const appSpecs = await registryManager.getApplicationSpecifications(mainAppName);
       if (!appSpecs) {
         throw new Error('Application not found');
       }
-      // operator stop persists so the reconciler does not restart it
-      await setAppOperatorStopped(appname, appSpecs, true);
-
+      ({ ids, actuated } = await setAppOperatorStopped(appname, appSpecs, true, { awaitPass: true }));
       // eslint-disable-next-line no-restricted-syntax
-      if (appSpecs.version <= 3) {
-        // eslint-disable-next-line no-await-in-loop
-        appInspector.stopAppMonitoring(appname, false);
-        appRes = await dockerService.appDockerStop(appname);
-      } else {
-        // For composed applications (version > 3), stop all components in reverse order
-        // eslint-disable-next-line no-restricted-syntax
-        for (const appComponent of appSpecs.compose.reverse()) {
-          appInspector.stopAppMonitoring(`${appComponent.name}_${appSpecs.name}`, false);
-          // eslint-disable-next-line no-await-in-loop
-          await dockerService.appDockerStop(`${appComponent.name}_${appSpecs.name}`);
-        }
-        appRes = `Application ${appSpecs.name} stopped`;
-      }
+      for (const id of ids) appInspector.stopAppMonitoring(id, false);
+      stoppedName = appSpecs.name;
     }
 
-    const appResponse = messageHelper.createDataMessage(appRes);
+    // A pass that completed is not a container that stopped - docker being
+    // unreachable completes by deferring. Probe rather than infer, so a stop
+    // that has not happened yet is never reported as one that has.
+    const outcome = actuated
+      ? await containersReachedStopped(ids)
+      : { settled: false, reason: 'no reconcile has run yet' };
+
+    if (!outcome.settled) {
+      // Accepted, not applied. The intent is durable and the reconciler will
+      // converge, so an error here would be false - the old direct call threw
+      // in exactly this case, after the lock had already been written.
+      const pending = messageHelper.createDataMessage(
+        `Application ${stoppedName} will be stopped: ${outcome.reason}`,
+      );
+      return res ? res.json(pending) : pending;
+    }
+
+    const appResponse = messageHelper.createDataMessage(`Application ${stoppedName} stopped`);
     return res ? res.json(appResponse) : appResponse;
   } catch (error) {
     log.error(error);
--- a/ZelBack/src/services/appManagement/appController.js
+++ b/ZelBack/src/services/appManagement/appController.js
@@ -721,32 +721,8 @@ async function appPause(req, res) {
 async function appUnpause(req, res) {
   return deprecatedPauseResponse(req, res);
 }
-/**
- * Docker restart app (internal function)
- * @param {string} appname - Application name
- * @returns {Promise<void>}
- */
-async function appDockerRestart(appname) {
-  try {
-    // mainAppName extracted for potential future use
-    // eslint-disable-next-line no-unused-vars
-    const mainAppName = appname.split('_')[1] || appname;
-    const isComponent = appname.includes('_'); // it is a component restart. Proceed with restarting just component
-    if (isComponent) {
-      await dockerService.appDockerRestart(appname);
-      // Note: startAppMonitoring would need to be injected or called separately
-      log.info(`Component ${appname} restarted successfully`);
-    } else {
-      // ask for restarting entire composed application
-      // This would need getApplicationSpecifications from registryManager
-      log.info(`Restarting entire application ${appname}`);
-      await dockerService.appDockerRestart(appname);
-    }
-  } catch (error) {
-    log.error(`Docker restart failed for ${appname}: ${error.message}`);
-    throw error;
-  }
-}
+// Nothing in this module drives a container. Every route here records the
+// operator's desired state and lets the reconciler actuate it.
 
 /**
  * To stop all non Flux running apps. Executes continuously at regular intervals.
@@ -799,6 +775,5 @@ module.exports = {
   appKill,
   appPause,
   appUnpause,
-  appDockerRestart,
   stopAllNonFluxRunningApps,
 };
--- a/ZelBack/src/services/appManagement/appController.js
+++ b/ZelBack/src/services/appManagement/appController.js
@@ -182,6 +182,16 @@ async function operatorTargetIds(appname) {
   }
   const appName = installedRes.data[0].name;
   const ids = await appReconciler.componentIdsOf(installedRes.data);
+  // An installed app with nothing to address is not an app that needs nothing
+  // done to it - it is one whose parts this node cannot work out: a spec that
+  // will not decrypt, falling back to a docker listing that is empty or that
+  // failed outright. Acting on the empty list wrote no intent, settled
+  // vacuously against nothing, and answered that the command had succeeded.
+  // Refused rather than reported, in the operator's terms: what they need to
+  // know is that nothing happened.
+  if (!ids.length) {
+    throw new Error(`Application ${appName} was not changed: this node cannot determine its components`);
+  }
   if (!appname.includes('_')) return { ids, appName };
   // A component is addressed by name, and a name that is not one of this app's
   // components addresses nothing. Taking it verbatim wrote a durable operator
--- a/ZelBack/src/services/appManagement/appController.js
+++ b/ZelBack/src/services/appManagement/appController.js
@@ -170,8 +170,17 @@ async function setAppOperatorStopped(appname, appSpecs, stopped) {
     : [appname];
   // eslint-disable-next-line no-restricted-syntax
   for (const id of ids) {
+    // Written through the reconciler's per-key slot rather than straight to the
+    // store. A pass reads the lock and acts on that answer once docker has
+    // replied, so a write landing in between is not seen: the pass starts a
+    // container the operator has just stopped and the next pass stops it again.
+    // applyIntent waits out any pass deciding for this id, holds the key while
+    // the write lands, and enqueues on release - so the two cannot interleave,
+    // and the next pass reads what was just written.
     // eslint-disable-next-line no-await-in-loop
-    await appsRuntimeState.setOperatorStopped(id, stopped);
+    await appReconciler.applyIntent(id, async () => {
+      await appsRuntimeState.setOperatorStopped(id, stopped);
+    });
     // A stop retracts the controller's desire as well as taking the lock. The
     // lock only suppresses the reconciler while it is held; a desire left
     // standing is reconciled against the stopped container the moment the lock
--- a/ZelBack/src/services/appManagement/appController.js
+++ b/ZelBack/src/services/appManagement/appController.js
@@ -4,7 +4,6 @@ const serviceHelper = require('../serviceHelper');
 // Removed verificationHelper to avoid circular dependency - will use dynamic require where needed
 const messageHelper = require('../messageHelper');
 const dockerService = require('../dockerService');
-const registryManager = require('../appDatabase/registryManager');
 const appsRuntimeState = require('./appsRuntimeState');
 const appReconciler = require('../appMonitoring/appReconciler');
 const fluxNetworkHelper = require('../fluxNetworkHelper');
@@ -162,16 +161,40 @@ async function executeAppGlobalCommand(appname, command, zelidauth, paramA, bypa
  *
  * @param {string} appname app or component identifier
  * @param {object|null} appSpecs full app spec (null for a component command)
+ * The components are resolved from what the app is actually made of, never from a
+ * stored spec's `compose`: an enterprise app keeps its component names inside an
+ * encrypted blob, so reading compose yields an empty list and a whole-app command
+ * addresses nothing while reporting success.
+ *
  * @param {boolean} stopped
  * @param {object} [options]
  * @param {boolean} [options.awaitPass] hold until the reconcile pass has run
  * @param {boolean} [options.force] a stop is a hard kill, not a graceful stop
  * @param {boolean} [options.alsoRestart] raise the restart generation with the lock
  */
-async function setAppOperatorStopped(appname, appSpecs, stopped, { awaitPass = false, force = false, alsoRestart = false } = {}) {
-  const ids = (!appname.includes('_') && appSpecs && appSpecs.version > 3)
-    ? appSpecs.compose.map((c) => `${c.name}_${appSpecs.name}`)
-    : [appname];
+async function operatorTargetIds(appname) {
+  const mainAppName = appname.split('_')[1] || appname;
+  // eslint-disable-next-line global-require
+  const appQueryService = require('../appQuery/appQueryService');
+  const installedRes = await appQueryService.installedApps(mainAppName);
+  if (!installedRes || installedRes.status !== 'success' || !installedRes.data.length) {
+    throw new Error(`Application ${mainAppName} is not installed on this node`);
+  }
+  const appName = installedRes.data[0].name;
+  const ids = await appReconciler.componentIdsOf(installedRes.data);
+  if (!appname.includes('_')) return { ids, appName };
+  // A component is addressed by name, and a name that is not one of this app's
+  // components addresses nothing. Taking it verbatim wrote a durable operator
+  // lock under a component that does not exist - nothing clears one, and it holds
+  // the real component down if one is ever created with that name.
+  if (!ids.includes(appname)) {
+    throw new Error(`Component ${appname} is not installed on this node`);
+  }
+  return { ids: [appname], appName: appname };
+}
+
+async function setAppOperatorStopped(appname, stopped, { awaitPass = false, force = false, alsoRestart = false } = {}) {
+  const { ids, appName } = await operatorTargetIds(appname);
   // Components come up in compose order and go down in the reverse of it, so a
   // dependency outlives what writes to it: the database stops after the server it
   // serves, not before it. awaitPass holds each component's pass open before the
@@ -224,7 +247,7 @@ async function setAppOperatorStopped(appname, appSpecs, stopped, { awaitPass = f
     // the sync layer marks a component processed before it asks.
     if (stopped) appReconciler.clearControllerDesired(id);
   }
-  return { ids, actuated: allActuated };
+  return { ids, actuated: allActuated, appName };
 }
 
 /**
@@ -244,6 +267,10 @@ async function containersReachedStopped(ids) {
     // eslint-disable-next-line no-await-in-loop
     const actual = await appReconciler.dockerActual(id);
     if (!actual.reachable) return { settled: false, reason: 'docker is not reachable' };
+    // Nothing there is not the same as stopped. dockerActual distinguishes the two
+    // and this read the pair as one, so a command against a container that does not
+    // exist settled - and answered "stopped" for something that was never running.
+    if (!actual.exists) return { settled: false, reason: 'it is not installed on this node' };
     if (actual.running) return { settled: false, reason: 'the reconciler has not stopped it yet' };
   }
   return { settled: true, reason: null };
@@ -337,23 +364,9 @@ async function appStart(req, res) {
     //
     // awaitPass holds this handler until the pass has run, so a success still
     // means the container is running in the same wall-clock the direct call took.
-    const isComponent = appname.includes('_'); // it is a component start
-    let ids;
-    let actuated;
-    let startedName;
-
-    if (isComponent) {
-      ({ ids, actuated } = await setAppOperatorStopped(appname, null, false, { awaitPass: true }));
-      startedName = appname;
-    } else {
-      // Check if app exists before starting
-      const appSpecs = await registryManager.getApplicationSpecifications(mainAppName);
-      if (!appSpecs) {
-        throw new Error('Application not found');
-      }
-      ({ ids, actuated } = await setAppOperatorStopped(appname, appSpecs, false, { awaitPass: true }));
-      startedName = appSpecs.name;
-    }
+    // Which components this addresses - and whether the app or component even
+    // exists here - is resolved from what the app is made of, in one place.
+    const { ids, actuated, appName: startedName } = await setAppOperatorStopped(appname, false, { awaitPass: true });
 
     // A pass that completed is not a container that started - docker being
     // unreachable completes by deferring, and a synced component the election
@@ -439,23 +452,9 @@ async function appStop(req, res) {
     // Monitoring goes with the container, so the reconciler turns it off when it
     // stops one. Doing it here stopped the sampler for a container the stop had
     // not reached - an unreachable docker left it running and unwatched.
-    const isComponent = appname.includes('_'); // it is a component stop
-    let ids;
-    let actuated;
-    let stoppedName;
-
-    if (isComponent) {
-      ({ ids, actuated } = await setAppOperatorStopped(appname, null, true, { awaitPass: true }));
-      stoppedName = appname;
-    } else {
-      // Check if app exists before stopping
-      const appSpecs = await registryManager.getApplicationSpecifications(mainAppName);
-      if (!appSpecs) {
-        throw new Error('Application not found');
-      }
-      ({ ids, actuated } = await setAppOperatorStopped(appname, appSpecs, true, { awaitPass: true }));
-      stoppedName = appSpecs.name;
-    }
+    // Which components this addresses - and whether the app or component even
+    // exists here - is resolved from what the app is made of, in one place.
+    const { ids, actuated, appName: stoppedName } = await setAppOperatorStopped(appname, true, { awaitPass: true });
 
     // A pass that completed is not a container that stopped - docker being
     // unreachable completes by deferring. Probe rather than infer, so a stop
@@ -530,23 +529,9 @@ async function appRestart(req, res) {
     // the same request satisfied. Expressing it as a level rather than an action
     // is what removes the race: there is no window between this handler deciding
     // and the reconciler deciding, because only one of them decides.
-    const isComponent = appname.includes('_'); // it is a component restart
-    let ids;
-    let actuated;
-    let restartedName;
-
-    if (isComponent) {
-      ({ ids, actuated } = await setAppOperatorStopped(appname, null, false, { awaitPass: true, alsoRestart: true }));
-      restartedName = appname;
-    } else {
-      // Check if app exists before restarting
-      const appSpecs = await registryManager.getApplicationSpecifications(mainAppName);
-      if (!appSpecs) {
-        throw new Error('Application not found');
-      }
-      ({ ids, actuated } = await setAppOperatorStopped(appname, appSpecs, false, { awaitPass: true, alsoRestart: true }));
-      restartedName = appSpecs.name;
-    }
+    // Which components this addresses - and whether the app or component even
+    // exists here - is resolved from what the app is made of, in one place.
+    const { ids, actuated, appName: restartedName } = await setAppOperatorStopped(appname, false, { awaitPass: true, alsoRestart: true });
 
     const outcome = actuated
       ? await containersReachedRunning(ids)
@@ -605,23 +590,9 @@ async function appKill(req, res) {
     // with a mode: the lock, plus force. The reconciler reads the mode where it
     // stops the container, which keeps the choice of signal beside the decision
     // to stop rather than in a handler racing it.
-    const isComponent = appname.includes('_'); // it is a component kill
-    let ids;
-    let actuated;
-    let killedName;
-
-    if (isComponent) {
-      ({ ids, actuated } = await setAppOperatorStopped(appname, null, true, { awaitPass: true, force: true }));
-      killedName = appname;
-    } else {
-      // Check if app exists before killing
-      const appSpecs = await registryManager.getApplicationSpecifications(mainAppName);
-      if (!appSpecs) {
-        throw new Error('Application not found');
-      }
-      ({ ids, actuated } = await setAppOperatorStopped(appname, appSpecs, true, { awaitPass: true, force: true }));
-      killedName = appSpecs.name;
-    }
+    // Which components this addresses - and whether the app or component even
+    // exists here - is resolved from what the app is made of, in one place.
+    const { ids, actuated, appName: killedName } = await setAppOperatorStopped(appname, true, { awaitPass: true, force: true });
 
     const outcome = actuated
       ? await containersReachedStopped(ids)
--- a/ZelBack/src/services/appManagement/appController.js
+++ b/ZelBack/src/services/appManagement/appController.js
@@ -5,11 +5,11 @@ const serviceHelper = require('../serviceHelper');
 const messageHelper = require('../messageHelper');
 const dockerService = require('../dockerService');
 const registryManager = require('../appDatabase/registryManager');
-const appInspector = require('./appInspector');
 const appsRuntimeState = require('./appsRuntimeState');
 const appReconciler = require('../appMonitoring/appReconciler');
 const fluxNetworkHelper = require('../fluxNetworkHelper');
 const { extractIp, extractPort } = require('../utils/socketAddressUtils');
+const fluxEventBus = require('../utils/fluxEventBus');
 const log = require('../../lib/log');
 
 const { globalCmdDelayMs } = config.fluxapps;
@@ -163,8 +163,12 @@ async function executeAppGlobalCommand(appname, command, zelidauth, paramA, bypa
  * @param {string} appname app or component identifier
  * @param {object|null} appSpecs full app spec (null for a component command)
  * @param {boolean} stopped
+ * @param {object} [options]
+ * @param {boolean} [options.awaitPass] hold until the reconcile pass has run
+ * @param {boolean} [options.force] a stop is a hard kill, not a graceful stop
+ * @param {boolean} [options.alsoRestart] raise the restart generation with the lock
  */
-async function setAppOperatorStopped(appname, appSpecs, stopped, { awaitPass = false } = {}) {
+async function setAppOperatorStopped(appname, appSpecs, stopped, { awaitPass = false, force = false, alsoRestart = false } = {}) {
   const ids = (!appname.includes('_') && appSpecs && appSpecs.version > 3)
     ? appSpecs.compose.map((c) => `${c.name}_${appSpecs.name}`)
     : [appname];
@@ -180,7 +184,22 @@ async function setAppOperatorStopped(appname, appSpecs, stopped, { awaitPass = f
     // and the next pass reads what was just written.
     // eslint-disable-next-line no-await-in-loop
     const actuated = await appReconciler.applyIntent(id, async () => {
-      await appsRuntimeState.setOperatorStopped(id, stopped);
+      await appsRuntimeState.setOperatorStopped(id, stopped, { force });
+      // Raised inside the same slot as the lock, so a pass cannot read one
+      // without the other and bounce a container the operator meant to keep down.
+      if (alsoRestart) await appsRuntimeState.requestRestart(id);
+      // The operator's intent is the one desired-state write in this flow that
+      // announced nothing, so nothing could be ordered against it - and the
+      // failure it hides is an actuation on the PREVIOUS intent arriving after
+      // this one landed. Published from inside the slot: after the write, so it
+      // can never claim an intent that did not persist, and before the pass,
+      // which is what makes it the ordering point.
+      fluxEventBus.publish('app:operatorIntent', {
+        // The bare component id the reconciler publishes its own actuations
+        // under. An event carrying a different spelling of the same component
+        // cannot be ordered against them, which is the only thing it is for.
+        identifier: dockerService.getBaseAppName(id), stopped, force, restartRequested: alsoRestart,
+      });
     }, { awaitPass });
     if (!actuated) allActuated = false;
     // A stop retracts the controller's desire as well as taking the lock. The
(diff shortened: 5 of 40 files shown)


Cabecinha84 (commented):
PR #1780 — deep review

  fix/restart-pacing-crash-only → fix/syncthing-first-run-gate · 350 add / 33 del · 10 files · 5 commits · no prior
  reviews or comments (the PR is completely un-reviewed, so nothing here is re-litigating past feedback).

  What I actually verified

  - Ran the two affected unit files at head: 131 passing, 0 failing — matches the claimed 119 → 131.
  - Ran the full unit suite at head: 5105 passing, 5 pending, 23 failing — all 23 are in dockerService tests, which need
  a live Docker daemon and are untouched by this PR. Environmental, not caused by the change.
  - Traced every call site of restartWaitMs / recordRestart in ZelBack/ — there are exactly two, both in
  appReconciler.js, both updated. Both new params default to true, so nothing else changes.
  - Simulated the real reconcile loop against the actual (non-stubbed) appsRuntimeState module to confirm the emergent
  behaviour, rather than trusting the unit tests' arithmetic.
  - Cherry-picked the core commit onto origin/development to test the stacking claim.

  The core logic is correct

  - crashed is computed only after !actual.reachable (appReconciler.js:735) and actual.indeterminate (:744) have already
  returned, so it is never derived from a degraded inspect — the two error paths of dockerActual that return exitCode:
  null with no oomKilled field are unreachable at that point.
  - exitCode is gated on everRan (Status !== 'created'), so a never-run container is correctly not read as a fault.
  OOMKilled comes from the same inspect — no second Docker call, no TOCTOU.
  - The sliding window is right. autoRestartWindow is capped at RESTART_BURST_COUNT and burstExceeded reads recent[0] —
  the oldest of the last N — so stale entries can never false-trip. Verified: 4 restarts spread over hours plus one fast
  one does not trip.
  - setOperatorStopped(id, false) clears both arrays, and appController calls it before the Docker op on both appStart
  and appRestart, so a deliberate restart genuinely starts from clean.
  - The early return in restartWaitMs skips the stable-run ladder reset, which I initially flagged — but it self-heals:
  the ladder is cleared on the first crashed=true call. Verified: 3 crash rungs + weeks of clean restarts + one genuine
  crash → wait 0.
  - prepareCollection's twin-merge carries the new field with the correct cap.

  Test quality is genuinely high — the new unit tests are specific and the e2e suite 54 asserts the mechanism rather
  than the constant.

  Findings

  1. The burst ceiling has a 12.0-second cliff — and the PR body describes it as "a minute or two" ⟵ the one that
  matters

  burstExceeded is evaluated before the append, against 5 entries spanning 5 × interval. So the trip condition is 5d ≤
  60000 → d ≤ 12.0s. I measured it directly:

  spacing=12.000s -> paced after 6 restarts
  spacing=12.001s -> NEVER paced (60 restarts, 720s of hammering)
  spacing=13s     -> NEVER paced
  spacing=20s     -> NEVER paced
  spacing=60s     -> NEVER paced

  The PR justifies the gap as "A container that merely restarts every minute or two is not a host problem and is
  deliberately left alone." But the gap does not start at a minute or two — it starts at 13 seconds. An
  exit-0-laundering container dying every 13s now gets ~277 unpaced container creations per hour, forever, each one a
  full containerd shim + netns + veth + iptables + cgroup cycle. That container was previously paced to the 30-minute
  cap.

  The "Note for review" only owns the 77-second Palworld case. The actual behaviour change covers everything from 13s to
  the 10-minute stable-run window. I'd ask for either a second, wider rung (e.g. 20 restarts in 10 minutes also
  disposes into the ladder) or an explicit, quantified statement that 13s–10min unpaced is accepted — with the config
  comment corrected, because "cause-blind backstop" currently oversells what it catches.

  2. boot-lock.js now fails open where it used to fail closed

  if (queue.indexOf(ticket) < BOOT_LOCK_WIDTH) { /* acquired */ }

  indexOf returns -1 when our ticket is not in the queue, and -1 < 2 is true — so a missing ticket reads as "you hold
  the lock". bootQueue() returns [] whenever readdirSync throws, and run-parallel.sh:94 does rm -rf /tmp/e2e-boot-lock.
  If the directory ever disappears mid-gate, every waiter acquires simultaneously and the semaphore is silently defeated
  — which is precisely the starvation/contention failure the lock was written to prevent.

  The old queue[0] === ticket failed closed (kept waiting, eventually threw the explicit "queue is wedged" error).
  One-line fix:

  const pos = queue.indexOf(ticket);
  if (pos >= 0 && pos < BOOT_LOCK_WIDTH) { ... }

  Low probability with the current runner, but a silent over-admission is a bad failure mode for a lock and the fix is
  free.

  3. The stale comment on the failed-start path

  appReconciler.js:910-913 still says "pacing is free — the attempt was recorded above, so a persistent failure walks
  the backoff ladder instead of hammering." That is no longer directly true: a start failure on a never-run container
  has exitCode: null → crashed=false → no ladder entry. It now retries every MANAGED_RETRY_MS (5s) and is bounded only
  by the ceiling (5 × 5s = 25s ≤ 60s window, so it does still trip). It works, but the coupling is now implicit and
  undocumented: if restartBurstWindowMs is ever tuned below 5 × MANAGED_RETRY_MS — and both are config-driven precisely
  so the harness can compress them — a permanently failing start hammers forever. Worth updating the comment and,
  ideally, a test pinning that relationship.

  4. The stacking claim in the description is factually wrong

  ▎ "no file in this PR is touched by any branch below it, so this retargets development as the five below it land."

  Not so. Overlaps with the branches below:

  ┌──────────────────────────────────────────┬─────────────────────┐
  │                   file                   │   also touched by   │
  ├──────────────────────────────────────────┼─────────────────────┤
  │ ZelBack/config/default.js                │ #1774, #1777, #1778 │
  ├──────────────────────────────────────────┼─────────────────────┤
  │ appMonitoring/appReconciler.js           │ #1774, #1777, #1779 │
  ├──────────────────────────────────────────┼─────────────────────┤
  │ test-infra/runner/framework/boot-lock.js │ #1774               │
  ├──────────────────────────────────────────┼─────────────────────┤
  │ test-infra/runner/framework/test-env.js  │ #1774, #1779        │
  ├──────────────────────────────────────────┼─────────────────────┤
  │ tests/unit/appReconciler.test.js         │ #1777               │
  └──────────────────────────────────────────┴─────────────────────┘

  I cherry-picked 0a205a6ed onto origin/development: it conflicts on ZelBack/config/default.js. The conflict is trivial
  (adjacent config keys), and the intent — merge last, after the stack — is fine. But "retarget and go" is not accurate;
  this needs to merge in stack order or be rebased.

  5. 6f7016ad3's commit message contradicts the PR body

  matters

  burstExceeded is evaluated before the append, against 5 entries spanning 5 × interval. So the trip condition is 5d ≤
  60000 → d ≤ 12.0s. I measured it directly:

  spacing=12.000s -> paced after 6 restarts
  spacing=12.001s -> NEVER paced (60 restarts, 720s of hammering)
  spacing=13s     -> NEVER paced
  spacing=20s     -> NEVER paced
  spacing=60s     -> NEVER paced

  The PR justifies the gap as "A container that merely restarts every minute or two is not a host problem and is
  deliberately left alone." But the gap does not start at a minute or two — it starts at 13 seconds. An
  exit-0-laundering container dying every 13s now gets ~277 unpaced container creations per hour, forever, each one a
  full containerd shim + netns + veth + iptables + cgroup cycle. That container was previously paced to the 30-minute
  cap.

  The "Note for review" only owns the 77-second Palworld case. The actual behaviour change covers everything from 13s to
  the 10-minute stable-run window. I'd ask for either a second, wider rung (e.g. 20 restarts in 10 minutes also
  disposes into the ladder) or an explicit, quantified statement that 13s–10min unpaced is accepted — with the config
  comment corrected, because "cause-blind backstop" currently oversells what it catches.

  2. boot-lock.js now fails open where it used to fail closed

  if (queue.indexOf(ticket) < BOOT_LOCK_WIDTH) { /* acquired */ }

  indexOf returns -1 when our ticket is not in the queue, and -1 < 2 is true — so a missing ticket reads as "you hold
  the lock". bootQueue() returns [] whenever readdirSync throws, and run-parallel.sh:94 does rm -rf /tmp/e2e-boot-lock.
  If the directory ever disappears mid-gate, every waiter acquires simultaneously and the semaphore is silently defeated
  — which is precisely the starvation/contention failure the lock was written to prevent.

  The old queue[0] === ticket failed closed (kept waiting, eventually threw the explicit "queue is wedged" error).
  One-line fix:

  const pos = queue.indexOf(ticket);
  if (pos >= 0 && pos < BOOT_LOCK_WIDTH) { ... }

  Low probability with the current runner, but a silent over-admission is a bad failure mode for a lock and the fix is
  free.

  3. The stale comment on the failed-start path

  appReconciler.js:910-913 still says "pacing is free — the attempt was recorded above, so a persistent failure walks
  the backoff ladder instead of hammering." That is no longer directly true: a start failure on a never-run container
  has exitCode: null → crashed=false → no ladder entry. It now retries every MANAGED_RETRY_MS (5s) and is bounded only
  by the ceiling (5 × 5s = 25s ≤ 60s window, so it does still trip). It works, but the coupling is now implicit and
  undocumented: if restartBurstWindowMs is ever tuned below 5 × MANAGED_RETRY_MS — and both are config-driven precisely
  so the harness can compress them — a permanently failing start hammers forever. Worth updating the comment and,
  ideally, a test pinning that relationship.

  4. The stacking claim in the description is factually wrong

  ▎ "no file in this PR is touched by any branch below it, so this retargets development as the five below it land."

  Not so. Overlaps with the branches below:

  ┌──────────────────────────────────────────┬─────────────────────┐
  │                   file                   │   also touched by   │
  ├──────────────────────────────────────────┼─────────────────────┤
  │ ZelBack/config/default.js                │ #1774, #1777, #1778 │
  ├──────────────────────────────────────────┼─────────────────────┤
  │ appMonitoring/appReconciler.js           │ #1774, #1777, #1779 │
  ├──────────────────────────────────────────┼─────────────────────┤
  │ test-infra/runner/framework/boot-lock.js │ #1774               │
  ├──────────────────────────────────────────┼─────────────────────┤
  │ test-infra/runner/framework/test-env.js  │ #1774, #1779        │
  ├──────────────────────────────────────────┼─────────────────────┤
  │ tests/unit/appReconciler.test.js         │ #1777               │
  └──────────────────────────────────────────┴─────────────────────┘

  I cherry-picked 0a205a6ed onto origin/development: it conflicts on ZelBack/config/default.js. The conflict is trivial
  (adjacent config keys), and the intent — merge last, after the stack — is fine. But "retarget and go" is not accurate;
  this needs to merge in stack order or be rebased.

  5. 6f7016ad3's commit message contradicts the PR body

  The message ends with "NOT YET VALIDATED AT GATE SCALE." The PR body then reports a full 82/82 gate at that exact SHA
  in 70 minutes. The validation happened; the message didn't get amended. Anyone reading git log later inherits a
  warning that is no longer true.

  6. Observed behaviour the description doesn't mention (not a defect)

  Simulating the real loop for an exit-0 container dying every 2s gives waits of [0,0,0,0,0,0, 30s, 5m, 0,0,0,0,0, 15m,
  ...]. Each paced wait itself breaks the burst window, so the container earns another ~6 free restarts before it
  re-trips. It does escalate to the 30m cap, and steady state is ~6 restarts per 30 minutes — bounded and perfectly
  acceptable. But it is not the monotone ladder the description implies ("the next restart is paced 30s, and if the
  hammering resumes it walks to the 30m cap"). One sentence in the body would set expectations correctly for whoever
  debugs this from logs later.

  ---
  Verdict

  Not a blind ack — but nothing here is a correctness bug in the reconciler. The change is well-reasoned, the exit-code
  asymmetry argument is sound, and the test coverage is unusually good.

  What I'd ask for before acking:

  1. #1 needs an answer — either a wider second rung, or an explicit, quantified acceptance of the 13s–10min gap with
  the config comment corrected. This is the substantive one.
  2. #2 — the pos >= 0 guard. One line, take it.
  3. #4 and #5 — description and commit-message accuracy. Cheap, and #4 matters for whoever does the merge.
  4. #3 and #6 — comment/description updates; would not block on them alone.

  If David pushes back on #1 with "13s–10min unpaced is the intended trade and the customer outcome is worth it," that's
  a legitimate call and I'd ack on the strength of the rest — but it should be a stated decision, not a threshold that
  reads as tighter than it is.

MorningLightMountain713:
Thanks — the substantive one was right and is fixed, along with three of the others. Branch is now `3581b93ad`.

**1. The 12-second cliff — fixed by widening the window to 300 seconds** (`6704fb900`). Your arithmetic was correct: the ceiling is judged before the append against five entries spanning five gaps, so its reach is `window / count`, which at 5-in-60s was 12.0 seconds and nothing slower. I reproduced it against the real module with a control case that must trip, so a false "never paced" would have shown up.

The knob is really that one number. 5-in-300 puts the reach at 60 seconds, which is the band the description always claimed. Widening is safer than it looks in the direction that matters: `setOperatorStopped(id, false)` clears both arrays before the docker operation, so an operator restarting an app while debugging cannot walk into it, and a wrong trip costs a wait that the next genuine crash or operator start clears.

Framing worth stating, because it is the reason this mattered: the ceiling is not about "crashing". A non-zero exit or an OOM kill is recognised as a fault and paced by that path. A container exiting 0 is never treated as a fault at all, even when it is failing and hiding it — so for an image that launders its exit status, **the ceiling is the only thing that would ever pace it**. A 13-second invisible band matters more than it sounds.

Palworld sits outside the new reach: roughly 77 seconds at its worst against a 60-second threshold. That margin is 17 seconds on a number the note itself hedges, so the body now states the boundary and the observation together rather than leaving it to be rediscovered.

The reach had no test, and one derived from the same constants cannot notice them being retuned — which is how twelve seconds went unnoticed. Both are pinned now: the mechanism against `window / count`, and the shipped decision in seconds. The latter fails against the old window.

**2. `pos >= 0` — taken** (`3581b93ad`). Proven with a probe against the real module before and after: with the directory intact a full queue correctly refuses the waiter; with it removed the old code admitted one, and now both fail closed.

**3. The stale comment — corrected.** It now says what actually happens: a failed start never ran, so it is not a fault and does not walk the ladder directly; it reaches the ladder by filling the burst window. The relationship you flagged is written next to `restartBurstWindowMs`, where someone tuning it would see it. The 300-second window also takes that margin from 2.4x to 12x.

**4. The stacking claim — corrected, and it was worse than you scoped.** Six of the ten files are shared, not one: `ZelBack/config/default.js` (#1774, #1775, #1777, #1778), `appReconciler.js` (#1774, #1777, #1779), `tests/unit/appReconciler.test.js` (#1774, #1777), `test-env.js` (#1774, #1779), `boot-lock.js` (#1774) and `tests/unit/globalconfig/default.js` (#1774). You missed #1775 on the config file, #1774 on the reconciler test, and the globalconfig file entirely. The body now says it merges in stack order.

**5. The commit message — not changing it.** A commit message is a record of what was true when it was written; we don't amend to keep them current. The PR body carries the current state, and it does.

**6. The non-monotone ladder — documented.** Each paced wait breaks the burst window, so the container earns roughly six more restarts before it re-trips, settling at about six per 30 minutes. Bounded, as you say, but not the straight walk to the cap the description implied.

Verified at the top of the stack: **85 of 85 suites green** at `f6c6bc399`.

Cabecinha84 (commented):
PR #1780 — deep review at head afc31ccda

  Verdict: don't ACK yet. The 12-second cliff from the previous round is genuinely fixed, and the boot-lock fix is
  correct. But the burst path has a second-order defect that makes the PR's own headline claim about the bound false,
  and it re-opens a laundering bug that this very file's doc comment warns about. It's a small fix, not a redesign.

  What I verified (and how)

  I built a worktree at afc31ccda, ran the real modules, and simulated the actual reconciler loop order (restartWaitMs →
  if paced, wait and re-enter; else recordRestart → start → run → die) against the real appsRuntimeState with a fake DB
  — not the unit tests' arithmetic.

  - Unit suite: 5190 passing, 5 pending, 23 failing — all 23 are dockerService tests requiring a live Docker daemon,
  untouched by this PR. appsRuntimeState + appReconciler together: 134 passing, 0 failing.
  - Lint: clean (0 errors) on every changed lint-covered file. boot-lock.js and suite 54 are under an ignore pattern.
  - CI: build (ubuntu-22.04, 20.x, 7.0) passing.
  - Call sites: exactly two (appReconciler.js:874, :906), both updated; both new params default to true, so nothing else
  changes.
  - Classification: crashed is computed after the !reachable and indeterminate returns, so it never reads a degraded
  inspect. exitCode is gated on everRan; OOMKilled comes from the same inspect (no second call, no TOCTOU). Five
  table-driven tests cover it.
  - The ceiling's premise holds: dockerService.js:1034 creates containers with RestartPolicy: { Name: 'no' }, so every
  restart genuinely flows through the reconciler and lands in the window. Nothing restarts behind its back.
  - The reach is now exactly 60s, as claimed. Measured against the real module: 59s spacing → paced; 60s → paced; 61s →
  never paced, ever. Palworld at ~77s stays unpaced. The previous round's finding is properly fixed.
  - Mutation check: reverting restartBurstWindowMs to 60000 fails exactly one test — "paces a laundered-exit container
  dying every 30 seconds". The shipped-decision test is non-vacuous, as the author claimed.
  - boot-lock pos >= 0: correct and complete. Fail-closed is restored; a vanished ticket now ends at the explicit wedged
  error. bootQueue()'s sort is a stable (ms, pid) ordering, so width 2 is a real FIFO semaphore, not a race.

  Finding 1 — the burst path launders its own ladder and never reaches the 30-minute cap ⟵ blocker

  restartWaitMs resets the ladder when the previous run provably lasted STABLE_RUN_MS:

  const lastRestart = history[history.length - 1];   // restartHistory
  const lastDeath = Math.max(state.lastDiedAt || 0, lastFinishedAtMs || 0);
  if (lastDeath > lastRestart && lastDeath - lastRestart > STABLE_RUN_MS) { reset; return 0; }

  Before this PR, restartHistory received every restart, so lastRestart was always the real one and the difference was
  the real run duration. This PR makes recordRestart append to restartHistory only when crashed || burstExceeded. The
  automatic exit-0 restarts in between are now invisible to that check — so after any rung longer than 600s, the stale
  ladder entry sits more than STABLE_RUN_MS behind the current death and the code concludes the container ran stably for
  ten minutes. It did not: it restarted six times in twelve seconds.

  Traced against the real module, exit-0 container dying every 2s:

  t= 1250s wait=  0s ladder=[10,40,350] window=[342,344,346,348,350]
  t= 1252s ... 1258s  five unpaced restarts, none recorded in the ladder
  t= 1260s wait=  0s ladder=[10,40,350]  <-- STABLE-RUN RESET (ladder wiped)
  t= 1262s wait= 28s ladder=[1260]       <-- back to rung 1

  Steady state is a closed cycle 30s → 5m → 15m → reset → 30s, forever. The 30-minute rung is unreachable on this path.

  ┌─────────────────────────────────────┬──────────────┬──────────────────────────────────┐
  │   exit-0 container dying every 2s   │ starts in 2h │          rungs observed          │
  ├─────────────────────────────────────┼──────────────┼──────────────────────────────────┤
  │ base (fix/syncthing-first-run-gate) │ 7            │ 30s, 5m, 15m, 30m cap            │
  ├─────────────────────────────────────┼──────────────┼──────────────────────────────────┤
  │ this PR at afc31ccda                │ 78           │ 30s, 5m, 15m — cap never reached │
  ├─────────────────────────────────────┼──────────────┼──────────────────────────────────┤
  │ PR body's stated steady state       │ ~24          │ "escalates to the 30-minute cap" │
  └─────────────────────────────────────┴──────────────┴──────────────────────────────────┘

  So the body's "It does escalate to the 30-minute cap, and steady state is about six restarts per 30 minutes" is wrong
  in both halves — measured is ~19–20 per 30 min and the cap is never reached. That matters because for a laundered-exit
  image the ceiling is, by the PR's own argument, the only thing that ever paces it, so what it actually bounds to is
  the whole guarantee.

  The same applies to the failed-start retry loop (MANAGED_RETRY_MS = 5000): it trips the ceiling as the new comment
  promises, but then cycles the same way — ~39 attempts/hour rather than walking to the cap.

  Why no test catches it: every new unit test calls restartWaitMs(id, null, false) and never sets lastDiedAt, so
  lastDeath is 0 and the reset branch is structurally unreachable in the unit suite. E2E suite 54 asserts only the first
  backoff and stops. The interaction between the burst path and the stable-run reset is untested end to end.

  Fix — the stability comparison should use the last actual restart attempt, not the last ladder entry:

  const lastAttempt = Math.max(lastRestart, ...(state.autoRestartWindow || [0]));
  if (lastDeath > lastAttempt && lastDeath - lastAttempt > STABLE_RUN_MS) { ... }

  For a genuinely crashing container the two arrays agree, so nothing changes there; a container that really did run 10
  minutes still resets. Worth pairing with a unit test that passes death evidence across a full trip → pace → re-trip
  cycle, since that is the shape no current test drives.

  Finding 2 — the module default still says 60 seconds

  const RESTART_BURST_WINDOW_MS = config.fluxapps.restartBurstWindowMs ?? 60 * 1000;

  ZelBack/config/default.js ships 300000, but the in-module fallback was not moved with it. Any config that lacks the
  key silently reinstates the 12-second reach that the last round existed to fix — and it would fail silently, because
  both reach tests derive from the constant, not from the config. Should be ?? 5 * 60 * 1000. One line.

  Finding 3 — the PR body is internally inconsistent after the retune

  - Overview: "a container restarting faster than five times in sixty seconds" — that is the old framing; it is now five
  in 300 seconds, reaching 60s apart.
  - Changes table still lists restartBurstWindowMs: 60000; the shipped value is 300000.
  - The cited SHAs (3581b93ad, 6704fb900, f6c6bc399) no longer exist on the branch — it was rebased on 2026-08-11 and
  head is afc31ccda. The "85 of 85 suites green" attestation therefore points at a tree nobody can resolve. Worth
  re-anchoring to the current head, since that gate run is the only evidence for the harness commits.

  Not blocking, worth a line

  - The trip itself is unpaced. When the ceiling first fires, restartWaitMs falls through the early return, finds an
  empty restartHistory, and returns 0; pacing starts one restart later. Consistent with the body's [0,0,0,0,0,0, 30s,
  ...], so it looks deliberate — but the burstExceeded doc comment reads as though the crossing restart is the one
  paced, and it is not.
  - BOOT_LOCK_WIDTH defaults to 2, changing harness parallelism for everyone. Documented, measured, and
  E2E_BOOT_LOCK_WIDTH=1 reverts it. Fine.
  - Suite number 54 is free; seedTestApp genuinely supports { exitCode, exitAfterS } (reconciler-suite.js:418).

MorningLightMountain713:
All three findings are closed on the current head (`b93031a0b` — the branch has been rebased twice since `afc31ccda`: over the merged #1777, then onto #1779's review-round commits).

**Finding 1 — fixed, by a different rule than the one suggested, and the suggested patch is dead code under it.** The fix is in `recordRestart`: a rung is earned by `crashed || burstExceeded || the component already holds rungs`. Once anything has found fault, every restart lands in `restartHistory` — so the ladder's last entry IS the last actual attempt by construction, and the stable-run comparison measures a real run again. The `lastAttempt = max(lastRestart, ...autoRestartWindow)` patch was tried and removed: while rungs stand, every restart appends to both arrays, so the max cannot differ from `lastRestart`; while the ladder is empty, the reset branch is unreachable behind the empty-ladder early return. There is no state in which it changes the answer.

The cycle you traced (30s → 5m → 15m → reset, cap never reached) is pinned at both levels. Unit: "walks the full ladder to the cap on a continuous crash loop (waits honored)" drives the full trip → pace → re-trip shape with death evidence and asserts the cap is reached and sticky, alongside "resets after a recorded stable run" for the legitimate direction — and removing the holds-rungs clause fails exactly that test and nothing else. E2E: your closing hint was taken literally — suite `54-reconciler-burst-ceiling` seeds an exit-0 app with `{ exitCode, exitAfterS }` and drives the ceiling on a live fleet, including the rung that outlasts a stable run. The failed-start retry loop walks to the cap through the same rule.

**Finding 2 — fixed.** The module fallback is `?? 5 * 60 * 1000`, matching the shipped 300000.

**Finding 3 — fixed.** The body was rewritten — framing, config table, SHAs — and now dates every SHA-pinned result, with the gate on the current stack marked pending.

**Non-blocking.** The trip-unpaced behaviour is now stated in `burstExceeded`'s own doc: that restart is counted, not held — the one after it is the first to be made to wait. `BOOT_LOCK_WIDTH` stands as documented.

Unit on the current head: 5,585 passing / 18 pending / 0 failing. The full gate re-run on the rebased stack is pending and will cover suite 54 in sequence.

Cabecinha84 (commented):
Findings

  1. Composed-app stop and kill order was silently reversed — and the test that pinned it was weakened, not deleted

  Base appStop and appKill iterated appSpecs.compose.reverse(). setAppOperatorStopped now builds ids with
  appSpecs.compose.map(...) — forward order — and because awaitPass: true makes each component's pass complete before
  the next id is touched, the order is real and blocking, not incidental.

  The test at tests/unit/appController.test.js:277 is still titled "should stop all components for version 4+ apps in
  reverse order" but now only asserts that both components received the intent. So the behaviour changed, the test name
  still describes the old behaviour, and nothing fails.

  For a two-component app (say DB + app server) this inverts the shutdown order: the DB now goes down first. Nothing in
  the PR body mentions it. Either restore reverse order for the stop/kill direction, or state the decision and rename
  the test — the current state is the worst of the three.

  2. A failed stop or kill is the only actuation path with no retry

  appReconciler.js:886-908 — the !desired && actual.running branch has no try/catch. A throw propagates to
  runReconcile's .catch(), which logs and nothing else: dirty is not set, no scheduleRetry. Every sibling path in the
  same function paces its own retry — the failed start (:1049), the data clear (:868), the paused stop (:824), the
  restart generation (:936). This one drops to the hourly sweep.

  That was structurally true before, but this PR makes it load-bearing: /apps/appstop no longer stops anything itself.
  The operator now gets status: "success" with "Application X will be stopped: the reconciler has not stopped it yet"
  and the container keeps running for up to an hour. Previously the direct appDockerStop threw and they saw the error.

  A try/catch + scheduleRetry(identifier, MANAGED_RETRY_MS) around the stop would close it and match the rest of the
  function.

  3. recordRestartGeneration swallows write failures on the one path deliberately exempt from pacing

  appsRuntimeState.js catches and logs; the caller at appReconciler.js:940 proceeds regardless. Under read-succeeds /
  write-fails (Mongo disk full, for example), the next pass sees desiredGeneration > actuatedGeneration again and
  restarts again, every POST_START_VERIFY_MS, forever — and the comment right above it says this path is explicitly not
  paced by the ladder. Full Mongo failure is safe, because getState returns null and both generations read 0.

  Narrow, but it is the one loop with neither a ladder nor a retry bound. Letting this one throw (so the pass fails and
  retries paced) would be the cheapest fix.

  4. A state read failure downgrades a kill to a graceful stop

  appReconciler.js:892: const stopState = ... await appsRuntimeState.getState(identifier) — and getState returns null on
  a read error. forceKill becomes false and the container gets a graceful stop. The comment on operatorStopForce says
  durability means "a crash between the write and the stop cannot quietly downgrade 'kill now' to a drain" — true for
  that window, but not for a read failure, which does exactly that downgrade. The direction is safe; the comment
  overstates the guarantee.

  5. Error responses became success-shaped

  When docker is unreachable, all four endpoints now return status: "success" with "Application X will be stopped:
  docker is not reachable". The old code threw and returned status: "error". This is deliberate and defended in the
  body, and it is defensible — the intent is durable. But any UI or automation branching on status will read a hard
  infrastructure failure as a success. Worth an explicit line in the body and a heads-up to whoever owns the UI, rather
  than leaving it to be discovered.

  6. Two small ones

  - restartWaitMs's JSDoc still documents @param {boolean} crashed, which the function does not take — and the PR body
    specifically explains why it doesn't. Delete the line.
  - boot-lock.js: aheadOnArrival = Math.max(0, queue.indexOf(ticket)) reports 0 for a ticket that is not in the queue —
    the same -1 conflation that was just fixed one line below. Telemetry only, but it makes a defeated queue look like
    an empty one.

MorningLightMountain713:
Thanks — this was a good review. Five of the six are fixed; on the sixth I think the finding applies the old contract to the new one, and I've left it alone rather than change it quietly. Two of the five turned out to be worse than described and one turned out to be narrower, so details below.

## 1. The composed stop/kill order — fixed

Confirmed at the source: base `appStop` and `appKill` both iterated `compose.reverse()` while `appStart` iterated forward, which is the ordinary dependency shape — bring the database up first, take it down last. The new spread used `.map()` for every direction, and since all eight call sites pass `awaitPass: true`, each component's pass completes before the next id is touched, so the order was real and blocking, exactly as you said.

Reversed on the mapped ids when `stopped` is set. Worth flagging for anyone applying the same fix elsewhere: the ids are a fresh array from `.map()`, so reversing them never touches the spec — the base reversed `compose` itself, which mutates a document other callers hold, and restoring the old line verbatim would have reintroduced that.

**The test was worse than "weakened".** It asserted with `sinon.assert.calledWith`, which says nothing about order, so it passed in either direction — vacuous on the exact property its title named. Both stop and kill now pin the call index. I also pinned the forward direction on the start path, because otherwise a blanket reverse would satisfy the stop test while inverting startup and nothing would notice. Both drills bite: disabling the reverse fails stop and kill alone; reversing every direction fails start alone.

## 2. The unretried stop — fixed, but at the boundary rather than the branch

Confirmed, and the asymmetry is starker than the note suggests: there are 32 `scheduleRetry` sites in that file, and the paused stop, failed start, restart and data clear each wrap their docker call with a comment explaining this precise risk. The operator stop was the only actuation with neither. `runReconcile`'s catch only logged — it set neither `dirty` nor a retry — so it really did fall to the hourly sweep.

I didn't add the try/catch you suggested, because the sites that can throw are by definition the ones nobody thought to guard, and a per-site catch would miss the next one the same way. `runReconcile`'s catch now counts consecutive throws, publishes a `reconcileFailed` actuation event, and paces a retry for the first few before handing the component to the sweep.

Two things make that safe rather than reckless. A pass is level-based — it re-derives desired against actual rather than resuming half-finished work — so a retry is never a repeated action. And every deliberate no-retry decision in that function `return`s rather than throws, so a catch-level net cannot override any of them; the invalid-spec path behaves exactly as its comment says.

It is bounded because `scheduleRetry` is a flat five seconds with no escalation, and an unhandled throw may be permanent — retrying one forever is a log loop rather than a recovery.

**One thing worth recording:** docker being unreachable is *not* this case. That path is anticipated and defers (`docker unreachable … deferring reconcile`), so it never reaches the net — which is why a dockerd bounce cannot exercise it, and why suite 45 covers the deferral rather than this.

## 3. The swallowed generation write — fixed, and there were two sites

Confirmed, including your read-succeeds/write-fails precondition and that a wholly unreachable mongo is safe by accident. **There are two call sites, not one** — the second is the path where a start satisfies a pending restart, and its own comment states the consequence: "left pending, the pass that next finds it running would bounce a container the operator has just watched come up".

Rather than a local guard, this is a misclassification. `appsRuntimeState` splits its writes coherently: history swallows (`recordExit`, `recordRestart`) because losing one costs a log line, and intent throws (`setOperatorStopped`, `requestRestart`) because losing one loses an operator's instruction. `recordRestartGeneration` reads like a recorder and swallowed like one, but it behaves like intent — it is the only thing that stops the next pass bouncing the container. It is now on the correct side of a split that already existed.

**The subtlety your suggestion would have hit:** at both sites the write sat *before* the bookkeeping, so letting it throw as-placed would have discarded the actuated event, the peer notification and the post-start attachment check — for a restart that had already succeeded. Both sites now record last. It is the record that failed, not the restart.

You are right that this is narrow. It is also a symptom rather than a cause, and the broader handling is being designed separately; this change only stops the failure being silent.

## 4. The downgraded kill — fixed by deleting the second read

Confirmed, and your own caveat was the right one: the direction is safe and the comment overstated the guarantee.

It is narrower than the note implies, though. Reaching that branch requires `isOperatorStopped` to have already returned true — the same document, read moments earlier, through the same swallowing function. If reads were failing, the first would have returned null too and the branch is unreachable. So it needs one read to succeed and the very next to fail inside a single pass.

But the width stopped mattering once I looked at why there were two reads at all. The lock and the mode are two fields of one document, and the decision had already read it. A new `operatorStopState` answers both from a single read, and the flag travels with the decision that read it — so there is nothing later to disagree with it, the comment is true as written, and every operator stop costs one round-trip fewer than before. `isOperatorStopped` stays for its other caller.

The tests improved with it: the force cases used to set two stubs that could describe a state production cannot produce, and now set one. The regression test holds the later read empty — which is exactly what the failed read looked like — and asserts the kill still lands; restoring the second read fails it.

## 5. The success-shaped responses — no change, and here is why

This is the one I'd push back on. `success` on those endpoints means the request was accepted, not that the container is down, and the body says which of the two happened: `Application X stopped` when it has been verified, `will be stopped: <reason>` when it has not. The response is already honest about the distinction; only the enum is coarse.

The contract genuinely changed — it used to be synchronous, so success meant the container was down by the time you read it — and the body documents that ("every one of the four reports what happened rather than what was asked for"). But that is a contract change, not a defect in it.

I also checked the consumer question: FluxOS fans these commands out to peers itself, and it never reads the field. `executeAppGlobalCommand` is fire-and-forget and `deliverGlobalCommand` only inspects the HTTP status code, retrying a boot-gate 503 and logging anything else. So nothing in the product is misled.

If the UI does branch on `status` for these four I'll happily revisit, but I'd rather not change what operators' tooling sees on an inference.

## 6. Both small ones — fixed

`restartWaitMs`'s doc block documented a `crashed` parameter it does not take; removing that argument was this PR's own decision and explained in the body, so the doc contradicted both. Deleted.

The boot-lock telemetry is sharper than "telemetry only", so I took it more seriously than the note did. `bootQueue()` returns an empty list precisely when the lock directory cannot be read — the defeated-semaphore case the lock exists to expose — so flattening `-1` to `0` made a gate that was not working at all read as a gate with no contention. It reports `unknown` now, which is what the wedged-wait error two lines below already did. Worth noting the commit that introduced that reasoning fixed the same conflation for the acquire decision and the wait error and left the telemetry line behind it.

## Verification

**Unit: 5,603 passing, 18 pending, 0 failing.** Every fix has a mutation that kills its test and nothing else — including the two that discriminate rather than merely fail: unbounding the retry fails the bound test while leaving the retry test green, and restoring the two-read order fails the bounce path while the start path stays green.

**Harness: the full 87-suite gate is green on the top of the stack** — `suites_pass=87 suites_fail=0`, run with all six images digest-verified against the tree they were built from.

Cabecinha84 (commented):
One thing I'd want answered before merge

  The appkill privilege boundary is inconsistent with appremove. appKill now asks for appownerorfluxteam on the stated
  principle that "ending an app they only host is not theirs to order". But appUninstaller.js:1205 still gates removal
  on appownerabove — the node operator can uninstall the app, and the uninstall path itself calls
  dockerService.appDockerKill. So the operator is denied the lesser action and allowed the greater one, via a code path
  that performs the denied action.

  This isn't a security hole (it's strictly narrower than appownerabove, on a route that has never been reachable, so
  there's no regression), and the reasoning is defensible on its own terms — a hard kill mid-write corrupts data without
  removing the app. But it's a boundary someone will trip over, and it's a product call, not something the code
  settles. Also worth checking whether HomeUI exposes a Kill control to node operators.

  Non-blocking, worth saying in the review

  - The intent write now sits behind the reconciler's per-key slot, so a wedged pass blocks durability, not just
    actuation. applyIntent loops on await inFlight.get(id) before it writes. Its doc says the wait is "bounded by one
    pass of ONE component" — true only if passes terminate, and the dockerode calls in dockerActual/healDetachedNetwork
    carry no explicit timeout. Previously the lock write went straight to mongo and only the docker call could hang. The
    direction is safe (nothing wrong gets written), but the failure mode moved from "intent recorded, actuation stuck"
    to "intent never recorded". A bounded wait, or just a sentence in that doc comment, would record it.
  - The recordRestartGeneration loop is now rate-bounded, not time-bounded. Reads-work/writes-fail still bounces the
    container once per hourly sweep, forever, after the 3 fast retries. Much better than every POST_START_VERIFY_MS, and
    the author flags the broader handling as separate work — just don't let "fixed" read as "bounded".
  - configFallbacks.test.js only accepts numeric/array literals. A future config.fluxapps.someFlag ?? false fails with
    "make it a plain literal", which isn't achievable. Given the stated policy is that the fallback shouldn't exist at
    all, that's arguably intended — but the message should say so.
  - E2E_BOOT_LOCK_WIDTH defaults to 2, changing the shared gate for everyone off measurements from one 16-core host at
    MAXN=6. The lock exists precisely because overlapping boots starved each other and a healthy node blew its
    event-wait budget. Width 2 deliberately re-admits some of that; on a smaller runner a boot-starvation flake will
    read as a product failure. The =1 escape hatch is documented — just make sure the default is what the shared gate
    wants.
  - The headline overstates the invariant. "The reconciler is the only thing that starts, stops, restarts or kills a
    container" holds for the four operator endpoints. advancedWorkflows (redeploy), containerMountRecovery,
    appUninstaller, appInstaller and fluxNetworkHelper still call dockerService.appDocker* directly at 20+ sites. The
    body does acknowledge this under stoppingContainers, so it's not a contradiction — but the headline invites a
    stronger reading than the tree supports.
  - On your finding 5 (success-shaped errors): I agree with the author, for the reason he gave. I confirmed
    deliverGlobalCommand inspects only error.response.status (retries 503, logs anything else) and never reads the
    body's status field, and executeAppGlobalCommand is fire-and-forget. Nothing inside FluxOS is misled. The exposure
    is HomeUI and third-party tooling, and since it is a user-visible contract change on four endpoints, it deserves a
    heads-up to whoever owns the UI rather than a silent ship. Not a reason to hold the PR.

MorningLightMountain713:
Five of the seven are answered in code here, one is answered below without a change, and one is fixed in the frontend PR. The rest of the reply covers what changed beyond them.

## The appkill / appremove boundary

You are right that it was incoherent, and it is resolved the other way from the one you might expect: **the node operator loses `appremove` too**, rather than gaining `appkill`.

Hosting an app is not owning it. An operator who can remove one can script that against every install and keep a customer's app off their node indefinitely, which the customer experiences as an app that will not stay deployed and cannot diagnose. That is a worse failure than the one the narrower `appkill` gate was avoiding.

`appownerorfluxteam` admits exactly `{owner, fluxTeam, fluxSupport}` — the same set the vetted-app branch below it admitted. That branch existed only to keep the operator away from vetted apps, so with the operator out of the gate above it can no longer refuse anyone who reaches it. It goes, and with it two database reads and a vetted lookup on every uninstall.

Your point was that the boundary is inconsistent, and it was inconsistent in one more place than the two you named. `installapplocally` and `testappinstall` gated their by-name branch on `appownerabove`'s cousin `adminandfluxteam`, so the operator could still *place* an app. Closing remove and leaving that open would have left them able to put a customer's app on their node and unable to take it off — a worse boundary than either end state, and not one you should have to come back and point out. All three now admit the same set.

It is the same reasoning, not a wider one: that branch resolves an app **by name** from the marketplace, the global registry or a permanent message, so an operator reaching it chose *which* customer's app ran on hardware they control — and for a `g:`/`r:` app the new instance syncs that customer's data down to it. Placement is the spawner's decision.

Three things this deliberately does not touch, since they are what an operator and a developer actually use:

- **The temporary-message branch above it is untouched**, so an app under test is still installable by any logged-in user. It is addressed by hash rather than by name, never reaches the by-name gate, and expires on its own via `temporaryAppAllowance`.
- **The operator keeps every reversible control** — start, stop, restart — and `appstop` remains their lever against a misbehaving app.
- **The harness itself installs through this route** with the flux team key, which is why the full 87-suite gate is green on it rather than the change being argued from first principles.

You also asked whether HomeUI exposes a kill control to operators. It does not — `appkill` has no UI at all, on any branch. But the *uninstall* button is a node-operator control (`local.vue:57`, `privilege === 'admin' || 'fluxteam'`), so it is narrowed to fluxteam in RunOnFlux/fluxos-frontend#150.

## The intent write behind the reconciler's slot

Correct, and the doc comment claimed a bound it does not have. It now says the wait is one pass *provided the docker probe returns*, and names what a hang actually costs: not correctness — nothing wrong is recorded, and the caller's request hangs on a wedged daemon either way — but durability, because the write has not landed and so the intent does not survive a FluxOS restart during the hang.

Bounding that wait is not the repair. Giving up on it and writing anyway restores the interleave the slot exists to prevent. Bounding the docker calls is the repair, and that is fleet-wide work: dockerode is constructed bare, so no call FluxOS makes can time out, and every retry bound in the tree counts *answers* — a hang gives none. Written up internally rather than attempted here.

## recordRestartGeneration

Taken, and the comment now says "a rate, not a bound" in those words, with the arithmetic: `UNHANDLED_FAILURE_RETRIES` clears only on a pass that succeeds, so reads-work/writes-fail gives four bounces over fifteen seconds and then one per hourly sweep for as long as the condition holds. The broader handling stays separate work, and the note not to let "fixed" read as "bounded" was the right one to make.

## configFallbacks and the unachievable message

Taken, and further than suggested. Rather than keeping the refusal and rewording it, the guard now compares booleans, null and strings as well as arithmetic — a boolean drifts from config exactly as a number does, and declining to compare one leaves a whole class of fallback outside the net. Booleans and null are literal comparisons and a quoted string is a regex capture, so the digits-and-operators whitelist guarding `Function()` is unchanged. The unparseable sentinel is a Symbol rather than `undefined`, because `?? undefined` is writable and would otherwise be reported as unreadable instead of compared.

What stays genuinely unparseable — a named constant, an expression over another variable — now names the action that always exists: drop the fallback, since config ships the key.

## E2E_BOOT_LOCK_WIDTH

No change, and the evidence is in the PR body rather than on a bench.

The note reads the lock as having one job, and the width as re-admitting the thing it was built to stop. `boot-lock.js` gives it two, stopped by different mechanisms.

**Contention** is what the width governs. Overlapping boots make a healthy node blow its event-wait budget — suite 22's second fleet booted at load ~15 on 16 cores and mongo collection prep crawled at 7–17s a step. That is the cost width 2 accepts, deliberately, and it is what was measured.

**Starvation** is a different failure with a different cause: a protocol where every waiter races the same atomic create serves whoever wakes first, not whoever waited longest. In the 2026-08-06 gate, suite 13 sat thirty minutes through other suites' boots and was killed by the wall-clock backstop with 18 of its 19 tests already passed. What fixes that is **arrival ordering**, which the width commit does not touch — the queue still drains in the order it formed, which is what makes starvation structural rather than statistical. Width 2 changes how many of the queue's front are served at once, not who is served. So it cannot re-admit the starvation case, and the contention case is the one it was sized against.

On the measurement: the commit that introduced the width said, in capitals, that a two-fleet bench was not enough and that one full gate against the baseline would settle it. Two gates ran, matched on suite count, boot count and fleet-size distribution — the table in the PR body. Gate wall clock 4,978s → 4,192s, p50 queue wait 87.8s → 24.1s, total queue wait 11,697s → 3,747s. Boots do contend and each costs more; the overlap wins by 786 seconds.

The part of the note that stands is its last line: the right width is a property of the box, which is why it is a knob and not a constant. `E2E_BOOT_LOCK_WIDTH=1` restores strict serialisation without reverting anything, and it is documented at the definition.

## The headline

Fair, and it is fixed by making the claim true rather than by softening it. Body line 7 now reads:

> Every operator command is desired state the reconciler converges to. `appStart`, `appStop`, `appRestart` and `appKill` record the operator's intent and let the single actuator converge on it; no route in `appController` drives a container. Seventeen `appDockerStart`/`Stop`/`Restart`/`Kill` calls remain outside the reconciler — install, uninstall, redeploy and backup/restore, mount recovery, and the two-hourly sweep that stops containers FluxOS does not own — and those paths are not what this changes.

Your "20+ sites" was the right order of magnitude; the exact figure for run-state verbs is seventeen outside the reconciler, twenty-four including it. `fluxNetworkHelper` is no longer among them, for the reason below.

## Your finding 5, and a correction to what I told you

I said nothing in the product reads the `status` field on those four endpoints, and that if the UI did I would revisit. **It does, and I was wrong.**

`AppControl.vue`'s `handleAppOperation` branches on `response.data.status`, and on success it shows the toast and fires a **single** `/apps/listallapps` refetch with no polling and no retry. So where `success` now means "accepted" and the container has not transitioned yet, the panel shows a green toast beside a status that is still the old one, and nothing looks again.

What I checked and got right was the FluxOS side: `deliverGlobalCommand` reads only `error.response.status` and `executeAppGlobalCommand` is fire-and-forget, so nothing inside FluxOS is misled. The exposure is entirely HomeUI, which is what you said. Fixed in RunOnFlux/fluxos-frontend#150 — the panel now refreshes on a spreading schedule (0/2/5/10/20s) so it converges on what actually happened.

## The commits that answer a question you did not ask

`appController.appDockerRestart` is gone. Its only caller was the IP-change handler, and it resolves ONE container by name — a composed app has none under its bare app name, its containers are `<component>_<app>`, and the lookup reads `.Id` off an undefined result. So the first composed app threw a `TypeError` with nothing to catch it. The throw left the loop for `adjustExternalIP`'s outer catch: every app after it kept the old address, the `fluxipchanged` broadcast never went out, the confirmation transaction never ran and the geolocation update never happened. One log line was the trace.

The surviving apps are now handed over as durable restart requests, one per component, which survive a FluxOS restart part way through and queue behind the same slot as every other intent. Handed over through a seam rather than a require: `appUninstaller` requires `fluxNetworkHelper` and `appReconciler` requires `appUninstaller`, so reaching upward from the network layer closes a cycle — and twenty-odd app-layer modules import it, so it stays underneath them. `fluxNetworkHelper.setOnAddressChanged` mirrors `appUninstaller.setOnComponentRemoved`, and `serviceManager` wires it as it wires that one.

**The same loop no longer uninstalls the node's own apps.** Before restarting an app, `adjustExternalIP` asks whether it is already running at the address being moved to and force-removes it if so — sound in principle, since one instance per IP is what the host port mapping allows. The node's own registration is not another instance: it stores its own running-app row locally, at the address benchmark reports, so the row found at that address is usually itself, and removing on that took an app that was exactly where it belonged.

The address is still compared at IP granularity: that is the question being asked, and a UPnP sibling on another port holds the ports just as surely. Own-ness is the full socket address, which is what separates the two — the same `socketAddressesMatch` test the collision check in that file already uses.

Suite 56 covers the whole path end to end against a composed app. It changes the node's address for real — the new one onto its interface, the old one off — while benchmark's probe reports the move and the deterministic list still says where it was. Peers then fail to reach it because it genuinely is not there, which is what makes it ask benchmark whether it moved; the untouched list is what lets those same peers still accept the broadcast that follows. Both components restart on their own identifiers, and a peer accepts the `fluxipchanged` message, which is what distinguishes a handler that ran to the end from one that stopped at the first app.

Two harness assumptions went with it, both things an address change necessarily violates: the runner reached a node at a fixed address, so a node that moved went blind to its own client; and the stub identified a node by the address its requests arrive from, so a node that moved lost its identity and skipped its own availability check. Both are in this PR.

## Four more that are not in your notes

Two of them are this PR's own doing and you should not have had to find them.

**A stop is now honoured while a data volume is unavailable.** That branch defers everything except a stop, since a stop takes nothing from the app dir — but it read only the controller's stop. With these handlers no longer driving docker themselves, an operator's stop deferred with it, so the container ran on over the missing volume and the response said "will be stopped" for as long as the mount failed. The force flag is carried through too, so an `appkill` there is still a kill.

**The restart generation is raised by an atomic increment.** Read-then-write asked `getState` for the current number, which answers null both for "no record" and "the read failed" — so a failed read wrote a generation below the one already actuated, which reads as nothing pending. It also counts two concurrent requests as two, where the read-then-write lost one.

The other two are older and reachable through paths this changes. An operator command's components come from `componentIdsOf` rather than from a stored spec's `compose`, which is empty for an enterprise app and had it address nothing while reporting success — and a component name is checked against them rather than taken on trust, since one that is not the app's still wrote a durable lock. And `adjustExternalIP` defers while the node's own address is unknown, returning before the `userconfig` write so the change stays pending rather than being decided on a comparison against nothing.

Cabecinha84 (commented):
Findings

  1. componentIdsOf returns two different id spellings — component commands break on undecryptable enterprise apps ⟵ fix
  before merge

  appReconciler.componentIdsOf (commit 40fadc009) pushes bare comp_app for readable apps, but for unreadable ones it
  pushes raw docker names — and getAppIdentifier prefixes those with flux, so they come back as fluxcomp_app. The base
  enqueueAll got away with this ("canonicalised to the bare component identifier by enqueue"); the new consumer does
  not.

  appController.operatorTargetIds then does:

  if (!ids.includes(appname)) throw new Error(`Component ${appname} is not installed on this node`);

  The operator passes the bare comp_app, ids holds fluxcomp_app → every component-level
  appstart/appstop/apprestart/appkill against a v8 enterprise app whose spec will not decrypt is refused with a wrong
  message. That is a regression against base, which took the component name verbatim, and it bites exactly when
  benchmark is down — the state support most needs manual control in.

  Same root cause voids the composed reverse-stop order for that class (docker listing order is arbitrary, not compose
  order).

  Fix is one line in componentIdsOf: return ids.map((id) => dockerService.getBaseAppName(id)). Safe for enqueueAll too,
  which already canonicalises downstream.

  2. Empty ids reports success having written no intent ⟵ same commit, same class

  operatorTargetIds guards the component path but not the whole-app path. If componentIdsOf returns [] — enterprise app,
  decrypt fails, and dockerListContainers throws or finds no matching containers — then setAppOperatorStopped loops
  zero times, containersReachedStopped([])/ReachedRunning([]) settle vacuously true, and the handler answers Application
  X stopped. No durable lock was written. That is precisely the failure mode the PR's own doc comment says it exists to
  close ("addresses nothing while reporting success"). Narrow — it needs the containers to be gone, e.g. mid-redeploy —
  but it is a one-line if (!ids.length) throw.

  3. The appremove privilege narrowing is a product decision with no test and a cross-repo dependency

  The reasoning is sound and internally consistent (I checked: appownerorfluxteam admits exactly {owner, fluxTeam,
  fluxSupport}, which is byte-for-byte the set the deleted vetted-app branch admitted, so removing that branch genuinely
  cannot change an outcome). But:

  - No test pins it. tests/unit/appUninstaller.test.js is untouched and its only auth test is a generic
    verifyPrivilege.resolves(false) — reverting the gate to appownerabove fails nothing. E2E suite 60 exercises
    appremove with fluxTeamKey(), so it only proves the positive half. Contrast appkill, which got both halves driven
    from the node's own identity in suite 32. The install side got explicit calledWith('fluxteam', req) assertions; the
    uninstall side got none.
  - Deployment order matters. local.vue:57 still shows Uninstall to admin, so between this merging and
    fluxos-frontend#150 shipping, every node operator who presses it gets a 401. These two have to land together.
  - Node operators lose the ability to remove an app from their own hardware. They keep a durable appstop, so it is not
    the worst version of this, but it is a policy change that will generate tickets, and it is being made inside a
    reconciler PR. It should carry an explicit Flux-team sign-off and a release note, not just a merge.

MorningLightMountain713:
All three findings are answered in code, on `fc619642c`. Two of them are defects in this PR's own new code and one is a test that should have come with the change. Details below, including one part of finding 3 I am not taking.

## 1. componentIdsOf's two spellings — fixed at the source

Right, and the tell was already in the file: `setAppOperatorStopped` publishes its event as `getBaseAppName(id)` with a comment about spellings that cannot be ordered against one another. The list was known not to be canonical at one consumer and left uncanonical at the source.

Fixed in `componentIdsOf` rather than at `operatorTargetIds`, for the reason you gave — `enqueueAll` and `requestRestartOf` canonicalise downstream, so one spelling for everyone costs them nothing and leaves no second place to remember.

Scope is the membership test alone, and you did not overstate it: `applyIntent` and `appsRuntimeState` both canonicalise on ingest, and `dockerActual` re-applies the namespace through `dockerContainerInspect`, so the whole-app path, the durable lock and the settle probes were correct throughout.

Two things I would separate out from the finding.

**The composed stop order is not the same root cause, and the one-line fix does not address it.** An app whose spec will not decrypt has no compose to order by — docker's listing order is the only thing that exists, and renaming its entries does not recover an order that was never available. That is a property of enumerating from containers, not something this PR changed, and not something a fix can reach.

**On "it bites exactly when benchmark is down":** decryption runs through `checkAndDecryptAppSpecs`, which needs Arcane, the app's global message row and an owner/height lookup. More to the point, a node that cannot decrypt its own apps is a state we do not see, and it is not the state in which anyone reaches for these HTTP endpoints — a node in that condition is worked on directly. The fix stands because it is a regression against base and costs one line. The severity does not.

## 2. The empty list — fixed, wider than described, and worded for the operator

Confirmed exactly as set out: zero components all reach stopped by construction, `allActuated` is only ever lowered by an iteration, and the handler answered `Application X stopped` having written no durable intent.

It is wider than "the containers are gone". `componentIdsOf` wraps the docker listing in a catch that returns the list short, so a **listing failure** produces the same empty list — and that route answered success while docker was unreachable, bypassing the `if (!actual.reachable)` branch that exists for precisely that case, because the loop never ran.

The guard is where you suggested. The message is not. HomeUI's `handleAppOperation` reads `response.data.data.message` and shows it verbatim in an error toast, so the string is operator-facing, and "is not installed on this node" would be false — the app is installed. It now says the app was not changed and that this node cannot determine its components, which is true on both routes.

The same catch also drops undecryptable apps from the sweep and from the address-change restart, returning a list no caller can distinguish from a complete one. It stays lenient — one app's failure must not cost the readable apps their sweep — but it now names what it dropped, at error level.

Reachability, since it bounds what the finding is worth: every route to an empty list runs through a failed decrypt. A v1-3 app contributes its own name and a readable v4+ app contributes its compose entries, so only the undecryptable branch can contribute nothing. Your own stated trigger needs it too — for a readable app the list comes from compose, so absent containers do not empty it.

## 3. The appremove narrowing

**The gate is pinned.** One assertion, and it is the hop that was open rather than the policy, which was already covered either side of it: `verificationHelperUtils.test.js` asserts `appownerorfluxteam` is false for the node operator where `OrHigher` admits them, alongside true for the owner and for the team, and `verificationHelper.test.js` maps the string to that function. What was missing was that this route asks for it. It is asserted with the app name and as the only privilege check on the path — which also fails if the removed vetted-app branch returns.

You were right that the install side already had this, and it is stronger than you credited: one case satisfies the old gate, refuses the new one, and asserts the request is rejected.

**Deployment order — agreed, and it is in the body now.** `local.vue:57` renders Uninstall to admin until fluxos-frontend#150 lands, so the two ship together. That mismatch, rather than the policy, is what would generate a ticket.

**The sign-off — no.**

Hosting an app is not owning it. Operators are paid to run a customer's app on their hardware, and removing it is not a property right over that hardware — it is destroying a customer's instance and their volume data while continuing to be paid to host it.

There is no operator need on the other side of it. An app cannot exceed what was bought: `Memory`, `MemorySwap`, `NanoCpus` and `StorageOpt` are set on the container from the spec, so an app inside its allocation is spending cycles the operator sold, and an app outside it is a containment defect to be fixed in the limits. What an operator keeps is the control that is honest about its own cost — stop FluxOS, or drop the node — both of which give up the payment along with the obligation. A per-app control does not, and that is the shape that gets abused.

The argument does not stop at these two, and the rest is already decided. `appStart`, `appStop` and `appRestart` still ask for `appownerabove`; a follow-up PR takes the remaining app control away from the node operator entirely, with the frontend controls narrowed alongside it. It is separate only so that each gate ships with the control that exposes it. The body says so too.

"It will generate tickets" is a prediction with nothing behind it — neither of us has measured whether operators use that button. What can be said is that the change is right either way, and that a ticket asking why an operator cannot delete a tenant's data has an answer.

On sign-off: this is a FluxOS policy decision, taken by the people who own the policy, reviewed here in the open with the reasoning written down. There is no third party being bypassed. It belongs in the release notes, which is what a release note is for, and not behind a gate invented for it.

## Verification

Unit on the current head: **5,635 passing / 18 pending / 0 failing.** Every fix has a mutation that kills its own test and nothing else — reverting the canonicalisation fails the two identifier tests, removing the empty-list guard fails the vacuous-success test, and swapping the uninstall gate back to `appownerabove` fails the gate test. The full harness gate runs on the top of the stack.

Cabecinha84 (approved):
ack

PR #1781, opened 2026-08-09
Test harness isolation: the fleet installs from itself and reaches nothing outside — and four production fixes it turned up

MorningLightMountain713:
## Overview

The integration fleet reached the public internet on every run — roughly 1.2 million packets left the box per gate, to Canonical, Docker, NodeSource, Syncthing and GitHub. That made the gate depend on five third parties being up and honest, and it meant a node under test was not quite the node we ship. This branch closes that: the fleet now runs with its network sealed, installs its packages from a signed repository the image carries, and records anything that tries to leave.

Most of this is test harness. **But it is not only test harness, and the production changes should not be reviewed as if they were** — four of the seven commits are production fixes, three of which were found by building the harness rather than by looking for them, and two of the harness commits carry a production change alongside. The split is spelled out below before anything else, so nothing has to be inferred from a commit title.

Gated twice: **83/83 at `5cb0383b7`** (60m06s) and **84/84 at `5ec88dfbe`** (60m35s), the second with the fleet network sealed and tcpdump reporting `0 packets received by filter`.

## What is production, and what is harness

| commit | | what it changes |
|---|---|---|
| `0ac9563d3` endpoints become configuration | **production** | `config/default.js`, `cloudUIUpdateService`, `geolocationService`, `fluxService`, `imageManager`, `appSpecHelpers`, `systemService`, `scripts/update-cloudui.sh` |
| `85fa7b480` a boot sweep is over when it says so | harness **+ observability** | harness wait; `fileOperationRecovery` gains a `fileops:recovered` publication (+14 lines) |
| `0c03b6ef9` an apt failure resumes into the same failed task, forever | **production** | `fifoQueue`, `systemService` — no harness files at all |
| `5d5425ddc` a legacy node installs its packages from the fleet | harness **+ production** | the harness apt repository, and `monitorSystem`'s package checks become awaited rather than fire-and-forget `setImmediate`, publishing `system:packages-checked` |
| `5cb0383b7` one package apt cannot find stops a node installing the others | **production** | `fifoQueue`, `systemService` — no harness files at all |
| `aaf01e5c8` a node installs what a package needs, not what it suggests | **production** | `systemService` — no harness files at all |
| `5ec88dfbe` the fleet has no way out | harness only | the sinkhole, suite 95 |

By line count:

| | files | changed |
|---|---|---|
| **production code** (`ZelBack/`, `scripts/`) | 10 | +181 / −45 |
| **production unit tests** (`tests/unit/`) | 6 | +343 / −25 |
| harness (`test-infra/`) | 14 | +694 / −77 |

The production surface is four files carrying real behaviour change — `fifoQueue.js`, `systemService.js`, `cloudUIUpdateService.js` and `scripts/update-cloudui.sh` — plus five files where an endpoint literal became a config lookup, and one additive event publication. **`fluxEventBus` is test-only**, so both publications are observability and neither is on a production path.

## The apt queue defects

None of these were planned work. All three are invisible while apt succeeds, which is why nothing had found them — and closing the network is precisely what makes apt failure the normal case, so they were fixed one step before they would have started biting.

### A failed task resumes into itself, forever

`updateAptCache` passes `retainErrors: false` so that a failed `apt-get update` is dropped rather than left at the head of the queue. `queueAptGetCommand` built its worker options from `retries` alone, so `retainErrors` and `retryDelay` never arrived and the queue kept the task. `monitorAptCache` then resumes the queue on every failure, and it is async — it awaits several `runCommand`s first, so its `resume()` lands after `work()` has exited and `working` is back to false. With `retries: 0` there is no delay between attempts.

Measured against the real modules: **55,812 worker attempts in two seconds**, one core flat out, for as long as the mirror stays unreachable.

The fix is to forward every worker option the caller set.

### The queue died the moment that shortcut started working (review round three)

Fixing the payload above switched on a trapdoor the payload bug had been holding
shut. `runWorker` halted the queue **after** emitting `failed`, and an emit is a
synchronous yield: the listener runs right there. `monitorAptCache` resumes for a
failed update with no `await` ahead of it, so its `resume()` landed inside the
emit and the next line overwrote it. The loop broke out with work still queued,
`working` went false, and nothing restarted it - `push()` only calls `work()`
when the queue is not already working. Every apt task behind the update was
stranded, every one pushed later too, and `monitorSystem`'s `allSettled` never
settled: `system:packages-checked` was never published and chrony, syncthing and
netcat never installed until FluxOS restarted.

The fix is not to make listeners behave, because a queue cannot police them.
It is to stop depending on when they run: settle state first, then notify, and a
listener may reply instantly, slowly or not at all. Two consequences follow. A
resumed final failure reaches the retry sleep with no retries left, so a
ladder-over break stops it waiting out the full delay - 60s by default - holding
everything behind it. And `resume()` no longer reassigns `finished` when called
re-entrantly: `work()` returns at once while the loop is running, and storing
that resolved promise told `clear()` the queue was idle mid-flight, on the
apt-is-broken path `systemService` reaches from these very failures.

**An apt command completing is now a published fact**, which is what makes this
testable at all. A legacy boot runs a handful of them, so it is an event rather
than a cadence, and which completed - in what order - is the only direct evidence
the queue carried on past one that failed. Suite 94 gains the third regime a real
node boots into, packages missing *and* `apt-get update` failing:
`createTestEnv({ aptBadSource: true })` adds a source apt cannot reach alongside
the good one, so the update fails as it does behind a dead mirror while the work
behind it stays doable - swapping the good source out would fail that too and
prove only that a broken node stays broken. Proven both ways on a real fleet:
green 8/8 with the fix, and with the halt moved back after the emit all three
tests of that regime fail.

### The recovery routine could not see what had failed

`runWorker` emitted the whole payload as `options`, but for the `{commandOptions, workerOptions}` shape the command sits one level down — so `options.command` was always `undefined`, and the routine's opening shortcut, which treats an `apt-get update` failure as not worth panicking over, had never once fired. Every apt failure instead took the full recovery path: up to thirty minutes waiting on the dpkg lock, then `fuser -KILL` on it, then `dpkg --configure -a`, then `apt-get install --fix-broken` — for a node that simply could not reach a mirror.

The fix is to emit `commandOptions`.

### One package apt cannot find stops a node installing any of the others

A retained task went back to the **front** of the queue, so the next resume handed it straight back to the worker ahead of everything already waiting. Observed on a real node with an empty package index: `netcat-openbsd` could not be located and worked through its retry ladder, and when the ladder ended a resume granted it a fresh one. **chrony and syncthing, queued behind it, were never attempted at all** — and there was no ceiling, because each resume started the cycle again for the life of the process.

Retained tasks now go to the back, and a task is handed back at most `maxRetainCycles` (default 3) times before the queue emits `abandoned` and gives up. `systemService` logs that, because the caller already took its error when the first ladder ended and nothing else would ever learn the work had stopped.

### Two existing tests encoded two of these behaviours

One asserted that the failed task sits at the head of the queue, with a comment saying so, and five asserted the exact apt argument list. Both were updated deliberately — the expectations were the defect, not the guard.

## Every endpoint on the boot and app-lifecycle path is configuration

Endpoint literals scattered through `cloudUIUpdateService`, `geolocationService`, `fluxService`, `imageManager`, `appSpecHelpers`, `systemService` and `scripts/update-cloudui.sh` move into `ZelBack/config/default.js`. Configuration lives in the directory fluxbench hashes, so redirecting a node stays detectable tampering — which is why this is configuration and deliberately not an environment variable.

**The heading is scoped deliberately, because the sweep is not total.** Two endpoints outside that path remain literals and are follow-ups rather than oversights: `watchdogService`'s clone URL for `fluxnode-watchdog`, and the Azure provider's `login.microsoftonline.com` authority. Neither runs during boot or an app operation, which is why suite 95 stays green with them in place — that suite proves what the fleet reaches *while a gate runs*.

Two further literals are deliberately staying, and are not endpoints at all. The Google and Azure registry **scopes** (`.../auth/cloud-platform`, `containerregistry.azure.net/.default`) are permission identifiers those providers define, carried inside an auth request and never fetched — a node with a "configured" scope is a node that fails to authenticate. And `paymentService`'s `success_url` is handed to Stripe so it can redirect the customer's browser after checkout; FluxOS never opens a connection to it.

## The fleet serves its own packages

The dependency closure is resolved at image build into a signed apt repository the image carries. Nothing names a package or a version: whatever apt puts in the archive cache *is* the closure it resolved, against this image's own package state.

**One production change rides with this and is worth a reviewer's attention.** `monitorSystem`'s four package checks were fired and forgotten through `setImmediate`, so nothing could tell a check that ran and found its work done from one that never started. They are now awaited, and the result is published as `system:packages-checked` with the list of what was actually installed — published whether or not anything was, on the same principle as `fileops:recovered`, because presence alone cannot distinguish those two states on a seeded node. The checks still run through the same one-at-a-time apt queue; what changes is that `monitorSystem` now knows when they are finished.

Packages are pre-seeded at image build, so a normal fleet boots into the steady state a production node is already in — `monitorSystem()` runs for real, finds its prerequisites satisfied, and installs nothing. A per-suite `aptSeeded: false` flag purges the three packages before FluxOS starts, so the install path is covered too. Purge rather than remove, because a removed package leaves its config behind and reports as `deinstall ok config-files`, which is neither installed nor absent and which no real node is ever in.

Base packages are served over `file://` from the image; only syncthing's source goes over HTTP to the stub, because that is the one source FluxOS actually writes — so the keyring fetch, the source write, apt's HTTP transport and signature verification all run as they do on a node. Signing is load-bearing and proven by mutation: swapping the keyring for an impostor's makes apt refuse both distributions with `NO_PUBKEY` and the packages become unlocatable.

Syncthing now installs from a `.deb` rather than a release tarball. Its repository carries only the two most recent releases, so a `.deb` cannot pin an old version the way a GitHub release can, and the version therefore moves on every rebuild — `build-apt-repo.sh` records it to `syncthing.version` and the stub reads that file, so nothing downstream restates it. One side effect is worth naming: the package is built `[noupgrade]` and the tarball is not, so the `upgrades.syncthing.net` traffic existed *because* the harness installed the tarball, and switching closed it by construction.

## The fleet has no way out, and says so when something tries

The fleet network is `Internal: true` — no route off it at all. Nodes resolve through the external stub, which relays each query to the embedded Docker resolver at `127.0.0.11` (the one that knows every container alias on the network) and refuses anything it cannot answer within 300ms, recording the name and the node that asked. Relaying rather than answering from a list means aliases need no second copy in the stub and cannot drift from the ones the runner actually creates.

The redirection goes through `/etc/resolv.conf` deliberately: FluxOS resolves via c-ares, which does not read `/etc/hosts`, so a hosts entry would have covered nothing.

**Suite 95's third test is the reason the other two mean anything.** It deliberately reaches for `stats.runonflux.io` and asserts the resolver recorded it. On the first run that test failed while both hermeticity assertions passed — the sinkhole was wired to nothing, so "the fleet reached nothing outside" was being recorded against a resolver that recorded nothing at all. Any assertion that something did not happen needs a sibling proving the instrument can see it happen.

## systemService hygiene

`aptRunner` now passes `--no-install-recommends`. Without it, a node asking for four packages installed twenty-one, including `shared-mime-info`, dbus, glib and python3-gi — which is where ten legacy nodes sitting in uninterruptible sleep running `update-mime-database` came from.

**Release note.** This lands in `aptRunner`, so it applies to every apt invocation on every existing legacy node, not just the new ones — the change here most likely to surprise someone operationally. The four packages involved are safe without recommends, and an unseeded install was exercised at gate scale. Also operational: all three entry points now `delete process.env.NODE_CONFIG` at startup, so the `node-config` environment override no longer reaches FluxOS — nothing in the repo used it, but any external tooling that set it loses its overrides.

`addSyncthingRepository()` moves to where the install is. It previously ran before any version check and short-circuited only if the source file already existed, so a node whose syncthing was perfectly current still fetched a release key and wrote a keyring and an apt source it would never read.

## Measurement

Egress per full gate, captured on the host with tcpdump filtered to the harness subnet:

| destination | before | after endpoints | after own packages | after default-deny |
|---|---|---|---|---|
| **total off-box packets** | **~1,200,000** | **455,348** | **2,574** | **0** |
| Canonical | 421,196 | 425,030 | 0 | 0 |
| `archive`/`security.ubuntu.com` | 1,476 | 1,752 | gone | gone |
| `download.docker.com`/`deb.nodesource.com` | 266 | 296 | gone | gone |
| `apt.syncthing.net`/`syncthing.net` | 148 | 140 | gone | gone |
| `upgrades.syncthing.net` | 6 | 6 | gone | gone |
| `github.com` | 64 | 6 | gone | gone |
| `relays.syncthing.net` | 6 | 6 | 6 | gone |

The final capture's own tally is `0 packets captured, 0 packets received by filter`. Before anything was claimed from that zero, a container on the harness subnet curled `archive.ubuntu.com` through the same filter and produced 30 packets with the hostname visible — an empty measurement has two explanations, and the boring one is a broken instrument.

## Testing

- Full unit suite **5,610 passing, 18 pending, 0 failing**.
- Integration gate **87/87** at the current head, 59 minutes, with all six images digest-verified against the tree they were built from.
- Two new integration suites. `94-legacy-package-provisioning` covers all three regimes a legacy node can boot into — seeded, unseeded, and unseeded with `apt-get update` failing — where none were covered before. `95-fleet-reaches-nothing-outside` covers both node types, plus the instrument check described above.
- Every fix carries a mutation that kills its own test and nothing else. Two of the queue tests proved nothing on their first pass — one used a zero retry delay, so removing the ladder-over break changed nothing observable, and one asserted on a flag that holds either way — and were rewritten until they bit.
- One rebase note: #1779's round hardened the syncthing tarball verification in `Dockerfile.fluxos`, which this branch then deletes by design — syncthing arrives as a package from the image's signed apt repository, whose `signed-by` verification supersedes it.

## Notes for review

- Based on `fix/restart-pacing-crash-only` (#1780). That is sequencing, not a code dependency.
- The four production commits could have gone out on their own branch. Keeping them here was a deliberate call: three of them were found by this work, and the fourth is what the harness redirection is built on.
- If you want to review the production change on its own, `git diff 6f7016ad3..HEAD -- ZelBack scripts` is the whole of it — 10 files, +181/−45.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -392,3 +392,9 @@ test-infra/runner/test-logs/
 # mirrorImage). Content-addressed and re-fetchable, so only the first suite of
 # a gate pays for the download.
 test-infra/runner/.image-cache/
+
+# node-config loads this last, over everything in the pinned directory, so a
+# committed one would redirect a production node's endpoints. The harness writes
+# it inside the container at boot; it never belongs in the repo.
+ZelBack/config/local.js
+ZelBack/config/local-*.js
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -538,6 +538,10 @@ module.exports = {
     stallNudgeMaxIntervalMs: 900000, // nudge backoff cap (15min)
     stallRemoveMinWindowMs: 1200000, // 20min minimum evidence window before removal
     stallRemoveMinNudges: 3, // nudges that must have failed before removal
+    // Where a legacy node installs syncthing from. Arcane nodes ship it in the image and
+    // never reach either of these.
+    aptSourceUrl: 'https://apt.syncthing.net/',
+    releaseKeyUrl: 'https://syncthing.net/release-key.gpg',
   },
   // enterpriseAppOwners moved to helpers/enterprisenodes.json (synced from github every 6h, see enterpriseConfig)
   enterprisePublicKeys: [ // list of whitelisted nodes indentity public keys. Most trusted node operators that are publicly known, kyc. Eg Flux team members, Titan.
@@ -596,7 +600,22 @@ module.exports = {
   },
   geolocation: {
     ipApiBaseUrl: 'http://ip-api.com',
-    statsApiBaseUrl: 'https://stats.runonflux.io',
+  },
+  // The network's statistics service. One host, several paths: node location,
+  // marketplace listings, app USD pricing, and the minimum module versions a node
+  // checks its syncthing against at boot.
+  stats: {
+    baseUrl: 'https://stats.runonflux.io',
+  },
+  pricing: {
+    fluxRatesBaseUrl: 'https://viprates.runonflux.io',
+    // Consulted only when the rates service above is unreachable.
+    coingeckoBaseUrl: 'https://api.coingecko.com',
+  },
+  mongodb: {
+    // Where a replacement server signing key is fetched from when the installed one
+    // has expired. The version is appended: /server-<major.minor>.asc
+    signingKeyBaseUrl: 'https://pgp.mongodb.com',
   },
   analytics: {
     url: 'https://cloudaudit.runonflux.io', // analytics server URL (e.g. 'https://analytics.runonflux.io'). Empty = disabled.
--- a/ZelBack/src/services/appSecurity/imageManager.js
+++ b/ZelBack/src/services/appSecurity/imageManager.js
@@ -290,7 +290,7 @@ async function getUserBlockedRepositores() {
       return userBlockedRepos;
     }
     const usableUserBlockedRepos = [];
-    const marketPlaceUrl = 'https://stats.runonflux.io/marketplace/listapps';
+    const marketPlaceUrl = `${config.stats.baseUrl}/marketplace/listapps`;
     const response = await axios.get(marketPlaceUrl);
     console.log(response);
     if (response && response.data && response.data.status === 'success') {
--- a/ZelBack/src/services/appSystem/fileOperationRecovery.js
+++ b/ZelBack/src/services/appSystem/fileOperationRecovery.js
@@ -3,6 +3,7 @@ const log = require('../../lib/log');
 const { appsFolder } = require('../utils/appConstants');
 const executor = require('./volumeExecutor');
 const { sessionForMountedVolume } = require('./volumeSession');
+const fluxEventBus = require('../utils/fluxEventBus');
 
 /**
  * Reclaim what a FluxOS restart left behind from in-flight file operations.
@@ -25,6 +26,19 @@ const { sessionForMountedVolume } = require('./volumeSession');
  * @returns {Promise<{containers: number, removed: number}>}
  */
 async function recoverInterruptedFileOperations() {
+  const result = await sweepEveryMountedVolume();
+  // Published on every path that completes, including the one that found nothing.
+  // "The sweep ran and had nothing to do" is a different fact from "the sweep has
+  // not run yet", and the log line below cannot express the first because it only
+  // fires when there was something to report. Anything that restarts a node to
+  // exercise boot recovery needs to know the pass is over: without a signal it can
+  // only guess, and a pass that lands after the guess reaches into whatever is
+  // running by then.
+  fluxEventBus.publish('fileops:recovered', result);
+  return result;
+}
+
+async function sweepEveryMountedVolume() {
   const containers = await executor.reapOrphanedContainers();
 
   let mounts = [];
--- a/ZelBack/src/services/appSystem/fileOperationRecovery.js
+++ b/ZelBack/src/services/appSystem/fileOperationRecovery.js
@@ -26,15 +26,21 @@ const fluxEventBus = require('../utils/fluxEventBus');
  * @returns {Promise<{containers: number, removed: number}>}
  */
 async function recoverInterruptedFileOperations() {
-  const result = await sweepEveryMountedVolume();
-  // Published on every path that completes, including the one that found nothing.
-  // "The sweep ran and had nothing to do" is a different fact from "the sweep has
-  // not run yet", and the log line below cannot express the first because it only
-  // fires when there was something to report. Anything that restarts a node to
-  // exercise boot recovery needs to know the pass is over: without a signal it can
-  // only guess, and a pass that lands after the guess reaches into whatever is
-  // running by then.
-  fluxEventBus.publish('fileops:recovered', result);
+  // Published on every path that ENDS the pass - the one that found nothing,
+  // and the one that threw. "The sweep ran and had nothing to do" is a
+  // different fact from "the sweep has not run yet", and the log line below
+  // cannot express the first because it only fires when there was something to
+  // report. Anything that restarts a node to exercise boot recovery needs to
+  // know the pass is over: without a signal it can only guess, and a pass that
+  // lands after the guess reaches into whatever is running by then. A throw
+  // still propagates - a startup that throws is retried, and the retry
+  // publishes again, which is safe for the same reason the sweep is.
+  let result = { containers: 0, removed: 0 };
+  try {
+    result = await sweepEveryMountedVolume();
+  } finally {
+    fluxEventBus.publish('fileops:recovered', result);
+  }
   return result;
 }
 
--- a/ZelBack/src/services/cloudUIUpdateService.js
+++ b/ZelBack/src/services/cloudUIUpdateService.js
@@ -1,7 +1,7 @@
 const config = require('config');
 const fs = require('fs');
 const path = require('path');
-const { exec } = require('child_process');
+const { execFile } = require('child_process');
 const axios = require('axios');
 const log = require('../lib/log');
 
@@ -109,7 +109,11 @@ function runUpdateScript() {
   return new Promise((resolve) => {
     log.info('CloudUI: Running update script...');
 
-    exec('npm run update:cloudui', { cwd: PROJECT_ROOT, timeout: 300000 }, (error, stdout, stderr) => {
+    // The script is handed the API host rather than knowing one. Config is the single
+    // source of truth for every endpoint a node reaches, and it lives in the directory
+    // fluxbench hashes - a hardcoded URL in the script, or one taken from the
+    // environment, would be outside that protection and unreachable by the harness.
+    execFile('npm', ['run', 'update:cloudui', '--', config.github.apiBaseUrl], { cwd: PROJECT_ROOT, timeout: 300000 }, (error, stdout, stderr) => {
       if (error) {
         log.error(`CloudUI update script error: ${error.message}`);
         if (stderr) {
--- a/ZelBack/src/services/fluxService.js
+++ b/ZelBack/src/services/fluxService.js
@@ -815,9 +815,9 @@ function getEnterpriseAppOwners(req, res) {
 function getMarketplaceURL(req, res) {
   const userconfig = globalThis.userconfig;
   const development = userconfig.initial.development || false;
-  let marketPlaceUrl = 'https://stats.runonflux.io/marketplace/listapps';
+  let marketPlaceUrl = `${config.stats.baseUrl}/marketplace/listapps`;
   if (development) {
-    marketPlaceUrl = 'https://stats.runonflux.io/marketplace/listdevapps';
+    marketPlaceUrl = `${config.stats.baseUrl}/marketplace/listdevapps`;
   }
   const message = messageHelper.createDataMessage(marketPlaceUrl);
   return res ? res.json(message) : message;
--- a/ZelBack/src/services/geolocationService.js
+++ b/ZelBack/src/services/geolocationService.js
@@ -118,7 +118,7 @@ async function setNodeGeolocation() {
           dataCenter: ipRes.data.hosting,
         };
       } else {
-        const statsApiUrl = `${config.geolocation.statsApiBaseUrl}/fluxlocation/${localIp}`;
+        const statsApiUrl = `${config.stats.baseUrl}/fluxlocation/${localIp}`;
         const statsRes = await serviceHelper.axiosGet(statsApiUrl);
         if (statsRes.data.status === 'success' && statsRes.data.data) {
           storedGeolocation = {
--- a/ZelBack/src/services/systemService.js
+++ b/ZelBack/src/services/systemService.js
@@ -705,6 +705,7 @@ async function monitorAptCache(event) {
 async function monitorSystem() {
   if (isArcane) return;
 
+  let installed = [];
   try {
     // ensureChronyd answers whether chrony ended up configured, which is true
     // both when it installed it and when it was already there. What it installed
@@ -729,19 +730,24 @@ async function monitorSystem() {
       ),
     };
 
-    const outcomes = await Promise.all(Object.values(checks));
     const names = Object.keys(checks);
-
-    // Published whether or not anything was installed, because "checked and had
-    // nothing to do" is a different fact from "has not run yet" and the logs
-    // below only speak in the first case. A node that is already provisioned -
-    // which is every node after its first boot - is otherwise indistinguishable
-    // from one whose checks have not started.
-    fluxEventBus.publish('system:packages-checked', {
-      installed: names.filter((_, i) => outcomes[i]),
+    const settled = await Promise.allSettled(Object.values(checks));
+    // Named individually rather than as one failure, so a log says which check
+    // could not answer instead of only that something could not.
+    settled.forEach((outcome, i) => {
+      if (outcome.status === 'rejected') log.error(`monitorSystem - the ${names[i]} check failed: ${outcome.reason?.message ?? outcome.reason}`);
     });
+    installed = names.filter((_, i) => settled[i].status === 'fulfilled' && settled[i].value);
   } catch (error) {
     log.error(error);
+  } finally {
+    // Published on every path out of here, including the ones that failed.
+    // "Checked and had nothing to do" is a different fact from "has not run
+    // yet", and a check that threw is a third - tried and could not. All three
+    // are silence to a waiter, which then waits out its whole timeout for an
+    // answer that is never coming, and a node already provisioned looks the
+    // same as one whose checks never started.
+    fluxEventBus.publish('system:packages-checked', { installed });
   }
 }
 
--- a/ZelBack/src/services/systemService.js
+++ b/ZelBack/src/services/systemService.js
@@ -569,6 +569,13 @@ async function monitorSyncthingPackage() {
 
         if (upToDate) return false;
 
+        // After the up-to-date return - a current node is not handed a keyring
+        // and a source it will never read - and BEFORE the sources rewrite: the
+        // rewrite fails on a node with no source file at all, and its failure
+        // must not stand between that node and the file being created, or the
+        // node warns once a day forever and never upgrades.
+        await addSyncthingRepository();
+
         // The sources changed at version 2.0.0 from stable, to stable-v2
         const hasNewSources = serviceHelper.minVersionSatisfy(
           currentSyncthingVersion,
@@ -584,10 +591,9 @@ async function monitorSyncthingPackage() {
         }
       }
 
-      // Here, rather than before the version check: a node whose syncthing is
-      // already current has nothing to install, and writing it a keyring and an
-      // apt source it will never read is work it did not ask for. Only a node
-      // that is about to install needs somewhere to install from.
+      // The uninstalled case: nothing on the system, so there is no version to
+      // judge and nothing to rewrite - the source is simply ensured before the
+      // install.
       await addSyncthingRepository();
 
       const upgraded = await ensurePackageVersion(
--- a/ZelBack/src/services/systemService.js
+++ b/ZelBack/src/services/systemService.js
@@ -76,6 +76,7 @@ async function aptRunner(options = {}) {
   // any apt after 1.9.11 has the DPkg::Lock::Timeout option.
   const params = [
     '-y', // Auto-answer yes to prompts
+    '--no-install-recommends', // Only what the package needs, not what it suggests
     '-o', `DPkg::Lock::Timeout=${timeout}`, // How long to wait for a lock
     '-o', 'Dpkg::Options::=--force-confdef', // Use default for new config files
     '-o', 'Dpkg::Options::=--force-confold', // Keep old config files on conflict
@@ -530,8 +531,6 @@ async function monitorSyncthingPackage() {
   try {
     if (syncthingTimer) return;
 
-    await addSyncthingRepository();
-
     const versionChecker = async () => {
       const {
         data: { data },
@@ -571,6 +570,12 @@ async function monitorSyncthingPackage() {
           }
         }
       }
+
+      // Here, rather than before the version check: a node whose syncthing is
+      // already current has nothing to install, and writing it a keyring and an
+      // apt source it will never read is work it did not ask for. Only a node
+      // that is about to install needs somewhere to install from.
+      await addSyncthingRepository();
 
       const upgraded = await ensurePackageVersion(
         'syncthing',
--- a/ZelBack/src/services/systemService.js
+++ b/ZelBack/src/services/systemService.js
@@ -98,6 +98,14 @@ async function aptRunner(options = {}) {
     params: envParams,
   });
 
+  // A fact, not a cadence: a legacy boot runs a handful of apt commands, and
+  // WHICH of them completed - and in what order - is the only direct evidence
+  // that the queue carried on past one that failed. Published at the single point
+  // every apt command passes through, on both exits, so a reader sees the failure
+  // and the work behind it as two events rather than inferring the second from
+  // packages appearing on disk some time later.
+  fluxEventBus.publish('system:apt-command', { command: options.command, ok: !error });
+
   // this is so this command can be retried by the worker runner
   if (error) throw error;
 
--- a/ZelBack/src/services/systemService.js
+++ b/ZelBack/src/services/systemService.js
@@ -24,10 +24,18 @@ const isArcane = Boolean(process.env.FLUXOS_PATH);
 let syncthingTimer = null;
 
 /**
- * A FIFO queue used to store and run apt commands
+ * A FIFO queue used to store and run apt commands.
+ *
+ * Built complete. A queue that arrives without its worker has to be finished off
+ * later by whoever happens to get there first, and until they do it silently
+ * accepts work it cannot run - so its behaviour depends on call order, and a test
+ * that touches it inherits whatever the last one left behind.
  * @type {fifoQueue.FifoQueue}
  */
-const aptQueue = new fifoQueue.FifoQueue();
+// eslint-disable-next-line no-use-before-define
+const aptQueue = new fifoQueue.FifoQueue({ worker: aptRunner });
+// eslint-disable-next-line no-use-before-define
+aptQueue.on('failed', monitorAptCache);
 
 /**
  * For testing
@@ -130,7 +138,15 @@ async function queueAptGetCommand(command, options = {}) {
 
   const wait = options.wait || false;
   const commandOptions = { command, params, timeout: options.timeout };
-  const workerOptions = { retries: options.retries };
+  // Every worker option the caller set, not just retries. updateAptCache asks for
+  // retainErrors: false so a failed update is dropped rather than left at the head
+  // of the queue; dropping that request here left the queue holding a task it was
+  // told to bin, ready to run again the moment anything resumed it.
+  const workerOptions = {
+    retries: options.retries,
+    retryDelay: options.retryDelay,
+    retainErrors: options.retainErrors,
+  };
   return aptQueue.push({ commandOptions, workerOptions }, wait);
 }
 
@@ -673,9 +689,6 @@ async function monitorSystem() {
   if (isArcane) return;
 
   try {
-    aptQueue.addWorker(aptRunner);
-    aptQueue.on('failed', monitorAptCache);
-
     // don't await these, let the queue deal with it
 
     // ubuntu 18.04 -> 24.04 all share this package
--- a/ZelBack/src/services/systemService.js
+++ b/ZelBack/src/services/systemService.js
@@ -135,11 +135,16 @@ async function cacheUpdateTime() {
 
 /**
  * @param {string} command The command to run
- * @param {{params?: Array, timeout?: number, retries?: number, wait?: Boolean}} options
+ * @param {{params?: Array, timeout?: number, retries?: number, retryDelay?: number,
+ *   retainErrors?: boolean, wait?: Boolean}} options
  *
  * params: params to pass to command
- * timeout: how many seconds to wait (60 default)
- * retries: how many times to retry (3 default)
+ * timeout: how many seconds to wait (180 default)
+ * retries: how many times to retry (queue default, 5)
+ * retryDelay: how long to wait between attempts in ms (queue default, 60000)
+ * retainErrors: keep a failed task and try it again later rather than dropping
+ *   it (queue default, true). Every one of these is forwarded, so an option left
+ *   unset takes the queue's default rather than this function's.
  * wait: should the queue item be awaited
  *
  * @returns {Promise<Object | void>}
--- a/ZelBack/src/services/systemService.js
+++ b/ZelBack/src/services/systemService.js
@@ -522,8 +522,13 @@ async function ensurePackageVersion(systemPackage, requiredVersion, currentVersi
 
     if (!actualVersion) {
       log.info(`Package ${systemPackage} not found on system`);
-      await upgradePackage(systemPackage);
-      return true; // Package was installed/upgraded
+      // upgradePackage answers whether it FAILED. This branch used to discard that
+      // and report the install regardless, so an apt-get that could not find the
+      // package read back the same as one that installed it. The upgrade branch
+      // below has always read it.
+      const installError = await upgradePackage(systemPackage);
+      if (installError) log.error(`Package ${systemPackage} could not be installed`);
+      return !installError;
     }
 
     log.info(`Package ${systemPackage} version ${actualVersion} found`);
--- a/ZelBack/src/services/systemService.js
+++ b/ZelBack/src/services/systemService.js
@@ -37,6 +37,12 @@ let syncthingTimer = null;
 const aptQueue = new fifoQueue.FifoQueue({ worker: aptRunner });
 // eslint-disable-next-line no-use-before-define
 aptQueue.on('failed', monitorAptCache);
+// The queue has stopped handing this one back. Worth saying out loud: the caller
+// took its error long ago, so without this the work is simply never done again and
+// nothing ever mentions it.
+aptQueue.on('abandoned', ({ options, error, cycles }) => {
+  log.error(`Giving up on apt-get ${options.command} after ${cycles} attempts: ${error.message}`);
+});
 
 /**
  * For testing
--- a/ZelBack/src/services/systemService.js
+++ b/ZelBack/src/services/systemService.js
@@ -14,6 +14,7 @@ const log = require('../lib/log');
 const serviceHelper = require('./serviceHelper');
 const syncthingService = require('./syncthingService');
 const fifoQueue = require('./utils/fifoQueue');
+const fluxEventBus = require('./utils/fluxEventBus');
 const daemonServiceUtils = require('./daemonService/daemonServiceUtils');
 
 const isArcane = Boolean(process.env.FLUXOS_PATH);
@@ -548,7 +549,7 @@ async function monitorSyncthingPackage() {
           minSyncthingVersion,
         );
 
-        if (upToDate) return;
+        if (upToDate) return false;
 
         // The sources changed at version 2.0.0 from stable, to stable-v2
         const hasNewSources = serviceHelper.minVersionSatisfy(
@@ -560,7 +561,7 @@ async function monitorSyncthingPackage() {
           const updated = await updateSyncthingRepository();
           if (!updated) {
             log.warn('Failed to update syncthing repository sources, skipping syncthing upgrade');
-            return;
+            return false;
           }
         }
       }
@@ -576,13 +577,18 @@ async function monitorSyncthingPackage() {
         log.info('Syncthing upgraded, restarting to load new binary...');
         await syncthingService.systemRestart(null, null).catch(() => { });
       }
+
+      return upgraded;
     };
 
-    await versionChecker();
+    const upgraded = await versionChecker();
 
     syncthingTimer = setInterval(versionChecker, 1000 * 60 * 60 * 24); // 24 hours
+
+    return upgraded;
   } catch (error) {
     log.error(error);
+    return false;
   }
 }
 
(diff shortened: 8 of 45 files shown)


Cabecinha84 (commented):
PR #1781 — review

  Verdict: ACK with one small change requested. The production surface is small, well-reasoned, and the three apt-queue
  fixes are real bugs with real tests. I found one defect (harness-only impact, but its guard test is green for the
  wrong reason) plus a handful of notes. Nothing here justifies blocking.

  Note the base is fix/restart-pacing-crash-only (#1780), so this merges after that one.

  What I verified myself

  - Ran the six touched unit files with NODE_CONFIG_DIR=tests/unit/globalconfig: 189 passing, 0 failing (systemService
  69, fifoQueue 18).
  - npx eslint clean on the four changed production files.
  - Config keys stats / pricing / mongodb don't collide with existing top-level keys; all five services that now read
  them already require('config'); zero dangling references to the removed geolocation.statsApiBaseUrl.
  - Traced every caller of the apt path — monitorSystem (serviceManager.js:164, not awaited) is the only entry point;
  nothing in routes or other services reaches queueAptGetCommand.
  - Wrote a throwaway probe test to check the one claim I doubted (below).

  Finding 1 — request a change

  system:packages-checked is not published when a check throws, and the test that says it is doesn't test that.

  systemService.js:735 awaits Promise.all(Object.values(checks)) inside the try. Any rejection skips straight to catch,
  so fluxEventBus.publish never runs.

  The guard test — "publishes even when a check throws, so a waiter fails rather than hangs" (systemService.test.js) —
  does axios.get.rejects(...), but monitorSyncthingPackage's axios call carries its own .catch() that swallows it.
  Nothing ever throws, so the test passes without exercising the path it names.

  Probe (stubbing serviceHelper.runCommand to reject, so getPackageVersion genuinely throws):

  publish called times = 0
  AssertionError: expected +0 to equal 1

  Impact is confined to the harness — fluxEventBus.publish is a no-op unless testEventStream is set, so production is
  unaffected. But waitForPackagesChecked would burn its full 180s (480s in suite 94) instead of failing fast, which is
  precisely the failure mode restartFluxosAndAwaitRecovery was written to eliminate. Worth noting fileOperationRecovery
  got this right — its early-return path still publishes, and there's a test proving it. The two are inconsistent.

  Fix is one line either way: Promise.allSettled, or move the publish into a finally. And make the test throw somewhere
  the code doesn't already catch.

  Notes — not blockers

  --no-install-recommends changes behaviour on every existing legacy node. It lands in aptRunner, so it applies to all
  apt invocations, not just the new ones. The justification is solid (21 packages installed for 4 requested; ten nodes
  in D-state running update-mime-database), and the four packages involved are all safe without recommends. The 84/84
  gate exercised a real unseeded install. Just flag it in release notes — it's the change most likely to surprise
  someone operationally.

  addWorker now throws instead of no-op'ing. Behaviour change on a public method of a shared util. Only remaining call
  site is a test, so it's contained — but it's a hard failure where there used to be a silent one.

  systemService now attaches its worker and failed listener at module-load time, unconditionally, including on Arcane
  nodes where monitorSystem early-returns. I confirmed nothing reaches the queue there, so it's inert. Still:
  require('systemService') now installs an apt worker as a side effect.

  The "config is tamper-evident, environment is not" rationale is stated more strongly than the mechanism supports. It
  appears in cloudUIUpdateService.js and again in scripts/update-cloudui.sh's header. node-config honours NODE_CONFIG,
  NODE_CONFIG_DIR and custom-environment-variables.js — which is exactly how the harness overrides these same keys. So
  moving a literal into config makes those eight endpoints redirectable by anyone who controls the process environment,
  which a hardcoded string was not. That's a fine trade (env control ≈ game over anyway, and testability is a genuine
  win), but I'd soften the comment so nobody later leans on it as a security property.

  Nits: scripts/update-cloudui.sh still documents Usage: npm run update:cloudui and bash scripts/update-cloudui.sh —
  both now exit 1. And queueAptGetCommand's JSDoc doesn't mention retryDelay/retainErrors now that they're forwarded,
  and still says retries defaults to 3 (the queue default is 5).

  Possible flake: the sinkhole resolver refuses at 300 ms. Only names Docker's embedded resolver can't answer locally
  reach the stub, so the window is small — but a slow relay would be recorded as an escape attempt and returned NXDOMAIN
  to the node. Low risk, worth knowing when a suite-95 failure eventually shows up.

  What's genuinely good

  The fifoQueue work is the strongest part. I re-read runWorker line by line: the cycle counter on props[2] is sound
  (push() builds [payload, resolve], the list getter maps x[0], clear() reads task[1] — the third slot is free), the
  retain-to-back plus maxRetainCycles ceiling correctly bounds what was an unbounded retain-and-resume loop, and
  emitting commandOptions rather than the wrapper is what makes monitorAptCache's opening shortcut fire for the first
  time. The retainErrors: false forwarding is the actual fix for the 55k-attempts spin, and it's the kind of bug that
  only bites once the network closes.

  The two tests that encoded the old behaviour were re-baselined rather than deleted, with the reasoning written down —
  that's the right call and it's visible in the diff.

  exec → execFile with an argv array is a strict improvement, and the test deliberately stubs config with a
  non-production value so it can't pass by coincidence.

MorningLightMountain713:
Thanks — the requested change and the notes are all addressed, with one I'd push back on. Branch is now `f6c6bc399`.

**Finding 1 — `system:packages-checked`.** Fixed, and the guard test now throws somewhere the code does not already catch, so it exercises the path it names.

**`--no-install-recommends` — release note added** to the body, calling out that it lands in `aptRunner` and so applies to every apt invocation on every existing legacy node.

**`addWorker` throwing — answered, not changed.** The throw fires only when a worker is already installed, and the old behaviour was a silent wrong answer: the caller believes it installed a worker, the queue runs a different one, and nothing observable differs. It also has no production call site at all any more — both remaining callers are its own tests. I've left the method in place; an unused method isn't the same as an unwanted capability.

**Module-load side effect — fixed** (`f84f57abe`), but by moving the whole queue rather than just the worker. The existing comment was protecting something real: a queue that arrives without its worker silently accepts work it cannot run, so its behaviour depends on call order. Splitting them would have recreated exactly that. The queue and its worker are now built together on first use, so there is no moment when one exists without the other, and importing the module does nothing. There's a test proving the import is inert that fails if construction is made eager again.

**Doc nits — fixed** (`2cc65b851`), plus a third error neither of us had spotted: the JSDoc also said `timeout` defaults to 60 seconds where `aptRunner` uses 180. `retryDelay` and `retainErrors` are documented, and the numbers are named as the queue's defaults since that is whose they are. `update-cloudui.sh` now documents the two invocations that actually work.

**The tamper-evidence rationale — I don't think this one holds, and I'd rather not soften it.** The premise is that node-config honours `NODE_CONFIG`, `NODE_CONFIG_DIR` and `custom-environment-variables.js`. This codebase closes all three, in the first eight lines of `app.js`:

```js
process.env.NODE_CONFIG_DIR = `${__dirname}/ZelBack/config/`;
// ... NODE_CONFIG is the same door: the config package merges whatever
// JSON it holds over every file, after the directory is settled, so leaving
// it open redirects any endpoint without touching a hashed file - the one
// redirect tamper detection cannot see.
delete process.env.NODE_CONFIG;
```

The directory is pinned, `NODE_CONFIG` is deleted, and there is no `custom-environment-variables.js` in the repo. The comment even names the precise attack — a redirect that touches no hashed file — and shuts it.

The harness also doesn't work the way the note assumes, and for the same reason: it *can't* override through the environment, which is why `entrypoint.sh` copies config into the pinned directory instead. Its own comment says so — *"app.js hardcodes NODE_CONFIG_DIR to ZelBack/config/ (cannot be overridden from env — fluxbenchd hashes that directory for tamper detection)"*. Softening the comment would make it less accurate and invite someone to reopen a door that was deliberately closed.

**Possible flake on the sinkhole resolver** — noted, and worth knowing when a suite-95 failure eventually shows up.

---

**Two things worth reporting, because they came out of gating this rather than reviewing it.**

The gate caught a **real regression in this PR**. Moving the runner's overrides out of `NODE_CONFIG` and into a written `local.js` made the entrypoint build that file by *requiring* the per-node config, which spreads `shared.js` at that moment and freezes the result. Discovery autostart was enabled by a `sed` against `shared.js` applied afterwards — so it landed on a file nothing read again, and nodes silently never started discovery. Suites 02 and 04 were both red on it. Fixed in `55545fb5e` by having it travel in the same override JSON as everything else, which leaves one path into a node's configuration instead of two with an invisible ordering rule between them.

Separately, this PR now carries an image-provenance check (`1103448c9`, `e8be03dac`). A stale stub image cost a full 85-suite gate earlier today: a `syncthing-stub` four commits behind had no `/folder-writes-reset` route, so three restore suites died in a `before each` on an HTML 404 that reads exactly like a restore defect. `build-images.sh` now stamps every image with a digest of its build context and the runners refuse to start when it doesn't match the tree — an unlabelled image fails too, which is the difference between a guard and a suggestion. Every gate log now opens with `images verified against the tree`.

Verified: **85 of 85 suites green** at `f6c6bc399`, 62 minutes, `MAXN=6`.

Cabecinha84 (commented):
PR #1781 — review at current head (1db4be766)

  Verdict: request one change. One production regression that your earlier review couldn't have seen — it was introduced
  by a change that was in the version you reviewed, but it needs a probe to surface. Everything you flagged on
  2026-08-10 is genuinely fixed. Nothing else blocks.

  What I verified

  ┌───────────────────────────────────────────┬─────────────────────────────────────────────────────────────────────┐
  │                   check                   │                               result                                │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ Full unit suite, PR head                  │ 5206 passing / 5 pending / 23 failing                               │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ Full unit suite, base                     │ 5190 passing / 5 pending / 23 failing                               │
  │ (fix/restart-pacing-crash-only)           │                                                                     │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ Failure sets                              │ byte-identical — all 23 are the pre-existing dockerService tests    │
  │                                           │ that need the runonflux/website container. Net +16 tests.           │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ Six touched test files in isolation       │ 159 passing, 0 failing                                              │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ npx eslint on the 7 changed production    │ clean                                                               │
  │ files                                     │                                                                     │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ CI                                        │ build (ubuntu-22.04, 20.x, 7.0) passing                             │
  ├───────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ Dangling geolocation.statsApiBaseUrl refs │ none                                                                │
  └───────────────────────────────────────────┴─────────────────────────────────────────────────────────────────────┘

  Your Finding 1 is properly closed (101af676f): the publish moved into a finally, the checks are Promise.allSettled
  with per-check named logging, and the guard test now rejects serviceHelper.runCommand — where nothing catches —
  instead of axios, and pins { installed: [] } so it can't pass on a stale payload. I re-read it; it now exercises the
  path it names.

  The fifoQueue work still holds up on a second read. The props[2] cycle counter is safe (list getter maps x[0], clear()
  reads task[1]), abandoned is bounded, and emit('failed', { options: commandOptions }) is what finally lets
  monitorAptCache's update shortcut fire. The lazy getQueue() (72511b435) is a better answer to your module-load note
  than splitting queue from worker would have been — there's no window where one exists without the other, and import is
  now genuinely inert.

  Finding — request a change

  ZelBack/src/services/systemService.js — moving addSyncthingRepository() after the version check strands a class of
  legacy node permanently.

  addSyncthingRepository() moved from the top of monitorSyncthingPackage to after the hasNewSources branch. But
  updateSyncthingRepository() returns false when /etc/apt/sources.list.d/syncthing.list cannot be read
  (systemService.js:452 — "Unable to read syncthing sources"), and that false short-circuits the whole check:

  if (!hasNewSources) {
    const updated = await updateSyncthingRepository();
    if (!updated) return false;      // ← returns before addSyncthingRepository()
  }
  await addSyncthingRepository();     // never reached

  So on a node with syncthing installed at < 2.0.0 and no apt source file, the repository is never created and syncthing
  is never upgraded. It repeats identically every 24h — the node cannot heal, and the only symptom is one log.warn.

  I ran the same probe against both branches (syncthing 1.23.4 installed, fs.stat rejecting, cat returning empty, stats
  reporting min 2.0.7):

  base:  wrote syncthing apt source: true   (keyring + sources.list.d/syncthing.list written)
  PR:    wrote syncthing apt source: false  (stops at `cat /etc/apt/sources.list.d/syncthing.list`)

  That state is reachable in production: addSyncthingRepository() returns early without writing whenever addGpgKey fails
  (key fetch unreachable, keyring dir not writable) or addAptSource fails. Today those nodes retry daily and recover
  once the network comes back; after this PR they never reach the retry. Same for a node whose syncthing came from
  Ubuntu universe (1.19.2 on 22.04) rather than the FluxOS installer.

  Fix is a one-line move — put await addSyncthingRepository(); immediately after if (upToDate) return false;, before the
  hasNewSources branch. That keeps the entire stated benefit (a node already current writes nothing) and restores the
  old ordering for every node that actually needs to install. It also keeps the new test green, since that test stubs an
  up-to-date node. Worth asking for a unit test on the "1.x installed, no source file" path, since nothing currently
  covers it.

  Notes — not blockers

  1. delete process.env.NODE_CONFIG is an undeclared production change. It landed in 8a0b58715 after your review, in
  app.js, apiServer.js and homeServer.js. The reasoning is sound and I agree with him over your original note — the door
  really was open. But the PR body's production/harness table doesn't list it, and the release note only mentions
  --no-install-recommends. Nothing in the repo sets NODE_CONFIG (I grepped), so in-tree it's inert — but any external
  tooling or Arcane deployment that used it loses its overrides silently. It belongs next to --no-install-recommends in
  the release note.

  2. The gate evidence doesn't correspond to any commit on the branch. Every SHA cited in the body and in his reply —
  5ec88dfbe, 5cb0383b7, f6c6bc399, 55545fb5e, 1103448c9 — is gone. The whole branch was rebased at 17:02 on 2026-08-11,
  after the 15:15 "85 of 85 green at f6c6bc399" comment, and #1780 underneath it moved at the same time. The content is
  almost certainly the same (author dates all predate the comment), but the 85/85 ran against a different base — and
  #1780's replayed commits touch the reconciler burst ceiling and boot-lock admission, which suites 02/04 exercise.
  Worth asking for a re-gate at the current head before merge.

  3. fileOperationRecovery now has the inverse of the bug you found. recoverInterruptedFileOperations publishes
  fileops:recovered after sweepEveryMountedVolume() returns — but executor.reapOrphanedContainers() on that function's
  first line is unguarded. If it throws, no publish, and waitForFileOpsRecovered burns its full 120s. That's exactly the
  failure mode 101af676f fixed in systemService with a finally; the two files are inconsistent again, in the other
  direction. Harness-only impact.

  4. Sinkhole resolver, minor. refuse() calls upstream.close() and is also registered as upstream.on('error'), so an
  error on an unbound socket can throw out of the handler and take the stub down. Harness-only, low likelihood. Files
  under your existing 300ms-flake note.

  5. Stack depth. 1775 → 1778 → 1777 → 1779 → 1780 → 1781 → 1782. Six PRs deep before development. Nothing to act on,
  just merge-order awareness.

MorningLightMountain713:
**The finding — fixed (`71f979d41`).** Verified exactly as traced: `updateSyncthingRepository` answers false on an unreadable sources file, and that false stood between a 1.x node with no file and the file ever being created — one warning a day, forever. `addSyncthingRepository` now runs right after the up-to-date return (a current node is still handed nothing) and before the stable-v2 rewrite, which then finds the file it expects on the next pass. The test you asked for is in, written to fail on the old ordering: 1.x installed, no source file, the sources read failing → the source gets written. One honest sizing note: the population is the residue, not the fleet — the stable-v2 migration ran long ago, so this heals nodes already stuck in exactly this shape (a key fetch that failed in some earlier pass, or a universe install), which is small and was worth one line.

**Note 1 — declared.** The release note now carries the `delete process.env.NODE_CONFIG` change beside `--no-install-recommends`: nothing in-tree used it, external tooling that set it loses its overrides.

**Note 2 — the body now dates every SHA-pinned result** (the branch has been rebased twice since they were taken: over the merged #1777, then onto #1779's review-round commits), and the full gate re-run on the current stack is pending — agreed that it happens before merge.

**Note 3 — fixed (`d98b831f8`).** The inverse you spotted was real: the reap was the one step whose throw skipped the `fileops:recovered` publish, leaving a waiter to burn its whole timeout. Closed with the same `finally` shape as `101af676f`, the throw still propagating for the startup retry, pinned by a test that failed on the old code.

**Note 4 — fixed (`dbe548b6f`).** `answer()` doubles as the upstream's error handler and `close()` throws on a socket that never bound; an uncloseable socket is now left to the garbage collector and the refusal still goes out.

**Note 5 —** three of the six have merged since; the stack is four deep and merges in order.

Unit on the current head: 5,603 passing / 18 pending / 0 failing.

Cabecinha84 (commented):
PR #1781 — review

  Verdict: don't ack yet. One blocking regression, verified against the real modules, on exactly the code path this PR
  is about.

  The rest of the branch is in good shape, so this is one fix away from an ack.

  ---

  Blocking: a failed apt-get update now kills the apt queue permanently

  fifoQueue.js emits failed and then sets halted = true:

  resolve({ error });
  this.emit('failed', { options: commandOptions, error });
  this.halted = true;          // ← clobbers anything the listener did

  That ordering was harmless before, because options.command was always undefined (the payload bug this PR fixes), so
  monitorAptCache's update shortcut never fired and every resume came from the slow recovery path, i.e. after halted =
  true.

  Now the payload is correct, so monitorAptCache takes its shortcut — and that branch is fully synchronous, no await
  before it:

  if (options.command === 'update') { getQueue().resume(); return; }

  resume() runs inside emit, sets halted = false, calls work() (which returns immediately, working is still true) — and
  then this.halted = true on the next line undoes it. The queue breaks out, working goes false, and nothing will ever
  resume it again.

  Reproduced with the real systemService + fifoQueue (stub: apt-get update fails, everything else succeeds):

  after failed update:                  halted=true  working=false  length=0
  a task pushed after the failure:      NEVER RAN    (halted=true   length=1)

  Same scenario on the base branch: halted=false working=true — still churning, i.e. recovery in progress. On this
  branch it is dead for the process lifetime.

  Impact on a legacy node: monitorSystem now awaits its four checks, and addSyncthingRepository/ensurePackageVersion
  queue an apt-get update at boot. One update failure — unreachable mirror, DNS blip, expired repo key, broken sources —
  and:

  - every apt task queued behind it is stranded, and its wait: true caller never resolves;
  - every apt task pushed later is stranded too (push → work() → while (… && !this.halted) exits immediately);
  - Promise.allSettled in monitorSystem never settles, so system:packages-checked is never published and
    chrony/syncthing/netcat never install until FluxOS restarts.

  The harness can't catch it: the sealed fleet serves apt from a local signed repo, so apt-get update always succeeds
  there. The gate passing 84/84 is not evidence against this.

  Fix I verified (all 90 fifoQueue + systemService unit tests still pass, both repros go green):

  resolve({ error });
  this.halted = true;                                   // before the emit
  this.emit('failed', { options: commandOptions, error });
  …
  if (this.halted) { /* retain / abandon */ break; }

  // a listener may have resumed from inside the emit; the ladder is still over
  if (!retriesRemaining) break;

  The second break matters — without it a synchronous resume falls through to await sleep(retryDelay) and stalls the
  whole queue for the default 60s instead. Deferring the emit (setImmediate) also works but interacts with the
  fake-timer tests. Either way it deserves a regression test: fail an update with another task queued behind it, assert
  the second task runs.

  ---

  What I verified and found sound

  - fifoQueue retain-to-back + maxRetainCycles — stress-tested directly. A permanently-failing task no longer starves
    the queue: ran: bad,chrony,syncthing,bad,bad,bad, then abandoned, queue drains, wait: true callers all resolve. The
    counter lives in props[2], which list/clear don't touch. Correct.
  - Endpoint→config migration — all six modules already require('config'); config.github.apiBaseUrl exists;
    geolocation.statsApiBaseUrl has no leftover readers; npm run update:cloudui -- <url> → bash
    scripts/update-cloudui.sh <url> matches, and the script fails loudly with no argument rather than falling back to
    the public internet. execFile vs exec is behaviourally equivalent here.
  - monitorSyncthingPackage ordering fix (71f979d41) — correct. updateSyncthingRepository returns true when the file is
    already stable-v2, so the second addSyncthingRepository() on the outdated path is a no-op, not a trap.
  - delete process.env.NODE_CONFIG — the argument holds. The remaining node-config env paths (NODE_ENV,
    NODE_APP_INSTANCE) only select files inside the pinned, hashed directory, and there's no
    custom-environment-variables.js. .gitignoreing local.js/local-*.js is consistent with that.
  - Unit suite on the current head: 5595 passing, 5 pending, 21 failing — all 21 are dockerService tests needing a live
    Docker daemon (Cannot read properties of undefined (reading 'Id')), environmental, unrelated.
  - Harness — Internal: true doesn't cut the runner off (host→bridge routing survives, no published ports anywhere);
    suite 95's third test genuinely proves the instrument; verify-images.sh treating an unlabelled image as stale is the
    right call.
  - CI green, zero unresolved review threads.

  ---

  Non-blocking, worth raising

  1. fileops:recovered under-reports on the throwing path. result stays {containers: 0, removed: 0} if
     sweepEveryMountedVolume throws after reapOrphanedContainers() succeeded — the event says nothing was reaped when
     something was.
  2. DNS stub has no error handler on the server socket. A failing server.send emits error on a socket with no listener
     → uncaught exception → the resolver dies mid-gate, which would surface as a fleet-wide DNS failure. Cheap to guard
     now that answer() was hardened.
  3. --no-install-recommends is passed to apt-get update too, not just install. Harmless (apt ignores it), but the tests
     now assert it there, which pins a quirk.
  4. addWorker throwing is an API break on a shared util. Fine in practice — the only consumer moved to constructor
     injection — but it's a real behaviour change for anything out of tree.
  5. installed in system:packages-checked includes upgrades, not only installs, since ensurePackageVersion returns true
     for both. Naming only.
  6. The endpoint sweep isn't quite total: watchdogService.js still hardcodes the git URL, paymentService.js a
     home.runonflux.io redirect, and the Azure/Google auth providers their cloud scopes. None are boot-path, so 95 stays
     green — worth naming as a follow-up rather than letting "every endpoint is configuration" stand unqualified.
  7. The --no-install-recommends release note is right to be there. It lands in aptRunner, so it applies to every apt
     call on every existing legacy node. That's the change most likely to surprise operationally, and it's correctly
     flagged.

MorningLightMountain713:
Thanks — the blocker is real and you traced it correctly at every link, including the part that makes it ours rather than pre-existing. Fixed, with a broader fix than the one suggested and one extra defect found alongside it. Of the seven non-blocking notes, two are actioned, four are no-change with reasons, and on one I think three of the four examples are misidentified — details below.

## The blocking finding — fixed at the boundary, not at the branch

Verified independently rather than taken on trust: `runWorker` halts after the emit, `monitorAptCache`'s update shortcut is fully synchronous with no `await` ahead of it, `resume()` sets the flag and calls `work()` which returns at once because the loop is already running, and `push()` only starts a loop when the queue is not already working — so the queue really is dead for the process, with work still in it.

Your causal chain holds too. On `development` the emit carried the raw payload, so `options.command` was `undefined` and that shortcut could never fire. Fixing the payload switched on a trapdoor the payload bug had been holding shut, which makes this a regression of this branch even though the offending line is older.

**Why not the three-line version.** A queue cannot police its listeners. `monitorAptCache` is declared `async`, which reads as a guarantee of lateness and is not one — an `async` function runs synchronously up to its first `await`, and the update branch has none, so the same listener is synchronous on one path and deferred on every other. Rather than depend on which, the queue now settles its own state and then notifies: halt first, emit second. A listener may reply instantly, slowly or not at all and the result is the same.

**Your second `break` was necessary and I took it.** Without it a resumed final failure falls into the retry sleep with no retries left and holds everything behind it for the full delay — 60 seconds by default.

**One you did not mention, in the same re-entrancy family.** `resume()` also did `this.finished = this.work()`. Called from inside the loop, `work()` returns immediately and that already-resolved promise replaces `finished` — which `clear()` awaits to decide the queue is idle before wiping the list. `systemService` calls `clear()` on the apt-is-broken path these very failures reach, so it could empty the queue under a running worker. `resume()` now leaves `finished` alone when the queue is already working.

**Tests.** Four unit regressions, each killed by its own mutation and nothing else: halting after the emit again fails the draining pair, removing the ladder-over break fails only the retry-delay test, removing the re-entrancy guard fails only the `clear()` test. Worth recording that two of those tests proved nothing on their first pass — one used a zero retry delay, so removing the break changed nothing observable, and one asserted on a flag that holds either way. Both were rewritten until they bit.

**And the harness can now catch this class**, which answers your point that it could not. You were right about why: the fleet serves apt from a local signed repository, so `apt-get update` always succeeds there. `createTestEnv({ aptBadSource: true })` adds a source apt cannot reach *alongside* the good one — the update then fails as it does behind a dead mirror, while the work queued behind it stays doable, so a node that survives the failure still provisions itself. Swapping the good source out instead would fail the installs too and prove only that a broken node stays broken.

Suite 94 gains that third regime and asserts it off the event bus rather than by watching for packages to appear: an apt command completing is now a published fact, so the suite waits for the failed update and then for a command *after* it. Proven both ways on a real fleet — 8 of 8 with the fix, and with the halt moved back after the emit all three tests of that regime fail.

## Note 1 — `fileops:recovered` under-reporting: no change

The shape you describe is exactly right, and I traced every `await` in that sweep looking for a way in. There isn't one. `reapOrphanedContainers` catches the container listing failure and returns 0, and catches each removal individually, so it cannot throw at all. `listMountedFilesystems` can throw, but its caller catches it and returns early with the reaped count intact. The per-volume sweep catches per volume by design. What remains unguarded is a `.filter` over a value that is always an array.

So the stale `result` cannot be published. Left alone, because a fix here guards a door with no room behind it.

It did turn up something worth more than the note: a unit test asserted that a failed reap propagates, by stubbing `reapOrphanedContainers` to reject — which the real one cannot do. That test is gone. A green test for a contract nothing can exercise reads as coverage and is worse than no test.

## Note 2 — the DNS stub's socket: fixed, for the bind

Fixed, though scoped more narrowly than the note. The send case is not realistic: a bound UDP socket sending a DNS-sized packet to a container on the same bridge does not fail in any way that would surface — no `EMSGSIZE`, and ICMP errors are not delivered on an unconnected socket.

The bind is the real one. Port 53 already held, usually by the previous run's container on its way out, and without a listener the stub dies on a stack trace naming neither DNS nor the port while every node silently fails to resolve — a fleet-wide fault that is nothing of the kind and expensive to chase into FluxOS first.

One handler covers both, so the send path comes free rather than costing a line. A failed bind is fatal on purpose: a resolver that never bound is not a resolver, and the run should say so at startup rather than eighty suites later.

No test, and I would rather say that than imply one. The harness cannot make a bind fail on demand, and substituting the socket would assert against the substitute.

## Note 3 — `--no-install-recommends` on `update`: no change

Accurate, and it is not alone: `DPkg::Lock::Timeout` is equally inert for `update` — the code's own comment says `update` uses a different lock — and the two `Dpkg::Options` flags are dpkg settings `update` never reaches. Apt ignores all of them.

On the tests pinning it: I would argue that cuts the other way. These commands run as root, and an assertion on exactly what gets executed as root is worth keeping. It is only a nuisance to someone changing the flags, and changing what we exec as root on every legacy node to remove an inert argument is risk without a benefit anyone can observe.

## Note 4 — `addWorker` throwing: no change

Checked for the out-of-tree consumer you were worried about, and there is none: `FifoQueue` has exactly one consumer in the codebase, `systemService`, which passes its worker to the constructor and never calls `addWorker`. The class is not exported from the package and is not an entry point.

The direction is right regardless — silently keeping the first worker leaves the caller believing its own was installed while the queue runs somebody else's, which nothing downstream can detect.

## Note 5 — `installed` includes upgrades: no change

Correct, and naming only, as you say. Leaving it: that field is now the assertion in three harness suites including the new regime, so renaming it is a breaking payload change to make one word more precise. On the case the event exists for — a legacy node's first boot — every entry genuinely is a fresh install. `provisioned` would be the honest word if that payload is ever revised for another reason.

## Note 6 — the endpoint sweep is not total: the heading is now scoped

You are right that the unqualified claim should not stand, and the body heading is now scoped to the boot and app-lifecycle path, naming what is deliberately outside it.

On the specifics, I think three of the four are misidentified, and there is one more you did not name:

- **`watchdogService`'s clone URL** — agreed, a real endpoint and a real gap. On the follow-up list.
- **`paymentService`'s `success_url`** — not an endpoint. It is handed to Stripe so Stripe can redirect the customer's browser after checkout; FluxOS never opens a connection to it.
- **The Google and Azure scopes** — not endpoints either. `.../auth/cloud-platform` and `containerregistry.azure.net/.default` are permission identifiers those providers define, carried inside an auth request and never fetched. A node with a configured scope is a node that fails to authenticate.
- **`login.microsoftonline.com`**, in the same Azure provider, *is* a real endpoint and you did not list it. It joins the watchdog URL as a follow-up.

None are on the boot path, which is why suite 95 stays green with them in place — that suite proves what the fleet reaches while a gate runs.

## One change you did not ask for, which a reviewer will see

The diff removes about 345 lines of unit tests that touched `fluxEventBus`. `testEventStream` is false outside the harness, so `publish()` is inert in a unit run: the ring stays empty and nothing is emitted. Asserting on it there tests a mechanism that does nothing, and stubbing it substitutes for a dependency that already does nothing. Where a test existed only to assert an announcement it went with the assertion; where it also proved something independent, that part stays. `fluxEventBus.test.js` stays too — it tests the utility's own ring and counters rather than using the bus to test something else.

The coverage is relocated rather than lost: an announcement is a fact about a running fleet, and the harness is where one can actually be observed.

## Verification

**Unit: 5,610 passing, 18 pending, 0 failing.** **Integration gate: 87 of 87** at the current head, 59 minutes, all six images digest-verified against the tree they were built from.

Cabecinha84 (commented):
Change 1 — system:packages-checked reports installs that did not happen

  This is the one I'd actually push on, because it undercuts the evidence the branch is built on.

  ensurePackageVersion's not-found branch discards the result:

  if (!actualVersion) {
    log.info(`Package ${systemPackage} not found on system`);
    await upgradePackage(systemPackage);
    return true; // Package was installed/upgraded
  }

  upgradePackage returns Boolean(error) and nobody reads it. The outdated branch below gets this right (return
  !upgradeError).

  Probe — nothing on the system, every apt-get install failing with E: Unable to locate package:

  monitorSystem: settled
  packages-checked payload: [{"installed":["ca-certificates","netcat-openbsd","syncthing"]}]

  Three packages named as installed; none of them are. Only chrony is right, and only because its entry re-asks the
  package database afterwards (!chronyWasPresent && Boolean(await getPackageVersion('chrony'))).

  Suite 94's unseeded regime asserts on this payload, so an install that genuinely failed still reads green. That is the
  same shape as the sinkhole problem you called out yourself — an instrument that cannot see the thing it is asserting
  about. One line (return !(await upgradePackage(systemPackage))), or follow chrony's pattern and re-check the database.

  Pre-existing line, but this PR is what makes it load-bearing.

  Change 2 — drop or narrow 9fe077f14

  352 lines of unit tests deleted across 8 files, landed today, after the last review round. The premise doesn't hold:

  ▎ "testEventStream is false everywhere but the harness, so publish() is inert in a unit run: asserting on it there
  ▎ tests a mechanism that does nothing."

  sinon.stub(fluxEventBus, 'publish') replaces the method before the #enabled check ever runs. Those assertions were
  never testing the bus — they were testing what production code announces, with what payload, at what point. The bus
  being inert is irrelevant, because the stub intercepts the call. Same for the proxyquire-injected stubs in
  volumeExecutor and fileOperationRecovery, which captured payloads and asserted on their contents.

  What that removes:

  - monitorSystem now has zero unit coverage. The describe('monitorSystem tests') block was the only one, and this PR
    rewrites that function outright — the chronyWasPresent probe, allSettled, the installed computation, the finally
    publish. Codecov already flags 5 missing lines in systemService.js. Note that the deleted publishes even when a
    check throws test was the fix for my Finding 1 on 2026-08-10. Also note that if it had still been running it would
    not have caught Change 1 above — but it's the natural home for the test that would.
  - fileOperationRecovery's announces the pass even when the mount table could not be read — the guard for my Note 3
    from 2026-08-24, added in 8c7412e38 and deleted two commits later in the same PR.
  - Coverage of code this PR does not touch: appController's operator-intent contracts (announced while the slot is
    held, before the pass; silent when the write failed), fluxCommunication's withdrawn-vs-claim distinction,
    volumeExecutor's asked vs attempts regression guard — which, by its own comment, exists because that conflation was
    a real bug.

  Two parts of the commit are fine and worth keeping: the no-op stub removals in appInstaller / appSpawner (they
  asserted nothing), and the reap-throws test — I checked reapOrphanedContainers and it genuinely cannot reject, both
  failure modes are caught, so that one was asserting a contract nothing can exercise.

MorningLightMountain713:
One code change and one decline.

## Change 1 — the install that reports itself either way

Fixed in `620549ec9`, and the closing line of your note is the right reading of it: the branch was wrong for as long as it existed, and this PR is what made it load-bearing. `monitorSystem` used to be four `setImmediate` calls with the return values dropped on the floor; collecting them into `installed` is what first gave that boolean a reader.

It now returns `!(await upgradePackage(systemPackage))`, matching the upgrade branch ten lines below, and logs the failure. The unit tests stub the queue rather than driving it — a real failure walks five retries a minute apart, and what is under test is that the boolean says what happened, not the retry ladder. Restoring the old line fails the failure case and leaves the success case green.

Two corrections to what follows from it.

**Suite 94 is not blind to a failed install.** Take your own probe payload — `["ca-certificates","netcat-openbsd","syncthing"]` — and put it through the assertion you say reads green. It fails: `have.members` is set equality, chrony is absent, and ca-certificates is extra. Chrony is absent precisely because it is the entry you correctly identify as honest, so a total install failure cannot produce a payload matching that set. For a partial failure the payload assertion does pass, and the test immediately after it — `ends up with the same packages a seeded node starts with` — reads `dpkg-query` for all three and fails. The third regime does the same after a failed `apt-get update`.

**Nothing in production reads it.** `fluxEventBus.publish` is inert outside the harness, and the only other consumer of that boolean is syncthing's restart, guarded by `currentSyncthingVersion &&` — false exactly when this branch runs. So the value was a report with no reader, not a decision anything acted on. Worth fixing because a function should not lie about what it did; not worth calling evidence the branch rests on.

The real gap your finding points at is larger and is not this PR's: nothing acts on a failed install at all. `monitorSystem` runs once at boot from `serviceManager`, so a node that cannot install ca-certificates retries for twenty minutes, logs one line, and runs without a trust store until FluxOS restarts. That is written up internally, along with the position that a required package which cannot be installed should DOS the node, and the observation that nothing in FluxOS invokes `netcat` at all.

## Change 2 — the deleted assertions

Declining this one, though your reading of the commit message is fair: it argued the tests were inert because the bus is off, and that is not right. `sinon.stub` does intercept, so those assertions executed. That is a bad justification for a change I still think belongs.

The reason they went is simpler than the one the commit gave. `fluxEventBus` is harness observability. `publish` is a no-op outside a harness run, nothing in production consumes an event, and the bus exists so a suite can watch a running fleet from outside. `fluxEventBus.test.js` tests the bus itself and stays. Everywhere else, a stubbed `publish` is a unit test asserting on harness plumbing — it pins that a call was made to a mechanism that does nothing for the product, and it does that in place of testing what the function decided.

`monitorSystem` is the clearest case, and it is not uncovered. Suite 94 exercises it in all three regimes a legacy node boots into — seeded, unseeded, and apt-update-fails — and each one checks the packages with `dpkg-query` rather than taking the node's word for it. That is the same function, tested where the facts are real. It is also what caught the thing Change 1 is about, and a stubbed publish assertion never would have.

On the two you name as yours: both fixes are in the code and untouched. `monitorSystem`'s publication is still in a `finally`, and the recovery pass still announces on every path. What the deleted tests added on top of those fixes was an assertion against a stub.

The part of this I do take is scope. That commit reached into appController, fluxCommunication and volumeExecutor, which this branch has no other business in, and it did so without saying why in the message. The rule it was applying is a codebase-wide one and applying it in one pass is cheaper than three, but it should have been its own change and it should have said so.

Cabecinha84 (approved):
ack

PR #1782, opened 2026-08-11
chore(lint): the tree lints clean, by hand

MorningLightMountain713:
Eighth and last in the stack: `development ← #1774 ← #1775 ← #1778 ← #1777 ← #1779 ← #1780 ← #1781 ← this`. Based on `fix/external-endpoints-configurable` (#1781).

`npm run lint` reported **45 errors across 17 files**. None were introduced by the seven PRs beneath this one — this is pre-existing debt, separated out so it cannot hide a review of anything else.

Fixed **by hand, not with `--fix`**, because three of the categories needed a decision rather than a rewrite.

## What they were

**24 × `prefer-destructuring`.** Fourteen were the same line repeated in `fluxService.js`: `const userconfig = globalThis.userconfig`. The rest are individual. The reassignment cases take the parenthesised form, and only properties whose names already matched could be folded in — neighbouring lines assign under different names (`storedGeolocation = dbData.geolocation`) and are untouched.

**16 × `no-unused-vars`, and they are not all the same thing.** A `const x = sinon.stub(...)` still installs the stub, so only the binding is removed and the call stays. Deleting those lines outright would have quietly changed what the tests exercise — the point of not running `--fix`. Unused `require`s and unused named imports are removed entirely.

**2 × a rule that does not exist.** `// eslint-disable-next-line vue/max-len` on lines of 496 and 1612 characters — eslint reports an unknown rule as an error, so both the disable and the long line were failing. The intent was right and the rule name was wrong; they now name `max-len` and still apply.

**3 × `max-len`.** The same 304-character `sinon.assert.calledWithExactly` repeated verbatim in `systemService.test.js`, now wrapped.

**1 × a dead function.** `fixExplorer` in `explorerService.js` — 46 lines, neither exported nor called anywhere in the tree. Unreachable rather than merely uncalled, so removed; it is in the history if it is ever wanted.

## Verification

`npx eslint --ext .js ./` → **0 errors**.

Every test file touched, plus the tests covering every production file touched, run green: **793 passing** (`appSyncOrchestrator`, `appHashSyncService`, `arcaneAuthService`, `cpuBurstHelper`, `daemonServiceMiscRpcs`, `fluxPeerManager`, `watchdogService`, `networkHealthMonitor`, `systemService`, `fluxService`, `geolocationService`, `geolocationRule`, `globalState`, `upnpService`, `volumeConstructor`, `explorerService`).

One correction caught during the work and worth stating, because it is the argument against `--fix`: an early edit removed `const result = ` from the wrong occurrence — the text was not unique, and the earlier match's `result` was in use. eslint caught it as five `no-undef` errors. It was redone by line number.

### DIFF
--- a/ZelBack/src/services/analyticsService.js
+++ b/ZelBack/src/services/analyticsService.js
@@ -176,7 +176,7 @@ function analyticsMiddleware(req, res, next) {
     return;
   }
 
-  const zelidauth = req.headers.zelidauth;
+  const { zelidauth } = req.headers;
   if (!zelidauth) {
     next();
     return;
--- a/ZelBack/src/services/explorerService.js
+++ b/ZelBack/src/services/explorerService.js
@@ -1806,53 +1806,6 @@ async function reindexExplorer(req, res) {
   }
 }
 
-async function fixExplorer(height = 1670000, rescanApps = true) {
-  try {
-    const dbopen = dbHelper.databaseConnection();
-    const blockheight = serviceHelper.ensureNumber(height);
-    const database = dbopen.db(config.database.daemon.database);
-    const query = { generalScannedHeight: { $gte: 0 } };
-    const projection = {
-      projection: {
-        _id: 0,
-        generalScannedHeight: 1,
-      },
-    };
-    const currentHeight = await dbHelper.findOneInDatabase(database, scannedHeightCollection, query, projection);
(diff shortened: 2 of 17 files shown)


Cabecinha84 (commented):
wrong

Cabecinha84 (approved):
ack

PR #1783, opened 2026-08-13
fix(peers): a peer we just heard from is alive, whatever it sent

MorningLightMountain713:
## The problem

Liveness is credited **only** by a pong:

```js
onPingSent()     { this.missedPongs += 1; if (this.missedPongs >= this.maxMissedPongs) this.terminate(); }
onPongReceived() { this.missedPongs = 0; }   // the only thing that clears it
```

A pong is an ordinary frame. It queues behind everything else that peer sent us, and FluxOS reads that queue on one event loop shared with the API, broadcasts and app messages. So the busier a peer is talking to us, the later its pong is read — and `wsPingIntervalMs ?? 15000` × `wsMaxMissedPongs ?? 3` gives it 45 seconds before we terminate it.

**The heartbeat fires first on the connections we have the most evidence are alive.**

## What it does under load

Measured in `flux-peer-sim`, which models this window exactly. One node dying at 3,000 nodes:

| after the death | messages read | of which pings |
|---|---|---|
| +15s | 107,009 | **13** |
| +30s | 36,870 | **18** |
| +45s | 47,190 | 1,990 |
| +60s | — | **every node terminated all ~82,000 peer connections** |

Scheduling delay peaked at 0.1s — nothing was stalled, the readers processed 245,936 messages in that minute. The pongs were simply behind them.

It does not recover. Terminate → reconnect → the reconnection traffic refills the queues → 60 seconds later it happens again. From a single death.

## The change

`isAlive` also accepts a peer heard from within the window the pong count already allows:

```js
get heardFromRecently() {
  return this.lastMessageMono !== null
    && monotonicMs() - this.lastMessageMono < this.livenessWindowMs;
}
```

- **`livenessWindowMs` is derived**, not a new tunable — `pingInterval × maxMissedPongs`. A genuinely silent peer is dropped on exactly today's schedule; only a talking one is treated differently.
- **`lastMessageMono` starts `null`.** Seeding it at construction would give every new connection a free window in which no missed pong could terminate it. (The existing suite caught this — three tests failed until I fixed it.)
- **Credited before the rate limit.** A frame we decline to process still proves the peer is there; liveness is a question about the peer, not about how much work we accept from it.

Result on the same fleet, kill and seed: evictions from ~250,000 to **349**, socket ratio 1.00× instead of 5.47× and climbing, **with the congestion unchanged**. The queue isn't shortened — it stops being fatal.

## Monotonic time

Elapsed measurements move to a monotonic clock. Wall clock is adjusted by NTP and by hand; a backward step makes an elapsed time negative, a forward one ages a peer out by however far the clock jumped. Either can terminate a healthy connection.

`lastPingTime`, `lastPongTime` and `connectedAt` **stay on the wall clock** — they're reported over the API, where a caller wants a date rather than milliseconds since this process started. `latency` now derives from a monotonic pair instead.

## Tests

Four added. **Three fail against the pong-only behaviour.** The fourth passes there deliberately — it asserts a peer that has sent *nothing* is still terminated, which must not change.

199 passing locally (`NODE_CONFIG_DIR=$PWD/tests/unit/globalconfig mocha tests/unit/fluxPeerManager.test.js --exit`).

## Scope and caveats

- **This is a code-shape finding, not an observed production incident.** I have not seen the fleet do this. I have seen a faithful model of it do it, reliably, triggered by the kind of correlated failure node-down certificates are designed to handle — and certificates will make such storms considerably more common.
- `tests/unit/fluxPeerManager.test.js` has a pre-existing `no-unused-vars` error at the `onclose` test. Present on `development`, untouched here.

### DIFF
--- a/ZelBack/src/services/utils/FluxPeerSocket.js
+++ b/ZelBack/src/services/utils/FluxPeerSocket.js
@@ -1,4 +1,5 @@
 const config = require('config');
+const { performance } = require('perf_hooks');
 const WebSocket = require('ws');
 const log = require('../../lib/log');
 const serviceHelper = require('../serviceHelper');
@@ -10,6 +11,21 @@ let _fluxNetworkHelper;
 function getFluxNetworkHelper() {
   if (!_fluxNetworkHelper) _fluxNetworkHelper = require('../fluxNetworkHelper');
   return _fluxNetworkHelper;
+}
+
+/**
+ * Milliseconds from a clock that only moves forward.
+ *
+ * Every elapsed measurement below uses this rather than Date.now(). Wall clock is adjusted by NTP
+ * and by hand, and a backward step makes an elapsed time negative while a forward one makes a peer
+ * look silent for however far the clock jumped — either can terminate a healthy connection.
+ * Timestamps that are REPORTED stay on the wall clock, because a caller reading connectedAt or
+ * lastPongTime wants a date, not milliseconds since this process started.
+ *
+ * @returns {number}
+ */
+function monotonicMs() {
+  return performance.now();
 }
 
 const CLOSE_CODES = Object.freeze({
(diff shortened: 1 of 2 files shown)


Cabecinha84 (approved):
ack

PR #1787, opened 2026-08-15
Validate an app owner as an identity a signature can verify against

MorningLightMountain713:
## The problem

**One signed update can burn an app permanently, and FluxOS accepts it.**

`owner` is the field that says who controls an app, and FluxOS only ever checked that it was a string — there is no address validation on it anywhere in the service tree.

Set it to a username, or to an address with one character mistyped, and the app is not transferred to anyone. It is destroyed. No private key in existence produces a signature that verifies against `"TrippleCore"`, so the app has no controller at all — not its former owner, not FluxOS, not us. It cannot be updated, renewed, restarted or reassigned by anybody. It keeps running until its paid time runs out, and then it is gone, along with its name and its data.

This is not theoretical. On 2026-08-14 a customer's app `satisfactory1785685226558` had its `owner` changed to the literal string `TrippleCore` by a signed update 27 minutes after registration. The app was healthy with 18 days of paid life left and nobody on earth could touch it.

**It was only recoverable by accident.** The marketplace branch of the update verifier lets the FluxSupportTeam key sign for any app whose name merely *contains* a number greater than `Date.parse('2020-01-01')`. `satisfactory1785685226558` qualifies because of its timestamp suffix, so we could re-sign the owner back. **An app without a big number in its name would have been unrecoverable** — there would have been no path at all.

A sweep of all 1060 live global specs found exactly one bad owner: this app. The nine `0x…` owners are legitimate EVM logins. So one casualty so far, one that happened to be saveable, and nothing stopping the next.

## Why registration is safe and update is not

| message | signature verified against | result for a garbage `owner` |
|---|---|---|
| `fluxappregister` | `appSpec.owner` — the owner **in the message** | impossible: no signature verifies against `"TrippleCore"` |
| `fluxappupdate` | `previousAppSpecs.owner` — the **old** owner | accepted: the new value is never used as a key |

Registration is protected only as a side effect of the owner doubling as the verification key. On update the incoming owner is pure payload — it is inside the signed bytes, so the customer genuinely authorised it, but nothing asks whether the value is an address.

## What this changes

`signatureVerifier.isValidSigningIdentity()` accepts what `verifySignature` can actually verify against: a Flux ID (`1…`, base58, **checksum decoded**) or a `0x` Ethereum address.

That rule is not new — it already existed inline in `idService`'s login path (`:259-272` on `development`) and nowhere else. It now lives once, beside the verifier it describes, and both login sites call it too.

The checksum is deliberate. An address one character off its own is exactly as unsignable as a username and produces the identical unrecoverable state, so shape alone is not enough.

## Why there is no fork height

The check applies to **live submissions only** — the five call sites passing `verifyAppSpecifications(spec, height, true)`, all API paths sitting below a "timestamp over 1 hour old" guard:

- `registryManager.js:1821` (`/apps/appregister`)
- `advancedWorkflows.js:2751` (`/apps/appupdate`)
- `appValidator.js:1388`, `:1456`, `:1611` (the verify-spec endpoints)

The four sites that pass nothing are gossip and chain replay — `messageStore.js:167,175` and `appHashSyncService.js:317,322` — and they are untouched.

A message being submitted right now is not history, so there is nothing to grandfather and no constant to carry forever. Applying the rule to the replay path instead would make a node syncing from scratch reject the height-2826803 message: `appHashSyncService.js` catches per-message, so it counts `failed += 1`, never inserts it, and never marks the hash — leaving that node permanently divergent from every peer that stored the message before this shipped.

Because the verify-spec endpoints are included, the frontend gets the error *before* it prompts for a signature. A user can no longer sign a message that orphans their app.

The residual is that a modified node could still gossip a bad owner in. That only lets someone brick their own app.

The third parameter of `verifyAppSpecifications` is renamed `checkDockerAndWhitelist` → `liveSubmission`, since it now gates two things and the old name would be wrong about one of them.

## Testing

Full ZelBack unit suite: **4883 passing, 0 failing, 18 pending**.

New `tests/unit/signatureVerifier.test.js` covers the predicate, including that a hex public key is rejected. `verifySignature` does accept one — but that branch exists for node-to-node messaging, where a FluxNode's identity is its collateral pubkey, and no login or app owner can ever be one.

Five new `appValidator` cases. The one that matters is the last: an owner of `"TrippleCore"` **still validates** when replayed without the live flag. That is the regression guard for chain history — if someone later moves this check, that test goes red.

Fixtures in `appValidator.test.js` and `idService.test.js` moved off placeholder strings (`1owner`, `1Z1234341Z…`) onto a synthetic but checksum-valid address. They were fake data that never represented an identity.

## Not in this PR

The marketplace fallback at `messageVerifier.js:269-278` is untouched. Its test for "is this a marketplace app" is only *does the app name contain a number greater than `Date.parse('2020-01-01')`*, which is most timestamp-suffixed deploys — genuinely too broad. But it is currently the only recovery path for an already-orphaned app, and it is what recovered this customer. It should not be narrowed without a replacement.

The equivalent fix for v1–v9 specs in `flux-spec` is written separately, on the same reasoning.

### DIFF
--- a/ZelBack/src/services/appRequirements/appValidator.js
+++ b/ZelBack/src/services/appRequirements/appValidator.js
@@ -8,6 +8,7 @@ const daemonServiceMiscRpcs = require('../daemonService/daemonServiceMiscRpcs');
 const fluxCommunicationMessagesSender = require('../fluxCommunicationMessagesSender');
 const registryManager = require('../appDatabase/registryManager');
 const messageVerifier = require('../appMessaging/messageVerifier');
+const signatureVerifier = require('../signatureVerifier');
 const imageManager = require('../appSecurity/imageManager');
 // const advancedWorkflows = require('../appLifecycle/advancedWorkflows'); // Moved to dynamic require to avoid circular dependency
 // eslint-disable-next-line no-unused-vars
@@ -1217,11 +1218,12 @@ function checkComposeHWParameters(appSpecsComposed) {
  * Validates specs including hardware requirements, architecture compatibility, and Docker compliance
  * @param {object} appSpecifications - Application specifications to validate
  * @param {number} height - Block height for validation context
- * @param {boolean} checkDockerAndWhitelist - Whether to check Docker, whitelist, and architecture requirements
+ * @param {boolean} liveSubmission - Whether the spec is being submitted now through the API, rather than
+ * replayed from a message already on chain. Gates the checks that only hold for a spec being accepted today.
  * @returns {Promise<boolean>} True if validation passes
  * @throws {Error} If validation fails (e.g., incompatible architectures, missing requirements)
  */
-async function verifyAppSpecifications(appSpecifications, height, checkDockerAndWhitelist = false) {
+async function verifyAppSpecifications(appSpecifications, height, liveSubmission = false) {
   if (!appSpecifications) {
     throw new Error('Invalid Flux App Specifications');
   }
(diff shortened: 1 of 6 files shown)


Cabecinha84 (approved):
ack.

My only question if we merge this, is what will happen to that app that the appowner have updated and put wrong app owner on it for the new nodes on the network that are syncing the apps?

MorningLightMountain713:
> ack.
> 
> My only question if we merge this, is what will happen to that app that the appowner have updated and put wrong app owner on it for the new nodes on the network that are syncing the apps?

I changed it back to the existing owner. The network will also still accept the "bad" update as this change only affects the ingestion points, so they will still accept the gossip / sync path

PR #1791, opened 2026-08-28
Hosting an app is not owning it: the node operator loses every app-scoped verb, a specification decrypts for its owner alone, and the check that decides who is asking is rebuilt underneath

MorningLightMountain713:
## Overview

Every app-scoped endpoint asks for `appownerorfluxteam` or `appowner`. The node operator is admitted to none of them, and `appownerabove` — the privilege that admitted them — is removed from the tree: the dispatch case and the verifier function are gone, so the string authorises nobody rather than waiting to be reached for again.

Underneath that, the machinery a route uses to decide who is asking has been rebuilt: a privilege is a named member rather than a spelled string, the check takes the identity it is checking rather than the request it arrived in, and no handler decides whether to authorise from whether an argument was supplied.

**This targets `development` directly.** The three PRs it was stacked behind — #1780, #1781 and #1782 — have all merged. It lands with RunOnFlux/fluxos-frontend#151, which is what stops the UI offering controls that now refuse.

## Why the operator is not a party to a customer's app

Hosting an app is not owning it.

On run state: an app cannot exceed what was bought — `dockerService` sets `NanoCPUs` and `Memory`/`MemorySwap` on the container from the spec — so an app inside its allocation is spending cycles the operator sold, and an app outside one is a containment defect to fix in the limits rather than to paper over on a single node with a button. The operator is paid whether the container runs or not, so a per-app stop withholds the service and keeps the payment. Stopping FluxOS forfeits the payment along with the obligation, which is what makes it the honest lever.

On reading: hosting is a reason to know what an app **costs** you, which `/apps/appsresources` answers unauthenticated and in aggregate. It is not a reason to read what is inside it.

## Decrypting a specification is the owner's alone

Both decrypt paths ask for `appowner`. Neither the node operator nor the flux team decrypts a customer's specification over the API; the team decrypts out of band, which keeps a decryption a deliberate act by a named person rather than a side effect of opening a page.

This changes how a specification is read. It is not a claim about everything a flux team caller can reach on a node they support.

A partly-redacted specification is neither usable nor safe: an update composed from one writes its blanks back over the customer's app, and the UI cannot tell a withheld field from an absent one.

## What the flux team may know instead

`GET /apps/appcomponentnames/<app>`, gated `fluxteam`: an app's component names, whether each is election managed, and its resource totals. Deliberately not specification-shaped, so nothing can mistake it for a spec or submit it as one.

The two fields are two different claims. `resources` is information v9 already treats as public — it keeps the totals outside the encrypted envelope so a node can judge whether it can host an app without reading it, and binds them into the AAD so a relayer cannot understate them. `components` is not: v9 seals the component list and publishes only a count, so handing over the names is an exception granted to an authenticated flux team caller on a node that already holds the plaintext, because the container tools address a component as `<component>_<app>` and cannot function without them.

Whether a component is election managed is answered by `mountParser.isGComponent`, the classifier the election itself uses. A sync flag counts only on the primary mount, so a substring search over `containerData` and the election disagree on a spec that carries `g:` anywhere else.

## `/apps/appexec` matches the terminal it duplicates

`appExec` moves from `appowner` to `appownerorfluxteam`. This widens it, deliberately.

The socket.io container terminal reaches the same component with more than `appExec` does — an interactive session, on a user the caller names. A caller admitted there is not meaningfully refused one non-interactive command, and the narrower gate described a boundary that was not there. The node operator is admitted to neither.

That terminal now authorises through `verifyPrivilege` like every other caller rather than reaching the underlying verifier directly, so it carries the privilege it requires and a search for who may open a shell in a customer's container finds it.

## A privilege is named, and names only what it reads

The most-used privilege in the tree was spelled `adminandfluxteam` and is an OR: it admits the node operator, the flux team or flux support, any one of them. At 105 call sites it read as the strictest gate in the system while being the widest of the non-user gates.

Every privilege is now a member of a frozen enum, and every compound reads `_OR_`, because none of them requires two identities at once. The **values are unchanged**: three of them are what `/id/checkprivilege` answers to clients and the frontend branches on those strings, so the names are ours to fix and the vocabulary on the wire is not. Those four response tokens are now named and pinned by a test rather than repeated as literals.

The check takes the value of the `zelidauth` header rather than the request that carried it. It read one field and could reach every other; the two callers that have no request had built one to fit. A privilege that resolves an identity is refused an app name, so a call cannot read as app-scoped when it is not.

## Authorisation does not depend on how the caller arrived

Sixty handlers decided whether to check by whether an argument had been supplied — six on the request, fifty-four on the response — so a caller who supplied neither was authorised by default. Three of the six run an `npm` script on the node.

They are pairs now, the shape already used elsewhere in the tree: the operation takes no request and consults nobody, and the endpoint beside it authorises before calling it. Where a handler only ever serves a request, the check is simply unconditional. The internal callers reach the operation, which is the expression they already used.

## A handler that fails answers the caller

Every route was registered as an arrow that invoked its handler and discarded the promise it returned, so express never saw a rejection and could not answer for one. All 491 registrations now hand it over, and a failure reaches express's error handler and answers `500` — the status text alone, since this branch also pins the node to run as a production server.

## No route that checks a privilege is served from a cache

`apicache` keys an entry on the request URL alone and answers from it before the handler runs. A privilege-checked route behind one therefore hands the first caller's response to whoever asks next without checking them, and whatever the handler would have *done* never happens — including the routes that delete a session row.

The cache is removed from every route that checks a privilege; the rest keep it. `routeWiring.test.js` takes that from the sources rather than from a list: it parses `routes.js` for every registration, resolves each handler to the module and function behind it, and parses that module for the functions which reach `verifyPrivilege` — following one function calling another within the module, which is the shape the `Api` halves use. **278 routes check a privilege and none of them is cached.** A list of paths would have been as good as the last person to update it, and it was already one short.

## A container listing answers what a container is doing

`/apps/listrunningapps` and `/apps/listallapps` returned docker's container object with three keys deleted. They return a view built from three of docker's own fields instead — `Names`, `State`, `Status` — keeping docker's names and forms. The response holds nothing else, and a field docker adds in a future version is absent by default rather than carried until someone notices.

An application's image, entrypoint, ports and build metadata belong to its specification. A listing answers what a container is *doing*; it is not a second route to what an application is made of, and a response that is constructed rather than filtered cannot quietly become one at some later exit.

`Names` stays docker's array, verbatim. FDM builds `/flux{component}_{app}` and matches `Names[0]` exactly, failing closed when it does not — its own comment records the consequence, *"no master found, dropped from haproxy"* — so a bare name, or one with the prefix stripped, would take every `g:` app out of routing.

`/flux/info` embeds the same listing and carries the same view, from the same function. A projection applied at one exit and not the other is the shape this avoids.

Callers inside this process keep the container objects: the spawner, the installer, the reconciler and resource accounting read `Names` and a count, and are unaffected. Split per this branch's convention — `listRunningApps()` and `listAllApps()` take no request and no response, and the `Api` halves own the express object.

The endpoints keep their cache, because their response no longer varies by caller — which is the property that makes this work rather than the caller-varying alternatives, none of which can sit behind a URL-keyed cache.

`/apps/listappsimages` answers the flux team. An image row names no application — nothing in it attributes a tag back to one, and the field that would, `Containers`, reads `-1` because docker does not compute it for a plain list — so there is no smaller answer available and the list is served whole or not at all. Gated rather than removed: nothing in this repository calls it, but a public endpoint on every node may have callers this repository cannot see, and a 401 is something they can act on where a 404 reads as a broken node. Its cache goes with the check.

These endpoints keep their names. Several under `/apps` say "apps" and mean containers or images; renaming a v1 route is a 404 for every caller, which is a louder break than a changed response and would land stacked on one. The misnaming is recorded in `routes.js` instead, and v2 gets the right names free in a new URL space.

Consumers were checked: the frontend's app list and manage page, FDM on `origin/main`, and the palworld and minecraft sites all read `Names`, `State` and `Status` and nothing more. RunOnFlux/fluxos-frontend#151 reads a component's image from the application's own specification, which the page already holds.

## Which code a node runs is the flux team's to choose

`/flux/entermaster` and `/flux/enterdevelopment` ask for `fluxteam`. They asked for `adminandfluxteam`, so this is a capability the node operator had and no longer has, and it is deliberate. The two endpoints that report where a node sits — `/flux/currentbranch` and `/flux/currentcommitid`, both new here — ask for the same.

The line is between updating and choosing. An operator still calls `/flux/updateflux`, `/flux/softupdateflux`, `/flux/softupdatefluxinstall` and `/flux/hardupdateflux`, all of which move their node along the branch it is on. Which branch that is, and which commit it sits at, is not theirs to set or to read.

`/flux/rebuildhome` is removed and `/flux/rebuildui` replaces it, with no alias. It is the only path this branch removes. The old name described building the UI on the node, which is not what happens — it is fetched as a published release — and `npm run homebuild`, the script the old endpoint ran, has not been in `package.json` for months, so every caller has been receiving an error rather than a rebuild. It refuses on ArcaneOS, where the watchdog owns CloudUI, and cannot run twice at once: the script removes the served directory before copying the new one in.

## Answers a caller can act on, and totals a node can act on

Five defects the work above put within reach, each fixed here.

**A cache never remembers a failure, in either shape a failure takes.** Handing every route's rejection to express turned a crash into a response, and apicache stores whatever the handler produced — so one transient rejection was served to every later caller for the whole window, up to a day on the longest of these, with a restart the only way to clear it. The store now takes `200` and nothing else: an allowlist, so a status nobody predicted is excluded rather than kept, and global, so a route added next month starts out right.

That covers what express writes, and a handler here does not write those. It catches its own error and answers `200` with `status: 'error'` in the body, which no status rule can see: `/benchmark/getstoredbenchmark` caches for an hour and reports no benchmark data for as long as a booted node has not benchmarked yet, so the node went on saying it for an hour after the benchmark landed. `/daemon/help` did the same whenever the daemon was still coming up. The payload is recorded on `res.locals` as the handler answers it — the only point at which it exists — and read back by apicache's own toggle, which it evaluates when the response ends. Public arguments, not internals. Refusing failures rather than admitting successes, and deliberately the opposite of the status rule beside it: a cached route may answer a payload this node did not build, and syncthing's carries no status field at all.

**A branch switch asks the question `git checkout` will answer.** `rev-parse --verify <branch>` never resolves a remote-tracking ref, so it refused a switch the node could make. And where the branch is genuinely absent — the installer clones `--depth 1 --single-branch`, so a node carries no reference to any branch but its own — it is fetched into a local branch and given its upstream by hand, which leaves the clone shallow and single-branch as the installer set it. The depth is passed only where the repository already is shallow, because a `--depth` fetch turns a full clone shallow and a legacy node would otherwise be truncated by a branch switch. Verified on an ArcaneOS node against a pristine clone.

**Every peer this node decides to drop leaves the map on the decision.** `evict()` had one caller; three other paths decided a peer should go and called `close()` alone. The gossip send-failure path is the sharp one, because it is reached exactly when `close()` can do nothing: `send()` returns false only for a socket that is not open, and `ping()` skips one, so the missed pong that would terminate it is never counted either. Until ws times the close out the peer holds a slot no reconnect is dialled for and is still offered as a sync source. That loop no longer iterates the peer map live, either — it awaits between sends, so the map is free to change under it from any path.

**The UI rebuild refuses where the UI is not ours, and never runs twice at once.** The script removes the served directory before copying the new one in. On ArcaneOS the watchdog owns CloudUI, which is why the periodic check stands aside there; reaching the script directly walked under that. It is still unconditional otherwise — repairing a CloudUI that is damaged rather than out of date is what it is for.

**An enterprise app is counted for what it reserves, not for nothing.** Its components are emptied in the local record — they are the customer's decrypted configuration — and the resource totals were counted straight off that record, so every enterprise app on a node counted as zero cpu, zero memory and zero disk. Those totals are what a node subtracts from its capacity before accepting more work, so a node holding them believed it was emptier than it was. The specifications are decrypted before counting, through the cached path. An app that still cannot be read is named rather than left silent, and the spawner stands down rather than taking on more — only the spawner, because refusing on the redeploy paths would freeze every other application over one unreadable one. **The endpoints are unchanged**: `/apps/appsresources` and `/flux/info` publish the same three totals, projected, and the list of unreadable apps does not leave the process.

## A socket listener that fails answers its own client

`asyncRoute` hands every HTTP route's failure to express. The socket.io side needed the same rule and did not have it: it was written down in comments in both handlers and enforced by hand three different ways, across nine listeners, four of which were covered.

socket.io calls a listener and discards the promise it returns, so a listener that throws produces a rejection nobody handles — it reaches `apiServer`'s uncaughtException handler and exits the process. Every listener registered through `FluxSocketIoServer` is wrapped once now: a throw or a rejection answers that one client and is logged, and nothing else on the node is disturbed. The same bargain `asyncRoute` makes for a route, and deliberately **not** the process-level `unhandledRejection` handler proposed earlier here and rejected — that one swallows every failure everywhere and leaves the node running in a state nobody chose. Removal by identity is translated through a `WeakMap`, so a handler that takes a listener off still takes off the one that was registered.

The container terminal needed it most. Its `exec` listener used two client-supplied arguments before reaching its own try block, and that namespace takes no middleware — the five arguments are whatever an unauthenticated client serialised, and nothing upstream makes them strings the way node's http parser does for a header. Both are checked at the top of the listener now, and answer what is actually wrong rather than a generic terminal failure.

**The terminal's session belongs to the connection, not to the message that opened it.** The stream, the analytics pair and the `cmd`, `resize` and `disconnect` listeners were all declared inside the `exec` listener, so every message brought its own copy and registered its own listeners on the shared socket — a client sending `exec` twice got two shells, and one keystroke written into both. Four listeners are registered once, at connection, and route to whatever session exists. A second terminal is refused because the session is occupied, which is a consequence of the session existing rather than a guard someone remembered to add, and every way the setup can fail gives the claim back — so the refusal can only ever be told to a caller that really has one.

## A handler that only ever serves a request stops asking whether one arrived

96 handlers ended `return res ? res.json(X) : X;` while reading `req.params` or `req.query` on their first lines, so a call without a request threw long before the hedge could matter. `routes.js` is their only caller and it always passes both, so the branch described an internal caller that does not exist — and it survived because the unit suite used it to read a return value, exercising a path production never takes. 171 of them are gone, and the tests that reached them pass a response and assert through `res.json` now.

Six keep theirs, hedges and tests both: `getBlock`, `getRawTransaction`, the three `getSingleAddress*`, and `viewDeterministicFluxNodeList`. Each has a real internal caller passing a fabricated request and no response, so there the hedge is the mechanism rather than the decoration. Removing one of those is what broke every internal `syncthingService` caller earlier on this branch, in silence, and it took a fleet run to find. They want the `Api` split that fixed that file, which is its own change.

A test was deleted only where an exact twin already asserted the same thing through `res.json`, decided by canonicalising both bodies and comparing them line for line rather than by matching titles — which is what caught the near-pairs a title match would have destroyed, including one where a parameter passed as the string `'false'` is the only coverage of the boolean coercion. Those were converted. So was everything else.

## The syncthing surface is not finished here

Two routes under `/syncthing` change privilege on this branch. `postDbIgnores` asks for `fluxteam` because `.stignore` decides what leaves the node for an app the operator does not own, and `/syncthing/config/gui` asks for it on both the read and the write, because syncthing's gui block carries the apikey that authenticates every syncthing call on the node — confirmed on a node rather than assumed. The operator can still read that key from their own disk; what goes is the network path to it, which needed only a session and no shell, and on ArcaneOS the operator has no shell at all.

**That is two routes, and the family is larger.** A number of `/syncthing` endpoints authorise nobody. It is pre-existing — the base does not gate them either — and this branch does not change it.

It is designed and scoped rather than left — the endpoints are inventoried, the one that is peer protocol is identified, and the privileges are chosen — and it ships as **its own PR in the release after this one**. Separate deliberately: gating a peer-protocol endpoint by mistake breaks syncthing across the network, so it needs its own fleet gate rather than a corner of this one. Anything gated there loses its cache with the check, and `routeWiring.test.js` derives that rule rather than reading a list, so it cannot be forgotten.

**Three smaller ones.** The app name reaches `/apps/appcomponentnames` either way its handler reads it — the route was registered without the optional marker, so the query form its own handler implements answered 404 while every sibling accepts both. Every git command in `fluxService` now runs in the node's own checkout rather than inheriting whatever directory the process was started in. Two of them only read; one writes. And a syncthing device read that cannot complete no longer withholds the folder list the same pass already read: the two are separate reads answering different questions, and a node that had read its folders was telling every peer asking "not ready" for as long as the device read kept failing.

## Also here

The node runs as a production server. `app.js` pins `NODE_ENV`, alongside the config directory it already pins — Express returns the exception stack instead of the status text to any caller otherwise. `NODE_CONFIG_ENV` is pinned beside it to the deployment node-config already resolves to, so the config directory's single file is still what loads and nothing is printed at startup. Assigned rather than defaulted: the environment is not hashed, so a value read from it changes how the node answers without touching anything tamper detection can see. `tests/unit/appEntry.test.js` loads the real entry point in a child process seeded with hostile values for all four variables and asserts what it leaves set.

`appTamperingDetectionService.js` carried two raw NUL bytes as the separator in a rollup key. They are escapes now — byte-identical at runtime, and the file is text again rather than classified as binary, which was making every `grep` over it return an empty result successfully.

## Verification

**Unit suite: 5,790 passing, 18 pending, 0 failing.** Lint clean.

**Full integration gate green: 89 suites, 0 failures.** Run on a dedicated 16-core box against images built from this branch and checked to carry it before the gate was allowed to start.

Every gate is pinned by a test asserting the privilege that route asks for, with `calledOnceWithExactly` so a second and wider check added beside it fails too. What the privilege admits is pinned on the other side of the hop in `verificationHelperUtils.test.js`, which asserts the node operator is refused by `verifyAppOwnerOrFluxTeamSession` **and** admitted by `verifyNodeOperatorSession` — the check that defines the node operator — so neither assertion can pass vacuously.

`tests/unit/privilegeCallShape.test.js` takes the shape of every privilege check from the sources rather than from a run, so it covers each of them rather than the subset a runtime test reaches: the privilege is a member, the second argument is the identity the request carries, an app name travels only as `{ appName }` and only where one is read, nobody reaches a verifier behind `verifyPrivilege`'s back, and no check is conditional on an argument being present. Anything it cannot classify fails rather than being skipped.

**Mutations run individually against every gate**, and against each pinned line added here: every one killed exactly its own test and nothing else. Reviving the `appownerabove` dispatch case fails the test that asserts the string authorises nobody. Readmitting the node operator to `verifyAppOwnerOrFluxTeamSession` fails the refusal test. Putting the cache back on a privilege-checked route fails the route-wiring assertion. Reverting any registration in `routes.js` to the form that discards its handler's promise fails the route-wiring assertion that reads all 491 of them out of the source, and names the ones that dropped it. Each of the call-shape assertions was killed by reverting one call site to its previous form.

`backupRestoreService` has a test file.

**Exercised live on a node running this branch, against a registered encrypted app**: as the flux team, component names and resource totals arrive and the per-component tools work; as the app's owner, the full decrypted specification renders, including a value that exists only inside the encrypted blob.

The fleet suite that drives the four run-state verbs now asserts the operator is refused all of them, with the container checked untouched afterwards and the owner's stop accepted on the same endpoint on the same node as the control.

## Deployment order: frontend first

The two are not symmetric.

**New UI against old FluxOS degrades.** The names endpoint 404s, so the flux team sees the encrypted state without resource totals and without the per-component tools until that node updates. Nothing breaks.

**Old UI against new FluxOS breaks.** The old page attempts the decrypt for a non-owner, is refused, and abandons the specification rather than falling back to what is public — so the flux team gets an unpopulated page on every enterprise app.

`cloudUIUpdateService` pulls the latest frontend release at boot regardless of the node's FluxOS version, so the two roll independently and the order is decided by which release is cut first. Cut the frontend, let it propagate, then release FluxOS.

Any tooling that drives an app-scoped endpoint with a node's own key stops being authorised by this change, and is refused explicitly with a 401 rather than failing quietly.


Cabecinha84 (commented):
Must fix before merge

  1. masterSlave uses the loose check the repo already retired. registryManager.js:857:

  masterSlave: Boolean(component.containerData && component.containerData.includes('g:')),

  mountParser.getComponentSyncMode/isGComponent exists precisely to replace this, and its own docblock
  (mountParser.js:408-415) names containerData.includes('g:') as the thing it replaces — it matches g: in an invalid
  non-primary position ('/data|g:/db'). Four call sites already use the canonical helper
  (advancedWorkflows.js:4178,4184,4238,4240, appUtilities.js:1150,1153,1168), and appReconciler.js:270 carries a comment
  explaining why the loose form is wrong. New code shouldn't reintroduce it. One line: masterSlave:
  mountParser.isGComponent(component.containerData || '').

  2. "It was the only route that was both cached and privilege-checked" is false. There are ~35+. The one that matters:
  /id/loggedsessions — cache('30 seconds') at routes.js:536, verifyPrivilege('user') inside, and it returns the calling
  zelid's own session rows (zelid, loginPhrase, createdAt, expireAt). URL-keyed cache ⇒ caller A's session list is
  served to caller B for 30 s. Also cached-and-checked: /flux/peerhistory, /flux/restart,
  /apps/tamperingevents/:appname, and the whole /syncthing/system/* + /syncthing/metrics set (all
  adminandfluxteam/fluxteam).

  This is pre-existing and out of scope for #1791 — but the claim should come out of the PR body, and /id/loggedsessions
  deserves its own issue. It's the same bug class the PR just fixed, one route over.

  Worth an answer before ACK

  3. The confidentiality half of the rationale isn't delivered. The PR argues the flux team must not read a customer's
  spec over the API — "the team decrypts out of band, which keeps a decryption a deliberate act by a named person." But
  the same fluxteam identity keeps, at appownerorfluxteam:

  - /apps/appinspect → dockerContainerInspect returns dockerode's object whole, including Config.Env — which is
    populated from environmentParameters (dockerService.js:317,1018). That is exactly the field the removed redaction
    blanked.
  - /apps/applog, /apps/apptop (argv), /apps/downloadfile, /apps/downloadfolder.
  - The socket.io container terminal — which this PR moves to appownerorfluxteam, i.e. it keeps the team's shell.

  And the REST twin /apps/appexec is appowner-only (appInspector.js:740), so after this PR the team gets an interactive
  shell over socket.io but is refused the same capability over HTTP.

  The update-safety argument for killing the redacted spec — blanks composed into an update get written back over the
  customer's app, and the UI can't tell withheld from absent — is sound on its own, and I'd take the change on that
  basis alone. But the "the team does not read a customer's spec over the API" framing overstates what actually changes.
  Either narrow the terminal to appowner to match appExec, or restate the rationale as update-safety.

  4. Operator policy — product sign-off, not code review. This removes every app-scoped control and read from the node
  operator on their own hardware: stop/start/restart/kill, logs, inspect, stats, process list, file browse/download,
  backup/restore, redeploy, terminal. What's left is /apps/appsresources (aggregate, unauthenticated) and stopping
  FluxOS. The argument is well made, but: an operator diagnosing which tenant is hammering my disk now has no per-app
  lever, and any third-party operator tooling authenticating with node keys breaks silently. Needs ops/product sign-off
  and a release note, not just a merge.

  One residual to check against the policy: the operator keeps /syncthing/system/browse and the syncthing folder config
  (adminandfluxteam), which enumerate app directories and <component>_<app> folder ids on the node — the very thing
  /apps/appcomponentnames restricts to fluxteam. Intended, or the next PR?

  5. Ordering is a hard gate. Frontend RunOnFlux/fluxos-frontend#151 must ship and propagate first (old UI + new FluxOS
  = unpopulated page on every enterprise app for the team). And this sits on a stack of three still-open PRs: #1780 ←
  #1781 ← #1782 ← this. It cannot merge before them.

  Nits (non-blocking)

  - verifyPrivilege('fluxteam', req, mainAppName) (registryManager.js:829) passes an appName the fluxteam branch ignores
    — reads as app-scoped when it isn't.
  - volumeSession.test.js:141 uses calledWith, not the calledOnceWithExactly the rest of the PR sets as the standard — a
    second, wider check added beside it would still pass.
  - operatorStopped / waitForOperatorIntent / "operator stop" naming across the reconciler is now a misnomer: the
    operator can no longer produce one. Suite 32's own filename included.

MorningLightMountain713:
Thanks — this was a useful review. All five are closed, and the fixes for two of them went wider than the report. Taking them in order.

## 1. `masterSlave` uses the loose check — fixed

`registryManager.js` now reads `mountParser.isGComponent(component.containerData || '')`. Right call: that field is what the election reads, so the classifier deciding it has to be the one the election itself uses, or a spec carrying `g:` in a non-primary position is election-managed by one and not the other.

## 2. The "only route" claim — the claim is gone, and so is the cache on every route that checks a privilege

The sentence has come out of the PR body. Rather than fix the routes a list names, the rule is now derived from the sources: `routeWiring.test.js` parses `routes.js` for every registration, resolves each handler to the module and function behind it, and parses that module for the functions reaching `verifyPrivilege`, following one function calling another within a module. **278 routes check a privilege and none of them is cached** — asserted, with three further assertions so it cannot pass on an analysis that resolved nothing.

`/id/loggedsessions` is uncached and closed here. It does not need its own issue.

You have the mechanism right, and it is not only the read: a URL-keyed cache answers *before* the handler, so for a route behind one, whatever the handler would have **done** never happens either. Both directions are closed by removing the cache rather than by reasoning about which routes are cheap enough to leave.

The same pass fixed a second fault in the cache. Handing every route's rejection to express turned a crash into a response, and apicache stores whatever the handler produced — so one transient rejection was served to every later caller for the window, up to a day on the longest of these. The store now takes `200` and nothing else, globally, so a route added next month starts out right.

## 3. The confidentiality rationale — restated, and the asymmetry removed

Both, rather than either. The PR body no longer claims more than the change delivers: it says this changes how a specification is read, and states plainly that it is not a claim about everything a flux team caller can reach on a node they support. The update-safety argument stands on its own, as you say.

And `/apps/appexec` now asks for `appownerorfluxteam`, matching the terminal it duplicates. You were right that the two disagreed; the fix is to widen the narrower one, because a caller admitted to an interactive session on a user they name is not meaningfully refused one non-interactive command. The socket.io terminal also now authorises through `verifyPrivilege` like every other caller rather than reaching the verifier directly, so a search for who may open a shell in a customer's container finds it.

## 4. Operator policy — decided, and the residual is scoped

The policy is the PR's premise and it is settled: hosting an app is not owning it. An operator is paid whether the container runs or not, so a per-app stop withholds the service and keeps the payment; stopping FluxOS forfeits both, which is what makes it the honest lever. The aggregate cost of what a node is carrying stays available unauthenticated.

Third-party tooling that drives an app-scoped endpoint with a node's own key is refused with a 401 rather than failing quietly, and that is called out in the PR body under deployment. We are moving fast here — a caller on an endpoint that has moved updates to the one that replaced it.

On the syncthing residual: two routes move on this branch. `postDbIgnores` asks for `fluxteam`, because `.stignore` decides what leaves the node for an app the operator does not own, and `/syncthing/config/gui` asks for it on both the read and the write, because syncthing's gui block carries the apikey authenticating every syncthing call on the node — confirmed on a node rather than assumed. The rest of that family is designed and scoped as a PR stacked on this one, not left open: `SYNCTHING_ENDPOINT_AUTHORISATION_DESIGN.md`. It is separate because at least one of those endpoints is peer protocol and gating it would take syncthing down across the network, so it needs its own fleet gate rather than a corner of this one.

## 5. Ordering — agreed, and already the plan

Frontend first. RunOnFlux/fluxos-frontend#151 ships and propagates before this, and the reasoning is in the PR body: new UI against old FluxOS degrades, old UI against new FluxOS breaks. `cloudUIUpdateService` pulls the latest frontend release at boot regardless of FluxOS version, so the two roll independently and the order is decided by which release is cut first. And yes — this is fourth in the stack and merges after #1780, #1781 and #1782.

## Nits

**`verifyPrivilege('fluxteam', req, mainAppName)`** — closed structurally rather than at that call site. A privilege that resolves an identity is now refused an app name, so a call cannot read as app-scoped when it is not, and `privilegeCallShape.test.js` takes that from the sources across every call site rather than from a run.

**`volumeSession.test.js`** — now `calledOnceWithExactly`, so a second and wider check added beside it fails too. That is the standard across the PR's gates.

**`operatorStopped` naming** — kept deliberately, and documented at the setter. The distinction the name draws is a *person* deciding a component stays down against the reconciler, the master/slave election or crash recovery bringing it back up; that reading is still correct now the node operator is not one of the people who can ask for it. The field is persisted and queried by name, which makes renaming it a migration rather than a rename.

## Also landed since the review

**An enterprise app is counted for what it reserves, not for nothing.** Chasing your point about `/apps/appsresources` turned up a live bug on production nodes rather than a wrong sentence. An enterprise app is stored locally with an empty `compose` — those components are the customer's decrypted configuration — and the resource totals were summed straight off that record, so **every enterprise app on every node counted as zero cpu, zero memory and zero disk**, including the fixed 10GB and 2GB swap per component. That total is what a node subtracts from its capacity before accepting more work, so a node holding them believed it was emptier than it was and kept taking work it could not run. Observed on a live node. Pre-existing — both halves are byte-identical on the base. The specifications are decrypted before counting, through the cached path, and the endpoints publish exactly the same three totals as before.

**Every peer this node decides to drop leaves the map on the decision.** `evict()` had one caller; three other paths dropped a peer with `close()` alone, and the gossip send-failure path is reached exactly when `close()` can do nothing.

**`/flux/enterdevelopment` works on nodes where it never has.** The installer clones `--depth 1 --single-branch`, so a node holds no reference to any branch but its own and could not check one out. Verified on a real ArcaneOS node against a pristine clone.

**`/flux/rebuildui` refuses where the UI is not ours and cannot run twice at once**, and every git command in `fluxService` now runs in the node's own checkout rather than inheriting whatever directory the process started in.

**`/apps/appcomponentnames` accepts the app name either way its handler reads it** — it was registered without the optional marker, so the query form 404'd while every sibling accepted both.

**A syncthing device read that cannot complete no longer withholds the folder list the same pass already read.** A node that had read its folders was telling every peer asking "not ready" for as long as the device read kept failing, and nothing bounds that.

Unit suite 5,790 passing, 0 failing; lint clean. A full integration gate runs on the final tip before this merges.

MorningLightMountain713:
Full integration gate green on this tip.

**`63104ae62` — 90 suites, 0 failures.** Run on a dedicated 16-core box against images built from this branch and checked to carry it before the gate was allowed to start. 437 assertions across the per-suite TAP output, none failing.

Two things landed since the reply above, both in the harness rather than the product.

**The fleet suite covering the syncthing folder/device split is verified by mutation.** It drives a five-node fleet, takes a node's device configuration down while its folder read keeps answering, restarts it so the failing read lands before the node has ever published, and asserts it publishes the folders it read anyway. Reverting the fix out of the image turns that suite red on exactly that assertion and nothing else; the recovery case stays green either way. The outage is read from the stub's own count of the reads it refused rather than from the node's log, so it cannot pass on an outage that failed to apply.

**The gate runner refuses on a host bar it cannot read.** Its two sysctl blocks raise a kernel pool the fleet depends on and abort when they cannot. Given a threshold that was not a number they did neither — `[ "$x" -lt "$y" ]` returns 2 on a non-numeric operand, and with `pipefail` but no `-e` the `if` reads that as false, so the raise and the abort inside it were both skipped and the gate ran on exactly the host it exists to refuse. Each refusal now carries its own status and is written to the run's log rather than only to the terminal.

Also: the unit suite no longer depends on how fast the machine running it is. Two tests left work running after they finished — a sixty-second sampling interval and a geolocation retry that reschedules itself every ten seconds — and both wrote through the real logger into whichever test happened to be running when they woke. Sixteen assertions across three files count logged errors exactly, so the suite was green or red on the host's speed. Both pre-existing; both closed.

Unit suite 5,790 passing, 18 pending, 0 failing. Lint clean.

The same-address port collision work is now open as #1795, stacked on this branch.

Cabecinha84 (commented):
The one thing to fix before ACK

  The "every route hands its failures to express" invariant is not pinned by any test.

  The PR body states: "Returning a route registration to the form that discards its handler's promise fails two of the
  three assertions written for it." It does not. I reverted two real registrations to (req, res) => { handler(req, res);
  } and all 72 tests across routeWiring, routes and routeGuards passed.

  asyncRoute is unit-tested as a function, and used in synthetic express apps inside those files — but nothing asserts
  that routes.js's 491 registrations use it. All 491 do today (I checked: 491 registrations, 491 asyncRoute(). So there
  is no live defect — only no guard against the next route added without it.

  This matters because it is exactly the failure routeWiring.test.js was written to prevent, in its own opening words:
  "a guard is worth exactly what it is wired to." The fix is a few lines against the recordRouteTable() the file already
  builds.

  Notes (not blockers)

  1. Pre-existing, but worth naming loudly: 31 /syncthing GET routes remain unauthenticated and cached — including
     /syncthing/db/browse/:folder, which returns a directory listing of any app's synced volume to any anonymous caller,
     and /syncthing/db/file. Unchanged by this PR (31 at base, 31 at head) and explicitly scoped to a follow-up. But
     given this PR's entire thesis is who may see inside a customer's app, I'd make that follow-up a release blocker
     rather than "next".
  2. workstreams/fluxos-core-cutover/SYNCTHING_ENDPOINT_AUTHORISATION_DESIGN.md is cited by path in the body but is not
     in this tree. Reads as if it lands here.
  3. The cache section slightly overstates itself. statusCodes:{include:[200]} only stops express's own 500s from being
     cached — which is precisely the new risk asyncRoute introduced, so it's correctly scoped. But handlers here answer
     200 with {status:'error'}, so /apps/listrunningapps (15s) and /flux/info (60s) still cache error bodies exactly as
     before. No regression; the heading "A cache never remembers a failure" just isn't literally true. (The body does
     admit this further down.)
  4. dockerTerminalHandler now passes the client-supplied zelidauth straight into verifyPrivilege, which throws
     TypeError for anything not a string. Caught by the listener's try/catch, so it fails closed — but a client sending
     the object form (which ensureObject used to accept) gets 'Error opening a terminal.' instead of 'Not authorized.'.
     test-infra sends a string; worth one glance at the frontend's terminal client before release.
  5. getApplicationComponentNamesAPI keeps the return res ? res.json(...) : ... shape this branch is otherwise removing.
     Harmless (its check is unconditional) but inconsistent for an API-only function.
  6. fluxService.js:429–430 — double blank line left by the const cwd removal. Lint doesn't flag it here.
  7. Scope creep: the mongo ensureIndexes batching in serviceManager.js and the NUL-escape fix are unrelated to the
     subject. Both safe (batch failure falls back to per-index, byte-identical key), but they widen an already very
     large PR.

MorningLightMountain713:
Thanks — this was a good review. The blocker is fixed, all seven notes are closed, and two of the notes were covering real defects rather than the nits they looked like. Taking them in order.

## The blocker: the invariant was not pinned, and you are right

Reproduced before fixing anything. I reverted two real registrations to `(req, res) => { handler(req, res); }` and ran all three files: **72 passing, 0 failing.** Nothing noticed.

The reason is narrower than "no test exists". `routeWiring.test.js` walks each registration looking for the module member expression behind it, and it searches *inside* each chain argument — so `asyncRoute(svc.foo)` and a bare arrow calling `svc.foo` resolve identically. The wrapper is invisible to the one test that reads `routes.js`.

`routeWiring.test.js` now records, per registration, whether the last thing in the chain is a call to `asyncRoute`, and asserts the list of registrations that are not is empty — naming any that fail. The AST walk that already read `routes.js` for the privilege assertions moved out beside `recordRouteTable()` so there is one reader rather than two. Mutation-checked: reverting those same two registrations turns exactly one test red and prints

```
these discard the promise their handler returns
-[ "get /daemon/getinfo", "get /daemon/getfluxnodestatus" ]
```

**The PR body sentence was also wrong as you read it**, and your reading is the natural one. It was true of the three `asyncRoute` unit tests and it sat one paragraph after "all 491 registrations now hand it over", which invites exactly the wrong conclusion. Rewritten to say what it is now true of.

## 1. The syncthing routes

Agreed, and it was already inventoried rather than waved at — the endpoints are enumerated, the one that is peer protocol is identified, and the privileges are chosen. Confirmed unchanged by this branch: the same count at base and at head.

It ships as its own PR **in the release after this one**, not as a corner of this one. Gating a peer-protocol endpoint by mistake is a syncthing outage across the network, so it needs its own fleet gate. That is a deliberate scheduling decision rather than a deferral: this PR is already very large, the defect predates it, and holding a finished change hostage to a pre-existing one is the wrong trade.

## 2. The design document cited by path

Removed. It was worse than reading as in-tree — it is a path in a private repository, so it named an internal document that no reader here can open. The paragraph now describes the state of the work without citing anything.

## 3. The cache section overstating itself

Correct, and we closed the gap rather than softening the heading.

You are right that `statusCodes: {include:[200]}` only stops express's own 5xx, which is precisely the risk `asyncRoute` introduced. What it cannot see is a handler that catches its own error and answers `200` with `status: 'error'`. The worst of those is longer than the routes you cited: **`/benchmark/getstoredbenchmark` caches for an hour**, and reports no benchmark data for as long as a freshly booted node has not benchmarked — so a node went on saying it for an hour after the benchmark landed. `/daemon/help` does the same whenever the daemon is still coming up. The six `1 day` caches are clean; they read local config and have no error path at all.

`cache()` now records the payload on `res.locals` as the handler answers it, and passes apicache a `middlewareToggle` that refuses to store a body reporting an error. Public apicache and express API, no internal fields. It moved into `routeGuards.js` with the `apicache.options` call so the whole rule lives in one place; the 152 call sites are unchanged. Refusing failures rather than admitting successes is deliberate and the opposite of the status rule beside it — a cached route may answer a payload this node did not build, and those carry no status field.

Three assertions, each mutation-killed separately: dropping the toggle, dropping the payload recorder, and emptying the status allowlist each turn exactly their own test red.

## 4. The terminal's zelidauth

I did the glance you asked for and the frontend is fine: it sends `localStorage.getItem('zelidauth')`, a string or `null`, and `verifyPrivilege` accepts both. The object form has no caller.

But the question was worth asking, because looking at it found something else. That namespace takes no middleware, so all five arguments are whatever an unauthenticated client serialised — and unlike an HTTP header, nothing upstream makes them strings. The listener used two of them before reaching its own try block. A throw there is a rejection in an async socket.io listener, which nobody handles: socket.io calls the listener and discards the promise, so it reached `apiServer`'s uncaughtException handler. Both are checked at the top of the listener now and answer what is actually wrong.

Underneath it, every listener registered through `FluxSocketIoServer` is wrapped once, so a failing socket listener answers its own client and is logged instead of reaching the process handler. That rule was already written in comments in both socket handlers and enforced by hand three different ways across nine listeners, four of which were covered. It is deliberately not the process-level `unhandledRejection` handler proposed earlier on this branch and rejected — this is bounded to one socket, the same bargain `asyncRoute` makes for a route. Five mutations, each killed by its own test.

And the terminal's session now belongs to the connection rather than to the message that opened it. The stream, the analytics pair and the `cmd`/`resize`/`disconnect` listeners were all declared inside the `exec` listener, so a client sending `exec` twice got two shells and one keystroke written into both. Four listeners registered once at connection; a second terminal is refused because the session is occupied. The two hard-won races in that file — a disconnect landing mid-setup, and pairing the analytics open with a close — each have a named test that goes red when its mechanism is removed, and the second of those had no test before.

## 5. `getApplicationComponentNamesAPI` keeping the `res ?` shape

Right, and it was not one function. A static sweep found **104** that take a request they always dereference and still hedge on `res` — 98 with no caller but `routes.js`, so the branch is unreachable. `getApplicationComponentNamesAPI` was the worst of them because this branch wrote it, in the shape this branch removes.

**171 of those branches are gone**, across 96 handlers, and the unit tests that reached them now pass a response and assert through `res.json`, which is what production does.

Six keep theirs deliberately, hedges and tests both: `getBlock`, `getRawTransaction`, the three `getSingleAddress*`, and `viewDeterministicFluxNodeList`. Each has a real internal caller passing a fabricated request and no response, so there the hedge is the mechanism, not decoration. Removing one of those is what broke every internal `syncthingService` caller earlier on this branch, silently, and it took a fleet run to find. They want the `Api` split that fixed that file, which is its own change. I also swept for callers an alias-based analysis would miss — destructured requires, computed access, spread re-exports — and found none.

A test was deleted only where an exact twin already asserted the same thing through `res.json`, decided by canonicalising both bodies and comparing line for line rather than by title. That caught several near-pairs a title match would have destroyed, including one where a parameter passed as the string `'false'` is the only coverage of the boolean coercion; those were converted instead.

## 6. `fluxService.js:429–430`

Fixed. The same function was also answering `res.json(...)` unconditionally two lines above the two that hedged, so both went with it.

## 7. Scope creep

Fair description, and both stay. The index batching and the escape fix were asked for on this branch deliberately rather than split off. Noted rather than argued.

---

**State.** This now targets `development` directly — #1780, #1781 and #1782 have all merged. The PR body has been rewritten for the corrections above.

**A full 94-suite integration gate is underway** on the tip of the branch stacked above this one, which contains this branch in full, so it is the first end-to-end coverage of everything here. Unit suite on this tip: 5,716 passing, 18 pending, 0 failing, lint clean. I will report the gate result here when it lands.

Cabecinha84 (approved):
ack

PR #1793, opened 2026-08-30
Refuse app installs onto ports another Flux node on the same IP already holds

TheTrunk:
## Problem

Where several Flux nodes share one public IP — a common multi-node setup — FluxOS can install an app
onto ports a sibling node already publishes. The router forwards each port to exactly one node, so
any install after the first is unreachable from the moment it starts, and nothing reports it: the
container binds fine in its own network namespace, healthchecks pass, and the instance is broadcast
to the network as live.

Both guards miss it:

1. **`ensureApplicationPortsNotUsed`** resolves a sibling's ports from its broadcast specification. A
   v8 enterprise app's `compose` is blank by design, so it contributes no ports and is dropped by
   `if (appPorts.length > 0)` — every enterprise app is exempt. (`assignedPortsInstalledApps` is
   per-node by nature: separate database, separate docker daemon.)

2. **`checkAppAvailability`** probes with `isPortOpen`, a bare TCP connect. It cannot tell the
   requester's test server from anything else answering at that address, so the sibling's existing
   app answers and the check passes — *because* of the collision.

## Fix

Two independent layers; either one closes the observed case.

**Identity in the port test.** `FluxHttpTestServer` echoes a per-run nonce; the peer fetches the port
over HTTP and verifies it (`isPortOwnedBy`). "The port is open" becomes "the port reaches me". A
requester that sends no token gets the old behaviour, and a tokenless test server still answers a
bare TCP connect — older nodes unaffected in both directions.

**Sibling scan.** For each other Flux node on our own IP, read `/apps/listrunningapps` and treat
published `PublicPort`s as taken (`ensureSameIpPortsFree`). Works for enterprise apps because docker
reports real host ports regardless of spec redaction — no decryption, no protocol change, no new
disclosure. Best effort with a 5s timeout, since layer 1 still has to pass.

## Verification

```
sibling holds the port : isPortOpen=true  isPortOwnedBy=false   old passes, new rejects
our own test server    : isPortOwnedBy=true                     free port still installs
legacy (no token)      : isPortOpen=true                        older peers unaffected
```

The `portManager` and `fluxNetworkHelper` unit suites were **not run** — they require MongoDB, which
was unavailable here. `node --check` passes on all five files. Please run
`npm run test:zelback:unit` before merging.

### DIFF
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -441,6 +441,9 @@ module.exports = {
     portTestPropagationDelayMs: 10000,
     portTestPeerTimeoutMs: 30000,
     portTestMaxAttempts: 5,
+    // Scan of other Flux nodes sharing our public IP. Short on purpose: these are LAN-adjacent
+    // hops, and a sibling that does not answer promptly is skipped rather than delaying an install.
+    sameIpScanTimeoutMs: 5000,
     spawnReconfirmDelayMs: 7500000,
     nonEnterpriseSpawnDelayMs: 120000,
     globalCmdDelayMs: 500,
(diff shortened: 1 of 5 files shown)


MorningLightMountain713:
superseeded by #1795

PR #1784, opened 2026-08-13
feat(dos): a residential node holds, evacuates, and only then goes to DOS

Cabecinha84:
A node on a residential connection is only fit to serve the network when it runs ArcaneOS. This puts such a node out of service in three stages, so that the destructive step is only ever reached on a node holding nothing.

```
HOLD      stops accepting NEW apps. Immediate, and it deletes nothing.
EVACUATE  gives up one app at a time, and only ones another host demonstrably holds.
DOS       once it runs nothing, and removeAllAppsLocally finds nothing to delete.
```

DOS >= 100 is not a mark: `nodeStatusMonitor` calls `removeAllAppsLocally` on it and `appStartupManager` does the same on every boot, so a node in that state has every app directory and volume on it deleted. Reaching it only on an empty node is the whole point of the staging.

**There is no in-FluxOS off switch, and the one that exists is better.** Enforcement acts only on a `RESIDENTIAL` verdict from the published table, so blanking `orgClasses` in `fluxos-network-policy` ends it fleet-wide within a day — no release, no node upgrades, nothing to roll back. A config flag would have needed all three.

**The trust boundary this inherits, stated rather than inherited silently.** The table is fetched unsigned from `fluxos-network-policy`; trust is TLS, GitHub, and write access to that repo, and this PR moves what those bytes can do from "placement quality" to "stage a non-ArcaneOS residential node out of service". That is a conscious trade. It is bounded — at most the 541 non-Arcane slots with readable bench, one app per 6 hours behind the full safety gate, DOS only on an empty node, self-healing on the corrected row — and it replaces a weaker root, not a stronger one: today's classification rests on plaintext `http://ip-api.com`, forgeable per node by any on-path party with no history to audit, while every byte the policy repo has ever served is in its git history behind branch protection (no force-push, admins bound, validated on publish — though merges do not yet require a second reviewer). Signed manifests (the fluxhashes scheme, in progress) are the successor; this artifact adopts that signing when it lands.

## What it acts on

Measured against `stats.runonflux.io/fluxinfo` (6,115 slots on 2,513 machines, 5,822 slots with readable bench) and `api.runonflux.io/apps/locations` (6,340 placements, 1,085 apps), classified by reading the published artifact through this branch's own `ipLocationStore` and `networkClassifier`.

Enforcement acts only on a `RESIDENTIAL` verdict from the published table. Of the 541 slots on 316 hosts whose bench confirms they are not ArcaneOS:

| verdict | hosts | slots | share | app placements |
[...]

PR #1795, opened 2026-08-31
Two Flux nodes on one public address can install onto the same port, and nothing refuses it

MorningLightMountain713:
## Overview

Several Flux nodes commonly share one public address — four behind one home router is an ordinary setup. Each is its own machine with its own docker daemon and its own application database, so each binds a host port successfully and nothing local objects. The router forwards that external port to exactly one of them.

Every other node's application is then unreachable from the moment it starts, **and nothing reports it**: the container binds fine inside its own network namespace, its healthchecks pass, and the instance is broadcast to the network as a live one. The owner is paying for instances that cannot be reached and the node believes it is hosting them.

Two things close it. The port test asks the peer to bring back what the port actually said, and checks it against a secret only this node could have put there. And before any of that, the other Flux nodes at this address are asked which ports they hold.

**Targets `development`.** #1784 stacks on top of this.

## Where the gap actually is

The install path already refuses this on a UPnP node, before anything is bound: `checkInstallingAppPortAvailable` maps each port before it binds its test server, `mapUpnpPort` returns false when the gateway rejects the mapping, and the caller throws.

| Node | Before | After |
|---|---|---|
| UPnP, sibling holds the port | the mapping fails and the install throws before binding | unchanged — already refused |
| **Statically forwarded** | `isUPNP()` is false, so no mapping is attempted and nothing fails | **refused** |

So the case being closed is the statically forwarded node. Two further weaknesses are closed with it. `ensureApplicationPortsNotUsed` resolves a sibling's ports from that sibling's globally broadcast specification, and a v8 enterprise `compose` is **sealed, not absent** — the ports are in there and a node can open them. Nothing on that path decrypted, so **every enterprise app read as holding no ports at all**. It decrypts now. And `checkAppAvailability` could not tell who answered.

## The port test asks for something only this node could have supplied

This node opens a `FluxHttpTestServer` on each port and asks a peer to test them. The peer connects, something answers, and it reports success — but *"something answered"* is not *"I answered"*. Behind a shared address the router forwards each port to one node, so the thing that answered can be a neighbour's application, and the peer cannot tell the difference because from outside there is none. The check passed **because of** the collision it exists to catch.

So the test server answers with a random token, generated per test run, and the peer is asked to **read each port and hand back what it found**. The comparison happens here.

| what comes back | meaning | verdict |
|---|---|---|
| our token | only this node could produce it | proven, proceed |
| something else | another application is on that port | refuse, **once a second peer agrees** |
| the peer names a port it could not reach | one peer's report about our reachability | refuse, **once a second peer agrees** |
| the peer will not answer at all | nothing about our ports — a stale node list, a clock, a key | ask another peer |
| no reading offered | the peer is on older code | ask another peer; see below |

**One witness to accept, two to refuse.** Our own token coming back is *proof* — a secret cannot be manufactured, so one peer settles it and a second adds nothing. Anything else is one observer's *report* about a third party, and a report can be wrong: a truncated read, something in the path, a peer having a bad moment. Refusing on the first of those leaves a node that quietly installs nothing until somebody goes looking. So a second peer is asked before refusing, spending a retry budget that already existed and was only ever used on peers too old to answer. A real collision corroborates itself for free — the router forwards that port to one node, so every peer that reads it sees the same thing — while an anomaly does not reproduce on a different peer.

**Running out is a decision, and it is made in one place.** Every way of learning nothing — no peer left to ask, no peer able to read a port back, no peer that answered, one witness and no second — resolves after the loop, publishes what it concluded and on what evidence, and proceeds on reachability alone. A verdict and "nobody has decided yet" are not the same value, so falling out of the loop cannot silently mean *refuse*.

**The peer asked is never a Flux node at our own address.** A node sharing our public address is not outside it — its packets never leave the router — so what it can reach says nothing about what the internet can reach, which is the only question being asked. That holds whatever the router does. The network state can be asked for an *external observer* rather than merely another node, and the callers that ask a peer what this node looks like from outside use it: this test, the UPnP keep-alive, the applications availability check, and `checkMyFluxAvailability`, where a neighbour's honest "I cannot reach you" was two points of `dosState`.

**The token is deliberately not sent to the peer.** This check exists precisely because a peer is not in a position to judge — so it is not asked to. It returns raw material; this node decides. A peer that is out of date, broken or dishonest cannot manufacture a secret it was never given, and to produce the token at all it has to genuinely reach us.

**It ships in one go.** A peer on older code returns no reading, which means *nothing was learned* — not *the ports are bad*. The node asks another peer, and if none can read a port back it proceeds exactly as it did before this change. That path is what stops the first nodes to take the update refusing every install while the rest of the network catches up, and it stops happening on its own within hours as the fleet rolls. There is no activation height and no second release.

**The peer's reading is capped at 256 bytes** — it relays what it read from a stranger's port, and an uncapped relay is a node that can be asked to shuttle a payload.

## The address a node is asked to test is the one it is talking to

`/flux/checkappavailability` took the address to probe from the request body. A caller that names it chooses where the node connects, and the reading above is handed back — so a body-supplied address is a fetch primitive aimed at whatever the caller likes, this node's loopback and the private side of its own router included. The range guard is not one: `portMin` is 1 and `portMax` is 65535, and `bannedPorts` names this node's own services rather than a database's.

The honest caller never needed it. It is asking whether **its own** ports are reachable, so the address it means is the one it is connecting from — which is the rule `/flux/addpeer` already applies, and the identity every inbound peer is already held to. The address now comes from the socket. Naming one by hand is a Flux team privilege and nothing else: skipping the signature and choosing the address are one question, so the handler asks it once, with `Privilege.FLUX_TEAM`. **A node operator can no longer drive this endpoint with `zelidauth` and no Fluxnode signature** — `development` allowed it. Both product callers sign, so FluxOS is unaffected; operator tooling outside this repo would need the change. A request is bounded at 50 ports, which is `maxComponents` times ports per component.

`/flux/keepupnpportsopen` — the keep-alive a UPnP node asks a peer to run against its ports — follows the same rule through the same function: the address comes from the socket, naming one is Flux team, and a login alone does not drive it. Its port list is bounded by what an honest one holds, every installed application's ports plus the four node service ports the caller adds, and is deliberately **not** filtered by range or by the banned list: the caller's own service ports sit inside the banned 16100–16299 block, and the filter that is right on the availability endpoint would silently end the keep-alive. A test pokes them.

## Asking the siblings, before anything is opened

```
POST /flux/portsinuse
  →  { status: 'success',
       data: { pubKey, ports: [31000, 31001, 32115], askedAt, signature } }
```

Every host port this node's installed applications hold, deduplicated and sorted. Asked of the other Flux nodes at this address before the firewall is opened and before any mapping is attempted, so a refused install costs nothing to unwind.

- **It counts enterprise applications**, because a node decrypts its *own* specifications. The globally broadcast specifications are decrypted too now, so this is no longer the only check that can see them — but this one still sees what the other cannot, below.
- **It counts applications that are installed but not running.** An installed application holds the router's forward whether its container is up or not, and this is the only check that can see that — the port test asks whether a port answers *now*, which is a different question.
- **Siblings come from the network state**, filtered to entries sharing our address, rather than from the list of nodes already running a global application — a node with no applications yet is the one most likely to be installing.

**It is not the disclosure the container listing was.** What an encrypted specification withholds is *which application uses what*. A bare list of ports names no application, and the identical fact is available to anyone who port-scans the address without any endpoint at all.

**Both directions are signed.** The ask is verified against the deterministic node list — the same check `/flux/checkappavailability` already makes — and carries a timestamp, so a captured signature is not a permanent bearer token for the endpoint. The window is checked before the signature, so an ask outside it is refused for nothing, whoever sent it; inside it, a captured ask yields the port list and nothing else. The endpoint answers any listed Fluxnode, not siblings alone — which ports a node holds is a fact about that node, and a sibling is one caller for it. The answer is signed too, and verified as the node **at the address that was dialled**: a node's own record of what it has installed is the truth about which ports are spoken for here, but only once it is that node saying it rather than whatever is listening on the address. The answer signs the ask's time back, so it answers one question rather than every later one. That verification is extracted rather than copied — two copies of a signature check is the one duplication that must not drift.

Signing the ask also closes a fail-open: `signMessage` catches its own failures and answers with nothing, so a node whose key was briefly unavailable would have sent an unsigned ask, had every sibling refuse it, and read that back as "no sibling holds the port". It says so and asks nobody.

The cost of answering matters, because answering means reading and decrypting this node's own specifications. `assignedPortsInstalledApps` reached past the cached decrypt path to the primitive underneath it — `checkAndDecryptAppSpecs`, which holds no cache and costs two `globalAppsMessages` queries and a benchd RSA decrypt **per enterprise app, per call**. It decrypts through `enterpriseAppDecryptionCache` like everything else now: keyed on `spec.hash`, seven-day TTL, one in-flight attempt shared between concurrent callers. It reads specifications unformatted, because the formatter strips the hash the cache keys on; enterprise `compose` carries the ports either way, which is the shape every version-4-and-above branch already reads.

An app that cannot be read refuses the whole answer, which is what the per-spec call did by throwing. A port list missing an app's ports is worse than no list: every caller is asking which ports are taken, and a hole reads as free.

**The thirty-second cache is on the value, not the response.** A route cache stores whatever the handler produced — including a `200` carrying an error body — so a transient failure would be served to every caller for the rest of the window after the condition cleared, which `routes.js` says at the top must never happen. `portsInUse` caches the answer instead, as a named entry beside every other TTL cache in `cacheManager`, and a throw produces no value to remember.

**It is advisory, and it depends on the router.** A sibling is reached at the public address it shares with us, so asking it leaves the router and comes back in. Where a router does not do that, siblings are silent, this check contributes nothing, and the port test carries the load — which is why silence is never read as clearance, and why a node now says when siblings exist and none of them answered.

## A port a sibling holds is an answer, not a fault

The spawner stands down, exactly as it does for the unreachable-ports check beside it. The node simply cannot host this application while the sibling holds the port, and that is an answer, not a fault. What separates the two is how it is told, not what it costs the app: an answer is one log line and a published deferral the fleet can observe; a throw is an error with a stack trace and no event. The app holds its spawn-cache entry for the same time either way — it is filed at selection, before any check runs.

The application keeps the entry every selection takes in the spawn cache, so this node stops considering it until that expires. That is correct rather than incidental: nothing changes here until the sibling gives the port up.

## Carried with it

**A lost sync request was permanent.** Not a port defect, and it predates this branch. It is in scope because a node that hits it never becomes `READY`, and `READY` is the state every check here runs from.

The sync orchestrator recorded a peer as asked and *then* sent the request. A send into a socket that is closing does not throw — it is accepted and discarded — so nothing surfaced the loss. The peer stayed marked as asked, every later pass filtered it out, and the node reported that there were no new eligible peers to ask and asked nobody. It stayed `SYNCING` and never became `READY`, permanently, out of a window a few milliseconds wide. The set was cleared only on a degraded network, so in ordinary operation never.

The fix is a deletion. `FluxPeerManager` already tracks exactly this and already forgets it when the connection goes. The orchestrator was keeping a second copy of the same fact, and only the peer manager's copy followed reality. It reads the owner's copy now, so a peer whose connection goes away becomes askable again by construction, rather than by a listener keeping two records in step.

**An outbound dial no longer asks benchmark who we are.** The dial refuses this node's own address — a self-connection is not merely a wasted socket, it takes a peer slot, is offered back as a peer to gossip and to sync from, and answers every question with what the node already knows. Getting a current address for that check by calling `getLocalSocketAddress` costs an uncached benchd RPC on every attempt, on five paths including `/flux/addpeer` and the reconnect queue, and makes all five depend on benchd answering. The peer manager already holds this fact — it is told every refresh from the one place the node learns what it is, and already uses it to keep the node out of its own sync draw — so the dial reads the owner's copy, and asks only when the node has not been told yet.

**A node that cannot sign announces nothing, and records nothing it did not announce.** The broadcast signer read both key accessors and checked neither, so a node whose key was unavailable put every gossip message on the wire with null where the key and signature go, and every peer refused it without a word. It signs through the same `nodeSigner` as the rest of this change and answers nothing when the node cannot sign as itself; each of its callers sends nothing on that. The running-apps announcement is one fact recorded twice — the location table, and the event log that peers sync from — and sent once, so a node that cannot sign records it nowhere, asked before the first write: its own view of where it runs stays the network's view rather than diverging from it.

## What this does not close

**A local install reaches neither check.** `ensureApplicationPortsNotUsed` and `checkInstallingAppPortAvailable` are both called only from `trySpawningGlobalApplication`, today and after this change. This work sits where the existing checks sit rather than widening that, which is its own decision.

**Reading the router's mapping table was considered and rejected.** From an 820-gateway fleet survey: ports the walk reported free still returned `718` on 4 routers, and 27 gateways end the walk with `402` rather than `713` — the client terminates on the first error, so a transient fault returns a *truncated* table that reads as "these ports are free", failing open silently. It is also one SOAP call per existing entry at a 1.8s timeout each. A plain HTTP GET to a machine the operator owns is more reliable than SOAP to a consumer router, and it works whether or not UPnP exists.

## Supersedes #1793

This replaces the approach in #1793, which is closed by it.

## Ports in the test harness come from one place

Part of this diff is test infrastructure rather than product, and it changes how every suite gets a port.

A hand-picked host port is how two suites end up on one port, and that surfaces minutes later as an install refused with `already used with different application` — a product fault to read, and not one.

A port is allocated where a specification is **built**, before it is signed and hashed, by the two functions every seeded or registered application passes through. It is stable per application and component, so a specification rebuilt to update an application keeps the port it had rather than carrying a change nobody asked for. Nothing hand-picks a port any more except the one suite whose subject is two applications wanting the same one, and a port declared at or above the allocator's own range is refused when the specification is built, naming both applications — so the hand-written and allocated spaces cannot drift back into each other.

## Verification

**Unit suite: 5,815 passing, 18 pending, 0 failing. Full integration gate on `f98651845`, the head of the product changes: 91 suites, 0 failing.** Lint clean.

Each rule has a test that goes red when the rule itself is reversed: the corroboration count, the distinct-witness redraw, the single-node peer draw, the signed-request guard, and the spawner standing down rather than faulting the application. That last one is worth saying — a refactor turning that path into a throw passes the entire tree without it.

The two functions the decision rests on are asserted directly. `portNotOurs` is covered on our token coming back, on a port answered by something that is not us, on a port the peer got nothing from, on one it would not have read, and on this node having published no token at all. `refusedPort` holds on one witness, refuses once a second **distinct** peer agrees, and does not count one peer asked twice as two — which is the whole reason the tally is keyed by peer rather than by reading.

The loop around them is asserted on every way of running out, including the two this change added: a peer that rejects the request is not a witness, and a peer that could not reach a port is exactly one. The address derivation is asserted on a body naming somewhere other than the caller, on the honest caller, on an IPv4-mapped remote address, and on the operator's carve-out. `siblingHoldingPort` is asserted on an answer signed by a Fluxnode at a different address, one not signed at all, and one answering a different question than the one asked — each of which must return "no sibling holds it" without that becoming a blanket clearance. The endpoint is asserted on an ask older than its window and one carrying no time at all.

**Driven through a real five-node fleet.** Every peer that could be drawn is set to pass a port test *without ever connecting* — which is what the asker receives when the router forwarded its port to a sibling — and the node refuses the install. The control is the point: a **fresh** application of the same shape, same node, same fleet, same port range, with the peers now really connecting, installs. A fresh one because an application a node stands down over deliberately keeps its spawn-cache entry and is not reconsidered.

The suite also covers the sibling ask end to end: the endpoint reports the ports the node has installed, signed, and an install onto a port a sibling at the same address holds is refused with that sibling named. And the case the corroboration rule exists for — exactly one peer can be drawn, it dissents on every attempt, and the install proceeds anyway. Deterministic on purpose: with several peers and one dissenting, the draw decides whether the old rule refused, and a test that is only sometimes red is not evidence.

**The harness asserts on events, not on the wording of log lines.** The spawner's two stand-downs publish the reason they stood down, and the port test publishes what it concluded and on what evidence — `ports:notOurs` when a port answered with something that was not ours, `ports:unproven` when the check learned nothing and proceeded on reachability alone, each naming the peers it asked and the peers it heard from. A log line is not a contract, and a corrected message should not turn a suite red while the behaviour is right. It also expresses the rule properly: corroboration is asserted as two peers that are *distinct*, which no match against a sentence could say.


Cabecinha84 (commented):
What I'd ask for

  1. A same-IP peer can still be drawn as the prober — and the precedent to exclude it is 100 lines below.
  checkInstallingAppPortAvailable picks its peer with getRandomSocketAddress(localSocketAddress), which excludes only
  the exact ip:port (networkStateManager.js:339-370), not siblings. On exactly the topology this PR is about, drawing a
  sibling makes the verdict meaningless: either the router doesn't hairpin (connection fails, peer reports closed) or it
  does and source-NATs (the connection arrives with the router's LAN address, never matching askingIP). Either way,
  refusal. Probability is small (~3/13000 per attempt), but callOtherNodeToKeepUpnpPortsOpen in this same file already
  does if (localIp === askingIP) and redraws (portManager.js:~893). One line, existing pattern.

  2. The new refusal branch duplicates provably dead bookkeeping (portManager.js:800-806). portsNotWorking,
  originalPortFailed and nextTestingPort are write-only across the whole function — nothing ever reads them. The
  pre-existing branch above does it too, so this is copying a wart rather than creating one, but it also carries an
  eslint-disable no-unused-vars for a variable that is never read. The new block only needs the log.warn and portsStatus
  = false.

  3. The question I'd actually want answered: what happens on a node whose inbound path rewrites the source address?
  This change turns the port test from "did something answer" into "did this specific peer's address arrive at my
  socket". Any path that SNATs inbound traffic — NAT reflection, a VPS-fronted tunnel, some firewall configurations —
  now fails permanently, and finished = true is set on the first mismatch, so a single anomalous peer ends the attempt.
  The retry budget (portTestMaxAttempts: 5) exists and is currently only spent on unreachable peers. Retrying with a
  second peer before refusing would distinguish "my address is being answered by a sibling" (every peer fails) from
  "that one peer is odd" at near-zero cost, and would make the change much safer to roll out. The failure direction is
  safe (refusal, and it's logged distinctly), but a node that quietly stops installing anything is a long afternoon for
  whoever debugs it.

  Smaller things, non-blocking

  - specifiedPorts (portManager.js:438) duplicates appUtilities.getAppPorts, and appSpawner already computed appPorts 40
    lines above the siblingHoldingPort call. Two port-extraction functions that can drift.
  - The unauthenticated endpoint is defensible but not the only option. /apps/installedapps is already public and
    already exposes non-enterprise ports, so the true increment is enterprise apps' host ports (not identities) plus
    ports of installed-but-stopped apps, which a scan cannot see. The description's port-scan argument mostly holds. But
    this codebase already has a signed-peer pattern — checkAppAvailability verifies pubKey against the deterministic
    node list plus a signature — and reusing it here would delete both the disclosure argument and the cost argument at
    once. Worth asking why not, rather than requiring it.
  - A transient failure gets pinned for 30 seconds. portsInUseApi catches and answers 200 with an error body, which
    passes the statusCodes: { include: [200] } allowlist — so a benchd blip is cached and served to every caller for the
    window. Impact is nil (advisory answer), but it sits awkwardly next to the "an error must never be remembered"
    comment at routes.js:60.
  - No unit test for the spawner stand-down. appSpawner.test.js adds only a stub; the "sibling holds the port → return
    shortDelayTime" path is covered by harness suite 98 alone.

MorningLightMountain713:
Serviced — all seven. Six were real and are fixed; one was a question and is answered; half of one describes a mechanism that no longer exists, and I'll be specific about which half. Closing them surfaced six further defects, also fixed here, one of them a product bug that predates this branch.

## The three you'd ask for

**1 — a same-IP peer can still be drawn as the prober.** Taken, and it was worse than the three call sites you named. There are **four**, and the fourth is the one that bites: `checkMyFluxAvailability`. A neighbour behind our own router honestly reporting "I cannot reach you" was **two points of `dosState`** and a step toward the address-change handler — on exactly the topology this PR exists for.

Rather than repeat `if (localIp === askingIP)` a fourth time, the picker now answers two different questions instead of one question with a flag: `getRandomSocketAddress` means "another node" and is unchanged for its six callers, and `getRandomExternalObserver` means "a node that can see me from outside". Both run on one walk that takes a rule and returns nothing when nothing matches. The port test, the UPnP keep-alive, the applications availability check and `checkMyFluxAvailability` all ask the second question now.

That also closes a crash on a path nobody had hit. Landing on itself, the draw took "the one before, or else the next" without checking that either existed, so a single-node network threw `TypeError: Cannot read properties of undefined (reading 'ip')`. A one-node list now returns nothing, like any other list with no match in it.

**2 — the new branch duplicates dead bookkeeping.** Correct, and deleted from both branches rather than just the new one. `portsNotWorking`, `originalPortFailed` and `nextTestingPort` were write-only inside that function, and the `eslint-disable no-unused-vars` went with them. The refusal branch is now the log and `portsStatus = false`, as you said. (The identically named fields on `dosState` in the availability checker are a different mechanism and are genuinely read; those stay.)

**3 — a node whose inbound path rewrites the source address.** **The address comparison your review describes no longer exists.** Your review posted at 09:33 UTC and the commit replacing that mechanism landed at 09:42 UTC, so the two crossed. Your question is a good one and it deserves an answer against what is actually there, so here is the current shape.

The test server no longer judges by *who connected*. It answers with a random token generated per test run, and the peer is asked to read each port and hand back **what it found**. This node compares. So NAT reflection, a VPS-fronted tunnel, a firewall that SNATs inbound — none of them are visible to the check any more, because the source address is not what is being tested. A rewritten source address still delivers our token back, and the port proves.

**Your second half stood, and it is in.** One anomalous peer could still end the attempt, and a node that quietly installs nothing is exactly the long afternoon you describe. The rule is now **one witness to accept, two to refuse**: our own token coming back is proof, and a secret cannot be manufactured, so one peer settles it. Anything else is one observer's report about a third party, and a report can be wrong — so a second peer is asked before refusing, spending the retry budget that already existed and was only ever spent on peers too old to answer. A real collision corroborates itself for free, since the router forwards that port to one node and every peer reads the same thing; an anomaly does not reproduce on a different peer.

The token is deliberately never sent to the peer. This check exists because a peer is not in a position to judge, so it isn't asked to — it returns raw material and this node decides. A peer that is out of date, broken or dishonest cannot produce a secret it was never given, and to produce it at all it has to genuinely reach us. Its reading is capped at 256 bytes, because relaying what a stranger's port said is a shuttle if it is uncapped.

## The four smaller ones

**`specifiedPorts` duplicates `getAppPorts`.** Gone. `portManager` no longer parses specifications at all — the caller passes the ports it already computed.

**"Why not the signed-peer pattern?"** — the right question, and the answer is that there was no good reason. `/flux/portsinuse` is now a signed POST verified against the deterministic node list: the same check `/flux/checkappavailability` already makes, on the endpoint this feature's own port test already posts to. An operator asking by hand is accepted on the usual privilege. The verification is **extracted rather than copied** — two copies of a signature check is the one duplication that must not drift.

You were right that the disclosure argument was not the load-bearing one. The cost argument is: answering means reading and decrypting this node's own specifications, and left open an anonymous caller could ask for that as often as it liked. Signing deletes both at once, as you said it would.

**A node that cannot sign the question asks nobody.** `signMessage` catches its own failures and answers with nothing, so a node whose key was briefly unavailable would otherwise send an unsigned ask, have every sibling refuse it, and read that back as "no sibling holds the port" — a blanket clearance out of a broken key.

**A transient failure pinned for 30 seconds.** Fixed, and by construction rather than by patching the allowlist. The cache moved off the response and onto the value: `portsInUse` caches the answer as a named TTL entry beside every other one in `cacheManager`. **A throw produces no value, so an error can no longer be remembered.** The awkwardness you spotted next to the comment at the top of `routes.js` was the comment being right.

**No unit test for the spawner stand-down.** Three now: that a sibling holding the port produces the stand-down delay, that it does **not** file the application in the pre-install error cache, and the control — same application, same everything, no sibling — which does not stand down. Each goes red when the rule it covers is reversed. You were right that without them, a refactor turning that path into a throw passes the whole tree.

## What your findings uncovered

Six more defects, all fixed here. They are listed because they are as much yours as the seven:

- **The redraw excludes the peers already asked.** The corroboration tally counts distinct witnesses; the draw beside it was random and could return the peer just asked. One witness, redraw, same witness, tally still one, install proceeds — so the refusal your point 3 is about could not fire.
- **The two ways of proceeding without proof are distinguishable.** Excluding already-asked peers makes a second state reachable: every peer there was has been asked and disagreed, against there being no peer to ask at all. They shared one sentence, and the first of them read as the second while a peer had in fact disagreed. Each is published as its own event now, naming the peers asked.
- **The ports endpoint reads the parsed body.** `express.json()` is global on that path, so a handler reading the raw request stream is waiting for an `end` event that has already been and gone, and the request hangs rather than answering. The older handlers there survive it only because the product posts a pre-stringified body that axios does not label as JSON, so the parser skips it — two patterns, each working for one kind of caller, and `req.body` works for both.
- **A specification's port is allocated without writing into the shared default.** The builder spreads a module-level default and a spread is shallow, so an allocated port landed in the default itself and came back on the next build as a hand-picked port inside the allocator's own range.
- **A node that has not started yet is not counted as a dialer** when the harness waits for a fleet to finish peering.

The sixth is a product bug with nothing to do with ports:

- **A lost sync request is no longer permanent.** The orchestrator recorded a peer as asked and *then* sent. A send into a closing socket does not throw — it is accepted and discarded — so the peer stayed marked as asked, every later pass filtered it out, and the node reported "no new eligible sync peers to ask" and asked nobody. It stayed `SYNCING` and never reached `READY`, permanently, out of a window a few milliseconds wide. The fix is a deletion: the peer manager already tracks this and already forgets it when the connection goes, so the orchestrator's second copy of the same fact is gone rather than kept in step by a listener.

## Where it stands

Unit suite 5,848 passing, 18 pending, 0 failing. Lint clean. The full integration gate ran on the tip of the stack — the branch above this one, rebased onto this branch's head — so this head is covered by it: 94 suites, 93 green. The one red is a teardown race in the harness's own fixture shutdown rather than in anything under test: every test in that suite passed and only the `after all` hook exceeded its budget. It is being fixed on that branch.

Thank you — the same-IP draw and the retry budget were both real holes, and the fourth call site was worth the whole review on its own.

Cabecinha84 (commented):
What I'd ask for

  A. The echo turns every node into a 256-byte content oracle for an attacker-chosen address — blocking

  ZelBack/src/services/fluxNetworkHelper.js:522 — portAnswered(ip, port) where ip and ports come straight from the
  request body, and nothing binds ip to req.socket.remoteAddress.

  Before this PR, isPortOpen(ip, port) on the same body already gave a signed caller an open/closed oracle against an
  arbitrary ip:port — bad, but only one bit. This PR upgrades it to content: a caller can now ask any Flux node to
  connect to 127.0.0.1, 10.x, 172.17.0.1, or any external host, send an HTTP GET, and return the first 256 bytes.

  The range guard is not a guard: portMin: 1, portMax: 65535 in config/default.js, and bannedPorts covers only
  16100-16299, 26100-26299, 30000-30099, 8384, 27017, 22/23/25, 3389, 5900/5800, 161, 512/513, 5901, 3388, 4444, 123,
  53. Redis 6379, MySQL 3306, Postgres 5432, docker 2375/2376, Elasticsearch 9200, 8080 are all reachable, and 256 bytes
  is enough for a banner or a full HTTP header block — that is internal-network reconnaissance across the whole fleet.

  The gate is a valid Fluxnode signature, so it is not anonymous. But "one node key ⇒ read the internals of all 13k
  nodes" is exactly the pivot that matters in a P2P network, and the fix is one line: require ip to match the connecting
  socket's remote address when echo is set (the legitimate flow already satisfies this — the requester's public IP is
  the peer's remoteAddress), keeping the arbitrary-ip form for the operator-privileged path if anything needs it.

  Related, same handler: ports is unbounded and each echo costs up to 5s sequentially — pre-existing shape, but worth a
  length cap while you're in there.

  B. The port-test loop can fall out of the bottom and refuse silently — should fix

  ZelBack/src/services/appNetwork/portManager.js:830-979. portsStatus starts false, and there is no post-loop decision:
  if the while exits on i exhaustion with finished === false, the function returns false and the install is aborted with
  no log line and no event.

  That is reachable in a way that defeats the rule this PR just introduced:

  - attempt 1: peer A disagrees → refused === null && i < max → continue (portsStatus still false)
  - attempts 2–5: every drawn peer is unreachable (axios rejects, resMyAppAvailability is undefined, neither branch
    runs)
  - loop exits → return false → refused on one witness, with nothing in the logs saying why.

  Every other exit is handled explicitly and publishes ports:unproven / ports:notOurs. This one — and the
  all-peers-unreachable case, which is pre-existing but equally silent — is the only one that isn't. Given the whole
  point of the corroboration rule is "a node that quietly stops installing anything is a long afternoon for whoever
  debugs it", the fall-through deserves the same treatment: decide explicitly after the loop, apply the same
  single-witness rule, and publish an event.

  C. The corroboration rule protects one exit and not the other — worth a decision

  portsStatus = false; finished = true on data.status === 'error' (portManager.js:872-889) still refuses on one peer's
  word, and the echo change made that branch materially easier to trip: isPortOpen needed only a TCP handshake,
  portAnswered needs actual bytes back within 5s, otherwise the peer throws Failed port: N.

  For the case this PR targets (a sibling's app holding the forward) the refusal is correct. But a slow read, a
  congested path, or a peer having a bad moment now produces the same verdict from a single observer, while a token
  mismatch — strictly stronger evidence — requires two. That asymmetry is the reverse of the design's own argument. The
  blast radius is bounded (deferral + spawn-cache entry, not a permanent stop), so this is a "decide and say so" rather
  than a blocker — but treating an echo-mode error as a disagreement rather than a verdict would cost one line and make
  the rule consistent.

  D. A same-IP node's /flux/portsinuse answer is trusted without proof

  siblingHoldingPort (portManager.js:531) takes a sibling's port list at face value and stands the spawner down. A node
  on the deterministic list declaring the victim's IP could answer with every port in range and defer every app on that
  node indefinitely — no proof required, unlike the port test right below it. The doc calls the check "advisory", but
  the effect on the app in hand is not advisory.

  The exposure is limited (the attacker needs a confirmed Fluxnode declaring that IP), and the failure direction is a
  deferral rather than an error-cache entry, so I would not block on it — but it belongs in the PR description as a
  known trust boundary, or bound it (e.g. ignore an answer that claims an implausible number of ports).

  Smaller things, non-blocking

  - initiateAndHandleConnection now does a benchd RPC per dial. fluxCommunication.js:1043 unconditionally calls
    getLocalSocketAddress() where the old code did it once (if (!myPort)), and benchmarkService.getBenchmarks() is an
    uncached RPC with a DB-write side effect. fluxDiscovery already holds localSocketAddr at line 1435 and could pass
    it; as written each discovery cycle adds ~20 extra benchd round trips. Also makes every new outbound connection
    hard-dependent on benchd answering — for discovery that's already true, but /flux/addpeer (line 1241) and the
    reconnect queue now inherit it.
  - The signed /flux/portsinuse ask has no nonce or timestamp ({ pubKey, asking: 'portsInUse' }), so one captured
    signature is replayable against any node forever. Impact is small given what is disclosed and the 30s value cache,
    but it also means the endpoint is readable by any signed node anywhere, not just a same-IP sibling — which is the
    only legitimate caller.
  - refusedPort counts two peers disputing different ports as corroboration. The test at portManager.test.js:884 says
    this is deliberate ("the peers need not have tripped on the same port"), which is defensible — but the log line then
    reads port N ... as read by 2 peers, which is not what happened. Naming both ports would cost nothing.
  - assignedPortsInstalledApps switched to formatSpecs: false. I checked: enterprise.compose carries ports, and five
    other callers in the base already use the unformatted form, so this is consistent — but it does change what feeds
    restoreAppsPortsSupport's firewall rules, which is worth a conscious nod rather than a comment about cache keys.
  - Docs/cosmetics: the portNotOurs JSDoc block is orphaned above PORT_TEST_CORROBORATION (portManager.js:596-604) and
    portNotOurs itself ends up undocumented; siblingHoldingPort's @param {object} appSpecFormatted (line 526) no longer
    matches its signature; fluxHttpTestServer.js picked up two stray double-blank-line runs.

MorningLightMountain713:
Serviced — all four, and all the smaller ones. Two went wider than you asked and one narrower, and I'll say which and why. Head is `18a705431`.

## A — the echo oracle

Taken, and **widened to both modes**. You scoped the bind to `echo`; the address now comes from `req.socket.remoteAddress` whenever the caller is a signed peer, so the one-bit `isPortOpen` oracle closes with the 256-byte one.

The reason is the bug's own history. `echo` was added to a handler that already let the caller choose the address, and it inherited that freedom for free — a guard scoped to today's mode is one the next mode forgets. So the address stops being an input rather than being checked: nothing to guard, and whatever gets added to that handler next inherits safety instead of the hole.

Your range-guard point is exactly right — `portMin` is 1, `portMax` is 65535, and `bannedPorts` names this node's own services rather than a database's. Worth adding that it is a **write**, not only a read: the probe sends a request, so it is a forgery primitive against anything with a GET side effect, not merely a banner grab.

Both product callers already satisfy the bind — the port test and the availability checker each send their own address and post from it. The operator carve-out you suggested is kept. `ports` is capped at 50, which is `maxComponents` times ports per component.

## B — the loop falling out of the bottom

Taken, and it is **a decision rather than a log line**. You are right that it is reachable and right about how; the part worth naming is that the verdict was *backwards*. Everywhere else the rule is "nothing learned, proceed on reachability alone" — that is the no-observer branch, the no-reader branch and the single-witness branch. A peer that never answered taught us nothing either, and it landed on refuse.

`portsStatus` was initialised to `false`, which is also what refuse looks like, so a verdict and "nobody has decided yet" were the same value. It is tri-state now, and every way of running out — no peer left, no peer able to read, no peer that answered, one witness and no second — resolves once after the loop, publishes what it concluded, and proceeds. That deleted the two `if (i < max) continue; else <copy>` blocks inside it, so the case you found is unrepresentable rather than newly handled.

## C — the asymmetry

Taken, and **wider than the finding**. The caller checked `status === 'error'` and never looked at what the error said, so a peer rejecting the request outright — a stale node list, a clock, a key briefly unavailable at its end — landed in the identical branch and was logged as `could not reach the ports`. That is a broken conversation recorded as a failed port test.

They are separated on the discriminator that was already there and already parsed for the log line. A peer that **names** a failed port has read something at this address: it joins the same corroboration map and answers to the same rule, one witness to hold and two to refuse. A peer that names no port has told us nothing: it is asked past, like an unreachable one, and logged as what it was. No wire change, so a peer on either version behaves the same.

On the cost of making an error need two witnesses: if the ports genuinely are closed every peer says so, corroboration is immediate and the verdict is unchanged. It only diverges when peers disagree with each other, which is precisely where one peer's word should not have decided.

## D — trusting the sibling

Taken, but **not the bound you suggested** — and I want to be straight about why, because the disagreement is the useful part.

Capping the number of ports claimed achieves nothing: a claimant only needs the specific ports of the application being installed, and those are on chain. The problem was never the size of the claim, it was that we authenticated who was **asking** and accepted an unauthenticated answer from whatever was listening on the address.

So the answer is signed, and verified as the node **at the address that was dialled** — not merely as some listed Fluxnode, which would let a key from anywhere answer for a sibling. The verifier is extended rather than copied: an optional address binds a signer to a place, which an answer needs and a request does not.

What I did **not** do is make the check demand proof of the fact itself, because that fact is unobservable from outside. The sibling asserts that a port is *allocated*, which is true of an application installed but not currently serving anything — and that is the one case this check exists for, since the port test only ever answers whether a port responds *now*. There is no proof available for it. A node's own record of what it has installed is the truth here; the fix is making sure it is that node saying it.

The real limit is elsewhere, and it is in the description now: a sibling is reached at the public address it shares with us, so the ask leaves the router and comes back in. Where a router does not do that, siblings are silent and this check contributes nothing. It is why silence is never read as clearance, and a node now says when siblings exist and none of them answered rather than making that look like having no siblings.

## The smaller ones

**benchd per dial.** Taken. Not by passing the address down, though — that fixes the two discovery paths and leaves `/flux/addpeer` and the manual add needing their own rule. `FluxPeerManager` already holds this node's address, is told every refresh from the one place the node learns what it is, and already uses it to keep the node out of its own sync draw. The dial reads the owner's copy and asks only when the node has not been told yet.

Reverting to the old `if (!myPort)` would not have been right either: it read the port once per process and never again, so after an address change a node announced a stale one for as long as it ran. The owner's copy is refreshed once per discovery cycle, so staleness is bounded where it previously was not.

One correction: `getBenchmarks` does not write to the database per call. `storeBenchmarkToDb` returns immediately unless a day has passed. The RPC is real and uncached; the write is not.

**Replay.** Taken. The ask carries a timestamp and is refused outside its window — as *stale*, not as unauthentic, because those want different fixes and a node whose clock has drifted should read which from one line. An operator is not asked for one. The same field binds the answer: the sibling signs the ask's time back, so a captured answer cannot be replayed into a later question either.

**`refusedPort` naming one port.** Taken — the line names each peer's own reading, since two peers tripping on different ports is still corroboration and naming only the first port describes something else.

**`formatSpecs: false`.** Your check was correct, and it is a conscious nod in the description now rather than a comment about cache keys.

**Docs and cosmetics.** `portNotOurs` has its documentation back — the block had drifted above `PORT_TEST_CORROBORATION`, so the constant had two and the function none. `siblingHoldingPort`'s `@param` matches its signature. The blank-line runs are gone.

## One thing you did not raise

The comment justifying the same-address exclusion claimed most consumer routers do not hairpin. That is both disputed and unnecessary: a node at our address is not *outside* it, its packets never leave the router, so what it can reach says nothing about what the internet can reach. That holds whatever the router does, and it is what the code says now — in both places it appeared.

Thank you. The echo escalation was the right thing to lead with, and the loop exit is the kind of defect that only ever shows up as a node that quietly stops installing.

Cabecinha84 (commented):
The one finding I'd actually ask for in this PR

  The echo response fits under the 256-byte cap with 48 bytes to spare, and the failure mode is fleet-wide.

  I measured it against the real FluxHttpTestServer rather than counting headers:

  raw length: 218    token ends at byte 208    cap: 256
  "HTTP/1.1 200 OK\r\nContent-Type: application/json\r\nDate: ...GMT\r\n
   Connection: close\r\nTransfer-Encoding: chunked\r\n\r\n48\r\n
   {"status":"success","data":{"token":"88bfcb3b...9772"}}\r\n0\r\n\r\n"

  It works today. But the token is the last thing in the response, so it is the first thing lost, and ~48 bytes of
  additional headers truncates it. What makes that worth a line of code: this failure corroborates itself. Every peer
  reads the same truncated bytes, so portNotOurs fires on peer 1 and peer 2 alike, refusedPort reaches its threshold
  immediately, and every install on every node is refused. The two-witness rule is designed to filter anomalies that
  don't reproduce on a different peer — this one reproduces by construction, which is precisely the class the rule
  cannot catch.

  Cheapest fixes, either is one line: raise MAX_ECHO_BYTES to 512, or put the token where headers can't push it out (a
  response header emitted right after the status line, or make the body the token alone). I'd rather see this in the PR
  than tracked.

  Should fix, non-blocking

  1. routes.js:351 contradicts itself on a security-relevant endpoint. The block still opens with "Unauthenticated and
     reachable by anyone, and EVERY answer is expensive… The cache is the only thing bounding that rate" and refers to
     "the cache keys on the request URL … hence the guard" — a rejectQueryParameters guard that isn't on this route.
     Only the appended last three sentences are current. The endpoint is now signature-gated, and the first thing the
     next reader sees says it isn't. This is my second merge condition because it's the kind of comment someone "fixes"
     in the wrong direction.
  2. siblingHoldingPort can still throw into the 6-hour pre-install error cache — the exact outcome its own doc-comment
     exists to prevent. The .catch(() => null) guards only the axios call; verifySignedFluxnodeMessage →
     deterministicFluxList sits inside the same Promise.all unguarded, and appSpawner's outer catch (line 952) sets
     trySpawningGlobalAppCache with oneHour * 6. Low probability, one-line fix: catch inside the per-sibling mapper.
  3. getFluxNodePublicKey returns the Error object instead of throwing (fluxNetworkHelper.js:875). So the fail-closed
     handling you got added for signMessage has a sibling hole: a briefly-unavailable key yields pubKey: {} after
     JSON.stringify, every sibling refuses, and it reads back as "no sibling holds the port". It is surfaced now by the
     new "none of them answered" warning, so it's not the silent fail-open — but it's the same class, and it's a
     pre-existing idiom worth naming.
  4. Privilege.NODE_OPERATOR_OR_FLUX_TEAM is not node-local, contrary to the comment "that privilege is already
     node-local". The Flux team key retains the arbitrary-address probe against every node in the fleet. The PR narrows
     the hole enormously (any signed Fluxnode → operator/team), so this is a comment correction, or
     Privilege.NODE_OPERATOR if node-local is what's meant.
  5. assignedPortsInstalledApps and assignedPortsGlobalApps still hand-roll the extraction getAppPorts was just
     consolidated into — in the same PR whose new doc-comment says "The one place this is derived. A second extraction
     living somewhere else is two answers to one question." Both reduce to { name: app.name, ports: getAppPorts(app) }.
  6. The blank-line runs in fluxHttpTestServer.js are still there (lines 14–15 and 56–58), despite being reported as
     fixed. Lint-clean, so purely cosmetic.

MorningLightMountain713:
All seven were right. Head is `989d69d2f`, and the PR body is updated to match.

**Lead — the echo response fits under the 256-byte cap.** `dd0e4b0af`. The response is now written in full by this file: `sendDate` off, an explicit `Content-Length` so nothing is chunked, our own `Connection`, and the token in a header immediately after the status line. Node emits `writeHead` headers before the ones it adds itself, so that position is a fixed offset. Raising the cap would not have held: the margin was being set by whatever `Connection` the reading peer sent, so it moved between 20 and 80 bytes with the caller.

**1 — `routes.js` contradicts itself.** `0a2263416`. The block described a `rejectQueryParameters` guard that is not on that route. It is gone, and the endpoint is described once, beside the code that enforces it.

**2 — `siblingHoldingPort` can throw into the six-hour cache.** `d8e2fce16`. The contract is held at the function boundary rather than a line at a time, so no line added inside it later can file an application in the pre-install error cache. Two tests, red against the previous code: a sibling that cannot be verified no longer costs the answers of the siblings that replied, and a failure before any sibling is reached answers "no information".

**3 — `getFluxNodePublicKey` returns the Error object.** `2e28d229c`. Both accessors return `null`, and `getFluxNodePublicKey` logs where the failure happens rather than only at the far end. A new `nodeSigner()` answers `{ pubKey, sign }` or `null` for the five call sites that took a key and never checked it. No message shape moved: `pubKey` keeps its position in every body, because the two signing conventions on the wire are verified by different verifiers and one of them re-stringifies what it received. The consumer-side guard in `enterpriseNetwork.js` stays, with a comment saying why — a consumer refusing a value it cannot use is a stronger guarantee than a producer promising not to send one.

**4 — `NODE_OPERATOR_OR_FLUX_TEAM` is not node-local.** `6ec055530`. `/flux/checkappavailability` now has a single gate, `Privilege.FLUX_TEAM`, used for both the signature bypass and the address choice. Those were always one question: the bypass exists so a person can drive the endpoint by hand, and naming the address is the only thing a person gets from it — asking whether the ports are open on the machine you are sitting at answers nothing anyone wants.

Behaviour change worth flagging: a node operator can no longer call that endpoint with `zelidauth` and no Fluxnode signature. Both product callers sign and go through the signature route, so FluxOS itself is unaffected, but operator tooling outside this repo would need the change. A test asserts the handler is called with `FLUX_TEAM` and never with the old constant, so it cannot drift back unnoticed.

**5 — `assignedPortsInstalledApps` / `assignedPortsGlobalApps` hand-roll `getAppPorts`.** `04e22577f`. Both derive ports through `getAppPorts` now.

Your "both reduce to `{ name, ports: getAppPorts(app) }`" was close but not exact, and the difference was worth more than the deduplication: `assignedPortsGlobalApps` read the stored specifications without decrypting them. A version 8 specification seals `compose`, every port lives inside it, so an enterprise application produced no ports and was dropped by a `length > 0` filter — the pre-install check read those ports as free. It decrypts now, through the same cached path the installed list already used.

The two differ on what to do with a specification that will not open, deliberately rather than by accident: the installed list refuses, because `portsInUse` publishes it to a Flux node sharing our address and a hole there tells a sibling a port is free when it is not; the global list answers short and logs the count, because every node stores specifications sealed to keys it was never meant to hold.

`isPortAvailable` and `findNextAvailablePort` went with it — no caller anywhere.

**6 — the blank-line runs are still there despite being reported fixed.** `3ac2ce0a4`. You were right on both halves. None of round two's commits touched `fluxHttpTestServer.js`; the runs removed then were in `portManager.js`.

Also in this push, `aea70ec56` and `989d69d2f`: names only, no behaviour. `extractIp(localSocketAddr)` was being recomputed in four places inside one function under three different names.

Unit suite 5,802 passing, 18 pending, 0 failing. Full integration gate running on this head.

MorningLightMountain713:
Gate result, following on from the comment above: **91 suites, 0 failing**, on head `989d69d2f`.

```
RESULT  suites_pass=91  suites_fail=0
```

That is the first full integration gate on any of this PR's commits. The previous one, 94/94, was on #1784's head and contained this branch only as far as `18a705431` — everything from `dd0e4b0af` onward was covered by unit tests alone until now.

Unit suite alongside it: 5,802 passing, 18 pending, 0 failing. The PR body's Verification section is updated to match.

Cabecinha84 (commented):
5. Non-blocking follow-ups (post-merge, your call)

  1. fluxBroadcastHelper was not converted. serialiseAndSignFluxBroadcast is the one remaining path that reads the key
     directly and checks nothing — it now emits "pubKey": null, "signature": null instead of {}. Same outcome (every
     peer refuses it silently), so not a regression, but it is the highest-traffic instance of exactly the failure
     nodeSigner's own docstring was written about.
  2. /flux/portsinuse has no rate limit, and verifySignedFluxnodeMessage is called without a socketAddress bind — so any
     listed Fluxnode may ask any node, not just a sibling, and a captured ask is replayable for its 60s window. Each
     request costs a fresh ECDSA sign (the port list itself is cached). Low severity; the disclosure argument in the PR
     body is sound.
  3. /flux/keepupnpportsopen still takes a body-supplied ip now that /flux/checkappavailability no longer does.
     Pre-existing and out of scope, but the pair is now inconsistent.
  4. The "stand down rather than throw" rationale is largely presentational. By the time the sibling check runs,
     appSpawner.js:397 has already parked the hash in trySpawningGlobalAppCache for 12h, so the catch block's set(...,
     6h) is a no-op and both paths return shortDelayTime. The event and the log line are the real difference — which is
     fine, but the comments claim more than the code does.

  6. Release note worth carrying

  /flux/checkappavailability now requires Privilege.FLUX_TEAM (was NODE_OPERATOR_OR_FLUX_TEAM) on the zelidauth path,
  and otherwise takes the probed address from the socket. Both product callers sign, so FluxOS is unaffected — but any
  operator tooling outside this repo driving that endpoint with zelidauth will break.

MorningLightMountain713:
All four taken, in this PR rather than after it. Head is `8e3ea916e`; the product changes end at `f98651845`, and everything above it is test-only.

**1 — the broadcast signer.** `c053c1efc`. It signs through `nodeSigner` and answers nothing when the node cannot sign as itself; each of its five callers sends nothing on that, and the Flux Storage request refuses rather than sending a header carrying null. The running-apps announcement needed a decision rather than a guard: it is one fact recorded twice — the location table, and the event log peers sync from — and sent once. A node that cannot sign now records it nowhere, asked before the first write, so its own view of where it runs stays the network's view instead of diverging from it. Each guard has a test that is red without it.

**2 — the ports endpoint.** `d4abebec3`. The premise in the comment was wrong, not the code: which ports a node holds is a fact about that node, and a sibling is one caller for it, so the endpoint answers any listed Fluxnode and the comment now says so. What changed is the order — the window is checked before the signature, so an ask outside it is refused for nothing, whoever sent it — and the comment states what a captured ask yields inside the window: the port list, signed to the ask's own time, and nothing else. No bind to the socket: a hairpinned ask arrives carrying either the public address or the sibling's LAN address depending on whether the router source-NATs, so that bind would refuse the honest caller. No rate limit, because no HTTP route on the node has one and this is not the cheapest place to make a node work.

**3 — the keep-alive endpoint.** `f98651845`. Taken into this PR: the rule about where a probed address comes from should hold for both routes rather than leave its twin as an exception. Same shape, through one shared function rather than a second copy: the address comes from the socket, naming one is Flux team, and that is the same single privilege as skipping the signature. Its port list is bounded by what an honest one holds — every installed application's ports plus the four node service ports the caller adds — and is deliberately not filtered by range or by the banned list: the caller's own service ports sit inside the banned block, and the filter that is right on the availability endpoint would silently end the keep-alive. A test pokes them. The release note covers both endpoints now.

**4 — the stand-down rationale.** `8415ee093`. You are right. The hash is filed in the spawn cache at selection and the catch adds nothing for a hash already there, so a throw and a stand-down leave the app in the same place — which also means round three's item 2 could not have happened as described, and the comments that said so are corrected. They now say what actually differs: an answer is one log line and a published deferral, a throw is an error with a stack trace and no event. The test guarding that asserted a cache neither path writes, so it could not go red on a throw; it asserts the deferral event instead, which a throw does not publish, and it is red against the branch turned into a throw.

Unit suite 5,815 passing, 18 pending, 0 failing. Lint clean.

Full integration gate on `f98651845`, the head of the product changes: 91 suites, 0 failing.

The fleet suite for this change, `98-same-ip-port-collision`, is 8 of 8 green, including the case that drives `/flux/keepupnpportsopen` against a real node through `express.json()` and a real socket.

Cabecinha84 (approved):
ack

PR #1796, opened 2026-09-03
feat(auth): the support team is a list of Flux IDs, not one

Cabecinha84:
## What

Support is held by four people, and the code could only name one of them. Two places said so:

- `fluxSupportTeamFluxID` in config, a single string that four privilege checks compared against with `===`. It decides who gets a `fluxteam` session and, through it, app logs, inspect, stats, files and the lifecycle controls.
- `fluxapps.teamSupportAddress`, which says who may sign a marketplace app update. One address per fork, so nobody but that address could sign.

Both now name a list.

### `fluxSupportTeamFluxID` is a list (`6018265f5`, `73c89d9f2`)

Two helpers in `verificationHelperUtils` — `fluxSupportTeamZelids()` and `isFluxSupportTeamZelid()` — replace the `===` in `verifyFluxTeamSession`, `verifyNodeOperatorOrFluxTeamSession` and `verifyAppOwnerOrFluxTeamSession`, and in the two places `idService` resolves a login to `fluxteam` (HTTP and websocket).

The list is read defensively. A node whose own config still carries the single string this used to be must keep working: reading that string as though it were an array matches nothing, and support would be locked out of the node with no error to say why. An empty or missing value yields an empty list, which grants no one.

The four Flux IDs replace the single seed address, which was the value this config already carried rather than a decision about who support is.

### A support fork names the whole team (`da53344eb`)

`teamSupportAddress` forks now carry `addresses` instead of `address`, and any address in the fork in force can sign. `supportAddressesOf()` reads either shape and `verifySignatureFromAny()` tries each, stopping at the first that signs — so a message signed by the first address costs exactly what it cost before.

The new fork takes effect at **2951000** — about a week past the tip as of the last commit (2931010, 7th September 2026, at the ~30s blocks the chain is producing) — and carries the existing address forward: a fork replaces its predecessor rather than adding to it, so leaving it out would have stopped it signing.

It is dated ahead rather than at the tip because of the rollout below: a node on an older FluxOS rejects a message signed by any of the four new addresses, so the fork has to fall after the network has had time to update. The harness chain start moves with it (2920000 → 2952000), since `tests/unit/harnessChainStart.test.js` fails if any gate in the production config sits at or above it — and for the same reason it fails: every suite would otherwise run the pre-fork branch of the rule this PR adds.

The fork already in force is left exactly as it is. A message is verified against the fork at *its own* block (`messageStore.js:189` passes the message's `block`, not the tip), so rewriting that entry to the new shape would change which signatures the past accepts.

## What this does not change

`getApplicationSpecificationAPI` still requires `Privilege.APP_OWNER`. Enterprise spec decryption stays the owner's alone and the flux team continues to decrypt out of band, as the comment there describes. This PR does not touch it.

## Two things to know before merging

**Rollout.** Nodes on an older FluxOS read only the fork at 1851659. A message signed by one of the four new addresses will not verify for them until they update, and they will reject it. `16iJqiVbHptCx87q6XQwNpKdgEZnFtKcyP` verifies under both forks, so it should keep signing until the network has updated.

**Frontend.** `GET /flux/fluxids` now reports `fluxSupportTeamFluxID` as an array. The frontend compares against it and needs the matching change to `.includes()`.

## Testing

Six new tests:

- `verificationHelperUtils.test.js` — a list of ids, a bare string read as a one entry list, and unset/empty/falsy granting no one.
- `messageVerifier.test.js` — a signature from any address in a fork that names a list, and that the fork in force at the block is the one read, not the newest. Both were confirmed to fail without the change.

```
messageVerifier + verificationHelperUtils + idService + fluxService + registryManager + appsService: 359 passing
eslint: clean on every changed file
```

Fork selection was also checked against the real config: at 2950999 only the existing address is in force; at 2951000 all five are.

https://claude.ai/code/session_011PvQJyYykeM3vCgrDSvBZG

### DIFF
--- a/ZelBack/config/default.js
+++ b/ZelBack/config/default.js
@@ -347,9 +347,24 @@ module.exports = {
       multiplier: 1, // multiplier in case we want to increase prices globaly
       minUSDPrice: 0.99, // min. usd price that can be paid with stripe/paypal.
     },
+    // Who the support team is, from which height. Only the latest fork at or below
+    // a message's own block is consulted, so entries are only ever appended: an
+    // edit to one below the tip changes which signatures the past accepts.
     teamSupportAddress: [{
       height: 1851659, // height from which address is valid
       address: '16iJqiVbHptCx87q6XQwNpKdgEZnFtKcyP',
+    }, {
+      // From here a fork names a list rather than one address. The address above
+      // is carried forward - a fork replaces its predecessor rather than adding
+      // to it, so leaving it out would stop it signing.
+      height: 2919141,
+      addresses: [
+        '16iJqiVbHptCx87q6XQwNpKdgEZnFtKcyP',
+        '16dNCFf7nR3nx5iwn2RQMBw6KcJXkE3JC1',
+        '15c3aH6y9Koq1Dg1rGXE9Ypn5nL2AbSJCu',
+        '1NGqYirE4T9wzd1ZcGrw3HjETiuCkt6Sgy',
+        '13BBPcpHxwCaC61vjQgK6qeDcprFJEGVkP',
+      ],
     }],
     usersToExtend: ['1MCBJn6qsy3YRY2YasdYMYdJcdhy1ev8Rd'], // addresses that can extend applications on behalf of app owners (expire-only updates) addresses cannot be deleted over time, just adding new ones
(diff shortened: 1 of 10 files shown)


MorningLightMountain713 (approved):
Ack. Assume we are going to change the fork height to the future?

PR #1794, opened 2026-08-30
fix(apps): a missing container must not crash FluxOS, a failed redeploy must not uninstall the app, and a log poll must lose nothing

Cabecinha84:
> **The full harness gate is green at `ccff2d043`: 99 suites, 0 failures.** The unit suite is green at the current tip (6349 passing, 0 failing) and lint is clean. The streaming path is proved in a browser against a live node, and every documented request shape is verified against that same node — see **Testing**.
>
> Sections 19-22 landed after that gate run, and **a full gate run is in flight at the current tip** — its result will be posted here. They are unit-covered, harness-covered in suite 84, and proved end to end in a browser against a node running this tip.

Two failures observed on a live node, three hours apart, both reached from the same line of code. Fixing them opened five more in the same paths, and one of those turned the log endpoint itself into a piece of work: it answered every poll on a 1500 ms timer, lost lines written between polls, and made docker re-read the whole log to do it.

**This is the node side of a four-repo change.** It is additive — a client that knows nothing about it gets the old behaviour exactly — and it must ship first. See **Deployment order**.

---

## 1. FluxOS restart-looped until a browser tab was closed

`GET /apps/applogpolling/<component>/100` for an app whose container no longer exists killed the FluxOS process. systemd restarted it, the page polled again, and it died again — one exit per poll.

```
21:34:45  "OPTIONS /apps/applogpolling/rustoxide_<app>/100"  204
21:34:46  "GET     /apps/applogpolling/rustoxide_<app>/100"  304
21:34:47  fluxos.service: Main process exited, code=exited, status=1/FAILURE
```

`dockerContainerLogsPolling` rejected, and `appInspector` called it from inside a `new Promise` executor without awaiting the promise it returns — so the rejection had nothing attached to it. The route is privileged (`appownerabove`), but any app owner could take down every node hosting their app by leaving the log page open across a removal.

The fix is now **structural rather than a guard**: section 7 rewrote that function to be a plain `async` function that returns its lines or throws, called from a `try/catch`. No callback, no second reporting channel, no un-awaited promise left to go unhandled.

## 2. A live app was uninstalled by a soft redeploy that never removed it

Changing one environment variable on a running app removed it from the node and broadcast the removal:

```
19:23:41  masterSlaveApps: starting docker component:rustoxide_<app>
19:23:43  REMOVAL REASON: Soft redeploy failure - failed during soft redeploy:
          Cannot read properties of undefined (reading 'Id') (softRedeploy)
[...]

PR #1797, opened 2026-09-04
A node answers a state-sync request for itself, and every outstanding request has its own clock

MorningLightMountain713:
## What this changes

A node that has just started does not yet know what the network is running. Until now it answered a peer's app-state sync request anyway, with whatever it happened to hold — and an empty answer and a complete survey are the same three fields on the wire, so the asker counted it as one of the peers it needed. A node could therefore finish its state sync, mark itself ready and start placing apps having learned nothing, without any peer having said anything false.

Alongside that, the request path had no clock of its own. It asked three peers under one shared deadline, so a peer that simply never replied was invisible until the whole budget expired, and nothing re-examined the pool unless something external happened to poke it.

This makes a node answer for itself, gives every request its own clock, and counts peers rather than answers.

## The changes

**A node that cannot answer says so.** A sync response carries `refused` when the responder's own app-state store is not authoritative — its own sync completed, or its block fallback elapsed. The asker records no completion, sets that peer aside, and asks someone else. An empty store that *is* authoritative sends no such field: a network with nothing running is a real survey, and reading it as a refusal would stop such a fleet syncing at all.

**A completion is a peer, and a peer is every stream it was asked for.** Completion requires `appSyncMinCompletions` *distinct* peers. Three answers from one peer used to satisfy a requirement whose entire point is three peers. A completion that cannot be attributed is refused and logged rather than credited.

A request asks each peer for four streams — running apps, installing apps, install errors, and pending registrations — and a peer is credited once all four have ended. Crediting on the three app-state types closed the request record while the fourth was still arriving, and the record closing is what stops this node accepting the rest of it. Pending registrations now report their completion and their refusal on the same terms as the other three. A peer refuses all four or answers all four, so nothing waits on a stream a peer was never going to send.

**Every outstanding request is one record with its own deadlines.** At most `appSyncMinCompletions` requests are outstanding at a time, each held by one peer. A request ends five ways — answered in full, declined, unverifiable, socket closed, deadline passed — and anything but the first frees it to be refilled from the remaining candidates. Two deadlines, because being quiet means two different things:

- `syncTimeoutMs / 12` — *never spoke*, armed when the request is sent
- `syncTimeoutMs / 4` — *spoke then stopped*, measured between arrivals

Both satisfy `deadline + S <= syncTimeoutMs` for a healthy peer's completion time `S`. Per-peer elapsed time is deliberately unbounded: a peer keeps its place for as long as it keeps delivering, because abandoning one most of the way through a large answer is the worst thing this could do.

A peer is credited with answering when its response **arrives**, not when this node finishes processing it. Everything one peer sends is processed through a single queue, so a peer that answered all four requests at once could otherwise be recorded as silent while its answers sat in hand.

**A chunk takes its place in the queue when it arrives.** Its envelope check is started rather than waited for, so the order chunks are processed in is the order they arrived and not the order their checks happened to finish. Chunks carry meaning by position: the sender orders by timestamp and an eviction carries none, so evictions arrive in the first chunk and clear a node's locations outright — a later chunk processed ahead of them has its rows deleted by an eviction that preceded it. The end-of-stream marker is positional for the same reason.

**A chunk this node cannot attribute ends that peer's request.** An answer known to be missing a piece is not a survey, and this node cannot say which piece, because the chunk it could not read is the one it cannot account for. Stepping over it and counting the rest credited a peer with a survey that had a hole in it. The request ends where a refusal or a stall ends it, and another peer is asked.

**The budget bounds the attempt.** When `syncTimeoutMs` expires the attempt is over: outstanding requests are closed, nothing further is asked, and the block fallback carries the node to readiness as it always has. A peer already tried is not tried again. Dropping below the peer threshold and recovering is a restart of the sync, not a continuation, and clears the record.

**One request record, one writer.** "There is a request outstanding to this peer" was recorded in three places — an asked-mark on the peer manager, a slot on the orchestrator, a flag on the socket — maintained by different code and cleared by different rules. It is now a single record holding the connection it was sent on and how it ended. The pool counts it, the deadline hangs off it, the candidate filter reads it, and the response path asks it whether an arriving answer is still wanted, so the gate that admits a response and the deadline that gives up on one cannot disagree.

The request path is a reconciler over that record set. The peer threshold, a peer joining or leaving, a refusal and a deadline all say the state changed; none of them decides anything. A call arriving while a pass is running marks the table dirty and returns, and a pass that fails is reported and ends there — nothing holds the reconciler's promise and one of its callers is a timer, so a throw inside a pass was a rejection with no owner, which node raises to the process handler and answers by exiting.

**A connection ending is announced, and it is a different fact from a peer leaving.** `peerAdded` fires for every connection established, including one replacing a dead socket at an address already held, where no peer was added and the count did not move. `peerRemoved` fires when an address leaves the map. They were never a pair, and the case neither covered is a peer dialling back in over a socket only this node still believed in — which left a request written into a connection that no longer exists, found by sweeping the record set for connection ids that had changed underneath it.

`peerConnected` and `peerDisconnected` replace them, both naming the connection they are about. The replace path announces the socket it drops. `peerRemoved`, the connection-id accessor and the sweep are gone, because the fact they were reconstructing is now stated where it happens.

**A peer whose connection dies mid-attempt has had its turn.** Its record closes rather than being discarded, so the slot it held is offered to a peer that has not answered yet rather than back to one that has just shown it cannot hold a socket long enough to reply. This is what ends a request that was written into a socket closing as the bytes left — the send does not throw, so nothing else says the answer is never coming.

**A node not accepting connections refuses the upgrade.** It answers the WebSocket upgrade with `503` rather than completing the handshake and closing with `4019`. Its capabilities ride in the upgrade response headers, so a completed handshake is enough for the dialling node to construct a peer, count it and write to it before this side has run a line of its own. Scoped to `/ws/flux*`; browsers reach `/ws/id`, `/ws/sign` and `/ws/payment`. `NODE_UNCONFIRMED` also becomes retryable — it says "still booting", which is a statement about *when* rather than about the caller, and it sat in the same bucket as policy violations. The refusal is written to a socket this server now owns: an error listener is attached before writing, and the status is flushed before the socket is destroyed.

**One block fallback for every node.** The enterprise variant is removed. `appSyncFallbackMinutes` is 125 minutes because that is `locationTtlS` — one full lifetime of a running-app location record, and therefore the point at which every holder has had to announce itself at least once. The shorter enterprise value followed the spawner's enterprise deferrals, but those are a priority — how long before a node may compete for an app — and halving one grants an advantage. This is how long before a node assumes it knows what the network looks like, and there is no advantage in assuming it sooner: at the fallback a node begins installing apps *and* begins answering other nodes.

**A degraded node stops asking.** Below its peer threshold a node has judged its own gossip unreliable, so a survey gathered from the peers it has left is not one to complete a sync on. Authority already earned is not revoked — a node past its fallback holds a full location lifetime's view, and losing peers does not erase it.

## Mixed fleet

`refused` is signed as part of the payload and read back the same way, so a build that does not know the field verifies the message and ignores it, reading the empty final batch as the completion it would have sent itself. A `503` at the upgrade reaches an older dialler as an ordinary connection error, which is better than the `4019` it used to get, since that was never retried.

`appStateSyncRefusal` joins `FLUX_CAPABILITIES`. A peer advertising it is asked whatever its uptime; a peer that does not keeps the `appSyncMinPeerUptime` bar, because an older build cannot say it has nothing worth surveying — it answers with an empty batch. Dropping the bar for those peers would reintroduce the original behaviour during exactly the window that makes it likely: a rolling upgrade, when many nodes are young at once. The bar and `getPeerFluxUptime` retire with the last build that cannot refuse.

Older builds answer the pending-registration request and mark the end of that stream, so they are credited on all four like anyone else.

## Test harness

Three defaults changed, each affecting every suite:

- **A fleet boots into a network that is already up.** `createTestEnv` boots the minimum number of nodes another node needs for its sync to complete as already-synced. A fleet whose nodes all boot together has nobody who can answer anybody, which is a network-wide cold start and not what a booting node meets. The cold start is asked for by name with `syncedNodes: []`.
- **A stub peer is a peer of the nodes that need to hear it.** A stub is a server with no client: it cannot dial, and it cannot honour `/flux/addoutgoingpeer` either, so which nodes ended up peered with one was whichever happened to reach it. Every stub now asks the nodes it is declared a peer of to dial it, and those peerings are read back from each node's own peer list before the fleet is handed over. A suite that needs a node blind to a stub narrows it by name.
- **A placement order is the order the stamps come out in.** The election order is ranked on `runningSince`, which is written when a container starts, and a node still syncing defers that start until its own sync completes. `placeGAppInOrder` brings every holder to READY before the first placement and waits for each holder's record to carry a stamp before placing the next, so each holder's start falls inside its own step at any load.

Two failures a fleet of real nodes cannot produce are now reachable. `unverifiableSyncPeers` gives a stub that answers a sync request with a correctly shaped response signed over a payload it does not carry — the shape a forgery has on the wire, where a random signature would be caught by the parser instead. `disconnectStub` takes a stub off the network mid-run, so a connection dies under a request that is open. Every real responder signs correctly with its own key, so without these the only answer failures a suite could reach were silence and a refusal.

Suite `83-app-state-sync-liveness` is new: a fleet where nobody can answer completes nothing, the one node that can is asked the moment it joins, a peer that answers unverifiably is set aside, a peer whose connection dies has its slot offered to someone who has not answered, and a second fleet at `appSyncMinCompletions: 3` measures the deficit arithmetic at the production requirement.

## Testing

- Unit: `appSyncOrchestrator` 103, `fluxPeerManager` 226, `fluxCommunication` 82, `fluxCommunicationMessagesSender` 63, `peerCodec` 32, `socketServer` 13, `fluxServer` 8. Every guard added here is proved red by patching the product line and running the one describe.
- Integration: suite 83 is 9 of 9 on a fleet, covering the four-stream credit, the unverifiable answer and the connection death end to end.


Cabecinha84 (commented):
Findings (none blocking, ranked)

  1 — 83-app-state-sync-liveness.js, "asks no more peers at once": the aggregate assertion is fragile.
  The test accumulates peers across every ephemeralSync:requested in the buffer and asserts asked.size <= 3. But the suite's
  own comment says DECLINER is "a seventh candidate that has to be replaced to reach them" — the moment a replacement round
  lands in the buffer before it's read, asked.size is 4 and the test fails. It only passes because the buffer is read
  immediately after the first event arrives live; waitForEvent also checks the pre-existing buffer first, so a fast fleet
  can hand back a buffer that already contains the whole sequence. The per-event peerCount <= 3 check is the real cap
  assertion and is correct. I'd drop the aggregate, or bound it to the peers named by the first event.

  2 — A refusal drops that peer's temp-message stream as collateral.
  respondWithTempMessages is deliberately not gated by appStateAuthoritative — correct, temp messages have nothing to do
  with app-state authority, and a booting peer still holds valid pending registrations. But the first of the three refusals
  closes the request as declined, and from then on isSyncResponseWanted is false, so any temp chunk still in flight is
  dropped at FluxPeerSocket.js:467. The same narrowing exists on the success path: #closeRequest(peerKey, 'answered') fires
  when a peer completes its three app-state types, which can precede its temp stream (that one does an unindexed
  sort({receivedAt: 1}) over the whole collection). This is a pre-existing hole — clearSyncRequested() cut every temp stream
  at collective completion — but the window is now per-peer and earlier. Low impact, cheap to close by letting
  fluxapptempsync through regardless of outcome.

  3 — Progress is credited before envelope verification.
  dispatchSyncResponse (fluxCommunication.js:776) emits EPHEMERAL_SYNC_PROGRESS on arrival; verifyFluxBroadcast runs later
  inside processSyncChunk. A peer we asked can therefore hold its slot for the entire syncTimeoutMs by streaming chunks that
  never verify. Bounded by the budget and only reachable by a peer this node chose to ask, so severity is low — but
  "per-peer elapsed time is deliberately unbounded" is the one place the deadline design has no answer, and that's worth
  stating rather than leaving implicit.

  4 — #reconcile() is fire-and-forget from five call sites, one of them a timer.
  signMsg(...) at appSyncOrchestrator.js:657-660 sits outside the try/catch that wraps nodeSigner(). A throw from
  signer.sign rejects #reconcilePass → #reconcile, and nothing consumes that promise. There is no
  process.on('unhandledRejection') anywhere in the tree, so Node raises it to the uncaughtException handler at
  apiServer.js:232, which calls logErrorAndExit. The same shape existed before (#requestSyncs()), but it is now reachable
  from #onRequestDeadline's setTimeout, #onPeerRemoved, #onEphemeralSyncRefused and peerAdded. One .catch() closes it.

  5 — Operational consequence worth a decision, not a change.
  A whole-network simultaneous cold start now costs ~125 minutes of paused spawning: nobody is authoritative, everyone
  refuses everyone, #syncBudgetSpent latches after 120s, and only a degrade-and-recover reopens it. That is the correct
  trade against the old behaviour (READY with a confidently empty view), and rolling upgrades don't hit it. But it lands in
  the same PR as enterprise nodes going 62 → 125 minutes, and the budget/fallback asymmetry is stark — the node gives up
  permanently after 2 minutes when the recovery road is 125, and peerAdded is explicitly ignored past that point. I'd want
  the team to confirm they accept the worst case rather than discover it.

  6 — Minor. tests/unit/globalconfig/default.js picks up appSyncMinPeerUptime: 7500 but not appSyncFallbackMinutes; the ??
  125 default and the proxyquire-based fallback tests cover it, but the fixture no longer mirrors the shipped config. And
  stop() leaves globalState.appStateAuthoritative and peerManager.syncResponseWanted set.

MorningLightMountain713:
## Integration gate

`20260905T125147Z`, the full 95-suite gate on this tree: **94/95**. The one failure was `96-surplus-trim-and-the-writer`, which is fixed — suites 68, 69, 70, 96 and 97, every user of the placement helper, are green.

**96.** The fixture needs the syncthing seed to be the newest copy. The stamp that ordering rests on is written when a container starts, and a node still syncing defers that start until its own sync completes — so the masterSlave primary, which is the holder placed first, could start last and re-stamp itself newest. `placeGAppInOrder` brings every holder to READY before the first placement, so each holder's start happens inside its own step and the order out is the order asked for at any load.

**89 is not this PR's.** It is a production cold-start standoff in the syncthing leader election: two empty folders, neither elected, containers down indefinitely. Three candidate gates were silent about which one a folder lost at, so the folder state machine now records the reason on the edge where it changes. The root cause is unknown and the subsystem needs an owner.

## Review findings

**`respondWithTempMessages` carries the `appStateAuthoritative` guard** and refuses on the same terms as the three app-state types. The risk is completeness, not authenticity — every temp message is independently re-verified, but a booting node holds an unknown fraction of the network's pending registrations and cannot say which it is missing, which is exactly the partial answer those three refuse to give. Refusing costs a node nothing on its own road to authority: pending registrations count toward no completion.

**The envelope is verified where the response arrives**, ahead of the deadline it renews. Whether a peer signed what it sent is a statement about the peer; storing its payload is a statement about us. Unverifiable bytes are not progress, so a peer cannot hold one of the answer slots for a whole attempt by streaming inside every stall window.

**A round publishes how many requests are outstanding.** Nothing outside can reconstruct it: one pass can publish two events, and a decline is answered without reaching the event stream at all, so the peers named across events cannot be added up.

**A request is opened only once its signature exists.** Signing answers `null` rather than throwing, and that `null` reaching the codec is a throw inside a reconcile that nothing awaits, with records already open and deadlines armed against requests never sent. The signatures are taken before the first record opens and a missing one logs and returns, so a pass that cannot sign costs nothing and its peers stay eligible for the next one. `encodeSignedSyncRequest` refuses a request carrying no key or signature; the decoder already refused both.

**A stopped orchestrator holds no authority.** `stop()` drops `appStateAuthoritative` — this node's claim to know what the network runs, read by the two guards that answer a peer's sync request.

## Open

- **89's root cause.** Production, another subsystem. The diagnostic added here has not yet been seen firing on a real failure.
- `syncChunkQueues` is keyed by `peerKey` rather than `connectionId`. Raised in review, judged harmless, untested.

Cabecinha84 (commented):
Conditions for ACK

  1. Full 95-suite gate re-run at head. The reported 94/95 gate 20260905T125147Z predates the final commit 7c9e9710c
  (15:25 BST), which is the placeGAppInOrder fix for the one failure. Only the 5 placement-helper suites were re-run
  after it. That matters more than usual here because 16efb6df1 changes createTestEnv defaults — every fleet now boots
  with the top N nodes authoritative from block zero — which is a premise change for all 95 suites, not just the new
  one.

  2. Product sign-off on two ops-visible policy changes, not a code change:
  - Enterprise block fallback 62 → 125 minutes (#ensureBlockThreshold deleted). The PR's argument — a fallback is a
    measurement, not a priority — is sound; it's still a doubling of enterprise time-to-READY on that road.
  - The give-up/recover asymmetry. I confirmed the mechanism: #syncBudgetSpent latches at 120s, #reconcilePass returns
    on it, and the only reset is #resetSyncState() from #onPeersDegraded() — which requires DEGRADED, which requires
    peersBelowThreshold, which FluxPeerManager emits only below appSyncDegradedThreshold (4). So a node that burns its
    120s budget and then sits with 5–11 healthy authoritative peers has no road back: it stays SYNCING with the spawner
    paused for the full 125 minutes, and every peerAdded is ignored. This is the correct trade against the old behaviour
    (READY with a confidently empty view), but it should be an explicit decision, not a discovered one.

  Prior review findings — status at head

  ┌─────┬──────────────────────────────┬────────────────────────────────────────────────────────────────────────────┐
  │  #  │           Finding            │                                   Status                                   │
  ├─────┼──────────────────────────────┼────────────────────────────────────────────────────────────────────────────┤
  │ 1   │ Fragile aggregate assertion  │ Fixed — aggregate dropped; per-event peerCount and the new outstanding     │
  │     │ in suite 83                  │ field asserted instead                                                     │
  ├─────┼──────────────────────────────┼────────────────────────────────────────────────────────────────────────────┤
  │ 2   │ Refusal drops the temp       │ Half fixed (below)                                                         │
  │     │ stream as collateral         │                                                                            │
  ├─────┼──────────────────────────────┼────────────────────────────────────────────────────────────────────────────┤
  │     │ Progress credited before     │ Fixed — verifyFluxBroadcast now runs in dispatchSyncResponse before        │
  │ 3   │ envelope verification        │ PROGRESS is emitted and before queueing; skipped for refused; unit-tested  │
  │     │                              │ red                                                                        │
  ├─────┼──────────────────────────────┼────────────────────────────────────────────────────────────────────────────┤
  │ 4   │ #reconcile() fire-and-forget │ Mitigated, not closed (below)                                              │
  ├─────┼──────────────────────────────┼────────────────────────────────────────────────────────────────────────────┤
  │ 5   │ 125-minute cold-start worst  │ Not addressed — this is condition 2                                        │
  │     │ case                         │                                                                            │
  ├─────┼──────────────────────────────┼────────────────────────────────────────────────────────────────────────────┤
  │ 6   │ Config fixture drift; stop() │ Fixed — fixture carries both keys; stop() drops appStateAuthoritative      │
  │     │  leaves state set            │                                                                            │
  └─────┴──────────────────────────────┴────────────────────────────────────────────────────────────────────────────┘

  Findings still live (none blocking)

  A. The answered path still cuts the temp stream. The refusal half is genuinely closed — respondWithTempMessages now
  carries the same appStateAuthoritative guard, so all four streams refuse together. The other half isn't. The four
  responders run concurrently, each in its own setImmediate (fluxCommunication.js:859-892), so a temp stream can finish
  after the three app-state types. The moment those three complete, #closeRequest(peerKey, 'answered') fires and
  isSyncResponseWanted goes false, dropping any remaining temp chunk at FluxPeerSocket.js:467. On development the window
  ran until collective completion; it is now per-peer and earlier. Low impact — pending registrations self-heal via
  broadcast, and the temp collection is normally the fast stream — but the commit message's "on the same terms as the
  rest" only covers the refusal path.

  B. #reconcile() still has no .catch() at any of its five call sites (appSyncOrchestrator.js:186, 461, 483, 528, 538),
  one of which is #onRequestDeadline's setTimeout. The specific reachable throw is gone — all four signatures are now
  taken before the first record opens (:665-673), and I checked that nodeSigner never returns a truthy object with a bad
  pubKey and that verificationHelper.signMessage catches internally and returns null, so the new
  encodeSignedSyncRequest throws are unreachable from this path. But there is still no process.on('unhandledRejection')
  in the tree, so any future throw in #reconcilePass (an injected callback, an event-bus listener) reaches
  apiServer.js's uncaughtException handler and calls logErrorAndExit. One line closes the shape.

  C. Verification moved out of the serialized queue (new, minor). Chunks are now verified concurrently and enqueued in
  completion order rather than arrival order. In practice the only await is getFluxnodesByPubkey with the same pubkey
  for every chunk of a stream, so resumption is FIFO and ordering is preserved — but it is no longer structurally
  guaranteed. A done chunk overtaking a data chunk would credit that type complete early, and once all three completed
  the queued remainder would be dropped.

  D. syncChunkQueues keyed by peerKey, not connectionId (the author's own open item — confirmed reachable). If a peer
  reconnects while an old drain is still running, a chunk on the new connection is pushed into the old state object and
  processed against the old peerSocket, so isSyncResponseWanted rejects it; and the old drainer's
  syncChunkQueues.delete(peerKey) then removes the entry the new connection is using. Needs a reconnect inside an active
  drain, and the deadline recovers it. Keying on connectionId closes it.

  E. Pre-existing, but now load-bearing. #onPeersDegraded() sets DEGRADED only from READY or SYNCING. A node in
  INITIALIZING (threshold crossed before the first block) or RESYNCING (a second peer drop after a recovery) keeps
  reconciling from a peer set it has already judged too small. The new "a degraded node stops asking" guard reads
  #state, so it silently doesn't apply in those two windows. Unchanged from development — a follow-up, not a change
  here.

  F. Scope. Commit 77da64ee5 (syncthing folder state machine) is pure logging/diagnostics for issue 89 and unrelated to
  sync liveness; same for the treeSize/verifyPool timeout bumps. Zero behavioural risk, but it makes the PR harder to
  bisect.

MorningLightMountain713:
## Where the findings landed

**A — the answered path cut the temp stream.** Taken, and the cause was wider than the temp stream. A request asks each peer for four streams and the tally counted three, so a peer was credited while its fourth was still arriving, and the record closing on that credit is what stopped this node accepting the rest of it. All four are counted now; the pending-registration stream reports its completion and its refusal on the same terms as the other three. Since a peer refuses all four or answers all four, nothing waits on a stream that was never coming.

Two corrections to the finding itself. The leftover chunks were not dropped at the socket — that branch falls through to the gossip pipeline, where they are hashed, cached, envelope-verified and then logged as an unrecognised type. And the temp query is index-backed: `serviceManager` creates `{ receivedAt: 1 }` on that collection, so the round-one characterisation of it as an unindexed sort was wrong.

**B — `#reconcile()` fire-and-forget.** Taken. The catch is in `#reconcile()` rather than at the five call sites, so a sixth caller cannot forget it; a failed pass is logged and ends there, which is what the two existing early-return paths already do. Not closed with a process-level `unhandledRejection` handler — that was proposed and rejected on this codebase before, and the reason is recorded in `socketIoServer.js`.

Your clearance of the specific throw checks out: `nodeSigner` returns `null` rather than an object with a bad key, so the `encodeSignedSyncRequest` throws are unreachable from that path.

**C — verification moved out of the serialized queue.** Taken, and it is more serious than "minor". The consequence is not only an early completion. Chunks carry meaning by position: the sender orders by timestamp, an eviction carries none, so evictions arrive in the first chunk and clear a node's locations outright. A later chunk processed ahead of them has its rows deleted by an eviction that preceded it, and that node reads as running nothing until its next broadcast.

A chunk now takes its place in the queue when it arrives, with its envelope check started rather than awaited, so the order is the arrival order again — structurally, as it was before, rather than as a property of a shared latch.

A chunk that fails that check now ends the peer's request instead of being stepped over. An answer known to be missing a piece is not a survey, and this node cannot say which piece, because the chunk it could not read is the one it cannot account for.

**D — `syncChunkQueues` keyed by `peerKey`.** Reachable, as you say, with one correction that raises it: the deadline does not recover it. Progress is credited at arrival on the new connection's own record, so the deadline keeps being renewed while the data is discarded — the peer holds a slot and delivers nothing until the old drain finishes.

Closed by the change below rather than by keying the queue. A peer whose connection dies mid-attempt has had its turn, so it is not asked again on its new connection, so it never holds an open request while its old connection's work is still draining.

**E — DEGRADED reachable only from READY or SYNCING.** Confirmed, and not fixed here. This is already fixed on `feat/fluxos-v9` by `bbee66891`, which replaced the scattered from-state lists with one evaluator computed from levels; its message records the RESYNCING hole as something suite 108 caught on a fleet. `development` has not received that work. A third variant of the fix on this branch would collide with the evaluator when the lines converge, so this belongs to that convergence rather than to this PR.

**F — scope.** Not changed. The syncthing commit alters no control flow: every value it logs is already in scope where it lands, and it follows the edge-based rule — a line when the reason changes, silence while it holds. It is also the only instrument anyone has for an open production standoff, so removing it from this branch delays that for eight lines of logging.

## Also in this revision

A connection ending is announced, and it is a different fact from a peer leaving. `peerAdded` fired for every connection established — including one replacing a dead socket at an address already held, where nothing was added and the count did not move — while `peerRemoved` fired only when an address left the map. They were never a pair, and the case neither covered is a peer dialling back in over a socket only this node still believed in. `peerConnected` and `peerDisconnected` replace them, both naming the connection; the replace path announces the socket it drops; `peerRemoved`, the connection-id accessor and `#sweepRequests` are deleted, because the fact they were reconstructing is now stated where it happens.

## Conditions for ACK

**Enterprise 62 → 125 minutes** — confirmed as intended. The shorter value was a defect rather than a policy: that timer measures when information has certainly arrived, and a node cannot be given information sooner by shortening it.

**The give-up/recover asymmetry** — confirmed as designed. Worth noting it is not new: on `development` the request path had a single trigger and nothing re-entered it after the timeout either, so a node that burnt its budget also waited out the block fallback.

**The full gate.** #1794 is now stacked on top of this branch, so the gate runs there, above both sets of commits. #1797 merges first.

Unit at head: `appSyncOrchestrator` 103, `fluxPeerManager` 226, `fluxCommunication` 82, `fluxCommunicationMessagesSender` 63, `peerCodec` 32, `socketServer` 13, `fluxServer` 8. Suite 83 is 9 of 9 on a fleet, covering the four-stream credit, the unverifiable answer and the connection death end to end. The harness gained the two failures a fleet of real nodes cannot produce: a peer that answers with an envelope nobody can attribute to it, and a stub taken off the network under an open request.

Cabecinha84 (approved):
ack

PR #1801, opened 2026-09-12
Narrow the node's HTTP surface, give fluxbench a health endpoint, and fix six correctness defects

MorningLightMountain713:
Twenty-eight commits that narrow the node's HTTP surface to what it actually owns, give the node a dedicated place to answer "is this node fit?" other than an auth primitive, and fix six correctness defects found along the way.

`ZelBack/src/services/syncthingService.js` sheds 1,272 lines against 163 added; `routes.js` loses 55 of its 60 `/syncthing/*` GET routes.

---

## 1. The syncthing API was proxied over HTTP, and nothing consumed it

Syncthing is an internal implementation detail of a node. FluxOS was re-exporting its read and control API as 55 GET routes — folder and device config, the node's own filesystem browser, logs, the event stream, the GUI config (which carries syncthing's own api key), plus restart, shutdown, upgrade, reset, pause, resume and discovery. Some of those controls are destructive: `reset` erases syncthing's index database and forces a full resync. Several carried no privilege check at all.

Deleted rather than gated, because nothing consumed them as routes:

- no internal call site — where FluxOS needs this data it already calls the underlying function in-process (`getConfigFolders()`, `getHealth()`, `systemVersion()`, `getDbStatus()`, `systemRestart()` …), never the route;
- no reference in `fluxos-frontend`, `flux-domain-manager`, `fluxstats`, the watchdog or the backup middleware;
- measured peer traffic to `/syncthing/*` is **exclusively `deviceid`**;
- none of them is published in flux-spec.

The underlying functions are untouched — only the HTTP wrappers, their route registrations and their exports go. **Removing the routes removes no capability.**

Three of the deleted routes (`debug/cpuprof`, `debug/heapprof`, `debug/file`) had never worked in the first place: they called bare `axios` with a relative path and no `baseURL`, so every request failed at URL parsing and answered `ERR_INVALID_URL` on any node, in any configuration, since 2023.

Four functions were left exported to nobody once their routes went. Three are still called inside the module, so only the export is removed; `getConfigGui` is among them, and it is the call that reads syncthing's api key out of the config. `systemResetFolderId` had no caller at all — not internal, and never referenced by `routes.js` in its history — so the function goes with the export. It erases a folder's index database and restarts syncthing.

**What stays:** `deviceid` (a FluxOS peer-protocol endpoint, called unauthenticated by other nodes so the fleet can learn each other's device ids), the four FluxOS-owned computed endpoints — `metrics`, `metrics/health`, `metrics/history`, `peer/diagnostics` — which return our own contract rather than a raw proxy, and the write routes.

Anything worth surfacing in future should be a FluxOS-owned endpoint that queries syncthing internally and returns our contract, not a passthrough.

## 2. A syncthing blip could fail a node's benchmark, and a broken syncthing could not fail anything

`syncthingStatusOk` — the flag `isRunning()` read — was set as a **side effect of `getDeviceId`**, which ran on every peer request to `/syncthing/deviceid`. That probe is three sequential calls on a 5s timeout behind a lock, and a single failure flipped the flag false. So a peer asking this node for its device id while syncthing was momentarily busy could flip the node's health. On Arcane nothing ever refreshed it back: the only refresh path was gated behind `!isArcane`.

The blip then propagated: `isRunning` false → `loginPhrase` throws → the health probe polling it fails → the node's benchmark fails. That is the shape behind "Error on FluxOs response" benchmark failures on nodes with no lasting fault.

- **`probeSyncthing()`** is the raw, side-effect-free probe. It parses the meta body inside its own try, so a body that answers but does not parse is a failed probe rather than an exception thrown at whoever asked.
- **`getDeviceId()`** returns a cached id and no longer touches health. The device id is SHA-256 of syncthing's certificate — immutable for the life of the install (verified against syncthing's `lib/protocol/deviceid.go`) — so peers now cost a lookup rather than three requests. The cache is cleared in `stopSyncthing`, the only window a new certificate could appear.
- **Health is derived, not stored.** `refreshSyncthingHealth()` records `lastHealthyProbeAt` on success and writes nothing on failure; `isRunning()` answers whether a probe succeeded inside a five-minute window. The sentinel calls it every loop, on Arcane and non-Arcane alike.

Deriving it is the point. A stored verdict has to be lowered by somebody, so every path that fails to reach the lowering line reports health with nothing behind it — and the paths that fail to reach it are exactly the ones where syncthing is broken. A throw out of `ensureSyncthingRunning` (whose first act is to *stop* syncthing) skipped it; so did a sentinel that never starts because the binary is missing, and a probe that threw rather than returned. Those are not four defects to patch: they are one default asserting something it has not measured. Staleness cannot be skipped. The window also absorbs the consecutive-failure counter it replaces, which was approximating a freshness bound and only counted the passes that actually ran.

A blip can no longer fail a benchmark. A real outage still does, and now so does a sentinel that has stopped running.

**Health has three states, not two.** `healthState()` answers `ok`, `unhealthy` or `unmeasured`, and unmeasured refuses nobody. Before the sentinel has taken the question on there is no reading, and inventing one is wrong whichever way it is invented: a healthy default asserts a probe that never ran, an unhealthy one blames syncthing for a check this node has not made. The sentinel declares itself on its first line, above the wait for the syncthing binary — a legacy node missing that executable never leaves the wait, and that node has to read unhealthy rather than sit unmeasured for ever. It starts once, because boot retries itself every 15s after a throw and restarting the window on each retry would be the same hiding place.

**The sentinel starts above the daemon wait.** It has no daemon dependency: it talks to syncthing's own API, env and userconfig. Below the wait, a node whose fluxd was slow to warm up never measured syncthing at all, and `/id/loginphrase` answered "Syncthing is not running properly" while syncthing was fine. Both stamps are on the monotonic clock — this is boot, which is when ntp steps the wall clock, and a window measured on `Date.now()` then expires on the spot or never. The window and the sentinel's cadence are config (`syncthing.healthWindowMs`, `syncthing.sentinelIntervalMs`); the window is four missed passes of that loop, so shortening one alone lets a healthy node go stale between two good probes.

**Whether FluxOS supervises syncthing is one question, asked in one place.** `supervisesSyncthing()` answers it — syncthing is on this host and this is not Arcane — and `idService` asks it for the fitness gate rather than re-deriving it from the node type. It used to be decided from `SYNCTHING_PATH`, alone among eighteen `isArcane` definitions in the repo: a path override that happens to be Arcane-only today, so a legacy operator relocating syncthing's config had FluxOS take the node for ArcaneOS and stand back from a daemon nobody else supervises. `config.syncthing.ip` is `127.0.0.1` on every node on the fleet, so nothing moves there; the harness is where the two answers differed.

**`SYNCTHING_PATH` is a location, and one resolver answers it.** `getConfigFile` honoured it while the repair path built `~/.config/syncthing` directly, so an operator who set it had FluxOS create, chown and spawn into one directory and read the api key — which authenticates every syncthing call — out of another install's `config.xml`. `syncthingHomeDir()` answers for all of them. `syncthingOwnedDirs()` answers the related one: the FluxOS user owns the syncthing home always, and `~/.config` above it only while the home is still in there — following the home wherever an operator put it would hand the user something like `/var/lib`.

## 3. `GET /flux/health` — fitness has its own endpoint now

`loginPhrase` had grown a node-fitness battery (db, syncthing, docker, hardware, DOS, apps-DOS) for one reason: the node's health is probed through `/id/loginphrase` and there was nowhere else to ask. An auth primitive should not be the fitness gate.

`checkNodeFitness()` holds that battery as one shared function. `GET /flux/health` returns its result — success with a per-check data object, or the first failing check in **the same error envelope `loginPhrase` produces**, so a probe can move across with a URL change and identical gating. `loginPhrase` now calls the same function, so the two cannot drift.

**Syncthing decides fitness wherever FluxOS owns the daemon.** That check answers "has the operator broken syncthing", which is why the whole repair path — stop, install, configure, spawn — and the ownership dance ask the same question. Where we do not own it, FluxOS has no way to restart it, so failing the node there would take it off the network for a fault it neither caused nor can repair. `/flux/health` reports `syncthing: 'degraded'` there instead of failing, and `'unmeasured'` where no probe has landed yet — neither refuses the node. The app path is unaffected either way: mount-safety, folder state and the `g:` election each hold their own syncthing readiness and stand down on it without help from here.

Worth stating plainly, because it is a behaviour change nobody has run: before this branch, an Arcane node's health flag moved only when a peer asked for its device id, so a node hosting nothing synced was never checked at all. It is checked now.

**Three dead branches went with the move.** `getDOSState` answers with `createDataMessage`, which hardcodes `status: 'success'`, so the branch returning "Unable to check DOS state" could not run and the `status === 'success'` test around the rest was always true. It answers with `dosState` and `dosMessage` and nothing else — `nodeHardwareSpecsGood` is emitted by nothing in this repo, fluxbench or the frontend, so `undefined === false` made both conditions reading it permanently false, including the one returning code 100 for hardware. That question is already answered live four lines above by `confirmNodeTierHardware`. The two tests covering those branches went too: both stubbed `getDOSState` to return a shape the real function cannot produce, so each passed by asserting how unreachable code behaves. Removing them exposed that the gate's two live terms were never covered separately — every remaining case set a score above the bar *and* a message — so there is now one test each.

**Failures are logged on the transition, not on the condition.** Moving the battery into a shared function dropped the diagnostics `loginPhrase` used to produce, and restoring them at the point of refusal would have been worse than the silence: `/id/loginphrase` is uncached and fluxbench retries it on a 30s sleep while a node is failing, so that is roughly 2,880 lines a day for a state that has not moved, burying the one line that says when it started. The verdict is remembered and a line is written only when it changes — one when the node stops being fit, naming the check, one when it is fit again. Keyed on the check and its message, never the DOS code, which climbs as the score accrues.

One intentional behaviour change: the `firstLoginPhraseExecution` one-strike grace is removed. It existed to absorb syncthing probe blips; §2 handles that at the source.

This is the **additive** first step: nothing is taken away from `loginPhrase` here. Stripping it to a pure phrase primitive comes once the health probe has moved to the new endpoint.

## 4. A ping in flight is not a missed pong

`missedPongs` was incremented when a ping was **sent**, so it counted pings sent since the last pong rather than pongs missed. A peer answering every ping perfectly still read as having missed one for the whole round trip after each ping, and every reader of the counter believed it.

`getEligibleSyncPeers` believed it, and `pingAll()` pings every peer in one loop — so for that window **every** peer failed its `missedPongs !== 0` bar at once and the pool of peers worth asking for app state was empty however healthy the fleet was.

That window is not the RTT-sized race it looks like. A silent peer is terminated from inside `onPingSent`, which runs inside that same `pingAll` loop, so the disconnect fires at the moment every surviving peer has just been pinged and none has been credited with a pong yet. On a node that is **20ms** from "missed pongs, terminating" to "No peer left to ask, 1 state-sync answer is still needed": the one moment the node must find a replacement peer is the moment it is structurally blind to all of them.

**The fix is to the counter, not to its readers.** A ping going out marks the peer as owing a pong; the *next* ping decides the previous one was missed, because by then it has had a full interval to arrive. A peer answering normally sits at 0 permanently, so `getEligibleSyncPeers`' existing bar is correct as written and is not touched.

A silent peer now takes `maxMissedPongs` whole intervals to drop rather than one short — ~45s at the production interval — which is what `peerResponsiveness()` already documents ("keeps a full peer list for ~45s", "missedPongs moves on the first missed round"). Neither was true of the old counter, which dropped at 30s and moved at 0s: the code was running a round early and the comment described the intent.

## 5. The reinstall pass could destroy the app it was updating

`reinstallOldApplications` is started **fire-and-forget** from the block scanner, so a pass that outlives the gap between blocks is simply run again on top of itself.

Two passes on one app destroy it. Two passes 10.1s apart on one app: the first has soft-uninstalled the component and is sleeping out `composedDelay` before reinstalling it; the second starts on the same app and asks docker to remove the container the first has already begun removing —

```
(HTTP code 409) unexpected - removal of container 376b1d1d1732 is already in progress
```

That throw lands in the redeployment catch, whose cleanup **force-removes the entire app**. The first pass then woke, tried to install, and was refused with "Another application is undergoing removal". The app was left deleted with nothing to put it back. **Changing a specification is an ordinary thing to do to a running app.**

The obvious guard is the flag every neighbour already respects — `globalState.reinstallationOfOldAppsInProgress`, which `forceAppRemovals` and the spawner stand aside on. It cannot do this job. That flag is a *signal to other subsystems*, so it is raised late and deliberately, only once the pass commits to destructive work on an app; raised at the door it would hold the spawner and `forceAppRemovals` off for every scan, including the ones that find nothing to do. A re-entrancy lock has the opposite requirement — it is worthless anywhere but the first line. One flag cannot be raised at two different moments, and guarding on it leaves the whole setup window open: a daemon sync call, a database read and one spec lookup per installed app. A second pass starting inside that window sees the flag down, and nothing stops it afterwards either, because the per-app guards ask `operationHolding('reinstallation')` — which excludes that very flag so a pass can ignore its own, and therefore ignores the other pass's too.

So it is two mechanisms. The signal keeps its timing. A module-scope `AsyncLock` is taken on the first line, checked and registered with no await between them, and released in a `finally` that also clears the signal — so no return or throw inside the body can leak either.

Skipped rather than queued: the scanner runs the pass again on a later block and the app is still obsolete then, so declining loses no work.

## 6. Three API answers that were a TypeError or the wrong privilege

- **`restartFluxOS` was gated on the node operator alone**, the only endpoint in its family to be. Every sibling admits `adminandfluxteam`, including `restartBenchmark`, `restartDaemon` and `hardUpdateFlux` (which deletes `node_modules`, `ZelFront` and `package-lock.json`). The value had been `'admin'` since the endpoint was written and the later rename to `Privilege.NODE_OPERATOR` carried it across unchanged — nothing ever chose it. The asymmetry made the team's own update path unfinishable: `softupdateinstall` advances the checkout but nothing could then restart the process, so the node ran old code off a new tree — and the watchdog, which compares the tree to the remote rather than to what is running, saw them equal and stopped trying.
- **`removeAppLocallyApi` read `appname.includes('_')` above the guard that rejects an absent appname**, so omitting it answered a `TypeError` instead of "No Flux App specified". The guard moves above the dereference. The existing test asserted only that *something* was answered and *something* logged — both true of the crash — so it now pins the message.
- **`appExec` used `containers.find(...)` without asking whether it found one**, so a caller naming an app this node does not run got `Cannot read properties of undefined (reading 'Id')`. An owner may ask any node about their app and most nodes do not run it, so that is the ordinary answer, not an edge case. The wording now matches what `appController` already says for the same condition.

---

## Verification

- **Full unit suite: 6,472 passing, 0 failing, 14 pending.**
- **Two harness suites cover the health state on a real fleet**, each on both node types with a syncthing FluxOS owns, so the fitness gate is live rather than carved out. One waits for a reading before asserting, so a syncthing that comes up broken fails it; the other reproduces the daemon warm-up by answering every RPC from the node with `-28 Loading block index...`, which is what `waitForDaemonRpc` loops on.
- **Mutation checks, each branch broken on its own rather than reverting a change wholesale.** The reinstall lock: dropping the door check, dropping the registration and dropping the release each turn a different test red. Syncthing health: ignoring the freshness window, not recording a success, moving the meta parse back outside its try, fabricating health at load, treating unmeasured as a fault, the sentinel not declaring itself, the window back on the wall clock, unmeasured never expiring, a second start restarting the window, the recovery line, and the recovery line firing on a first probe each turn a different test red. The fitness gate: applying the syncthing check on Arcane, and not reporting `degraded`, each turn a different test red. The fitness logging: logging every refusal, keying on the DOS code, dropping the refusal line and dropping the recovery line each turn a different test red.
- **The pong counter and the original reinstall guard**: 8 tests and 1 test respectively go red when reverted, each with a neighbouring control that stays green either way.
- `GET /flux/health` is covered against a genuinely booted node in the harness — real db, docker, hardware, DOS and syncthing — pinning the claim the refactor rests on: `/flux/health` and `/id/loginphrase` gate on the same checks, so a fit node clears both together.
- A node ran the API-surface commits of this branch continuously for three days with `NRestarts=0` and nothing at error level in its journal: `/flux/health` returned every check `ok`, the removed routes 404'd, and `deviceid` and `metrics/health` answered 200. That predates the rebase and the commits after it.
- Lint clean on every file touched.

`privilegeCallShape`'s call-site floor drops from 250 to 200. It is a canary against the static analysis silently breaking, not a count of record; removing the syncthing handlers took the real count to 238, and 200 still catches a broken sweep (which would find near zero) without tripping on ordinary route churn.


Cabecinha84 (commented):
Blocking findings

  1. The reinstall re-entrancy guard has a residual window it doesn't close

  ZelBack/src/services/appLifecycle/advancedWorkflows.js:4154

  The new guard reads globalState.reinstallationOfOldAppsInProgress on entry — but that flag is only set at line 4189,
  deep inside the per-app loop, after generalService.checkSynced() (a daemon RPC), getInstalledAppsFromDb(), and one
  getStrictApplicationSpecifications() per installed app.

  Two passes launched inside that window both clear the guard. Neither then stops the other: the per-app guards at
  4265/4354/4389 call operationHolding('reinstallation'), which deliberately excludes this very flag. So the exact
  sequence the PR describes — 409 removal … already in progress → redeployment catch → removeAppLocally(…, true, true,
  true) on the whole app → first pass wakes to "Another application is undergoing removal" → app gone — can still
  happen.

  The PR's own observed incident was two passes 10.1s apart. On a node with several apps where the obsolete one isn't
  first in the list, the pre-flag window is plausibly that long. The fix removes most of the bug, not all of it.

  Fix (cannot broaden what other guards see):
  let reinstallPassRunning = false;   // module scope

  async function reinstallOldApplications() {
    if (reinstallPassRunning) { log.info('…already running…'); return; }
    reinstallPassRunning = true;
    try { /* existing body */ } finally { reinstallPassRunning = false; }
  }
  Using a separate flag rather than setting reinstallationOfOldAppsInProgress at entry matters — setting the shared one
  early would make forceAppRemovals and operationHolding stand aside for the whole scan, not just the destructive part.

  2. The syncthing health flag now fails open on the path the PR is hardening

  ZelBack/src/services/syncthingService.js:1688-1700

  In development, ensureSyncthingRunning called getDeviceId(), which set syncthingStatusOk = false as a side effect on a
  failed probe — before the repair path ran. The PR replaces that with side-effect-free probeSyncthing() and moves flag
  maintenance to refreshSyncthingHealth(), called after ensureSyncthingRunning inside the same try:

  try {
    if (!isArcane) await ensureSyncthingRunning(installed);   // stop → reinstall → configure → spawn
    await refreshSyncthingHealth();                            // ← skipped if the above throws
    …
  } catch (error) { log.error(error); return 2 * 60 * 1000; }

  If stopSyncthing(), installSyncthingIdempotently(), configureDirectories() or the spawn throws — which is precisely
  the state where syncthing is broken (disk full, download failure, fs error) — control jumps to the catch and the flag
  is never updated. syncthingStatusOk initialises to true, so a node that has never had a clean probe reports syncthing
  ok indefinitely, through isRunning(), /id/loginphrase and the new /flux/health — while the sentinel has just
  deliberately stopped syncthing. Development lowered the flag on every such pass.

  Fix:
  try {
    if (!isArcane) await ensureSyncthingRunning(installed);
  } finally {
    await refreshSyncthingHealth();
  }

  ---

  Process condition

  The harness gate was not run at this head. The PR body cites "100 suites, 100 passed, run against 7d3413ea1". That
  commit is the pre-rebase copy of the tip and is not an ancestor of b9b41d609. I diffed the two trees: 45 files,
  ~2930/1034 lines of other people's work landed in between (#1802, #1803) — including fluxPeerManager.test.js,
  appSyncOrchestrator.test.js, the new peerSetStabilityService, and harness suites 14/19/55/69/70/97 plus two new ones.

  Good news: I confirmed the PR's own delta is byte-identical before and after the rebase (the only differences are hunk
  line offsets), so nothing was lost. But the gate result doesn't transfer, and the overlap is not incidental —
  development's new peer-set-collapse and sync-readiness work reads getEligibleSyncPeers, whose peer.missedPongs !== 0
  bar is exactly what §4 changes. By inspection the interaction is benign or beneficial (silent peers now linger ~15s
  longer, so dips are if anything less likely, and healthy peers stop flapping to 1), but the harness is the right place
  to prove it.

  Ask for a gate re-run at b9b41d609.

  ---

  What I verified as correct

  - Unit suite green at this head, run locally: 6454 passing, 0 failing, 1 pending. Lint clean on all 15 touched files.
  - No dangling references to any of the 55 removed exports anywhere in the repo (I enumerated them mechanically; the
    three apparent hits are unrelated local identifiers).
  - No consumer of any removed route — in this repo, or in fluxosuidevelopment, fdmmain, flux-apps-dns-manager,
    fluxstatsmaster, watchdogupdateui/watchdogfixui, fluxbenchmaster, fluxview, fluxmonitormaster (all checked out
    locally). Only /syncthing/deviceid is consumed, at syncthingMonitorHelpers.js:29. The UI's Syncthing version tile
    comes via /flux/info → fluxService.js:1353 calling systemVersion() in-process — untouched. The claim holds.
  - syncthingCallShape (the static Api/internal separation invariant) and routeWiring both still pass — the refactor
    didn't break the separation those exist to protect.
  - Device-id cache is safe. refreshSyncthingHealth re-writes cachedDeviceId on every successful probe (every 60s), so
    even on Arcane — where stopSyncthing never runs — a replaced certificate self-heals within a minute. The
    getDeviceIdApi change (serving a cached id while syncthing is down) is an improvement: the id is still correct, and
    its only consumer treats it as data, not as a liveness signal.
  - Pong counter. Terminate now fires at maxMissedPongs intervals (~45s) instead of one early (~30s), matching what
    peerResponsiveness() already documented. getEligibleSyncPeers' bar is untouched and is now correct as written.
  - Privilege widening on restartFluxOS is consistent, not arbitrary. hardUpdateFlux, softUpdateFluxInstall, rebuildUi,
    restartBenchmark, restartDaemon are all adminandfluxteam; what stays NODE_OPERATOR-only is node configuration
    (adjustKadenaAccount, adjustRouterIP, adjustAPIPort, reindexDaemon). Still — it is a deliberate widening of who can
    restart FluxOS fleet-wide, so get an explicit sign-off from whoever owns the privilege model rather than letting it
    ride in on a 15-commit PR.
  - /flux/health leaks nothing new. It is unauthenticated and returns the same error envelope and the same dosMessage /
    db / docker error strings that unauthenticated /id/loginphrase and /flux/dosstate already return.

  ---

  Non-blocking

  - checkNodeFitness JSDoc says "never throws" — it isn't true. confirmNodeTierHardware(),
    fluxNetworkHelper.getDOSState() and appInspector.getAppsDOSState() are all called outside any try inside it. Both
    callers wrap it, so the behaviour is fine; the comment is the thing that's wrong, and this codebase treats comments
    as load-bearing.
  - Lost diagnostics. A non-ok fitness result is no longer logged in either loginPhrase or nodeHealth. Development
    logged it (db/docker failures hit the outer catch's log.error, and the apps-DOS branch had an explicit
    log.error(errMessage)). These are exactly the "Error on FluxOs response" diagnostics §2 is about — a log.warn on the
    failing check is worth keeping.
  - /flux/health sits behind cache('30 seconds'). Safe by the repo's own stated rule (that rule is about
    privilege-checked routes; this one is unauthenticated with a uniform response), and consistent with /flux/dosstate
    directly above it. But /id/loginphrase is uncached, so the new endpoint is up to 30s staler — worth being explicit
    about, since the PR's claim is that a probe "can move across with a URL change and identical gating".
  - Dead exports left behind. getConfigGui, getConfigOptions, getConfigDefaultsFolder, systemResetFolderId,
    syncthingController, getAxiosCache, getConfigFile now have no caller outside syncthingService.js. Consistent with
    "the underlying functions are untouched", but getConfigGui in particular is the one carrying syncthing's own API key
    — worth a follow-up sweep.
  - Pre-existing comment drift at 4265/4354/4389: "this pass set reinstallationOfOldAppsInProgress before the loop" —
    it's set inside the loop at 4189. Not introduced here; cheap to correct while you're in the file, and it's the same
    misreading that made finding #1 easy to miss.

MorningLightMountain713:
Thanks — both blocking findings were real. Fixed at `3bd1cd11a`, along with most of the non-blocking ones. Two I'm pushing back on, with reasoning.

## 1. Reinstall re-entrancy — agreed, fixed differently

You're right that the guard leaves the setup window open, and right about why nothing stops the second pass afterwards: the per-app guards exclude `reinstallation` so a pass can ignore its own flag, and therefore ignore the other pass's too.

Your fix is sound and your reason for a *separate* flag is the important part — setting the shared one at entry would make `forceAppRemovals` and the spawner stand aside for the whole scan. Implemented with the repo's existing `AsyncLock` rather than a new boolean, since `asyncLock.js` already documents this exact "refuse rather than queue" case and `peerNotification.js` already uses a module-scope lock for the same shape. Checked and registered with no await between them, released in a `finally` that also clears the signal.

One thing worth flagging: the old canary set the flag and called once, which proves a guard reads a flag and would have passed just as happily against the window you found. It now parks the first pass inside `checkSynced` and starts a second.

## 2. Syncthing health failing open — agreed, and it was wider

Your diagnosis is right. Two things it didn't cover:

- `getDeviceId` was a *second* prover before this branch — every peer hitting `/syncthing/deviceid` lowered the flag. Making it side-effect-free left the sentinel as the only one.
- `startSyncthingSentinel` blocks in its binary-wait loop, so a node with no syncthing binary never runs `refreshSyncthingHealth` at all and reports healthy from boot. No throw required.

So the `finally` closes one of three paths. The flag is now derived instead: `refreshSyncthingHealth` records `lastHealthyProbeAt` on success and writes nothing on failure, and `isRunning()` asks whether a probe succeeded inside a five-minute window. Nothing has to write a failure, so nothing can fail to — including a sentinel that has stopped running entirely, which no patch to the lowering line could have covered. The window also absorbs the consecutive-failure counter, which was approximating a freshness bound and only counted the passes that ran.

`probeSyncthing` also parses the meta body inside its own try now, so a body that answers but doesn't parse is a failed probe rather than an exception at the caller.

**Related change you should look at:** the syncthing check now decides fitness on legacy only. It answers "has the operator broken syncthing" — which is why the whole repair path and the ownership dance are `!isArcane`. On Arcane we ship syncthing and FluxOS cannot restart it, so failing the node there takes it off the network for a fault it neither caused nor can repair. `/flux/health` reports `syncthing: 'degraded'` there instead. The app path is unaffected either way — mount-safety, folder state and the `g:` election each hold their own readiness.

## Process condition — agreed, gate run

Your analysis of the rebase was exactly right, including that the delta is byte-identical across it. The gate has now run at this head.

It earned its keep: it caught a defect we introduced in this round, in the same module. `isArcane` in `syncthingService` read `SYNCTHING_PATH` while the other seventeen definitions in the repo read `FLUXOS_PATH`, and unifying them looked safe. It isn't the right question either way — all three uses ask whether *this process owns the daemon*, and that is two facts: Arcane supervises its own, and a syncthing on another host cannot be stopped, reinstalled or chowned from here whatever the node type. The second was never checked. That is what the condition says now.

## Non-blocking

**Taken:**

- *Lost diagnostics* — restored, but on the transition rather than the condition. A line per refusal is ~2,880 a day, since `/id/loginphrase` is uncached and fluxbench retries on a 30s sleep, and it buries the line that says when the node stopped being fit. One line when the verdict changes, one when it recovers.
- *Dead exports* — removed. `systemResetFolderId` had no caller at all, internal or otherwise, and `routes.js` never referenced it in its history, so the function went with the export.
- *Comment drift at 4265/4354/4389* — you're right it's pre-existing (`92260528f`, already in development). Corrected here since the file was open.

**Not taken, with reasons:**

- *`checkNodeFitness` JSDoc "never throws"* — I think this one is wrong. `confirmNodeTierHardware` wraps its whole body in try/catch and returns `false`, and its two awaits carry `.catch()` handlers on top of that. `getDOSState` and `getAppsDOSState` are synchronous object builders — no await, nothing that can fail. The comment is accurate as written.

  There is a real problem two lines away that this turned up: `getDOSState` answers with `createDataMessage`, which hardcodes `status: 'success'`, so the `status === 'error'` branch could not run. And `nodeHardwareSpecsGood` is emitted by nothing — not this repo, not fluxbench, not the frontend — so `undefined === false` made both conditions reading it permanently false, including the one returning code 100. Two tests covered those branches by stubbing shapes the real function cannot produce. All removed. Deleting them exposed that the gate's two live terms were never covered separately, so there's one test each now.

- *`/flux/health` cache staleness* — keeping it. `/id/loginphrase` is uncached because it mints a random phrase per caller and writes it to the DB; cache it and two callers get the same phrase. That says nothing about how fresh a health verdict should be. `/flux/health` is a status read whose underlying state moves on the order of minutes, and the cache also stops every poller driving a docker image list and a DB read. The claim in the description is about gating, which is identical — both call `checkNodeFitness` and return the same envelope.

## restartFluxOS privilege

Deliberate, and the rationale is in the branch. `restartFluxOS` was the only endpoint in its family the team could not reach — every sibling admits `adminandfluxteam`, including `hardUpdateFlux`, which deletes `node_modules`. Without it the team's own update path is unfinishable: `softupdateinstall` advances the checkout and nothing can restart the process, so the node runs old code off a new tree and the watchdog, comparing disk to remote, sees them equal and stops trying. The value had been `'admin'` since the endpoint was written and the later rename carried it across unchanged — nothing ever chose it.

Cabecinha84 (commented):
Non-blocking findings (worth a comment)

  1. Boot-time false "Syncthing is not running properly" (legacy nodes only).
     - The health timestamp is set when the module loads (syncthingService.js:75).
     - But the sentinel only starts at serviceManager.js:608, after waitForDaemonRpc(), which retries every 5 seconds
       with no upper limit.
     - So if the daemon takes more than 5 minutes to warm up or come back, /id/loginphrase fails with a syncthing error
       even though the real cause is the daemon, and fluxbench reports "Error on FluxOs response – Syncthing…".
     - It fixes itself about a minute after the daemon's RPC comes up (fluxbench keeps retrying), but the reason shown
       is wrong. development didn't have this, because its flag started as true.
     - Suggested fix: set the timestamp when startSyncthingSentinel() starts, and/or start the sentinel before the
       daemon wait.
  2. Stale comment in test-infra/runner/framework/test-env.js:1158: it still says SYNCTHING_PATH decides who supervises
     syncthing.
  3. PR description is out of date:
     - The title says "four" defects; the body says six.
     - §2 says isArcane now reads FLUXOS_PATH, but supervision is actually decided by the IP-based supervisesSyncthing
       rule from the last commit.

MorningLightMountain713:
All three answered, at `b2c7bdfd3`. Findings 1 and 2 were both real, and finding 2's comment sat above a live defect.

---

## 1. Boot-time false "Syncthing is not running properly" — real, fixed

Confirmed end to end: the API listens at `apiServer.js:268`, *before* `startFluxFunctions()` at `:289`, so the node answers throughout; the sentinel started at `serviceManager.js:608`, behind a `while (true)` on `waitForDaemonRpc` with no upper bound. A daemon slow to warm outlived a stamp taken at module load, and the node reported a syncthing fault with nothing having looked at syncthing.

**Both halves of your fix landed.** The sentinel stamps itself when it takes the question on, and it now starts above the daemon wait — it has no daemon dependency at all (syncthing's own API, env, userconfig; mongo is up 200 lines earlier).

**One thing on top.** Stamping at sentinel start alone still answers a question the node cannot yet answer: for one window it reports `syncthing: ok` on the strength of a probe that has not run. So health has three states rather than two, and "nothing has measured this" is one of them:

- `measurementStartedAt === null` → **unmeasured**. Refuses nobody. A node that refuses itself over a check it has not run is reporting its own startup as somebody else's fault.
- measuring, no success yet, inside the window → **unmeasured**. Bounded, not open-ended.
- measuring, no success, window elapsed → **unhealthy**. This is the arm that catches a legacy node whose syncthing binary never appears — it never leaves the binary wait, which is why the stamp goes *before* that loop and not after.
- fresh success → **ok**.

`/flux/health` reports `unmeasured` as its own value rather than dressing it as either verdict.

**Two related defects closed with it:**

- **The window was on `Date.now()`.** It is elapsed time and nothing outside the process reads it, so it is on `process.hrtime.bigint()` now. Wall time costs exactly the case the window exists for — a node boots, ntp steps the clock, and the window expires on the spot or never.
- **The transition log compares against what was last reported**, not against a second reading taken inside the same probe. A derived state changes with the clock and the window falls due *between* passes, so the old comparison was of one reading with itself and the line could essentially never fire. It is the same bookkeeping `idService` keeps for the fitness verdict, and it also catches `unmeasured → unhealthy` — a syncthing that never once answered.

The window and the sentinel cadence are config (`syncthing.healthWindowMs`, `syncthing.sentinelIntervalMs`), because the window is four missed passes of that loop and shortening one alone lets a healthy node go stale between two good probes.

## 2. Stale comment in `test-env.js` — real, and there was a defect under it

The comment was wrong, and the `delete` it justified is still right for a different reason, so both changed.

**A second one you did not flag is worse.** `entrypoint.sh:126` asserts the claim as an invariant — *"read the same variable FluxOS reads, not a different one that happens to be set alongside it"* — and stopped being true one commit after it was written. It is not the same variable and should not be: FluxOS supervises a syncthing that is on this host and not Arcane, binary mode settles the first half, so the entrypoint decides the second. It is the **complement** of that rule, not a copy of its input. Both comments now say what is true.

**Underneath them, a real one.** `SYNCTHING_PATH` says where syncthing's config and identity live. Three places needed that answer and one looked: `getConfigFile` honoured it while `configureDirectories` and the spawn built `~/.config/syncthing` directly. An operator who set it had FluxOS create, chown and spawn into one directory and then read a different install's `config.xml` — which is where the api key for every syncthing call comes from. One resolver now, all three call it. Nothing changes on a node that has not set it, which is every node on the fleet.

## 3. Description

Noted. It has moved further out of date with these four commits and will be brought in line with the branch.

---

## Collateral: one you did not raise

`idService` decided the syncthing gate on node type alone, while `syncthingService` decides ownership on *syncthing is on this host and not Arcane*. The comment above the check gives the reason for the carve-out — "a fault it neither caused nor can repair" — and that reason covers both cases while the condition covered one. `idService` asks `syncthingService` now. Nothing changes on the fleet, where `config.syncthing.ip` is `127.0.0.1` on every node; the harness is where the two answers differ.

## Verification

- **Unit: 6,469 passing, 0 failing, 14 pending.** Lint clean on every file touched.
- **Suites 1701 and 1702 on a real fleet: 2/2, `suites_pass=2 suites_fail=0`.** Images verified against the tree, node image md5-matched to source.
- **1701 covers both node types**, each on a syncthing FluxOS owns, so the gate is live rather than carved out. It waits for a reading before asserting, so a syncthing that comes up broken fails it.
- **1702 reproduces the defect rather than imitating it:** `rpcFailures` makes the daemon stub answer `-28 "Loading block index..."`, which is what `waitForDaemonRpc` loops on.
- **Reverting the fix turns all three of 1702's assertions red** — assertion 1 with the node answering `error` and the body reading *"Syncthing is not running properly"* while its daemon loaded the block index.
- **Unit mutations, each branch broken on its own:** fabricating health at load, treating unmeasured as a fault, the sentinel not declaring itself, the window back on the wall clock, unmeasured never expiring, fitness refusing on anything not `ok`, the transition log compared within one pass, and the log ignoring a syncthing that never worked — each turns a different test red, three of them exactly one.

Cabecinha84 (commented):
Before merging

  - Re-run the full gate at b2c7bdfd3. The author ran only suites 1701 and 1702 at this head; the 102/102 result is from
    before these commits.
    - This matters because boot order changed. On development, the sentinel starts after cleanupCrontabAndMounts
      (serviceManager.js:569 → :608). Now it starts at :479, before app volumes are mounted.
    - So after a host reboot, a legacy node now starts syncthing before its loop mounts are up.
    - The design should cover this: the .stfolder marker lives inside the volume, so syncthing halts those folders, and
      the first monitor pass (:661, after mounts) re-checks every mount. The old order was still never tested under the
      full gate.

  Non-blocking items

  - noteMeasurementStarted() can be called more than once. If boot throws after :479, startFluxFunctions retries every
    15s and resets the window each time. While syncthing has never answered, the node then stays unmeasured, which never
    fails it.
    - Separately, if boot throws before :479 every time, the sentinel never starts and the node stays unmeasured
      forever.
    - Neither is a regression: on development, syncthingStatusOk defaults to true, so those nodes pass the check there
      too. A measurementStartedAt === null guard fixes the first case.
  - The "marking syncthing not running" log line can be skipped. It is written only inside refreshSyncthingHealth, which
    never runs when ensureSyncthingRunning throws. The generic log.error still fires every 2 minutes. There is also no
    log line when syncthing recovers.
  - configureDirectories changes ownership of ~/.config even when SYNCTHING_PATH points elsewhere. --home
    ${syncthingHome} is also passed unquoted into a shell command. The path is set by the operator, so this is cosmetic.
  - Response change: /flux/health can now return syncthing: 'unmeasured'. Whatever fluxbench does with that field needs
    to accept this value.
  - The PR body is stale:
    - It says 23 commits; there are 27.
    - It still says the syncthing check decides "legacy only"; it's now "wherever FluxOS owns syncthing".
    - It doesn't mention the unmeasured state or the earlier sentinel start.
    - The title still says "four" defects while the body lists six.

    The author said it will be updated.
  - Still needs sign-off: widening restartFluxOS to NODE_OPERATOR_OR_FLUX_TEAM is a policy decision, unchanged since the
    last review.

MorningLightMountain713:
All of it answered at `21953b40d`, and the gate has run.

## The gate

Full harness gate on chud at `21953b40d`: **103 suites, 102 passed, 1 failed**, each accounted for by its own `DONE` line rather than inferred from the aggregate. Boot order is the thing you wanted covered, and 102 suites boot through the new sequence and pass.

The failure is suite 85, and it is the harness rather than the branch. The suite drives blocks with `driveUntil` while `bootAndPeer`'s ticker is also advancing the chain, so it moves two blocks per drive, every other block is stale on arrival, and the node under test comes to rest on one parity of heights. The periodic-pass modulus is even for every roll of the period, so the reinstall pass cannot fire at any budget. Measured from this run: node-00 processed 109 blocks as tip, **all 109 odd and none even**, and the chain moved 216 blocks for the 108 the suite budgets. Fixed by `e5de22b50` on #1804, which has not reached `development` yet.

## 1. `noteMeasurementStarted()` re-entrancy — real

`startSyncthingSentinel` returns immediately once it has started. The loop and the metrics interval already refused a second start; the stamp and the binary wait above them did not, so the guard belongs on the function rather than on the stamp — which also stops a binary waiter accumulating on every boot retry.

## 2. The transition log

The asymmetry was real: there is a recovery line now, on the unhealthy → ok edge only, since a first probe landing after boot is a node starting up rather than recovering.

The skipped edge is not. Edges are detected against the health last reported, not against a reading taken inside the pass, so a pass that throws before the refresh defers the line rather than losing it. It goes unsaid only if `ensureSyncthingRunning` throws on every pass, and that case writes the underlying error every two minutes.

## 3. `configureDirectories` — real, and there were two of them

`getConfigFile` chowns the same pair on every config read, with `~/.config` hardcoded the same way. One rule now, `syncthingOwnedDirs()`: the syncthing home always, and `~/.config` above it only while the home is still inside it. Following the home wherever an operator relocates it would hand the FluxOS user something like `/var/lib`.

`--home` and `--logfile` are quoted. Both reach a `shell: true` command line, where a path with a space stops syncthing starting. Nothing else needed it: `runCommand` is `execFile` over an argv array, and the daemon cannot go through it because it awaits the process to exit.

## 4. `/flux/health` returning `unmeasured`

No fluxbench change is needed. `CheckFlux` reads the top-level `status` and, on an error, `data.message`; it never looks at `checks.syncthing`. And `unmeasured` cannot produce `status: error` — fitness refuses on `unhealthy` alone, and only where FluxOS owns the daemon.

## 5. The description

Rewritten: title, commit and unit counts, the three health states, the sentinel's position in the boot sequence, `supervisesSyncthing()` as the one ownership question, `SYNCTHING_PATH` as a location with one resolver, and `unmeasured` named as a value the endpoint returns.

## 6. `restartFluxOS`

Signed off. The reasoning is in `92e89044a`: it was the only endpoint in its family gated on the node operator alone, while `restartBenchmark`, `restartDaemon` and `hardUpdateFlux` already admit the team — and the asymmetry left the team's own update path unfinishable, since `softupdateinstall` advances the checkout and nothing could then restart the process.

## Verification

- Unit: **6,472 passing, 0 failing, 14 pending.** Lint clean.
- Mutations, each branch broken on its own: removing the start guard, removing the recovery line, widening it to fire on a first probe, and always chowning `~/.config` each turn a different test red.

Cabecinha84 (approved):
ack

PR #1802, opened 2026-09-12
A banned application is judged on what it is; the spawner stops drawing it; the operator's own lists retire

MorningLightMountain713:
## A banned application is judged on what it is

`checkApplicationsCompliance` asks the identity questions — hash, name and owner — on the raw stored record for every installed application, before any decryption. The decrypt then serves only the question that needs it, which images the components run. An application whose specification cannot be read is still judged on what it is, and is deferred for images with a warning naming it.

Previously the sweep decrypted first and judged only what came back readable. An enterprise application carries its components inside an encrypted blob, so one that failed to decrypt was dropped from the list entirely, and a ban on its hash, name or owner went unexamined rather than unmatched. A second path produced the same outcome even when decryption succeeded: `decryptEnterpriseApps` defaults to `formatSpecs: true`, and `specificationFormatter` rebuilds the object from a fixed field list that omits `hash`, so the hash comparison was testing `includes(undefined)`.

Measured against the live network, of the 126 banned applications holding a current specification:

| | still running | removed |
|---|---|---|
| **enterprise** (60) | **56** | 4 |
| **non-enterprise** (66) | 1 | **65** |

The single non-enterprise survivor runs an older hash that is not on the list, which is correct — the ban is hash-exact.

## The spawner does not draw what it cannot place

Blocked applications are filtered out of candidate selection, as a named `afterBlocklist` stage in the existing candidacy tally.

The aggregation asks one question — which applications have fewer instances than they want — and a blocked application answers yes forever: it has none and can never be given one. Without the filter it is drawn, refused at the install-time compliance check, and drawn again every time the node's error cache expires. Blocked applications were **72 of the 173** applications the network reported short, and **241 of the 595** missing instances. `appSpawnErrorCache` has a seven-day TTL, so the burst recurs weekly.

Only what an application *is* can be judged at selection: the aggregation projects no repotags and an enterprise application carries none in the clear, so image and namespace bans remain the install-time check's to make.

These two changes belong in the same release. The first removes the 56 banned applications still running, which drops all 72 blocked ones to zero instances — more attractive to the aggregation, not less. Shipping the first alone makes the candidate pool worse.

## The operator's blocked ports and blocked repositories are retired

A node operator could refuse ports and docker repositories in `userconfig`, and the node enforced both: a blocked port turned an application away at spawn and removed it from the availability test, and a blocked repository was refused at install and swept off the node if already running. Both are gone, along with `POST /flux/adjustblockedports`, `POST /flux/adjustblockedrepositories` and the two GETs that read them back, the DOS states a node entered for holding more than 100 ports or 10 repositories, and the fields themselves from every template that writes a userconfig.

The vetted list goes with them. `isAppVetted` existed to exempt an application from these two checks and had no other caller, so with them gone it decided nothing; `getVettedRepositories` was its only reader. `vettedrepositories.json` stays in the policy bundle for releases that still read it.

**The network blocklist is untouched.** `blocklist.json` and `blockedrepositories.json` still decide what a node will run.

## The parallel harness runner reaps what no live run owns

`run-parallel` sweeps harness leftovers before and after the gate, scoped by ownership rather than by what is on the box. Every container, network and volume the harness creates carries `flux-e2e-run=<RUN_LABEL>`, and a run records that label in the `/24` claim it holds, so the labels live runs hold are readable from outside them and anything else carrying the key belongs to a run that is gone. A concurrent `run-all.sh`, or a second gate, keeps its own objects.

The claim root and the boot lock are not swept. Both already reclaim per entry on owner liveness — `_scan_and_claim` takes a dead run's `/24` under flock, and `bootQueue` drops a ticket whose pid is gone on every read — so removing either wholesale reclaims nothing that is not already reclaimed, and takes coordination state that live runs are holding.

Networks are taken by name as well as by label, because `runLabels()` is empty without `E2E_RUN_LABEL` and a suite run outside `run-all.sh` leaves a `flux-test-` network no filter can find. A live fleet's unlabelled network survives regardless: one with attached endpoints refuses removal, and its containers are unlabelled too.

A live claim naming no run label cannot be identified, so the sweep stands down and names it. That is the only case that still needs the box to itself.

## Rollout

**fluxos-frontend#164 deploys before this merges.** The administration page reads both retired GET endpoints on mount, so a frontend still carrying those controls against a node running this branch gets 404s on load and two buttons that post to routes which no longer exist.

## Evidence

| | |
|---|---|
| Unit suite | **6392 passing, 0 failing** |
| Harness suite `1601-blocklist-enforcement` (new) | passes, and **fails 3/3 on unfixed code** — mutation-tested by reverting the two product files and building real images from that tree |
| Spawner suites | 9/9 green: 1601, 09, 16, 21, 27, 39, 76, 78, 98 |
| Full gate | **99 pass, 1 fail** |
| Sweep against a live fleet | takes nothing: 9 harness containers before and after, network, `/24` claim and boot lock all intact |
| End to end | administration page against a node running this branch: the section is gone and nothing adjacent to it changes; both retired routes answer 404 while neighbouring routes are unaffected |

The single gate failure is `85-app-spec-update-reinstall`, which is a defect on `development` rather than in this branch: `reinstallOldApplications` has no mutual exclusion with itself, so two passes ten seconds apart race and the second force-removes the application the first is installing. Confirmed from that run's own node logs, and fixed by `7d3413ea1` in #1801.

## Notes for review

`getBlocklist()` keeps `null` ("the list could not be obtained") and `[]` ("nothing is blocked") distinct, and two tests pin it. A null list skips the verdicts that need the document and nothing else, so the sweep still applies what it can decide without one rather than enforcing nothing that pass.

`blockedrepositories.json` remains the fallback and its entries keep their older meaning, one bare string tested against hash, owner, repository and namespace alike, so nothing already in force stops being enforced. Its entries are marked `legacy` rather than guessed at, because guessing a kind is the ambiguity the typed document exists to remove.

An unknown `kind` refuses nothing, which is what lets a newer policy document ship before the reader that understands it.

Companion: RunOnFlux/fluxos-frontend#164


Cabecinha84 (commented):
1. E2E_SWEEP_ANYWAY=1 prints that it swept, then doesn't — test-infra/runner/run-parallel.sh:250

  [ -n "$claims" ] && echo "WARNING sweeping past a live claim because E2E_SWEEP_ANYWAY is set" >&2
  if [ -n "$(live_claims)" ]; then      # ← re-tests, still true
    log "SKIP    post-gate sweep ..."   # ← so it skips

  The refusal text advertises E2E_SWEEP_ANYWAY=1 as the override for known-dead claims. With it set, the run proceeds,
  announces it is sweeping past the claim, and then skips the sweep — leaving the lock root the pre-gate sweep exists to
  clear. The message and the behaviour contradict each other.

  2. The guard landed on the pre-gate sweep; the post-gate one is still unconditional — run-parallel.sh:251 vs :346

  The skip message says "post-gate sweep", but it sits at the pre-gate call site. Line 346 still calls
  sweep_harness_leftovers with no guard at all. So the PR body's claim —

  ▎ "After the gate the sweep is skipped rather than refused, since our own claims are released by then."

  — is not implemented. And this is the case that actually bites: the refusal text itself says "A single suite
  (run-all.sh) CAN share the box". If someone starts a run-all.sh alongside a running gate, run-parallel's post-gate
  sweep force-removes every flux-e2e-run container, every flux-test- network, and rm -rfs the lock root out from under
  it — precisely the destruction this PR set out to prevent. The guard block looks like it was written for line 346 and
  placed at 251.

  3. No negative caching on the typed document — imageManager.js:232

  getBlocklist() caches only on the success path. When blocklist.json is absent, empty, or non-array, every call issues
  a fresh axiosGet — and the spawner now calls it every pass (30 s enterprise / 60 s otherwise,
  enterpriseNetwork.js:135), where before this PR the network was only touched from the install path. Moot today because
  the document is live and caches for 6 h. But if it is ever unpublished, renamed, or rate-limited, the whole fleet
  polls raw.githubusercontent.com every 30 s indefinitely, and the fallback's own failure isn't cached either. Worth a
  short negative-cache TTL on the miss.

  Nits

  - imageManager.js:14 — orphaned // Cache for blocked repositories comment, its variable is gone.
  - imageManager.js:452 — double blank line before return true.
  - blockedReasonFor does .map(...).find(...), evaluating all 310 entries per subject instead of short-circuiting.
    Negligible, but a for loop with early exit is strictly less work on a path that now runs per candidate per pass.

  Rollout gates (not defects, but they decide the merge order)

  - fluxos-frontend#164 is still OPEN (af3239291, mergeable). The PR body correctly states it must deploy first — the
    administration page reads both retired GETs on mount. Confirm the deploy, not just the merge.
  - The reported gate result is 99 pass / 1 fail, attributed to a reinstallOldApplications self-race fixed by 7d3413ea1
    in #1801. I checked: #1801 is open and unmerged, and 7d3413ea1 exists only on origin/fix/api-refinements. The
    attribution is honest and plausible, but that suite will keep failing on re-run until #1801 lands — so don't treat a
    green gate as a precondition here, or land #1801 first.
  - Intended network effect on merge: ~56 currently-running banned enterprise applications get removed fleet-wide.
    During rollout, nodes still on the old release will keep drawing and refusing those apps (their flat document
    matches by hash/owner) until they upgrade — transient extra churn, self-resolving.
  - One-way change to operator files: every userconfig-rewriting endpoint (adjustKadenaAccount, adjustRouterIP,
    adjustAPIPort, adjustExternalIP, adjustPGPidentity) now drops blockedPorts/blockedRepositories from an existing
    config/userconfig.js. Nothing reads them any more so it's harmless, but the operator's values are gone for good once
    any of those routes is hit.

MorningLightMountain713:
Both harness findings were right, and they were one mistake: the guard block was written for the post-gate call site and placed at the pre-gate one. Rather than move it, the sweep no longer needs it.

The sweep is now scoped by ownership. Every container, network and volume the harness creates carries `flux-e2e-run=<RUN_LABEL>`, and `run-all.sh` records that label in the `/24` claim it holds, so the labels live runs hold are readable from outside them and anything else carrying the key is a leftover. The two lock directories are not swept at all, and nothing replaces that: `_scan_and_claim` already reclaims a dead run's `/24` under flock and `bootQueue` already drops a ticket whose pid is gone on every read, so deleting them wholesale reclaimed nothing that was not already reclaimed and was the only part of the sweep that could destroy a live run. `live_claims`, the refusal and `E2E_SWEEP_ANYWAY` are gone with the constraint, and both call sites are plain unconditional calls. `run-parallel` and `run-all.sh` now share a box in both directions, and so do two gates.

One case that fell out of testing it: a claim written by a clone predating the label file names no run label, so a live run would have been invisible and swept. The sweep stands down and names the claim when it finds a live run it cannot identify.

Verified against real docker on a box with one live run and one dead one: the dead run's container, network and volume went, an unlabelled `flux-test-` network went, the live run's three survived, and both lock directories were untouched. Then against a live fleet mid-suite, where it took nothing — 9 harness containers before and after, network, claim and boot lock intact. Suite 1601 passed 3/3 over it.

On the negative cache: not doing it, because the fetch it would protect does not survive the stack. #1804 sits 61 commits directly on this tip and replaces `getBlocklist` with a synchronous read of a locally held signed bundle — peers before the published source, one fetch in flight at a time, no boot fetch for a node that restored from disk, and a 24-hour poll on a slot derived from a hash of the node's collateral so the fleet is spread across the period rather than ticking together after a release wave. The 30-second spawner call that made the finding worth raising becomes a local memory read. A short negative TTL is a weaker answer to the same question, and it would be deleted at that merge.

Worth noting for anyone reading this later: caching the miss would also have been wrong on its own terms. A null blocklist makes `checkApplicationImagesCompliance` throw, and that is on the live app registration path, so a cached miss refuses users for the length of the TTL including after the document returns.

All three nits are fixed. `blockedReasonFor` now names the per-entry verdict and returns at the first refusal instead of mapping all 310 entries and picking from the array — same order, same answer.

On the rollout notes, all four are accurate. The userconfig rewrite dropping `blockedPorts` and `blockedRepositories` is intended rather than incidental: the fields are retired, and a node that rewrites its config stops carrying them.

Cabecinha84 (approved):
ack

PR #1803, opened 2026-09-12
Readiness is time spent with peers, and a node whose peer set keeps collapsing is taken out of service

MorningLightMountain713:
## Readiness is time spent with peers, not uptime

A node reaches spawning readiness only while it holds peers.
`#blocksSinceSyncStarted` advances only while the peer threshold is met and is
zeroed when it is lost, so a peerless node's timer never expires. The invariant
falls out of the counter rather than being enforced by a gate, and every consumer
of that timer inherits it without a second condition to forget.

Required, because `#checkReadiness()` has no peer condition: every gate peers would
satisfy is waived by the block fallback, and `#canSendMessages` is on-chain
confirmation, which says nothing about peers. A node that never joined the network
reached readiness on elapsed time alone and began spawning. `peersBelowThreshold`
cannot catch it — it fires only `if (#aboveThreshold && size < degraded)`, and
`#aboveThreshold` latches on the way up, so the demotion path covers a node that
fell off the network and not one that was never on it.

Two supporting changes:

- **`#peersReady` is a level**, mirroring FluxPeerManager's latch, so it is the
  hysteretic pair — up at `appSyncPeerThreshold`, down at
  `appSyncDegradedThreshold`. A node oscillating between the two does not reset.
- **`#onPeersDegraded` zeroes the counter before its state check**, not inside it.
  The method acts only from READY and SYNCING, so a node losing peers again while
  RESYNCING would otherwise keep its credit — the one road to READY with an empty
  peer set that survives every other guard.

A degrade also revokes state-sync authority: a node below the degraded threshold is
not hearing announcements, so it must not answer peers as though it held a full
view. That is per node and the fleet would feel it together — an event putting many
nodes below the threshold at once takes the fleet's app-state authority with it, and
none of them answers a state sync again until it has held peers for a full fallback.
It is the intended reading rather than a cost: a node that has stopped updating its
view still answers confidently, and the asker counts that as a completed survey. What
it costs is a paused spawner. Authority gates only the state-sync responder, no node
loses apps — `dbReady` comes from the chain-sourced hash sync, which completes with
every peer unreachable — and there is no deadlock, since readiness is a completed
sync OR an expired timer and every node reaches the second on its own clock.

A node that never crosses the threshold now says so. It reaches neither peer event, so
a budget that never advances is its only symptom, and the orchestrator announces that
on the edge in both directions — once when the budget stalls for want of peers, once
when it advances again. Nothing about such a node changes otherwise: it keeps the apps
it runs, and takes no new ones.

## A node whose peer set keeps collapsing is taken out of service

Five dips in a rolling two hours sets the sticky DOS, which stops the spawner and
removes the node's apps. The threshold, window and sweep interval are config.

The numbers come from the fleet: a deterministic sample of 250 nodes, 228
answering, ~20,700 node-hours, 890 connectivity diagnoses. Peers held are p10 26,
median 36, p90 52, and the lowest observed at any diagnosis is 7 — never at or
below the degraded threshold. A dip is the loss of roughly ninety per cent of a
peer set, so the rule is inert on a healthy fleet. Two alternative signals were
rejected against the same data; one would have fired on ~9% of nodes.

A teardown the node performs on itself is excluded — `disconnectAll()` on
confirmation loss crosses the same edge, and an unconfirmed node already loses its
apps. Release requires the peer set to have been up and dip-free for a full window, tracked
as a stamp that every fall clears — a node with no peers cannot dip at all, so its
quiet is quiet for want of anything to observe. A teardown the node performed on
itself clears the stamp like any other fall: it stays out of the tally, so the node is
not punished twice for one fault, and the minutes it spent with no peers are still not
stability it demonstrated. Coming back needs the rise threshold rather than the
degraded one, because between the two a node produces no edge and can neither dip nor
show that it has stopped.

## Partitioning suites run at production's miss count

`PARTITION_PEERS` states `wsPingIntervalMs` and `wsMaxMissedPongs` together, and
all five partitioning suites use it, so naming one half without the other is not
expressible. `peerDeathMs()` derives peer-death time from the effective config and
refuses a production fallback — production's pair gives 60s where the harness
answer is 6s, and an over-long budget passes every wait it is given.

Required because `configOverrides` layer on the harness config, not on production:
suites that set only the interval inherited a compressed miss count of 2 instead of
3, leaving two thirds of the intended margin in the suites most exposed to load.

## Evidence

Full parallel gate on a five-node-per-suite fleet: **99 of 101 suites**, at `186117683` — this
branch, below its current head. The commits above it are the suite 55 work described below; they
touch `55-residential-node-evacuation.js`, `coupled-knobs.js` and its unit test, and nothing else.
Suite 55's affected tests run green at the head. #1804 gates all three files unchanged.

Two suites failed in that run, neither of them behaviour from this change.

**83 — `app-state-sync-liveness`.** The node held four peers, could ask only one, and reported
`No peer left to ask` when that one was cut. `getEligibleSyncPeers` skips any peer whose
`missedPongs` is non-zero, and `missedPongs` is incremented when a ping is *sent*, so `pingAll`
puts every peer in the fleet inside that window at the same moment. #1801 fixes the counter in
`cd36c4e07` and merges ahead of this branch.

**55 — `residential-node-evacuation`.** Two faults, both in the suite rather than the product,
and both fixed here.

Its app was an image that installs no signal handler and runs as PID 1, where Linux discards
unhandled signals — so `docker stop` spent its full ten-second grace before the kill. A departure
therefore took longer to become visible than the gap between two holders' decisions, and both
holders of one app handed it back. Production has three orders of magnitude more margin on the same
inequality. The app now honours the stop, and the bound is asserted rather than implied.

Removing that grace then exposed a second test which had been sampling the container's state either
side of a stand-down — a transition that now completes between the two samples. It reads the
reconciler's own `desiredChanged` event instead, which carries the reason as well as the state, so
it distinguishes a stand-down from a crash and cannot be satisfied by an empty answer.

Unit suite green.


Cabecinha84 (commented):
Condition

  The cited gate evidence is not from this branch. The body claims "104 suites, 0 failures at 1c5f8cac2". That commit is
  on origin/fix/enterprise-placement and is not an ancestor of this PR at all.

  It is materially applicable — I checked: 18 of the 21 touched files are byte-identical between 1c5f8cac2 and
  550a6ff8e, including all of peerSetStabilityService.js, appSyncOrchestrator.js, FluxPeerManager.js, suite 1504, and
  the test-env.js peerless-override block. The 3 that differ (config/default.js, serviceManager.js, test-env.js) differ
  only by unrelated enterprise/policy work carried on that branch; the peerSetDip* keys and the
  peerSetStabilityService.start() call site are identical.

  But the run also predates development's #1802 merge, which changed run-parallel.sh — the gate runner itself. Ask for a
  gate re-run at 550a6ff8e, or at minimum correct the Evidence section to name the branch it actually ran on.

  Non-blocking findings

  1. #reconcilePass() — dead guard with a comment that now argues the opposite. appSyncOrchestrator.js:707: if
  (this.#state === STATES.DEGRADED) return; is unreachable. DEGRADED is only entered from #onPeersDegraded, only called
  from #peersBelowHandler, which sets #peersReady = false first — and line 693 already returns on !#peersReady. Worse,
  the comment above it still reads "A level, and the reason the latch is not enough on its own", describing #peersReady
  as a latch — exactly what this PR stopped it being — three lines after the new comment correctly calls it a level. In
  a codebase where comments are load-bearing, this one now defends the opposite of the change.

  2. peerSetStabilityService.evaluate() — docstring claims continuity the code doesn't enforce. Documented as requiring
  "the peer set to be UP and dip-free for the whole window"; the code is remaining === 0 && peerSetUp, where peerSetUp
  is instantaneous. A deliberate teardown sets peerSetUp = false and records no dip, so a confirmation-flapping node
  punches holes in the window and still releases on "up right now". Harmless in practice (unconfirmed ⇒ no apps), but
  the claim overstates the rule.

  3. Same docstring, and the suite-1504 comment: "A node that is out of service loses confirmation, drops every peer."
  It doesn't. As verified above, no confirmation path reads DOS state. The release rule is still right for the reason
  given (quiet is not evidence), but the stated mechanism isn't the one that produces it.

  Two things worth flagging to the team (not review objections)

  A new systemic coupling. This PR does two things at once that were previously separate: a degrade now revokes
  state-sync authority (the old comment explicitly preserved it — "Authority already earned is not revoked here") and
  zeroes the 125-minute budget. Fleet-wide, an event that pushes many nodes below 4 peers therefore takes the fleet's
  app-state authority away simultaneously and requires 125 continuous peered minutes before anyone can answer a state
  sync again — during which arriving nodes can't complete sync and must wait out their own 125 minutes. The measured
  floor of 7 peers across 890 diagnoses makes this remote and the semantics are defensible, but it's new and the body
  doesn't discuss simultaneity.

  The 5–11 peer band stays silent. The invariant makes crossing appSyncPeerThreshold (12) a hard prerequisite for ever
  reaching READY. A node that settles at 5–11 peers gets neither readiness nor an explanatory DOS: the rise latch never
  sets, so peersBelowThreshold never fires and peerSetStabilityService never sees a dip. The service's own docstring
  frames itself as "the half that says why" — for that node it isn't, and appStartupManager's sync_timeout will strip
  its apps with a generic reason. Worth a follow-up (a distinct hold reason or log), not a blocker; such a node is
  genuinely unhealthy.

MorningLightMountain713:
Gate re-run at `186117683` on this branch: **99 of 101 suites**. The Evidence section is rewritten to cite it, and states the delta from that tree to the head rather than leaving it to be found.

On the ancestry: `1c5f8cac2` is not an ancestor of this PR, but this PR's head was an ancestor of `1c5f8cac2` — that gate ran a tree containing every line of this branch with #1804's on top. So the evidence was applicable, and it was also concealing a failure. The file comparison could not have found it: it compares the files this PR touches, and the greenness depended on one it does not. Suite 14's test 7 degrades a node a second time and waits for the transition, which happens only from READY or SYNCING. This PR zeroes the block counter on degrade, so the node no longer returns to READY on the next block — it sits in RESYNCING and the wait times out at 30s. The fix was written on #1804. It is now the first commit on this branch, and the general rule it cost us is that a gate is evidence for the tree it ran on, so the delta from that tree to the head belongs in the Evidence section.

`run-parallel.sh` did change in the #1802 merge, but only in the sweep of leftover docker objects before and after a run; suite execution is untouched. That was not the reason the re-run was needed.

Two further corrections to the Evidence section, ours rather than yours: the suite count belonged to #1804's tree, and 1501 to 1503 are #1804's policy suites which have never existed on this branch. Only 1504 is here.

**1 — the dead guard.** Confirmed. Removed rather than re-commented. Keeping it would contradict the argument this PR makes for the counter: the invariant falls out of the level, and every consumer inherits it rather than carrying its own copy of the condition. A second road into DEGRADED must clear `#peersReady`, because everything downstream reads it — a leftover `return` in `#reconcilePass` would have protected one reader and left the rest. The stale comment went with it. The test named for the guard now holds the property: dropping `#peersReady = false` from the degrade handler fails it, and it did not fail before.

**2 — the release rule.** Confirmed, and the code moved rather than the wording. `peerSetUp` is now `upSince`, a stamp cleared by every fall, so the release requires the peer set to have held continuously for the whole window — which is what the docstring always claimed. A teardown the node performed on itself clears it like any other fall: leaving it out of the tally is what stops the node being punished twice for one fault, but those minutes are not stability it demonstrated. The docstring also now states the part that was never written down — coming back needs the rise threshold, not the degraded one, because between the two a node produces no edge and so can neither dip nor show that it has stopped.

**3 — the confirmation claim.** Confirmed and wrong in all three places it appeared. Confirmation comes from the daemon's own status; nothing in that path reads DOS state. Corrected in the service, the unit suite and suite 1504.

**On simultaneity.** A degraded node cannot be authoritative for app state sync — that is the intent, not a side effect. It is not hearing announcements, so an answer it gives is a view it has stopped updating, and the asker counts that as a completed survey; a confidently stale answer is worse than a refusal because nothing distinguishes it. The blast radius is one thing: `globalState.appStateAuthoritative` gates only the state-sync responder. No node loses apps, since `dbReady` comes from the chain-sourced hash sync and completes with every peer unreachable, and there is no deadlock, since readiness is a completed sync OR an expired timer. In the body now.

**On the 5 to 11 band.** Right, and now announced. The budget standing still is the only symptom such a node has, since neither peer edge has fired, so the orchestrator says it on the edge in both directions — once when the budget stalls for want of peers, once when it advances again. Behaviour is unchanged: the node already keeps its apps and already takes no new ones. What was missing was the reason.

Both gate failures are in the Evidence section with their mechanisms. 83 is the `missedPongs` counter, fixed in #1801 ahead of this branch. 55 was two faults in the suite itself, both fixed here — and the second only became visible once the first was fixed, which is the more useful half: a ten-second stop grace had been holding up an assertion that sampled a transition from either side.

Cabecinha84 (approved):
ack

PR #1799, opened 2026-09-07
Release 8.18.0: app ownership and tenant isolation, placement diversity, and the first gated release

Cabecinha84:
Release PR, `development` → `master`. 20 merged PRs since v8.17.1, 470 files, +89k/-17k, 57 new test files.

Cut following [`RELEASING.md`](https://github.com/RunOnFlux/flux/blob/development/RELEASING.md) — the process introduced by #1788 in this very release, so this is the first PR the gate judges.

| release step | state |
|---|---|
| `master` merged back into `development` (`13e92bf2a`) | done — the v8.17.1 fingerprint and the tampering blocklist entries that landed straight on `master` |
| the merged `ZelBack` is the PR head's `ZelBack` | yes — `master`'s two extra commits are both under `helpers/`, outside the hashed tree |
| `package.json` 8.17.1 → **8.18.0** (`0b7d1a390`) | done |
| `helpers/hashes.json` carries `f91c1a1ec2604642ddd69edc121d18b6` | done — the md5 over `./ZelBack` at this tip |
| the fingerprint is in the signed document central serves | verified — `hashes.runonflux.io/hashlist`, seq 41, valid Ed25519 signature |
| `npm run lint` | clean |
| CI on this PR | `gate`, `build (ubuntu-22.04, 20.x, 7.0)`, `build`, `deploy`, `publish` all green; **GitGuardian red** — see below |

**GitGuardian is red on a known false positive**, and it needs a maintainer dismissal rather than a code change. The flagged string is the WIF test key `5JTeg79...` in `tests/unit/fluxCommunicationMessagesSender.test.js`; #1795 added one more occurrence of a constant that already appears 20+ times in `development`. It is the same red that stood on #1795 itself. Every other release PR since 8.15.0 passed this check, so it is new with this content, not with this PR.

## ⚠️ Before merging: one cross-repo dependency is still open

**[flux-domain-manager#154](https://github.com/RunOnFlux/flux-domain-manager/pull/154) must merge and be deployed to every FDM server first.** #1777 in this release retires `apppause`, and a paused container was the only reason an operator-stopped `g:` master kept its load-balancer primary. FDM#154 replaces that with an explicit `GET /apps/heldcomponents` probe — an endpoint this release adds. Cutting the release before FDM has it means an operator-stopped `g:` app loses its primary and can come back on a different instance's copy of the shared volume.

[...]

PR #1804, opened 2026-09-12
Policy is one signed bundle distributed peer to peer, and a question is not news

MorningLightMountain713:
## Policy is one signed bundle, distributed peer to peer

Four consumers read from a single bundle verified once against pinned keys, rather than each fetching its own document. A change spreads from whichever node obtained it; the published source seeds it and is the floor, not the path.

Required because a per-consumer fetch cached whatever the response contained, for six hours, without checking it was a list — and a GitHub error page is truthy. Every read then threw `repos.forEach is not a function`: installs refused with a TypeError rather than the intended message, the removal sweep aborted, and in the spawner the TypeError did not match the one string that path treats as "service down", so apps were marked unspawnable for seven days. There is now no per-consumer fetch to answer wrongly and no cache to poison.

## A node that is behind asks the peer that arrived

Each arriving peer is asked directly whether it is ahead. The ask settles on that peer's reply rather than a fixed window — every peer answers, with a bundle if it is ahead, its sequence if it is not, null if it holds nothing — and it cannot reach the published source. That separation is what makes asking per join affordable: the source is rate-limited and shared by the fleet, so anything able to reach it runs on the backstop's timer instead.

There is one ask, and asking the network is that ask put to each peer. It returns when every one of them has answered or gone, so "no peer has anything newer" is a conclusion drawn from answers rather than from an interval having elapsed. A question that is not recorded cannot be answered, only outlived.

Peers are asked, and announced to, only if they advertise the capability. A node that does not has no handler for the ask, so including it buys a deadline's wait for a reply that cannot come and puts an unrecognised-type warning in its log once per adoption, on every node in the fleet.

Required because the alternative is being told, and being told is a broadcast a node must already be connected to hear: a peer that adopts in the seconds before it connects announces to nobody, and nothing afterwards revisits it.

**Reaching the published source is decided from the peer set, not from any one ask.** Three facts have to hold together: this node holds nothing, the set is above the threshold at which the network already considers a node worth gossiping with, and no question is outstanding. Every event that can change one of them re-evaluates the decision — a peer arriving, a peer leaving, an ask settling, and the peer count reaching the threshold, which is written after the arrival that crosses it. So a node whose peers all predate this protocol asks none of them and still reaches the source, which is the first wave of any rollout.

Boot does not reach the published source. A node restores what it last verified and takes its own slot in the backstop period, because agreement among nodes that restarted together is not evidence — each restored the same stale bundle, and a peer at the same sequence answers "not ahead", which is true and useless. Slots are spread across the period by node identity, so one node is always the next to look and announces whatever it finds.

**An ask names itself and its answer echoes that name**, so a reply settles the ask it names and nothing else settles anything. Without it the only thing tying a reply to a question was the socket it came back on, and a reply arriving after its own ask had timed out settled whichever ask was outstanding next — the node reading a stale sequence as the answer to a question it asked later.

An adoption announcement carries no name, and that is deliberate rather than an omission: it is what keeps the content address identical across the nodes that adopted the same sequence, which is what lets the filter collapse them into one. Being collapsible is also what makes it unfit to settle anything — whether a given peer's announcement arrives at all depends on whether an unrelated peer sent the same bytes first, and a fact about one peer cannot rest on that.

## Holding a bundle and being allowed to act on it are different

The acquisition gate is derived from both facts together: this node holds a verified bundle, and it has established that no peer it can reach is ahead of it. Recomputed from the pair rather than written by whichever arrives second — a latch on one of them records a conjunction it cannot track, and then the term it does not track can never open it.

**Confirmation is the peer set agreeing, not the first peer to answer.** A sequence is a claim the asker cannot check, so every capable peer's ask settles before the question is decided, and `minConfirmingPeers` of them must answer "not ahead". A peer holding nothing has said nothing about whether this node's policy is current, so it counts towards neither side.

**What the peer set cannot settle goes to the publisher, whose answer is signed.** That route is gated on confirmation rather than on holding something, so a node that restored a bundle off disk still reaches the source — and it is re-evaluated on every peer arrival, departure and settled ask, so a node that cannot yet be confirmed keeps asking rather than waiting out its own tick.

**Anything that would judge an app waits on that gate**, and so does anything that would destroy one it then has to rebuild. A redeploy uninstalls before it installs, and it is the install that needs the blocked-repository list to judge the image — so the refusal would otherwise arrive with the app already down, and only a successful install writes the local row back. The routes that pull an image answer 503 with a `Retry-After`, and the scanner's reinstall pass declines its block and asks again a few blocks later. Removal is not gated: it rebuilds nothing, and a node that cannot obtain policy is when clearing an app off it matters most.

**The ownership sweep is the one destructive path in the enterprise module**, so it runs only on policy this node has established is the network's, and it runs whenever that view changes — when the gate opens, and on every bundle change while it is open. An owner granted or revoked during the node's life is acted on when it happens rather than only at boot. Requests arriving during a pass coalesce into one further pass, and nothing downstream of boot waits on it.

## A message says how it travels

**Not policy-specific: this affects app-message sync today.**

Every message carries one of three intents, and the intent decides all three questions the transport has about it:

| intent | relay onward | deduplicate | retain |
|---|---|---|---|
| announce | yes, once | yes, on `hash(data)` | only if hash-announced |
| ask | no | never | no |
| answer | no | never | no |

Deduplication is the row that matters. An announcement is a fact about the network and reaches a node by many routes, so two copies are one fact. An ask and an answer are about the two nodes exchanging them — the sender is part of the meaning, and `hash(data)` is the one part of the message that throws the sender away. `{fluxapprequest, version, hash}` is what every node missing that hash sends, and `{fluxpolicyseq, seq}` is byte-identical from every peer holding that sequence, so a filter keyed on content alone delivers the first and silently drops the rest.

Delivery is a separate axis and is not the classifier: `fluxapprequest` is an ask delivered by broadcast, and `fluxpolicyseq` is an announcement when a node tells its peers what it adopted and an answer when it settles a peer's ask. A type belongs to two rows at once, so the intent travels with the message rather than being looked up from its name.

**Which intents a type may have is declared with its route, and the message chooses only where the type allows it to.** One type is sent both ways and takes its intent from the marker; every other declares a single intent, and a marker on one of those is ignored. A type that is relayed onward and also took its intent from its payload would let any node able to sign a message put an undeduplicated one into the network, and every honest node would announce it again to every peer for as long as its timestamp stayed valid.

The intent rides inside `data`, covered by the existing signature, and announcements carry no marker — so `hash(data)`, the content address both sides of the hash-announce protocol compute, is unchanged. A peer that sends no marker is read as it is today.

**Remembering a fact and being able to hand it over are separate stores.** A filter holds the hashes of every announcement and runs before signature verification, which is what bounds that cost — 26 peers deliver ~2.4k messages a minute of which ~135 are distinct. A second store holds the messages this node announced, which is exactly the set a peer can ask for, and is written after verification. Policy bundles, asks and answers occupy neither.

**One table says where each message goes** — its handler and the pipeline that carries it — read by the dispatcher, the socket and the chunk processor alike, so no part of the transport keeps a list of its own.

**One registry holds the requests this node has outstanding to peers**, identified by their own id and bound to the channel they went out on, settled exactly once with an outcome. What sits above it is each owner's policy: the sync orchestrator's record afterlife and completion pool, and policy's adopt-and-confirm.

## Announcements propagate by hash, on every node

A node announces the hash of what it holds and serves the message to whoever asks. The choice is no longer conditioned on a block height: the fork activated in October 2024, and the version floor disconnects anything old enough to need the legacy path.

Required because that height came from `isDaemonSynced().data.height` while `.data.synced` — the flag in the same object saying whether the number can be trusted — was discarded. A node that has not heard from its daemon reports height 0, which compares as "before the fork", so the least informed state produced the most obsolete behaviour.

## A pinned app runs where it says, or not at all

An enterprise owner may pin a v8 spec to named nodes, by address or by collateral; placement honours the pin or refuses to place. `datacenter: true` is the same kind of privilege and is held to the same rules.

Both are judged on live submission only — applying them to replay would have upgraded nodes rejecting messages already on chain that their peers accept — and both let an owner who has left the enterprise list carry an existing grant forward without acquiring a new one. Renewal is an update, so refusing every update from an ineligible owner expires the app rather than restricting it.

**A grant is carried forward by its owner, not by its name.** What the app was last registered as is looked up by name, the message log outlives the app, and a name is released once the app expires — so nothing is carried forward unless the previous specification's owner is the one submitting. Owners compare as the signature layer compares them: exactly for a Flux ID, which is base58, and without regard to case for an Ethereum address, where the capitals are a checksum over the same twenty bytes.

A spec that names this node by address is honoured whether or not the daemon can be reached, since the address needs no daemon to check. A collateral pin that cannot be resolved matches nothing.

**Re-pinning an app with synced data: back it up first.** Changing an app's `nodes` list is now enforced on the nodes that lose it — each one uninstalls the app cleanly and tells the network, instead of tearing it down and then discovering it may not rebuild it. The departures are not ordered against the arrivals, because the spawner counts an app's locations without asking whether its pin names them: an app at its instance count on the wrong nodes is short nowhere, so no newly named node installs it until a copy has already left. That means there is a window with no copy of the volume anywhere. Back up a `g:`/synced app before re-pointing its pin. Ordering this properly needs placement to count eligible copies rather than location rows, and is v9 work.

## This node's identity resolves at boot

The key identifying this node is in `flux.conf`, and deriving it involves no daemon. That file is parsed when the key is wanted rather than as a side effect of the first RPC, so identity is settled at boot instead of on the first attempt that happens to find the config loaded. The spawner is gated on that identity, and a node with its key on disk all along spent up to five minutes unable to acquire anything.

## Harness

Three suites tested mechanisms this branch replaces, and their premises are corrected rather than their assertions relaxed. One asserted fleet-wide behaviour on nodes that held zero peers after a restart, which also left its final test unable to fail.

1502 covers two peers answering the same question with the same content — the mirror of two nodes asking it, and the case a single answering peer cannot show, because the first answer always gets through. The peer stub answers a policy ask as a real peer does, so two stubs told the same sequence put identical content on the wire from different senders.

It also covers the rollout: a fleet whose only peers advertise no policy capability. The node asks none of them, announces to none of them, and reaches the published source anyway. The stub advertises the capability it implements, and a flag produces the peer that predates the protocol.

And it covers the rung the peer set cannot carry: a fleet that can field two answering peers on a node told it takes three. What its peers say is true, unverifiable and not enough, the bundle is on disk throughout, and the node goes on refusing to act on something it already holds until the publisher settles it.

## Evidence

Full parallel gate on a real fleet, five nodes per suite, at `66a10c08c`: **106 suites pass, 0 fail**, 100 minutes.

Full unit suite green: **7050 passing, 0 failing**, 17 pending.

The unit command now matches every test under `tests/unit`, including the ones in subdirectories, which its glob did not descend into. The pattern is quoted so mocha expands it: `/bin/sh` runs npm scripts and has no globstar, so an unquoted `**` reads as `*` and would silently match a fraction of the suite.

Suites 1501–1504 cover the bundle lifecycle, peer distribution, pinned placement and peer-set collapse; 1502 includes two nodes asking the same question, two peers answering it, and a node catching up from a peer that arrives.


Cabecinha84 (commented):
Blocking: policyStore.markConfirmed() burns its latch on an empty store

  ZelBack/src/services/policyStore.js:151

  function markConfirmed(source) {
    if (confirmed) return;
    confirmed = true;      // ← latched here
    if (!current) return;  // ← guard runs after
    globalState.policyReady = true;

  If markConfirmed is reached while current === null, the one-way latch is consumed with nothing to show for it. Every
  later call — including the one inside adopt() — hits if (confirmed) return and globalState.policyReady can never open
  for the lifetime of the process. appSpawner.trySpawningGlobalApplication then returns installDelay forever and the
  node stops acquiring apps entirely. Only a restart clears it.

  Two reachable triggers. I proved both with probe tests against the real module:

  A. A backstop 200 that does not verify, on a cold node. refresh() calls markConfirmed('the published source')
  unconditionally after consider() (policyStore.js:322-327), and consider() returning false for a body that fails
  verification is not distinguished from "valid, nothing new". A captive portal, a transparent proxy, an injected ISP
  page — anything that answers 200 with non-bundle bytes. This is the start() path for a node with no stored bundle,
  which on the release that ships this is the entire fleet.

  B. One peer sending seq: 0. notePeerSeq(0, key) → 0 <= getSeq() (0) → markConfirmed('peer') with current === null.
  {type:'fluxpolicyseq', version:1, seq:0} is one signed broadcast from any FluxNode in the deterministic list, it is
  not relayed so the blast radius is direct peers, and it wedges every peer that has not yet adopted. Cheap, targeted,
  and it survives the victim subsequently obtaining perfectly valid policy.

  Probe results (before fix):

  A: backstop serves an unverifiable body, then a peer serves a good bundle
     → getSeq() === 7, policyReady === false     (expected true)
  B: garbage fetch, then a good one from the backstop
     → getSeq() === 3, policyReady === false     (expected true)
  C: peer sends seq 0, then a good bundle arrives
     → getSeq() === 11, policyReady === false    (expected true)

  Fix is moving one line:

    if (confirmed) return;
    if (!current) return;
    confirmed = true;
    globalState.policyReady = true;

  I applied exactly that locally: all three probes pass and the 48 existing policyStore tests stay green. Worth adding
  one of the probes as a regression test — the suite currently has no case where markConfirmed is reached with an empty
  store.

  Non-blocking findings

  1. respondWithPolicy JSDoc contradicts its body. fluxCommunicationMessagesSender.js:126-135 still says "Answered only
     when this node holds something strictly newer. Silence is the answer for 'nothing newer'." The body now answers in
     all three states — that was the point of the "a peer with no policy says so" commit. The stale header is the first
     thing a reader sees.
  2. "A peer that lies low is ignored" is false. Same file, inline comment: "we only act on a HIGHER claim". A lower
     claim calls markConfirmed('peer'), which is acting on it. Verified: a node that restored seq 2 while the network is
     at seq 50 opens its confirmation gate on one peer claiming seq 1. So confirmation — the thing guarding "do not
     acquire on policy the network has moved past" — is forgeable by a single peer. The design may accept that (the
     claim genuinely cannot be checked), but the comment should say so rather than the opposite.
  3. minimumFluxOSAllowedVersion 8.13.1 → 8.18.0 (config/default.js:122). This disconnects every node below 8.18.0 via
     fluxNetworkHelper.js:431. It's the correct floor for deleting helpers/ — 8.18.0 already reads
     config.policy.baseUrl, not helpers/ — but it's a fleet-wide gate and 8.18.0 was itself still blocked on the FDM#154
     deploy. Needs an explicit ops decision on timing and the current fleet version distribution, not just a code
     review.
  4. Hard pinning is a live behaviour change for existing specs. appSpawner drops || app.version >= 8, and
     hwRequirements.checkAppNodesRequirements now runs for version >= 7 instead of === 7. Any v8 spec already on chain
     with a non-empty nodes[] and a non-enterprise owner becomes placeable only on its named nodes, and an instance
     already running elsewhere will throw is not allowed to run on this node on its next checkAppRequirements
     (redeploy/update path). The PR argues this is the honest outcome and I agree with the reasoning — but ask for the
     count of live v8 specs matching that shape before merging, because that number is the size of the disruption.
  5. Gate evidence is not at head. The body cites "104 suites, 0 failures at 1c5f8cac2". That commit is not an ancestor
     of 42854d3d6; three files differ, including production imageManager.js (a map().find() → early-return refactor in
     blockedReasonFor, semantically equivalent but production code) and run-parallel.sh/run-all.sh — the gate runner
     itself. Same condition as #1803: re-run the parallel gate at this head.
  6. fluxpolicyrequest bypasses the flood cache with nothing else bounding it. The DIRECTED_TYPES change is right for
     correctness, but dispatchFluxMessage has no per-peer rate limiting (rateLimitCache is HTTP-only), so one peer can
     now request the 127 KB bundle without limit and the flood cache no longer collapses identical repeats. Worth a
     per-peer counter on these two types.
  7. The peer rung is skipped below the latched threshold. refresh() uses peerCountIfAboveThreshold(), which reads 0
     until appSyncPeerThreshold (12). So a cold node with 1–11 peers goes straight to GitHub, contradicting the module
     header's "sixteen peers on the local network are asked before one request leaves for github" — and it's wrong at
     exactly the moment it matters. refreshOnce() single-flights the burst so the load is small; it's the doc claim
     that's off.
  8. Cosmetics: policyStore.getArtifact() has no production consumer yet (ipLocationSync still fetches
     config.policy.baseUrl directly); test-infra/config/shared.js:214 still declares the removed
     spawnDeferrals.targetedNodesMs; the PR template now says the blocklist is what "every node fetches directly", which
     this PR makes untrue.

Cabecinha84 (commented):
Condition for merge

  Re-run the full parallel gate at f904c9f51. The PR body cites "104 suites, 0 failures at 1c5f8cac2" — that head is 93
  commits and ~757 changed production lines behind, including policyStore.js itself (326 lines), the peer-distribution
  wiring, ipLocationSync artifact verification, and the comms handlers. That evidence no longer covers the code.

  Non-blocking findings

  1. fluxpolicyseq / fluxpolicy answers collide in the flood cache — fluxCommunication.js:667

  DIRECTED_TYPES covers the questions but not the answers. messageCache keys on object-hash(msgObj.data), and the answer
  payloads carry no sender and no timestamp: {type:'fluxpolicyseq',version:1,seq:null} is byte-identical from every
  empty peer, {...,seq:5} from every peer at 5, and {type:'fluxpolicy',version:1,bundle:raw} from every peer holding the
  same bundle. Within the 5-minute TTL only the first is dispatched; the rest never reach notePeerSeq/offerBundle, so
  their pendingPeerAsks entries fall back to the 3s PEER_WINDOW_MS timeout instead of settling on an answer.

  I tried to build a stuck state from this and could not — the first message of any given payload always either confirms
  or triggers a request, and seedIfPeersHaveNothing still fires once the timeouts elapse. So this is latency, not
  correctness. But it is the same bug class the PR fixes for fluxapprequest, and by the PR's own membership test ("does
  the sender's identity change what the message means") these qualify when they answer a directed ask — policyStore uses
  peerKey precisely to decide which ask settled. Worth either adding them or writing down why the answers are exempt.

  2. hwRequirements.checkAppNodesRequirements fails closed on a daemon hiccup for v8 — hwRequirements.js:298

  The guard widened from === 7 to >= 7, and obtainNodeCollateralInformation() is awaited unguarded. A v8 spec pinned by
  address to this node will now abort the install if the collateral RPC is unavailable, where v8 previously skipped the
  gate entirely. nodesNameThisNode already tolerates a null outpoint; the await above it does not. A .catch(() => null)
  would make the two agree.

  3. Two stale config comments — ZelBack/config/default.js:674 and :696 still say enterprisenodes.json is "fetched every
  6h, see enterpriseConfig". enterpriseConfig no longer fetches anything. Notable in a PR with a commit named "comments
  state what the code does, not what it stopped doing".

  4. A restored node whose peers are all empty waits up to 24h to confirm. notePeerSeq(null) deliberately isn't
  confirmation, and seedIfPeersHaveNothing returns early because current is set — so the gate stays shut until that
  node's own phased tick. It follows from the stated design and should be rare, but it is the one configuration where a
  node holding perfectly good policy sits out a full day. Worth confirming it's intended.

  5. Transient validation rejections widen slightly. verifyTypeCorrectnessOfApp's datacenter check runs before the
  liveSubmission branch, so it now throws "network policy not yet obtained" on the replay path too. Recoverable —
  appHashSyncService only marks hashes on success — and pre-existing in kind, but the window is larger now that the
  on-disk helpers/enterprisenodes.json seed is gone.

MorningLightMountain713:
Answering both reviews in one place — 13 September, and the follow-up of 15 September.

## Merge condition

Full parallel gate on a real fleet, five nodes per suite: **104 suites, 0 failures, at `31c856218`** — the current head rather than the head the condition named, which the branch has since moved past. 65 minutes, `rc=0`, no suite reporting a non-zero code anywhere in the log.

Unit suite: **6634 passing, 0 failing.**

---

## 15 September

### 1. `fluxpolicyseq` / `fluxpolicy` answers collide in the flood cache

Fixed — though not by adding the two types to `DIRECTED_TYPES`. That list is gone.

A per-type list cannot express this. `fluxapprequest` is a question delivered by broadcast, and `fluxpolicyseq` is an announcement when a node tells its peers what it adopted and an answer when it settles a peer's ask. A type belongs to two rows at once, so the intent now travels with the message: announce, ask or answer, deciding relay, deduplication and retention together. Only an announcement is deduplicated, because only an announcement is a fact whose identity is its payload.

The marker rides inside `data`, covered by the existing signature. Announcements carry none, so `hash(data)` — the content address both sides of the hash-announce protocol compute — is byte-identical to what a node without the classifier computes for the same message. A peer that sends no marker is read exactly as today.

Two further consequences of the same diagnosis, which the finding did not reach:

- **An ask now names itself and its answer echoes that name.** Beyond making two peers' replies distinct, this fixes a second defect: a reply arriving after its own ask had timed out previously settled whichever ask was outstanding next, so a node could read a stale sequence as the answer to a question it asked later.
- **The cache was doing two jobs on one key.** It was both the flood filter and the content-addressed store served to peers requesting a hash — different keys in principle and different membership in fact, and it was written *before* signature verification, so an unverified message entered the store and would be served on request. These are now separate: hashes in the filter, which runs ahead of verification and is what bounds that cost; messages in the store, written after verification and holding only what this node announced.

### 2. `checkAppNodesRequirements` fails closed on a daemon hiccup for v8

Fixed. A spec naming this node **by address** needs no daemon to check, so a failure to resolve the collateral now narrows pin matching to addresses rather than refusing the install — the contract `nodePinning` states and the two sibling callers already kept. Narrowing is not widening: an unresolved outpoint matches nothing, so a spec naming only other nodes is still refused.

The function had no behavioural coverage before this; it now has five tests, and removing the guard or removing the outpoint arm fails different ones.

The wider observation behind the finding is real and scoped separately: this node's own collateral is read from nine places with four different error conventions and no caching. The obvious remedy — cache it, it does not change — is unsafe today, because re-collateralising a node does change it and a process-lifetime cache would then serve a dead outpoint. That needs an invalidation signal the daemon does not yet publish.

### 3. Two stale config comments

Removed — three lines, not two. Deleting the second exposed the two above it as the rest of one orphaned block describing the enterprise node list, which then read as though they described `cpuBurst`.

### 4. A restored node whose peers are all empty waits up to 24h to confirm

Working as designed; not changed.

The only route into that state is a restart of a node that was already confirmed: `current` is on disk, confirmation is in-process and deliberately does not survive, because yesterday's confirmation says nothing about today's network.

An empty peer is not weak evidence, it is the absence of evidence — a peer holding nothing may simply be newer than the policy system, which is indistinguishable from a network that has no policy and wants the opposite answer. And the state is self-clearing: a node holding nothing reaches the source as soon as its peers have answered, not on its tick, so every empty peer in that scenario is seconds from acquiring and announcing. The durable version requires the source to be unreachable, in which case the backstop tick fails too and this guard is not the constraint.

### 5. Transient validation rejections widen slightly

Fixed, and the inconsistency behind it closed.

The datacenter eligibility check has moved out of `verifyTypeCorrectnessOfApp` — which runs ahead of every `liveSubmission` branch — and now sits beside the node-pinning check under the same gate. Only the boolean type test remains in the function named for type correctness.

It also now grandfathers, as pinning does: an owner who leaves the enterprise list keeps an app that already had `datacenter: true` and cannot acquire it on one that did not. Renewal is an update, so refusing every update from an ineligible owner expired the app rather than restricting it — the outcome the pinning rule went out of its way to avoid, with nothing explaining why the two differed.

---

## 13 September

These were addressed in code at the time but never written up here.

**Blocking — `markConfirmed` consuming the latch with an empty store.** Real, and closed by removing the latch rather than guarding it: `policyReady` is now derived from both facts (`confirmed && current`) and recomputed whenever either is written. The suggested early return leaves the mirror case — a bundle arriving after confirmation — stranded in exactly the same way, because it still records a conjunction in a variable tracking one term.

1. **`respondWithPolicy` JSDoc contradicted its body** — corrected; it describes three answers for three states.
2. **"A peer that lies low is ignored" was false** — corrected, and now says the opposite: such a peer *is* acted on, with the reason that makes it safe.
3. **`minimumFluxOSAllowedVersion` → 8.18.0** — kept deliberately; it is the floor that makes the legacy paths unreachable.
4. **Hard pinning is a live behaviour change** — the count of matching live specs was measured before merge, as asked, and was small; rows are not published here.
5. **Gate evidence not at head** — see the merge condition above.
6. **`fluxpolicyrequest` bypasses the flood cache with nothing else bounding it** — the premise is the part to correct: the bound is not the cache and never was. Inbound frames are limited per peer at the socket, before dispatch, covering every type. The cache suppresses accidental duplicates, which a sender controls anyway by varying one byte. That reasoning is now written at the list.
7. **The peer rung below the latched threshold** — the accessor reads the network's own latched level with its existing hysteresis, so a late subscriber sees the level rather than needing an edge it missed.
8. **Cosmetics** — all three done: `getArtifact` has its consumer, the removed knob is out of the harness config, and the template no longer describes a direct fetch.

---

Also in this branch, from following finding 1 down: announcements no longer choose their propagation by block height (the fork activated in October 2024 and the version floor makes the legacy path unreachable), one table now says where every message goes rather than three lists, and one registry holds the requests this node has outstanding.

1502 gained the mirror of its two-askers test — two peers answering the same question with the same content, which a single answering peer cannot show because the first answer always gets through.

Cabecinha84 (commented):
Blocking: a signer-chosen intent skips dedup and makes some messages loop through the network

  messageIntent.intentOf() trusts data.intent for every message type. dispatchFluxMessage only records a message in
  announcementSeen when its intent is announce.

  That means a normal announce-type message, signed with intent: 'ask' or 'answer' inside data, is never marked as seen.
  Honest nodes then do this on every copy:
  1. Another peer announces its hash, and handleCheckMessageHashPresent requests it again, because the hash is not in
     announcementSeen.
  2. The message is handled again.
  3. announceToPeers announces it to all other peers again.

  Three handlers re-announce without any duplicate check, because they relied entirely on the old cache:
  - handleNodeSigtermMessage: re-announces whenever the IP has apps.
  - storeIPChangedMessage: returns true unconditionally.
  - storeAppRemovedMessage: returns true unconditionally.

  For these types every copy fans out to about 25 peers, and it only stops when the 240s timestamp window closes.
  - Who can trigger it: any registered node operator, and they can simply sign a new message every few minutes.
  - Impact: each node also repeats a database write per copy, and the per-peer limit of 120 msg/s is saturated
    network-wide, so legitimate gossip is silently dropped.

  Probe: I ran the same messages through the real dispatcher on both branches. The sequence was: dispatch from peer A,
  hash announced by peer B, then the same message again from peer B.

  ┌─────────────┬──────────────┬─────────┬─────────────────┬─────────────────┐
  │   branch    │    intent    │ handled │ announced again │ requested again │
  ├─────────────┼──────────────┼─────────┼─────────────────┼─────────────────┤
  │ development │ answer / ask │ 1       │ 1               │ 0               │
  ├─────────────┼──────────────┼─────────┼─────────────────┼─────────────────┤
  │ PR head     │ answer / ask │ 2       │ 2               │ 1               │
  ├─────────────┼──────────────┼─────────┼─────────────────┼─────────────────┤
  │ PR head     │ none         │ 1       │ 1               │ 0               │
  └─────────────┴──────────────┴─────────┴─────────────────┴─────────────────┘

  The comment in messageIntent.js says a wrong intent "costs that peer and not this node". That is not true: the intent
  is inside the signed payload, so honest nodes relay it too.

  Fix: honour a claimed intent only for the types that can legitimately carry one (fluxapprequest, fluxpolicyrequest,
  fluxpolicyseq, fluxpolicy), for example as a per-type setting in messageRoutes, and treat every other type as an
  announcement. As a second guard, announceToPeers could also write the hash to announcementSeen. Add a unit test for an
  announce type carrying intent: 'answer'.

  Medium: carrying a privilege forward is not tied to the owner

  verifyAppSpecifications lets an owner who is not on the enterprise list keep a pin or datacenter: true if the previous
  spec from getPreviousAppSpecifications had it. That function looks the previous spec up by name only, and the check
  never compares previous.owner. checkApplicationRegistrationNameConflicts frees an app's name once the app expires.

  So anyone who re-registers the name of an expired enterprise app inherits its privilege:

  - Datacenter: matching datacenter: true is trivial. My probe with a previous spec of { owner: ENTERPRISE_OWNER,
    datacenter: true } and an ordinary submitter was accepted.
  - Pins: the check only compares node lists, and the old node list is public.

  The pin half was already there at f904c9f51, and I missed it in my last review.

  Fix: require previous.owner === appSpecifications.owner before carrying anything forward. The existing tests pass a
  previous spec with no owner at all, which is why they don't catch this.


  1. Unmarked announcements can still settle an ask. settlePeerAsk falls back to matching by peer when there is no
     correlationId, so an ordinary fluxpolicyseq announcement settles an outstanding ask to that peer. It's harmless
     because it reports the peer's current sequence, but it contradicts the PR body's "a reply settles the ask it
     answers and no other".
  2. Two retry loops in scheduleIdentityResolution. A bundle change that arrives while a resolve attempt is still
     running starts a second attempt, and if both fail there are two retry chains. The only cost is extra logging and
     daemon calls.
  3. Mixed-fleet rollout. Nodes already running 8.18 still reject every datacenter: true spec from a non-enterprise
     owner, including replayed ones. Upgraded nodes accept carried-forward ones, so the two will disagree about those
     updates until the fleet has upgraded. This is the same window the author already accepts for pins.
  4. Stale number in the PR body: it says 6634 passing, and I get 6648.

MorningLightMountain713:
Answering the 16 September review.

## Merge condition

Full parallel gate on a real fleet, five nodes per suite: **106 suites, 0 failures**, `rc=0`, 62 minutes, no suite reporting a non-zero code anywhere in the log. Unit suite **6970 passing, 0 failing**.

The branch is rebased onto `development` now that #1801 has merged, so it stands alone rather than on a stack.

---

## 1. A signer-chosen intent skips deduplication — fixed

Correct, and the probe matches what the code did.

What a type is allowed to be is now declared beside its route, and both are required — `register` throws without them, so the signature is the guard rather than a test that has to remember. A message gets a say only for a type declared to be sent both ways, which is one type: `fluxpolicyseq`, news when a node announces what it adopted and an answer when it settles a peer's ask. `VARIES` is not a value a message can claim, and only a type nothing relays may be declared it. Every other type's declaration is the answer and a marker on it is ignored.

Not the suggested allowlist of four, because three of those four have a single role each and declaring them as both would grant a say that nothing needs. The second guard — writing the hash at announce time — is unnecessary once the only type whose payload decides is one nothing relays onward.

The blast radius is one handler wider than filed. Four rebroadcast without idempotence of their own, not three: the withdrawal path of `storeAppInstallingMessage` deletes the row its own timestamp guard matched on, so the second copy finds nothing to compare against and returns `true` again.

The requested unit test is there, and it is the mutation that matters: an announce type carrying `intent: 'answer'` still classifies as an announcement.

## 2. A privilege carried forward by name — fixed

Correct. The lookup is by name, the message log outlives the app, and a name is released once the app expires.

The owner test sits on the shared previous-specification accessor rather than inside each privilege, so a third one cannot be added without it.

`previous.owner === appSpecifications.owner` is too strict for an Ethereum owner. Signature verification compares recovered addresses without regard to case, so the same owner can legitimately appear in two capitalisations, and an exact comparison refuses their own renewal — which expires the app, the outcome carrying a grant forward exists to prevent. Owners now compare as the signature layer compares them: exactly for a Flux ID, where case is the identity, and case-insensitively for `0x`, where the capitals are a checksum over the same twenty bytes.

## 3. An unmarked announcement can settle an ask — fixed

Filed as cosmetic; the mechanism under it was worth the look.

An answer settles the ask it names and nothing else settles anything. Matching by socket cannot tell an answer from an announcement, and an adoption announcement carries no name by design — that is what keeps it content-addressable, and being content-addressable is what makes it unfit to settle a fact about one peer.

Behind that, the store had two ways of asking: a targeted ask that records the question, and a broadcast that recorded nothing and then waited a fixed moment to see whether the sequence had moved. There is one now. Asking the network is the targeted ask put to each peer, returning when every one has answered or gone, so "no peer has anything newer" rests on answers rather than on an interval elapsing.

Peers are asked, and announced to, only if they advertise the capability — a node that does not has no handler, so including it spends a deadline here and writes an unrecognised type in its log there.

## 4. Two retry chains in identity resolution — fixed

Real: a second attempt started beside one in flight, both could fail, and the second overwrote the first's handle. One attempt at a time now.

The timer is no longer the normal path. This node's key is in `flux.conf` and deriving it involves no daemon, but nothing parsed that file at boot — it was read only as a side effect of the first RPC, so the key read as absent, which looks exactly like a daemon that is not up. The parse is something a caller can ask for, and the key asks for it. The interval remains for a config that genuinely cannot be read.

## 5. Mixed-fleet rollout — accurate, no change

The window is real and is the one already accepted for pins. Nodes on the current release apply the datacenter check ahead of the live-submission branch, so they refuse replays too; upgraded nodes accept a carried-forward grant, and the two disagree until the fleet has upgraded.

## 6. 6648 rather than 6634 — does not reproduce

Counted two ways at the tip that review was written against, it gave 6634 exactly. The current figure is above.

What the check did surface is real: the unit command's glob did not descend into subdirectories, so tests in one were never run by it. The pattern is quoted so mocha expands it — `/bin/sh` runs npm scripts and has no globstar, so an unquoted `**` reads as `*` and silently matches a fraction of the suite.

---

Also in this branch, from following the first finding down: the decision to reach the published source is derived from the peer set rather than taken as a consequence of any single ask, and is re-evaluated by every event that can change one of its terms. 1502 covers the case that makes the difference — a node whose peers all predate this protocol asks none of them and still obtains policy.

Cabecinha84 (commented):
1. Blocker — the unit suite fails at this head

  CI result: build (ubuntu-22.04, 20.x, 7.0) shows 6983 passing, 3 failing, all in tests/unit/policyStore.test.js:
  - does not go to the source when a peer answered with a bundle
  - asks when peers exist
  - asks anyway when the transport cannot say how many there are
  - Reproducible: these fail locally on every run (3 of 3).
  - Cause: bisected to eb33d89d8 ("one ask, put to the peers that can answer it"). An ask now ends only when a reply
    carries its id, but these test stubs never send the id back. Each ask therefore waits out the 3s peerWindowMs, which
    is longer than mocha's 2s timeout.
  - Fix: the stubs should answer via notePeerSeq(…, key, id) / offerBundle(…, key, id), or the tests should use a short
    peerWindowMs.
  - The PR's evidence is wrong: the body and the latest comment say "6970 passing, 0 failing", which can't be true for
    this code. The fleet test gate is cited at 3a2e65e01, which is not on the rebased branch.

  2. High — during the rollout, a restarted node can stay unconfirmed for up to 24h, and redeploys then remove apps

  When it happens: a node restores its bundle from disk at boot, and none of its peers advertise the policyBundle
  capability. That describes every node running 8.18.x, i.e. the whole fleet during the rollout.

  What goes wrong:
  - considerSeeding() (policyStore.js:454) returns immediately whenever a bundle is already held (current is set).
  - There are no capable peers to ask, so nobody confirms the bundle.
  - policyReady therefore stays false until the node's own scheduled check (refreshIntervalMs = 24h).
  - Probe: 20 peers above the threshold, none capable, gave 0 asks, 0 GitHub fetches and policyReady=false.

  Why it matters: blocking new app spawns is intended, but the new check also blocks appInstaller.verifyAndPullImage
  (appInstaller.js:213):
  - The reinstall pass (advancedWorkflows.js ~4386–4398) runs softUninstallApplication / hardUninstallApplication and
    only then tries to install. The install throws, so the app is removed from the node and not rebuilt.
  - API redeploys and container health repair fail in the same way.

  Fix: also confirm a node that holds a bundle but has no capable peer to ask, e.g. by running one check against the
  published source when the node is above the peer threshold. Separately, the reinstall/redeploy paths should check
  policyReady before uninstalling anything.

  3. Medium — the ownership sweep acts on an unconfirmed bundle

  - The check: cleanupOwnershipViolations (enterpriseNetwork.js:222, called from serviceManager.js:740) only checks
    isPolicyKnown(). That returns true as soon as any bundle is held, including one restored from disk and not yet
    confirmed.
  - Probe: with a restored bundle, isPolicyKnown() is true while policyReady is false.
  - Why it matters: this is the one destructive path (it uninstalls apps and broadcasts fluxappremoved), and it
    contradicts the PR's own rule that holding a bundle is not the same as being allowed to act on it. If saving the
    newer bundle failed ("untidy rather than incorrect"), the sweep uninstalls customer apps based on the older bundle.
  - Fix: await globalState.waitForPolicyReady() before running the sweep.

  ---

  Verified as fixed

  - Message flood: each message type now declares its intent in messageRoutes. Only fluxpolicyseq lets the message
    choose (VARIES), and nothing relays it. All relayed types (sigterm, ipchanged, appremoved, …) are fixed to announce
    and are deduplicated again.
  - Privilege carry-forward: a pin or datacenter flag is only carried forward by the same owner (sameSigningIdentity,
    case-insensitive for 0x addresses).
  - Answers and asks: an answer settles only the ask whose id it echoes, and a peer that disconnects ends its ask
    without waiting for the timeout.
  - Identity resolution: only one attempt runs at a time, and flux.conf is read directly without needing the daemon.
  - Capabilities: they come from the handshake headers and are set before peerConnected, so capable peers are correctly
    recognised.
  - Lint: clean on the changed files.

  Non-blocking

  - datacenter now enforced less widely: development checked it on replay too (in verifyTypeCorrectnessOfApp); now only
    live API submissions are checked. The impact is low, since the flag only restricts placement to datacenter nodes.
    The comment at appSpawner.js:837 still claims it is always enforced, so it is now stale.
  - No backoff on the source fetch: a node holding nothing retries GitHub on every peer arrival or departure while the
    fetch keeps failing (only one fetch runs at a time).
  - More load from fluxapprequest: it is no longer deduplicated, so repeated identical v2 requests (up to 500 hashes,
    correctly recognised.
  - Lint: clean on the changed files.

MorningLightMountain713:
Answering the 17 September review. All three findings are addressed, including the second half of finding 2, which is now built rather than deferred.

## 1. Blocker — the three failing unit tests

Fixed in `946601408`.

Your bisect to `eb33d89d8` was right, and the mechanism under it: `policyStore.test.js` stubbed `serviceHelper.delay` to return instantly, with a comment claiming that was what kept tests out of the 3s peer window. `policyStore.js` references `delay` zero times at that head, so the stub was dead and the asks sat out the real window. Both fixes you suggested are in — the stubs that answer echo the ask id they were given, and a test that is not about the window runs under a short one while a test about the window passes its own.

On the evidence, you were right that it was wrong, and both citations in the body are replaced. The gate is a full parallel run at the branch head `3a02225e0`: **105 suites pass, 1 fails**. The unit figure is **7004 passing, 0 failing, 17 pending**.

The one failing suite is 84, and it is not this branch's. Its truncation assertion requires every notice to be a whole number of cut lines; docker's follow window opens at an entry boundary, so a viewer can join a 1.3MB line part-way through and the notice for it is honestly short by whole frames that never crossed the socket. That is diagnosed and fixed on #1808, which scopes the invariant to lines the viewer saw whole, and those commits are not on this branch or on `development`. It has failed the same way on the v9 lineage for the same reason.

## 2. High — a restarted node can stay unconfirmed, and redeploys then remove apps

Both halves are built.

**Confirmation** is now the peer set agreeing rather than the first peer to answer: every capable peer's ask settles before the question is decided, and `minConfirmingPeers` of them must answer "not ahead". A peer holding nothing has said nothing about whether this node's policy is current, so it counts towards neither side. What the peer set cannot settle goes to the publisher, whose answer is signed — and that route is gated on **confirmation**, not on holding something, which is the change that reaches your case: a node that restored a bundle off disk still goes to the source.

One correction on the 24-hour figure. It was true of the code you were reading, where `considerSeeding()` returned early whenever a bundle was already held, so nothing retried and the tick was the only way out. It is not true now: the source is re-asked on every peer arrival, departure and settled ask while the node is unconfirmed, so `refreshIntervalMs` is a freshness tick rather than a recovery path. The exposed state also needs a bundle on disk **and** too few capable peers together — a node with an empty store is not in it, since the same route is gated on confirmation rather than on holding.

**The redeploy half is now fixed**, and at the door rather than underneath. A redeploy uninstalls and then installs, and it is the install that needs the blocked-repository list to judge the image — so the refusal arrived with the app already down, and only a successful install writes the local row back. `/apps/redeploy`, `/apps/redeploycomponent`, `/apps/installapplocally` and `/apps/testappinstall` now answer **503 with a `Retry-After`** before the handler runs, and the block scanner's reinstall pass declines its block and asks again a few blocks later. Removal is deliberately not gated: it rebuilds nothing, and a node that cannot obtain policy is exactly when clearing an app off it matters most.

Asked at the front door rather than as a `policyReady` check copied into each destructive site — that shape rots, and it also puts the refusal on the wrong channel. A request that reaches a handler is answered in the body at 200; a refusal before a handler runs is a wire status. This is the second kind, and it sits beside `requireBootSettled`, which already refuses app-control routes the same way for the same reason.

## 3. Medium — the ownership sweep acts on an unconfirmed bundle

Fixed in `3d7d0bfbc`. You identified the right condition; the fix does not `await` it in the boot chain.

Taken literally, `await globalState.waitForPolicyReady()` there would hang boot. `AsyncGate.wait()` has no timeout, and in `serviceManager.js` the `setInterval(portManager.restorePortsSupport, …)` sits after that call in the same sequential `startDbDependentServices` chain — so port restoration would never start on exactly the nodes your finding 2 describes.

So the wait is detached: `waitForPolicyReady().then(() => requestOwnershipSweep())`, with the `policyReady` check also inside `cleanupOwnershipViolations` itself so the destructive path cannot run without it however it is reached. Everything below that line starts whether or not the gate ever opens.

The sweep also subscribes to bundle changes and re-runs on each one while the gate is open, which closes a second defect in the same path: it was one-shot at boot, so an owner granted or revoked during the node's life was never acted on.

## Non-blocking

`appSpawner.js:837` — fixed. `datacenter` is enforced as a privilege on live submission only, so a spec replayed from chain can carry it and that test is live.

**No backoff on the source fetch — this one is deliberate and we want to keep it.** A node without policy is inert: it cannot validate a submission, pull an image or acquire an app. Retrying the publisher on every peer-set change is what gets it out of that state, and it is the reason the 24-hour figure above no longer applies. It is bounded by one fetch in flight at a time and by how often the peer set actually changes, and adding a backoff would be a latch on the one path that ends the outage.

**Your last bullet is truncated** — it ends mid-sentence at "repeated identical v2 requests (up to 500 hashes," and is followed by two lines duplicated from your "Verified as fixed" section. I will answer what I think it is pointing at rather than ask you to repost: `fluxapprequest` is classified as an ask, and asks are never deduplicated because `hash(data)` is the one part of the message that throws the sender away — deduplicating it would deliver one peer's request and silently drop every other peer asking the same thing. The extra load is the intended price of that, and each request is bounded at 500 hashes. If you meant something else, say so and I will take it properly.

Cabecinha84 (commented):
Finding 1 — Blocking: three destructive redeploy paths are still ungated

  The PR states its own invariant clearly: "Anything that would judge an app waits on that gate, and so does anything
  that would destroy one it then has to rebuild." Three internal callers tear an app down and then hit the new gate on
  the way back up:

  - imageUpdateService.js:333 → advancedWorkflows.softRedeploy(appSpec, null) (periodic, first run 10–30 min after boot,
    then on interval)
  - volumeValidationService.js:232 → hardRedeploy(appSpecs, null) (fired from serviceManager.js:434, early in boot,
    before policyStore.start() at ~line 507)
  - appInspector.checkStorageSpaceForApps → softRedeploy on a storage violation (serviceManager.js:833)

  softRedeploy removes at advancedWorkflows.js:1410 and only reaches checkAppRequirements / softRegisterAppLocally at
  1427–1429. With policyReady false, appInstaller.verifyAndPullImage throws (appInstaller.js:213) — so the app is down,
  its local row is gone, and nothing reconciles an app with no row. That is precisely the failure the PR fixed for the
  reinstall pass and for the API routes, just reached by a different door.

  Failure scenario: node boots during the rollout with policyReady still false (peers below 12, or capable peers below
  minConfirmingPeers: 4 and the published source unreachable); volumeValidationService finds one bad mount,
  hard-redeploys, the install refuses, and the app is gone from that node with no row to rebuild it.

  Fix: put the check at the top of softRedeploy / hardRedeploy, before the removal — one place then covers the routes,
  the reinstall pass, the image updater, volume validation and the storage sweep, instead of five call sites that can
  disagree.

  Finding 2 — Should fix: the backstop fetch has no failure backoff

  considerBackstopFetch is re-evaluated on every peer arrival, departure and settled ask, and backstopFetchInFlight only
  serialises concurrent attempts — a failed fetch leaves nothing behind. Probed at head:

  A) restored: isReady=true policyReady=false
  B) rollout (no capable peers), after threshold:        backstopFetches=1   policyReady=false
  C) after 10 connect+disconnect pairs, github down:     backstopFetches=21  policyReady=false
  D) after github recovers, one peer event:              backstopFetches=22  policyReady=true  seq=6

  So while the source is unreachable, an unconfirmed node polls raw.githubusercontent.com once per peer event, forever.
  That is the exact traffic pattern the module's own comments say it exists to avoid ("the source is rate-limited and
  shared by the fleet, so anything able to reach it runs on the backstop's timer"), and it is worst in the one window
  where it matters: the first rollout wave, where most nodes cannot confirm from peers, with a positive feedback loop if
  the source starts answering 429. A minimum interval or exponential backoff on a failed backstop fetch closes it.

  ---

  Two decisions to confirm (not code defects — but they should be explicit, not incidental)

  3. Strict pinning is applied retroactively to apps already running. The privilege checks are correctly
  live-submission-only, but placement is not. hwRequirements.checkAppNodesRequirements went from version === 7 to
  version >= 7, and the spawner filter dropped the version >= 8 bypass. A v8 app with a non-empty nodes[] could
  previously be placed on, and installed by, any node — the API accepted pins from any owner, as the PR body itself
  notes. After this ships, the first redeploy of any such app on a non-named node uninstalls it and then throws is not
  allowed to run on this node. This is arguably the intended enforcement, but it is a one-shot fleet-wide teardown of
  running, paid-for apps and belongs in the release notes with a product sign-off, not as a side effect.

  4. minimumFluxOSAllowedVersion: '8.13.1' → '8.18.0'. Two things here. First, the PR's justification for deleting
  messagesBroadcastRefactorStart — "the version floor disconnects anything old enough to need the legacy path" — is not
  accurate: the floor is read in exactly one place, fluxNetworkHelper.isFluxAvailable:431, and there is no version gate
  anywhere in the peer handshake (FluxPeerManager.add / FluxPeerSocket). Nothing is disconnected. (The removal itself is
  still safe — the fork activated October 2024 — but the stated reason isn't the mechanism.) Second, what the bump does
  do is make every node running this build answer "not available" for any node below 8.18.0, which pushes them toward
  DOS/unavailable. Given 8.18.0 (#1799) was blocked on the FDM#154 deploy as of the last I have on it, this needs
  confirmed fleet adoption before it ships.

  ---

  Non-blocking

  - appValidator gates on two different facts. The pinning and datacenter privilege checks (lines 1283, 1310) read
    getEnterpriseAppOwners() === null, i.e. bundle held; the image check 45 lines below (1357) reads
    globalState.policyReady, i.e. confirmed. A node holding an unconfirmed restored bundle can reject a live submission
    with Pinning ... is only available for enterprise app owners when the honest answer is network policy not yet
    obtained — and by the PR's own argument that is a diagnosis of the wrong thing.
  - Stale references to renamed/removed functions: FluxPeerManager.js:638 cites policyStore.seedIfPeersHaveNothing (now
    considerBackstopFetch); fluxCommunicationMessagesSender.js:219 cites requestPolicyFromPeers (gone);
    ipLocationSync.js:22 cites policyStore.startSync() (it's start()).
  - GitGuardian is red — false positive. One secret: the WIF 5JTeg79… added at fluxNetworkHelper.test.js:466,479. That
    literal is already committed on development in seven files (34 occurrences in
    fluxCommunicationMessagesSender.test.js alone). Needs the usual dismissal, but it is a long-standing throwaway test
    fixture.
  - fluxpolicy is now the largest message type on the wire (1 MiB MAX_BUNDLE_BYTES) and there is no maxPayload on the
    websocket server — the cap is applied only after the frame has been fully received and parsed. Pre-existing
    infrastructure gap, newly relevant.
  - Adoption announcements dedup on hash({type, version, seq}), which is byte-identical across peers by design. Only the
    first peer's announcement of a given sequence is acted on within the 5-minute TTL. Covered in practice by the
    per-arrival ask, but it is a real narrowing worth knowing about.

MorningLightMountain713:
Thank you — nine items, all answered. Five fixed, one refuted, one closed by removing what it described, one is a dismissal on our side, one confirmed pre-existing and deferred. Twenty commits are now on the branch; the tip is `66a10c08c`.

## 1 — three destructive redeploy paths are ungated (blocking): fixed, `4b2ff85cc` + `fbdaa6d31`

Correct, and the sweep was wider than the three paths you named. `mayTearDownToRebuild(appSpecs, res)` asks both conditions before anything comes down, and every path that tears an app down calls it first: `softRedeploy:1461`, `hardRedeploy:1605`, `softRedeployComponent:1691`, `hardRedeployComponent:1859`, and the reinstall pass at `advancedWorkflows.js:4321`. The two component redeploys have the same shape as the whole-app ones — a spec pinned to other nodes does not belong here one component at a time.

The two answers differ. Policy is a wait: the app is left exactly as it is and the caller asks again on its own schedule. A spec naming other nodes is not a wait — the app does not belong here, so it is uninstalled and the network told, which withdraws this node's location record and frees the app to be placed where it is named.

The pin is read before the policy because it needs none: `nodes` is on-chain and signed, and removing an app outright is the one destructive act that does not depend on the bundle. A positive match stands on whichever identifier resolved; a negative one requires both. An entry names a node by socket address **or** by collateral outpoint, so an unresolved identifier does not narrow the answer — it removes this node's ability to recognise one whole form of the name. Read as "not named", an outpoint pin on a node whose daemon is unreachable would delete an app the spec names, and an address pin would do the same while benchmark is down.

`fbdaa6d31` covers the arming side you identified: volume validation, the image updater and the storage sweep each destroy an app in order to rebuild it and each was armed on a timer alone. Each now waits on the gate as well. Volume validation keeps its settling delay too — that one is about docker, mounts and crontab, which the gate says nothing about — so it holds both conditions rather than treating either as the other.

## 2 — the backstop fetch has no failure backoff: fixed, `197a77bdc`

Fixed as the first of the two things you offered: a minimum interval, not exponential backoff. The decision stays derived from the peer picture and re-evaluated wherever it changes, which is what keeps it free of a latch; what is paced is the retry, because a peer connecting or leaving is not evidence about whether the source is reachable.

`considerBackstopFetch` returns early while `monotonicMs() - lastBackstopAttemptAt < config.policy.backstopRetryIntervalMs` (60s), and the stamp is written in a `finally`, so a refusal stands as the answer for the interval exactly as a success does. It is a floor, not a backoff: it does not grow, it caps no number of attempts, and the next peer event after it passes tries again.

The stamp is monotonic. A node boots, ntp steps the clock, and an interval measured on wall time is spent on the spot — in exactly the window a restart wave makes both most likely and most expensive.

## 3 — strict pinning applied retroactively: fixed, `4b2ff85cc`

Your reading of the mechanism is right: placement is not live-submission-only, `checkAppNodesRequirements` is now `version >= 7` (`hwRequirements.js:301`), and the spawner filter no longer bypasses v8.

What a non-named node does is now a clean hand-back rather than a teardown followed by a refusal: it uninstalls the app and broadcasts the withdrawal, so its location record goes with it and the app is free to be placed where it is named.

The cost is real and is now in the PR body rather than left to be discovered. The departures are not ordered against the arrivals, because the spawner counts an app's locations without asking whether its pin names them — an app at its instance count on the wrong nodes is short nowhere, so no newly named node installs it until a copy has already left. Re-pinning an app whose data is synced is therefore destructive for the owner to back up before. Ordering it properly needs placement to count eligible copies rather than location rows, which is v9 work.

Measured against the live corpus before deciding: of 1,345 global specs, 11 carry a non-empty `nodes[]` and 10 of those are v7, which is already enforced today. There is exactly one v8 pinned app and it has zero live locations. `nodes` is not stripped from the enterprise envelope — only `compose` and `contacts` are — so that count is not hiding the 298 enterprise apps.

## 4 — `minimumFluxOSAllowedVersion` 8.13.1 → 8.18.0: both halves refuted

On the PR body's justification, the grep is right and the conclusion does not follow. The floor is read only in `fluxNetworkHelper.isFluxAvailable`, and nothing gates the peer handshake on version — but the sentence is about the floor that was already in force, not about the bump. `8.13.1` was set on 2026-06-03 in `f75f40ee2`, roughly twenty months after the October 2024 fork. Nothing the network will speak to is old enough to need the legacy path, which is what makes removing it safe. The sentence stands as written.

On fleet adoption, the network as counted:

| version | nodes |
|---|---|
| 8.18.0 | 6269 |
| 8.17.1 | 1 |
| unknown | 66 |
| **total** | **6336** |

8.18.0 is the network, not a target to reach.

## 5 — appValidator gates on two different facts: fixed, `e2be3f087`

Both privilege checks now read `globalState.policyReady ? enterpriseConfig.getEnterpriseAppOwners() : null`, folding unconfirmed into the null the rest of the module already means by *not known yet*. A node holding an unconfirmed restored bundle now refuses a live submission with "network policy not yet obtained", which is the diagnosis you said it owed the submitter.

The same commit closes a second way those lists answered wrongly. Membership was an exact string match, and capitalisation is not part of an Ethereum address — it is an EIP-55 checksum over the same twenty bytes, which is why `verifySignature` compares recovered addresses without it. An owner writing their own address in the other valid capitalisation was a stranger to every list while still signing as themselves: refused a pin, refused datacenter, refused cpu burst, and swept off an enterprise node as an app that does not belong there, with the removal broadcast. All seven comparisons go through one helper now, and a Flux ID falls through to an exact match.

## 6 — stale references to renamed or removed functions: fixed, `6203c1907`

Your three, and a fourth in `placementFeasibility.js`.

## 7 — GitGuardian: verified false positive, dismissing

Your reading is exactly right. That WIF is on `development` in seven files, 34 occurrences in `fluxCommunicationMessagesSender.test.js` alone. Nothing on this branch introduces it.

## 8 — no `maxPayload` on the websocket server: confirmed pre-existing, separate PR

Confirmed pre-existing rather than merely long-standing: `ZelBack/src/lib/socketServer.js` is byte-identical between this branch and `development`, same blob, so nothing here touches it.

Choosing a number needs the largest legitimate message measured rather than guessed — `appHashSyncService` chunks at 10,000 hashes, around 640KB of hash characters before JSON and the envelope — and a cap set too low closes peer connections with 1009 and breaks app sync. It goes in its own PR with that measurement behind it.

## 9 — adoption announcements dedup on content: closed, `d55daf87d` + `8fc17106d`

The narrowing you describe is gone, by removing the dedup rather than working around it. A message that is never relayed cannot arrive twice by different routes, so it does not belong in a content filter at all: `fluxpolicyseq` declares `INTENT.VARIES` in the route table and the duplicate filter claims a slot only for `INTENT.ANNOUNCE` (`fluxCommunication.js:686`). Every peer's announcement of a sequence is now acted on.

Worth recording that confirmation and the quorum were never exposed to it either way: `settlePeerAsk` returns unless the message carries a correlation id, announcements carry none, and only replies settle an ask. An announcement being collapsed could never have counted towards or against confirmation.

## Also in this range

Not attributable to your items, in the same push:

- `8d75b3373` — the spawner recorded "could not read the blocklist" as a seven-day per-app error; a blocklist that cannot be read now ends the pass instead.
- `693c9a933` — the tampering enforcer started at boot, before the bundle resolves, with its next tick 12 hours away and no subscription. It starts on the gate now.
- `e9fab75e9` — the hourly compliance sweep had no policy gate. Fail-safe is only one direction: a stale list also carries bans lifted since, and that sweep uninstalls with `sendMessage: true`.
- `a12f6fa70` — a log stream joined part-way through spliced docker's timestamps into the joined line.
- `29841682c` — the bundle size cap counted UTF-16 units; it counts bytes.
- `882b15d3c` — the periodic refresh bypassed the in-flight flag and the retry stamp, so it reached the source through a different door than `considerBackstopFetch`.
- `9ac2024fc` + `4ceb9231d` — `registerAppLocally` returns an outcome rather than a value that has to be read out of a message. `REFUSED` was three facts under one name — another operation holds the node, the attempt will not proceed, and the app is already installed — which demand three different reactions, so it is now `BUSY`, `DECLINED` and `ALREADY_INSTALLED`. Every outcome except `BUSY` is returned after the install hold is acquired, so concurrent attempts are serialised and the losers get `BUSY`.

## Evidence

Full parallel gate on a real fleet, five nodes per suite, at the tip `66a10c08c`: **106 suites pass, 0 fail**, 100 minutes.

Full unit suite: **7050 passing, 0 failing**, 17 pending. eslint clean.

The PR body is updated — the evidence section now describes this tip, and the re-pin note from item 3 is in the pinned-placement section.

Cabecinha84 (approved):
ack

PR #1806, opened 2026-09-15
fix: the node's own geolocation is a copy, and NodeJS has a version floor

MorningLightMountain713:
Two changes: a node silently refusing region-pinned apps, and a NodeJS version floor.

## The node's own geolocation is a copy

`getNodeGeolocation()` returned the module-level cache by reference. `getHostInfo` — the `/hostinfo` endpoint app containers call — removes `ip` and `org` before replying, so those fields were deleted from the node's own record rather than from a copy of it.

A node that no longer knows its own address cannot resolve itself in the published IP-location table, so `checkAppGeolocationRequirements` holds no region for it. An app pinned to a region names that region the way ip-api does; the published vocabulary resolves the name, and a resolved name turns off the self-reported fallback, leaving the table as the only thing that can grant a match. The node then refuses every region-pinned app while continuing to volunteer for them, because selection reads the node's address from the network layer and still places it correctly. The state outlives a restart: the record is written back in its edited form, and the address is only re-fetched on every fourth pass of a three-day loop.

The accessor answers with a copy, so the removals in `/hostinfo` shape that reply alone. The reply itself is unchanged — same two fields dropped, same fields sent.

`docs/flux-node-service.md` now lists what the endpoint actually sends. Its `geo` table named a `city` field that has never been in a payload and omitted `isp`, `asn`, `mobile`, `proxy` and `hosting`, which are. `static` and `dataCenter` there describe the range an address sits on; this node's own verdict on holding a fixed address is reached from evidence it observes about itself and is not what this endpoint carries.

## NodeJS has a version floor

`minimumNodeJsAllowedVersion` is 20.8.0, asked once during startup. A node on a runtime the code was not written against fails whichever paths need what it lacks, one at a time and silently, rather than reporting itself unfit.

The version is a property of the running process, so it is asked once rather than on a loop. The verdict is held in the sticky DOS slot: a successful availability pass ends in `setDosMessage(null)`, which would clear an ordinary message and let the node return to service on a runtime that cannot run the code, with nothing left to ask again.

The verdict is held under this check's own identity, so it stands whatever else has the node out of service and outlives the release of anything that has.

No floor configured allows rather than refuses. The check runs bare in `startFluxFunctions`, whose catch re-enters it, so a throw is a boot loop; and unsetting the key through a config layer is what takes the floor off a fleet, which a node held out of service by its absence could not act on.

A node below the floor reports `NodeJS Version Error. Current lower version allowed is v20.8.0 found v<version>`.

## A sticky DOS is held per owner

The sticky DOS slot held one string that was at once the reason an operator reads, the identity a writer recognises its own verdict by, and the latch that takes the node out of service. Four services need it to say this node is permanently unfit, and one slot expresses one of them, so each writer tested ownership by matching a prefix of the message and stood down when the slot was another's. A verdict that stands down is not recorded: the node returns to service on the other owner's release, for a condition that never lifted.

Verdicts are now keyed by owner, the shape `placementHolds` beside them already takes. The node is out of service while any owner holds one, the reported reason names every one of them, a release takes only its own, and an owner the module does not know is refused rather than minted.

That deletes the prefix match in each writer, the per-service flag tracking whether its verdict was still in the slot, and the tampering enforcer's deferred message and 60-second watch for the slot to come free. The watch is the reason this is a fix and not a tidy: it wrote DOS 100 - every app directory and volume on the node removed, and the removals broadcast - from a verdict cached at an earlier tick. Only an enforcement pass discarded that cache, and those are twelve hours apart, so a node taken off the blocklist in between was destroyed for a listing that no longer existed. A verdict is now written by the tick that reaches it, and there is nothing to wait for.

## The eviction shape is deliberate here, and is being reworked in v9

A node below the floor holds a sticky DOS of 100, and that does not merely take it off the network: `isNodeDos()` makes `appStartupManager` and `nodeStatusMonitor` remove every app the node holds, delete the volumes, and broadcast the removals so the network re-places those instances elsewhere.

The set of nodes this affects was measured against the live fleet, and the immediate removal is signed off. This is decided, not an open question.

`residentialNodeDosService` stages the same outcome rather than taking it in one step — hold placement, hand apps back one at a time and only to hosts that demonstrably have them, DOS once the node is empty. That machinery is welded to the residential verdict: the placement-hold owner, the settle marker, the departure pacing and the give-up pass each name it. Making it available to a second reason means a generic node eviction service, and that reaches across the DOS state, the sticky verdicts, three separate enforcing services and the harness suites that pin their behaviour. Those services are disparate enough that doing it here would be a change of a different size and shape to this one, so it is scoped as a v9 workstream.

The floor therefore ships with the immediate removal, and the staged path arrives with that service. One part of the staging would buy nothing here in any case: the settling window exists because a geolocation classification can be misread, and `process.versions.node` is a direct reading of the running process with nothing to confirm.

## Testing

7071 unit tests pass and lint is clean. The integration gate ran the full 103 suites on a five-node fleet: 102 passed, and the one red was suite 85, whose own fix has since landed on development with #1804.

Both branches of the accessor break independently: reverting the in-memory return fails the in-memory test alone, reverting the database-restore return fails that one alone. `/hostinfo` is covered end to end against the real accessor — removing the two deletions fails the reply tests and leaves the record tests green, and reverting the copy does the opposite. Turning the unset-floor branch into a throw fails only the two tests written for it, and a third loads the same way with a floor present so a config stub that never reached the module could not satisfy them.

Each property of the owner-keyed verdict breaks on its own: releasing every owner's verdict instead of one fails the two tests for a second owner alone; overwriting rather than joining fails the joined reason and the floor standing beside another owner; a hold that no longer reaches 100 fails nine; accepting an unknown owner fails the refusal; and dropping the idempotence guard fails only the test that the verdict is stated once however often startup re-enters the check.

The geolocation suite also passes under Node 16.20.2, so the accessor change carries no new runtime requirement of its own.


Cabecinha84 (commented):
Finding 1 (blocking-ish) — it bypasses the codebase's own staged-eviction policy

  residentialNodeDosService.js:1-17 states the house rule for a permanently unfit node, in its own words:

  ▎ HOLD … it stops accepting NEW apps. Immediate, and it deletes nothing.
  ▎ EVACUATE … it gives up one app at a time, and only ones another host demonstrably holds.
  ▎ DOS … once the node runs nothing, the sticky DOS goes on.
  ▎ "DOS >= 100 is not a mark: it makes nodeStatusMonitor and appStartupManager rm -rf every app directory and volume on
  ▎ the box. Reaching that state only on an empty node is the whole point of the staging."

  This PR introduces a new permanent-unfitness verdict and jumps straight to sticky DOS 100 at the top of
  startFluxFunctions. I traced the consequence:

  - appStartupManager.js:318 → removeAllApps('Node DOS') → removeAppLocally(name, null, true, false, **true**) —
    broadcast hardcoded on.
  - nodeStatusMonitor.js:65 → removeAllAppsLocally(…, 'DOS state >= 100'), broadcasting.

  So on the first boot after this release, 19 nodes wipe every app directory and volume they hold and broadcast the
  removals simultaneously. PlacementHoldOwner exists for precisely the "stop growing, keep the customer volumes" case
  and gets no new value here.

  To be fair to the PR: half the residential staging doesn't apply. The settle window exists because a geolocation
  classification can be misread; process.versions.node is authoritative and needs no window. But the evacuate-before-DOS
  half does apply, and for the same reason: these 19 nodes are demonstrably running those apps today, and the PR's own
  argument is that below-floor failures are "silent and path-specific" — i.e. they've been in this state for months. A
  few more days of staged hand-off costs nothing and avoids destroying volumes another host could take over gracefully.

  Ask: either register a PlacementHoldOwner.NODEJS_FLOOR and reuse the residential pacing, or get explicit team sign-off
  in the PR that immediate wipe is the intended trade — and say why the staging its own neighbour file documents is
  being skipped.

  Finding 2 — the ownership convention is declared but not followed

  NODEJS_DOS_MESSAGE_PREFIX is introduced with a comment about marking the slot "the way every other writer of the slot
  marks its own" — but it's only used for string concatenation. There's no isOurStickyDos(), and
  checkNodeJsVersionAllowed calls setStickyDosMessage unconditionally, without the if (sticky && !isOurStickyDos())
  return; guard that all three existing writers have (residentialNodeDosService:608, appTamperingBlocklistService:233,
  peerSetStabilityService:107).

  Today it can't clobber anything because it runs first at boot. But the constant's only purpose is ownership, and it
  isn't used for one — so if the check is ever moved, re-run, or called from a second site, it silently takes the slot.
  Three lines to make the declared convention real.

  (Related, cosmetic: on a below-floor node appTamperingBlocklistService will startSlotWatch() and poll every 60 s
  forever waiting for a slot that never frees. It's unref'd, so harmless.)

  Finding 3 — the privacy claim in the /hostinfo comment isn't what the code does

  The new comment says "The node's own address and the operator holding it are not the container's business." But the
  code only deletes ip and org, while storedGeolocation on the ip-api branch also carries isp, asn, proxy, hosting,
  mobile, static, dataCenter — all of which still reach the app container. And since org: ipRes.data.org ||
  ipRes.data.isp (:319), whenever ip-api returns no org, deleting org while keeping isp hides literally nothing.

  This leak is pre-existing, so it's arguably out of scope — but the PR is editing these exact four lines and asserting
  the opposite intent. Now that the accessor returns a copy, the clean shape is an allowlist (hostInfo.geo = {
  continent, continentCode, country, countryCode, region, regionName, lat, lon }) rather than a denylist, which also
  means a field added to storedGeolocation later isn't auto-exposed to containers. Either fix it here or drop the claim
  from the comment.

  Finding 4 — the actual bug site has zero test coverage

  There is no tests/unit/fluxNodeService.test.js, and grep -ri hostinfo across test-infra/ returns nothing. The new
  tests prove the accessor returns a copy; nothing proves that getHostInfo still strips ip/org from its reply. If
  someone later "optimises" the copy away or reorders the deletes, no test catches either half. A small getHostInfo test
  asserting (a) no ip/org in geo, and (b) getNodeGeolocation() still has them afterwards, closes the loop on the real
  bug.

  Finding 5 — test pollution hazard (minor but genuine)

  tests/unit/fluxNetworkHelper.test.js:775:

  Object.defineProperty(process.versions, 'node', { value: version, configurable: true });

  enumerable defaults to false, and the afterEach restore goes through the same helper — so after the first call,
  process.versions.node is permanently non-enumerable for the rest of the mocha process. fluxService.getNodeJsVersions()
  returns process.versions straight into res.json(), so any future test asserting the /flux/versions payload in the
  same run would find node missing. No test does today, so nothing is broken — but capture and restore the original
  property descriptor instead.

  Finding 6 — missing test for the !minimumVersion branch

  The comment justifies the fail-open path at length (a throw would boot-loop through startFluxFunctions's 15 s catch).
  Nothing exercises it. One test.

  Unverifiable from the repo

  The "19 nodes below the floor, nothing between 16.20.2 and 20.8.0" census is the load-bearing claim for the whole
  blast radius, and it can't be checked from the code. It needs an ops sign-off. One mitigation worth noting:
  minimumNodeJsAllowedVersion lives in config/default.js and the check fails open when unset, so node-config layering
  gives a real rollback lever if the census turns out to be wrong.

  (Small consolation on severity: DOS ≥ 100 is FluxOS-side only — it doesn't touch fluxd confirmation, so operators
  don't lose collateral. They lose their apps and any app revenue.)

MorningLightMountain713:
Rebased onto `development`, so line numbers below refer to the current tip.

**1 — staged eviction.** The immediate removal is intended, and the PR body now says so and says why the staging next door is not reused. The short version: that machinery is welded to the residential verdict — the placement-hold owner, the settle marker, the departure pacing and the give-up pass each name it — so a second reason needs a generic eviction service, which reaches across the DOS state, the sticky slot, three enforcing services and the suites that pin them. It is scoped as a v9 workstream rather than bolted on here. The settling window would buy nothing for this check regardless: it exists because a classification can be misread, and `process.versions.node` is a direct reading.

**2 — ownership convention.** Fixed. The slot is taken only while free, and the check recognises its own message before writing. Your reasoning for why it did not matter was the part that was wrong: `startFluxFunctions` re-enters itself 15s after any throw, and the other writers of the slot start partway through it, so every retry reaches this check with all of them live. Two tests, both of which fail if the guard is removed.

**3 — the privacy claim.** Correct, and it understated it: `hostInfo.ip` is set two lines above the deletions, and `/flux/geolocation` serves the whole record unauthenticated, so nothing is withheld either way. The comment no longer claims otherwise. The payload is unchanged — an allowlist would break any container reading `geo.isp` today, and that is a documented app-facing API. What was wrong is the documentation, now corrected: the `geo` table named a `city` field that has never been sent and omitted five that are, and `static`/`dataCenter` are described as what they are, properties of the range rather than this node's own verdict.

**4 — coverage at the bug site.** Fixed. New `tests/unit/fluxNodeService.test.js` covers `getHostInfo` against the real accessor rather than a stub, so both halves are caught separately: removing the deletions fails the reply tests and leaves the record tests green, reverting the copy does the opposite. A stubbed accessor could only ever see the first.

On the `serviceManager` call site — agreed, and it is not fixable here. Proving that boot wiring happened needs the boot sequence to be reachable from a test, and it is not: ~40 file-scope requires, two dependency loops that need a real mongo and a real docker, a `process.exit()`, and no seams. That is true of every service it starts, not this one check, and it is written up as its own piece of work. A test that reads the source to assert a call is present is not a test.

**5 — test pollution.** This one does not hold. `Object.defineProperty` on an **existing** property changes only the attributes named in the descriptor; the `enumerable: false` default applies to property *creation*. Measured on the runtime: before `{value:"22.17.1",writable:false,enumerable:true,configurable:true}`, after `{value:"16.20.2",writable:false,enumerable:true,configurable:true}` — identical but for the value, `node` still in `Object.keys`, still in the JSON payload. `/flux/versions` cannot be affected, so nothing changed here.

**6 — the unset-floor branch.** Fixed. It is also the lever that takes the floor off a fleet, so it earns a test more than most. Worth noting for anyone writing one: node-config seals its values once the module graph loads, so the key cannot be reassigned in place — the helper is loaded over a config carrying a different floor. A third test loads the same way *with* a floor and asserts the node still goes out of service, so a config stub that never reached the module could not satisfy the other two.

**The census.** Measured against the live fleet and signed off by the team before this was raised. It has also come out of the PR body: every node serves its exact runtime version publicly and unauthenticated, so naming the population and its support status here is a target list with a deadline on it.

Unit 6743 passing, 0 failing, lint clean. Every new test fails against the defect it covers.

Cabecinha84 (commented):
Minor, non-blocking notes

  1. Narrow edge case. If the slot is ever held by someone else when the check runs, the NodeJS verdict is never
     recorded. When that owner releases the slot, the node would return to service on an old runtime. Today this cannot
     happen: on first boot the slot is always empty, and on a retry this check already holds it. It would only matter if
     someone moved the call later in startup. Worth a note, not a change.
  2. Pre-existing and harmless. On a below-floor node that is also on the tampering blocklist,
     appTamperingBlocklistService keeps a 60 s poll (with unref) running forever, waiting for a slot that never frees.
  3. The boot call has no test. The author's reasoning (serviceManager can't be reached from unit tests) holds; it's the
     same for every other service it starts.
  4. The check can't protect against crashes that happen earlier. Code that fails while modules load on an old Node
     crashes before this line runs. That's inherent, not a defect.

  The one real condition before merge

  A node below the floor is wiped immediately. It gets a sticky DOS of 100, which triggers:
  - appStartupManager.js:318 → removeAllApps, which broadcasts the removals;
  - nodeStatusMonitor.js:65 → removeAllAppsLocally, which deletes the app directories and volumes.

  There is no staged hand-off like the one residentialNodeDosService uses (hold placement → hand apps over one at a time
  → DOS). On the first boot after the release, every affected node drops all its apps and volumes at once.

  - The author's reasoning is fair. A second reason for staged eviction needs a generic eviction service (planned for
    v9). The "settle window" part is pointless here, because process.versions.node is a direct reading of the runtime.
  - What's missing is the sign-off on record. The author says the census of affected nodes was measured and the team
    signed off, and deliberately removed the numbers from the PR body. That reason is sound: every node publishes its
    Node version. But the sign-off itself is not written anywhere in the PR, and this is the only irreversible part of
    the change.

MorningLightMountain713:
All five closed. The branch is rebased onto development and the tip is `e3452ba15`.

**The slot itself is gone, rather than the edge being documented.** Finding 1 was right about the mechanism and understated the consequence: `clearStickyDosMessage()` zeroed the state value as well as the message, and the floor is asked once at startup, so a verdict that stood down had nothing left to re-ask — the node returned to service on the other owner's release. Rather than comment that edge, the sticky DOS is now `owner -> reason`, the shape `placementHolds` twenty lines above it already used, and for the reason its own comment gives: a single slot cannot express two owners, so a second verdict either overwrites the first — stranding an owner that can no longer recognise, and so never release, its own DOS — or is dropped.

The node is out of service while any owner holds a verdict, the reported reason names every one of them, a release takes only its own, and an owner the module does not know is refused rather than minted. That deletes the prefix match in each of the three writers, their per-service claim flags, and the `NODEJS_DOS_MESSAGE_PREFIX` ownership constant this review asked to be made real in round two. The floor check is now eight lines with no ownership logic in it at all.

**Finding 2 was not harmless, and that word is why it survived.** The 60-second poll cost nothing in CPU; what it did on firing was write DOS 100 — every app directory and volume removed, and the removals broadcast — from a verdict cached at an earlier tick, with nothing re-checked at the moment of application. Only an enforcement pass discarded that cache and those are twelve hours apart, so a node taken off the blocklist in between was destroyed for a listing that no longer existed. On a hand-curated list, being delisted is a human correcting a mistake, which is the one case where destroying the node is least defensible. It is gone with the slot it was working around, and a verdict is now written by the tick that reaches it.

**Findings 3 and 4 stand as you put them.** The boot call is not asserted, and that is true of everything `startFluxFunctions` starts; it is tracked on our side rather than papered over here. A node that cannot load on an old runtime is off the network either way, so the floor being unreachable in that case costs nothing.

**On the census:** the affected set was measured against the live fleet and the immediate removal is signed off, which the body now says. The figures stay out of a public PR.

7071 unit tests pass and lint is clean. The integration gate ran all 103 suites on a five-node fleet, 102 green, and the single red was suite 85, whose own fix has since landed on development with #1804. Each property of the owner-keyed verdict was broken separately to check the tests can fail: releasing every owner's verdict instead of one, overwriting rather than joining, a hold that no longer reaches 100, accepting an unknown owner, and dropping the idempotence guard each kill a different set.

Cabecinha84 (approved):
ack

PR #1807, opened 2026-09-16
fluxapprunning is a placement claim, not a health report

MorningLightMountain713:
## The defect

`fluxapprunning` was derived from container run-state: a node announced an app only once all of its components were running. That makes the announcement a health report, but the spawner reads it as a **placement** record — it counts announcements against an app's `instances` target.

An app that could never start therefore silenced every node holding it. The registry saw a count far below the target and the spawner placed it again, without end. One app asking for **30 instances reached 618 nodes**, reserving hundreds of terabytes, and was still spreading when it was found. The signal that should have stopped the spawner was the one the breakage suppressed.

## A node announces what it holds

`applicationsToBroadcast = appsInstalled` — a node announces the apps it holds, whatever its containers are doing. A container that is down is recovered locally by the reconciler, not relocated, and whether an instance serves traffic is settled by the load balancer's own health check. A broken app now occupies exactly its `instances` and is visible at the right count, instead of metastasising.

`/apps/heldcomponents` already answers the ownership question, and `g:`/`r:` apps have been exempt from the run-state filter since 2024 — every app now behaves the way those already do.

## A node stops claiming an app when it decides to hand it back

A removal tells the network an app is gone, and the app stays installed until that removal finishes. An announcement built in that window named an app the node had given up, and because peers apply the two messages in arrival order, the late claim re-created the location row the removal had just cleared. The row then stood for its full expiry with nothing left to refresh it, and the spawner counted an instance the node no longer had.

`removeAppLocally` marks the app on entry under the name its removal message carries, and the announcement skips what is marked. Only a removal that tells the network marks: a removal whose containers are coming straight back keeps announcing, because silence there lets the row lapse and the app is placed a second time. The release is bound to the mark that took it, since the guards that refuse a removal return through the same `finally` without having marked anything.

The mark is **counted**. A forced removal skips the guard that admits one removal at a time, so two can run against a single app at once — a surplus trim and an expiry removal, or an app and one of its components, which share the name the removal message carries. Both mark; the first to finish would clear the mark outright and hand the announcement back to the removal still running, which is the window the mark exists to close.

The node's removal lock is released the same way. A call that finds the lock held returns through the same `finally` as the call holding it, and clearing it there frees the node while that removal runs — which an install reads as permission to start.

## A failed placement takes back what it claimed

An app's row in the local table is written **before** its image is pulled, and the announcement reports the apps installed on this node. A periodic announcement landing inside an install therefore says the node holds an app it has not got yet. That is harmless while the install is still running — it finishes, and the claim becomes true.

It is not harmless when the install fails. The spawner counts the claim against the app's `instances` target, and the location row stands for its full lifetime (7500s) with nothing left to correct it: `fluxappinstallingerror` clears the installing row and the installing broadcast, and never touches `appsLocations`. The app runs an instance short for two hours.

The teardown is the only thing that can retract it. The two paths that place an app this node does **not** already hold — the spawner, and the manual install API — now tear down with a broadcast removal, and only where a claim was made: the row is what the announcement reads, so an install that gives up before writing it has nothing for a peer to clear. The rebuild paths keep their silence deliberately: the claim there is one the node is keeping, and a removal would clear a row the app is coming back to.

### Why not simply stop announcing an app that is installing

Because on this lineage the announcement is the only thing holding the placement once the install passes 15 minutes.

The `fluxappinstalling` claim is what reserves a placement while an install runs, and the spawner counts it alongside running instances. On `development` it is broadcast once and never renewed, so it expires at `installingTtlS` (900s). Multi-component image pulls routinely outlive that. Today the announcement covers the gap; suppress it and a slow install loses its seat mid-pull, the app is placed elsewhere, and the first install completes into a surplus. That is a worse failure than the one being fixed.

`feat/fluxos-v9` renews the claim every 12 minutes (`startInstallingRenewal`, `installingRenewalS: 720`, 80% of the TTL) and can therefore do it the clean way. Backporting that is a protocol change — `development`'s message store rejects the v2 shape the renewal uses, and election ordering needs a stable `announcedAt` — so it is a separate piece of work, not this PR.

## A removal waits for the announcement it would contradict

An announcement names the apps this node holds; a removal says it holds one of them no longer. A cycle that took its list before the removal marked the app still names it, and a peer applies whichever arrives last — so a claim landing after the removal re-creates the location row it just cleared, and that row stands for its full lifetime with nothing left to correct it.

The order is settled where both messages come from. A removal that tells the network waits for the cycle in flight to finish sending: the claim is applied first and the removal clears it, and a cycle starting after the mark leaves the app out entirely. The wait is bounded, because a wedged cycle must not hold up removals — every install and redeploy queues behind one. A redeploy does not wait: it says nothing to the network, so it has nothing to contradict.

`broadcastedAt` is taken where the snapshot is rather than where the message is built, since every consumer reads it as the time these apps were installed here. A stamp taken after the per-app reads out-ranks a removal that happened while they ran.

## A test install claims nothing

A test install writes the app's row like any other install and throws it away, and the announcement is built from that table — so a cycle landing inside one names an app the node is only trying out, and the teardown that throws the row away says nothing to the network.

Marked for the length of the install and released once the teardown has deleted the row, which is what the announcement reads. Nothing is retracted, because nothing is claimed.

## A rebuild reinstalls as soon as its teardown reports it is done

A redeploy deletes the app's local row and re-inserts it, and the announcement is built from that table — so for as long as the node sits between the two it does not name an app it fully intends to keep. A node receiving that announcement directly leaves the row alone; one that is re-syncing prunes the sender's rows against the newest broadcast it has, and drops it.

The teardown confirms its own completion before it returns. Container removal is awaited, the network removal carries its own wait, force and retry and reports whether it succeeded, and `ensureAppDockerNetwork` meets a network that survives anyway by checking for one before creating it. The requirements check that follows sums the hdd each row in the local table declares, so deleting the row frees that space in the arithmetic outright rather than when the filesystem catches up.

Nothing in that sequence needs a pause, so the gap is the requirements check alone. Spreading a spec change across the fleet is the sibling `redeploy.probability`, which is untouched.

### What this leaves

Stated plainly rather than discovered later:

- **A node still claims an app while it is installing it.** The claim is wrong for the remainder of the install rather than for two hours after it, and it resolves either way inside the install's own lifetime — it becomes true, or it is retracted.
- **An install that outlives `installingTtlS` still loses its seat.** Pre-existing, and what the v9 renewal fixes.
- **If FluxOS dies mid-install nothing is broadcast**, so the claim lives its full expiry, and the local row says the app is installed. In-memory state cannot cover this; it needs a durable install state on the row.
- **The retraction is best-effort.** The removal message carries this node's address, which it asks fluxbench for as it sends; an answer it cannot use leaves nothing to broadcast. A node that cannot sign as itself is the same case. Both are logged, and in both the peers keep the claim until it expires.
- **A redeploy has a short window where the app is in neither place** — between its local row being deleted and re-inserted, which spans the requirements check. A node receiving the announcement directly leaves the omitted row alone, so the usual cost is one missed refresh; a node that is re-syncing its state prunes the sender's rows against the newest broadcast it has and drops that row outright.

`fluxModels/workstreams/placement-and-election/APP_PRESENCE_CLAIMS.md` carries the full model and both builds that close these.

## What this lineage allows

Several of the choices here are the best available on `development`, not the best available.

A node cannot stop announcing an app it is installing, because the installing claim is broadcast once and never renewed on this lineage — so the announcement is the only thing holding a slow install's placement.

Peers cannot settle two contradictory messages by what they say. The location table is the live mechanism here, and a removal deletes from it — discarding the timestamp that would let a peer reject a stale claim. So the contradiction is prevented at the sender instead: a removal waits for the announcement it would contradict. That holds for the peers this node sends to, and it is not the same guarantee as every peer being robust to any arrival order.

The node's operation locks are booleans any caller may set, so binding a release to its acquisition closes the case in front of it rather than making the class impossible.

Each of these is architectural, and each is what the v9 work exists to change: the installing claim gains its renewal, the event log becomes the source of truth peers fold rather than a record nothing reads, and a lock becomes something you cannot release without having taken it.

## Deliberate choices

- **An app that crash-loops forever on one node is never relocated.** The reconciler gives up only when it cannot create a container, not when one keeps dying, so the instance is recovered in place and the load balancer takes the dead one out of rotation. This is the intended consequence: a node-specific fault costs the owner an instance until they act, instead of the app metastasising. **Signed off explicitly.**
- **The reconciler boot-drain gate stays.** The snapshot waits for the reconciler's first pass over the apps held at boot: a pass that cannot recreate a container uninstalls the app, and a claim made before it would have to be taken back by a broadcast that is best-effort. The gate resolves immediately in steady state and is capped reconciler-side, so a wedged reconcile cannot hold up the node's network presence.
- **The broadcast is not filtered to decryptable specs.** The entry is `{name, hash, runningSince}`, all outside the enterprise envelope, so no decryption is needed — and filtering would strip an undecryptable app of its claim and cause the exact respawn this fixes.
- **`setOnContainerStarted` is removed entirely.** A container starting no longer changes what is announced, so nothing subscribed to it; every one of its four call sites already published the same fact on the event bus. Install completion still notifies: that is the claim event.
- **The heal-retry policy comment is corrected.** Never capping the container-repair retry rested on a broken container costing the node its claim and the app being placed elsewhere. It no longer does, which makes the heal this node's only route back to a working instance — a stronger reason not to cap it, and what the comment now says.

## The load balancer

Audited against this change on FDM `main`. For ordinary apps FDM builds its backends from the locations this announcement writes and health-checks every server it adds, so an instance that is not serving leaves rotation within seconds. Apps carrying app-specific checks are probed by FDM before they enter a backend at all. `g:` apps are elected by asking the node directly rather than from this announcement, and are unaffected — they have been exempt from the run-state filter since 2024. DNS records come from the app's registration, not from locations, so the record layer does not move. `runningSince` is preserved rather than restamped, so backend ordering and the reload-churn behaviour that depends on it are unchanged.

Liveness therefore stays where it belongs — with the load balancer's health checks and the app owner — while placement is settled by the announcement. That split is the point of this change, not a side effect of it.

## Testing

- Unit: **7097 passing, 0 failing**; lint clean.
- Suite **221** `unrunnable-app-keeps-its-instances`: an app asking for 2 instances lands on 4 against an image built without this change.
- Suite **222** `failed-placement-leaves-no-claim` (new): an app whose image pulls but whose container cannot start, so the install gets past writing the app's local row and then fails. It waits for `spawner:installFailed` — a node attempted the placement — then asserts a peer saw `network:appremoved` for the app, and that every node drops the location row for the sender that removal names. A node that tears down in silence broadcasts nothing, so the removal is the whole discriminator and no announce window has to be hit. The final check is keyed on the sender rather than on the app being absent everywhere, because an app whose container cannot start is offered to node after node and other claims come and go throughout.
- Suite **32** gains `keeps its claim on peers while the app is stopped`.
- Suite **55** `residential-node-evacuation` covers the departing case.
- Each fix has a mutation that breaks one test and a different one each time: the teardown forced to broadcast, forced not to, the departing mark reverted to set semantics, and the spawner's flag removed.
- A refused duplicate test install takes back nothing it did not raise: the teardown and the unmark are bound to the hold that set them. `leaves a running test install its mark and its app when it is refused` asserts the refusal first, so it cannot pass against an install that ran to the end and cleared its own mark; dropping `&& acquired` fails that test and only that test.
- The removal/announcement ordering is covered at both ends, each with its own mutation. Dropping the departing filter from the announcement breaks three tests: an app whose removal this node has broadcast is not announced, it is announced again once that removal ends, and a node whose every app is departing announces nothing. Dropping the mark from the removal breaks two: the mark stands for the length of a broadcast removal, and a removal that finishes does not unmark an overlapping one still running.
- Full gate at `163aa9439`: **107 of 108 suites pass**. The one red is suite `89`, which needs one of two instances to promote out of receive-only. Promotion defers to a running peer unless the node holds nothing, and the emptiness test is an item count that a volume's own scaffolding — `lost+found`, `appdata`, `.stfolder` — turns false on every node at once, so both instances defer and neither seeds. #1808 keys the deferral on whether a source exists and reads holdings from syncthing's `db/localchanged`.
- Suites **83**, **84**, **221** and **222** all pass.
- The dockerService unit tests assert against docker responses recorded by `tests/unit/fixtures/docker/record.js`, so they need no running container.


Cabecinha84 (commented):
Findings

  1. Medium, blocking — an app is now announced while it is still installing, and a failed spawner install leaves a
  phantom instance for up to ~2h

  - registerAppLocally writes the app's local database row before the image pull and container creation
    (appInstaller.js:551, "Database entry created … BEFORE Docker container creation").
  - The announcement now reads that installed set and only skips departingApps. So a periodic announcement (every 3600s)
    that fires during an install tells the network the node holds the app.
  - What happens if that install then fails:
    - The spawner installs with sendRemovalMessage=false (appSpawner.js:1067).
    - The cleanup in appInstaller.js:673 therefore tears the app down without sending fluxappremoved.
    - storeAppInstallingErrorMessage clears the "installing" records but does not touch appsLocations.
  - Result: a location row for an instance that doesn't exist, lasting up to RUNNING_EXPIRY_MS (7500s). During that time
    the spawner counts it and the app runs one instance short.
  - Before this PR: only g:/r: apps were exposed to this. For every other app it is new.
  - Suggested fix: mirror what the PR already does for removals, e.g. an installingApps set, marked right before the
    database insert, cleared in registerAppLocally's finally, and skipped by the announcement. After a successful
    install, onInstallComplete announces the app anyway.
  - It fixes itself after about 2h, but it is the mirror image of the removal window this PR handles, and the fix is
    about 10 lines.
  - This comes from reading the code path end to end; I did not reproduce it in the harness.

  2. Low — departingApps is a plain Set, not a counter

  - Forced removals skip the removalInProgress guard, so two forced removals of the same app can overlap with
    sendMessage=true. An example is the spawner's surplus removal (fire-and-forget) running at the same time as an
    expiry removal.
  - The call that finishes first unmarks the app while the other is still removing it, which briefly reopens the window
    commit 2 closes.
  - Rare. A Map<name, count> would fix it.

  3. Nits

  - appInstaller.js:696-704: the comment explaining why the announcement happens after the install lock is released
    refers to run-state and monitorAndRecoverApps force-including apps. Neither applies any more.
  - appReconciler.setOnContainerStarted / notifyContainerStarted no longer have any subscriber. The new comment mentions
    "an observer that needs the edge itself", but nothing uses it.
  - The listRunningApps stubs in peerNotification.test.js are now unused.

  4. Product decision to confirm explicitly (not a code defect)

  - An app that crashes forever on one node is now never moved elsewhere. There is no "give up after N crashes and hand
    the app back" path: the reconciler only removes an app when it can't recreate a container, not when the container
    keeps crashing.
  - Previously, a node-specific failure (GPU, disk, upstream port) got the instance re-placed on a healthy node. Now the
    app runs with fewer healthy instances until the owner acts, and the load balancer only takes the dead one out of
    rotation.
  - Given the 618-node incident this is the right default, but someone should sign off on it. A follow-up could add a
    bounded give-up that releases the app with sendMessage=true.

  Other points

  - Suite 83 (one red suite in the gate of 102): the PR says it is caused by existing code and fixed by #1804. But #1804
    is itself at REQUEST CHANGES (gossip-storm blocker), so the merge order the PR proposes ("#1804 first") is
    currently blocked. The last full gate also ran on the old base. After a rebase, the harness gate needs to be re-run
    against current development, and suite 83 needs an explicit decision.
  - Tests: the new suites 221 and 32 test the right thing (an app that stays down doesn't lose its claim). The fix in
    commit 3, where tests shared the globalState singleton and passed only because of load order, is correct and
    worthwhile.

MorningLightMountain713:
Thanks — all four land. Rebased onto current `development` and pushed, and the gate has run at the tip: **104 of 105 suites pass**, and the one red is suite 85, whose fix lives on #1804 rather than here.

**1. The phantom after a failed install — real, and the mechanism is worse than described.**

Traced it end to end and the chain holds. One thing to add: the announcement doesn't just create a claim, it **destroys the right one**. `storeAppRunningMessage` deletes the sender's `globalAppsInstallingLocations` row for every app it names, so naming a still-installing app promotes a provisional, retractable, 900s claim into a final 7500s one — which is exactly why the existing retraction machinery can't help. By the time the install fails, the row `fluxappinstallingerror` clears is already gone.

Fixed a path you didn't name, too: the manual install API (`appInstaller.js:1096`) has the same defect as the spawner.

**Fixed by retracting rather than suppressing, and that is deliberate.** The `installingApps` set you suggest is the right shape, and it is what `feat/fluxos-v9` can do — but not here. The `fluxappinstalling` claim is what reserves a placement while an install runs, and on `development` it is broadcast once and never renewed, so it lapses at 900s. Multi-component pulls routinely outlive that, and today the announcement is the only thing holding the placement afterwards. Suppress it and a slow install loses its seat mid-pull, the app is placed elsewhere, and the first install completes into a surplus — a worse failure than the one being fixed. v9 renews the claim every 12 minutes (`startInstallingRenewal`) and can therefore do it properly; backporting that is a protocol change, since `development`'s store rejects the v2 shape and election ordering needs a stable `announcedAt`.

So the failed placement retracts its claim instead, and the PR body states plainly what that leaves: the node still claims an app mid-install, an install past 900s still loses its seat, and a mid-install crash still leaves a claim nothing retracts. The full model and the two builds that close those are in `fluxModels/workstreams/placement-and-election/APP_PRESENCE_CLAIMS.md`.

**2. `departingApps` a set rather than a counter — fixed.** Counted, with a test that goes red if the counting is reverted. Worth noting `removalInProgress` in the same `finally` has the identical shape and is pre-existing.

**3. Nits — all three done, and two were more than nits.**

- The `appInstaller` comment explained the ordering by `monitorAndRecoverApps`, which no longer exists at all. Rewritten to the reason that still holds.
- `setOnContainerStarted` is deleted rather than documented. It had no production subscriber, all four of its call sites already published the same fact on the event bus, and its four unit tests registered a listener purely so they'd have something to assert. Removing it also invalidated a comment you didn't flag: `appReconciler.js` justified never capping the container-repair retry on the grounds that a broken container costs the node its claim and the app is re-placed elsewhere. This branch made that untrue, and it was the stated reason a cap was rejected. Corrected.
- The `listRunningApps` stubs are gone.

**4. The product decision — confirmed explicitly.** An app that crash-loops forever on one node is never relocated; the reconciler gives up only when it cannot create a container, not when one keeps dying. That is intended, it is signed off, and it is now stated in the PR body rather than inherited from the announcement change.

**Suite 83.** The `#1804` attribution is withdrawn — it was a symptom match nobody ran. There is a better-fitting candidate already merged: `getEligibleSyncPeers` skips any peer with `missedPongs !== 0`, and `ephemeralSync:requested` is only published when that pool is non-empty. `097c60ba3` records that `missedPongs` counted pings sent rather than pongs missed, and that `pingAll()` pings every peer in one loop — so every peer failed the bar at once and the pool was empty. Every observed failure of suite 83 predates that commit, and it is in this branch now. **Suite 83 passes on this gate**, which is what that predicts. The merge-order note is removed from the body — nothing here depends on that PR.

Cabecinha84 (commented):
Non-blocking items worth a comment

  1. Stale rationale on the boot-drain gate. peerNotification.js:145-148 still justifies waitForBootDrainSettled() with
     "a too-early snapshot misses apps whose containers are still being started." That can no longer happen — the
     snapshot is the installed set. Suite 46's comment was corrected to say precisely this; the source comment (and the
     PR body's "Deliberate choices" bullet) were not. The gate is harmless and capped, but it's now kept for a reason
     that no longer exists.
  2. "Rows are not deleted by the omission" is wrong for the sync path. True for direct gossip, but
     fluxCommunication.handleAppRunningSyncResponse:275 feeds the newest v2 broadcast per IP into
     messageStore.pruneAppRunningLocations, which does deleteMany { ip, name: { $nin: names }, broadcastedAt: { $lte: …
     } }. An omitted app's row is deleted on any peer that syncs. This is not a regression — before this PR the whole
     redeploy window omitted the app, so the exposure shrinks — and at a 3600s announce interval against a ~30s+pull gap
     the odds are a fraction of a percent. But the follow-up work in APP_PRESENCE_CLAIMS.md should be sized against the
     real mechanism, not "one missed refresh."
  3. The retraction is best-effort, and the limitations list doesn't say so. removeAppLocally swallows its own errors,
     and two paths skip the broadcast silently: throw new Error('Flux App not found') when the spec is in none of the
     four sources, and if (ip) when getLocalSocketAddress() returns falsy. Both are edge cases, but they belong next to
     "if FluxOS dies mid-install nothing is broadcast."
  4. Added gossip volume. Every failed placement now sends a network-wide fluxappremoved, including failures before the
     local row is written (tier lookup, requirements, port conflict), where the removal is a no-op at every peer.
     Bounded by spawnErrorsLongerAppCache, and far better than a 7500s phantom — just worth being aware of.
  5. Pre-existing, adjacent, cheap: removeAppLocally's finally sets globalState.removalInProgress = false
     unconditionally, so a refused non-forced removal releases the lock held by the removal actually running. This is
     the exact shape the PR just fixed for departingApps, in the same finally it just edited. Not this PR's job, but
     it's a one-line follow-up and the author is already standing there.

  ---

  The one red check — not the PR

  build (ubuntu-22.04, 20.x, 7.0)   fail   47s
    step: install syncthing
    Err:10 https://apt.syncthing.net syncthing Release  404 Not Found

  I reproduced this locally. Building the harness images from this tree fails at the identical fetch:

  E: Failed to fetch https://apt.syncthingcdn.net/pool/syncthing_2.1.5_amd64.deb  404 Not Found

  It's an upstream apt-repo outage. development was green at f5b11c08e two days ago with the same step. It needs a
  re-run once the repo is back, not a code change. It also meant I could not build the fleet images, so I could not
  independently re-run the harness gate.

MorningLightMountain713:
Thanks — all five land, and two of them turned out to be larger than they looked. Full gate at the branch tip: **104 of 105 suites pass**, and the one red is suite `85`, whose fix lives on #1804 rather than here. Unit 6496 passing, 0 failing, lint clean.

**1. Stale rationale on the boot-drain gate.** Correct, and the source comment was the load-bearing copy. The snapshot is `installedApps()`, a straight read of the local table that never touches Docker, so a container's start state cannot change what it returns. The gate stays, for the reason that does hold: a boot reconcile that cannot recreate a container uninstalls the app, and a claim made before that pass would have to be taken back by a broadcast that is best-effort. Comment and the Deliberate choices bullet both now say that.

**2. Rows are deleted on the sync path.** Right, and it changed the fix rather than just the wording. `handleAppRunningSyncResponse` feeds the newest broadcast per IP into `pruneAppRunningLocations`, so an omitted app's row is deleted on a peer that is syncing — the limitation now says so, naming which node does what.

Reading it that way surfaced the sharper case underneath: an announcement cycle decides its list, then spends time on a per-app read and a signature before it sends. A removal starting in that window is already marked, but the list predates the mark, so the app is named anyway — and on the peer the removal deletes the row and the claim re-creates it, unconditionally, because a deleted row has no timestamp left to compare against. That is the two-hour phantom this PR exists to remove, arrived at from the other direction.

It is settled at the sender: a removal that tells the network waits for the cycle in flight to finish sending, so the claim is applied first and the removal clears it, and a cycle starting after the mark leaves the app out. The wait is bounded — a wedged cycle must not hold up removals, which every install and redeploy queues behind. A redeploy does not wait, because it says nothing to the network and has nothing to contradict. `broadcastedAt` is also now taken where the snapshot is rather than where the message is built, since every consumer reads it as the time those apps were installed on that node.

**3. The retraction is best-effort.** Added to the limitations, and the silent case is no longer silent: a teardown that cannot name its own address logs that the app was removed without announcing it and that peers hold the row until it expires. The signing case was already logged by the signer.

**4. Added gossip volume.** Tightened rather than accepted. The claim is the app's row in the local table, so the retraction is now bound to it: taken when the row is written, and when a stale row is found and cleared, since that one has been announced. An install that gives up before writing the row broadcasts nothing, because there is nothing for a peer to clear.

**5. `removalInProgress` released unconditionally.** Fixed here rather than deferred — it is the same defect three lines below it in the same `finally`, and leaving one half bound to its acquisition and the other half not reads as deliberate. A call that finds the lock held now releases nothing it did not take.

**Two things beyond the review.**

A test install claimed the app and never gave it back. It writes the row like any other install and the announcement reads that table, so a cycle landing inside one claimed an app the node was only trying out — and its teardown is silent, so the claim stood for the row's full lifetime. Nothing is retracted now because nothing is claimed: the app is marked for the length of the test and released once the teardown has deleted the row.

The rebuild path had a fixed pause between tearing an app down and putting it back. The teardown already confirms its own completion — container removal is awaited, the network removal carries its own wait, force and retry and reports whether it succeeded, and an install meets a surviving network by checking for one before creating it. The requirements check that follows sums the disk each row in the local table declares, so deleting the row frees that space in the arithmetic rather than when the filesystem catches up. Nothing in that sequence needs a pause, and it sat in the one window where the local table does not name an app the node intends to keep.

**On the CI red** — agreed, and it was the upstream apt outage. The repository is back and the job is green on a re-run, with no code change.

**One framing worth stating plainly.** Several of these are the best available on `development` rather than the best available, and the PR body now says which and why. A node cannot stop announcing an app it is installing, because the installing claim is broadcast once and never renewed on this lineage. Peers cannot settle two contradictory messages by what they say, because the location table is the live mechanism here and a removal deletes from it. The operation locks are booleans any caller may set. Each of those is what the v9 work exists to change — the renewal lands, the event log becomes the source of truth peers fold rather than a record nothing reads, and a lock becomes something you cannot release without having taken it.

Cabecinha84 (commented):
Blocking finding

  testInstallingApps is unmarked by a refused duplicate test install

  ZelBack/src/services/appLifecycle/appInstaller.js:709

  } finally {
      if (acquired) globalState.installationInProgress = false;
      if (test) {
        try { await appUninstaller.removeAppLocally(appSpecs.name, null, true, false, false); } catch (e) { ... }
        globalState.testInstallingApps.delete(appSpecs.name);   // ← unguarded
      }
  }

  The add at appInstaller.js:410 happens after the removalInProgress / installationInProgress guards. The delete is in a
  finally that every path reaches, including a refusal.

  Failure scenario:

  1. POST /apps/testappinstall/myapp (call A) → takes the hold, testInstallingApps = {myapp}, writes the app's local
     row, starts a multi-minute image pull.
  2. POST /apps/testappinstall/myapp (call B) → refused at the installationInProgress guard, acquired === false, mark
     never added.
  3. B's finally runs if (test) and executes testInstallingApps.delete('myapp') — clearing A's mark.
  4. An announce cycle landing in A's remaining install window reads A's row and claims myapp as a placement. The
     location row lives its full RUNNING_EXPIRY_MS (7500 s) and the spawner counts it — the app runs one instance short
     for ~2 hours.

  That is exactly the defect af3a1cf68 exists to prevent, and exactly the shape this PR already fixed twice in the same
  review round: departingName (so a refused duplicate cannot unmark a live removal) and acquired (so a refused duplicate
  cannot release removalInProgress). testInstallingApps is the third instance of the same pattern and it was left
  unbound.

  testAppInstall has no guard of its own — the only serialization is inside registerAppLocally — so two concurrent calls
  are reachable from the API by an owner or the Flux team.

  Fix is a let testMarked = false; set where the add is, checked at the delete. (The adjacent forced removeAppLocally on
  the refused path is pre-existing and out of scope, but it lives in the same three lines.)

  ---

  Second condition: the gate has not been evidenced at this head

  The author's comment says "Full gate at the branch tip: 104 of 105 suites pass", posted 2026-09-19T10:56:15Z. Head
  b73e5bfee was committed 10:34:34Z — 22 minutes earlier. A 105-suite harness gate does not run in 22 minutes.

  That matters more than usual here, because b73e5bfee is precisely the commit that changes harness node boot to exit 1
  when memory is not in cgroup.subtree_control. The gate that produced "104/105" almost certainly predates it. Ask for
  the gate log at b73e5bfee specifically.

  ---

  What I verified and found sound

  - Removing redeploy.delay is safe for the requirements arithmetic. checkAppHWRequirements →
    resourceQueryService.appsResources() sums the rows in the local appsInformation collection, and nodeSpecs.ssdStorage
    is a cached fluxbench value (hwRequirements.js:36), not a live df. Deleting the row frees the space in the sum
    immediately; the pause bought nothing. softRemoveAppLocally does delete the row (appUninstaller.js:1162), so soft
    redeploy is covered too.
  - The PR body's claim about omitted rows is correct. messageStore.storeAppRunningMessage upserts per app with a
    newer-wins $cond and prunes nothing on the direct path; only the sync path (handleAppRunningSyncResponse →
    pruneAppRunningLocations) deletes $nin names. The body now says so.
  - softRemoveAppLocally has no unbound-release twin — it throws before its try, so its unconditional finally is
    unreachable on a refusal.
  - Component removal with sendMessage is still unreachable — removeAppLocallyApi rejects names containing _.
  - The empty-snapshot guard is intact (peerNotification.js:221), so the departing/test filters cannot make a node erase
    its own presence.
  - Dropping decryptEnterpriseApps from the announce path remains safe — the broadcast entry is {name, hash,
    runningSince}, all outside the enterprise envelope; dbSpecs empties compose/contacts but keeps name and hash.
  - claimed is set in both correct places — on a stale-row cleanup (that row has been announced, so retracting it is
    right) and after a successful insert. Both new sendRemovalMessage=true callers pass componentSpecs as
    null/undefined, so the isComponent branch that never sets claimed is not on their path.
  - readyTimeout cannot deadlock. The worst interaction is boot: the announce cycle holds the lock across
    waitForBootDrainSettled(), and a reconciler-driven uninstall (appReconciler.js:469, sendMessage=true) then pays the
    full 30 s. Bounded on both sides (boot drain caps at 2 min), and in that scenario the snapshot is taken after the
    gate opens, so it reads the mark anyway — correctness is preserved, it only costs latency.
  - New unit tests are good. They include real canaries (the wait test asserts the removal did eventually reach
    getLocalSocketAddress, so it cannot pass for a removal that never got that far), and the wedged-cycle test overrides
    ANNOUNCE_CYCLE_WAIT_MS through proxyquire rather than sleeping 30 s.

  ---

  Non-blocking

  1. Dead harness code from dropping suite 223. The /rpc-hold routes in test-infra/daemon-stub/index.js:868-890, the
     holdRpc/releaseRpc/releaseAllRpc helpers in daemon-control.js:346-363, and the 'app:announcing' subscription in
     node-client.js:207 now have no consumer — suite 223 was their only user. Either remove them or note why they are
     kept.
  2. The ordering fix has unit coverage only. The author's reason for dropping suite 223 is honest and correct (holding
     the address lookup arms it for eight unrelated callers, so what the suite observed was not the overlap it named).
     Worth stating in the body as a known coverage gap rather than leaving it implied by a deleted file.
  3. tests/unit/globalconfig/default.js:258 still carries delay: 30. Harmless dead fixture config, but it now drifts
     from ZelBack/config/default.js.
  4. The cgroup script writes all controllers in one shot (entrypoint.sh). subtree_control writes are all-or-nothing, so
     a single unrelated controller that cannot be enabled now takes memory down with it and hard-fails the node, where
     the old || : degraded. The 3× retry loop does not help a deterministic failure. Low likelihood (the container root
     only lists what was delegated to it), but the blast radius is "every harness node refuses to boot".
  5. PR body Testing section is stale — it still cites 6496 unit tests and does not mention that suite 223 was added and
     removed.

MorningLightMountain713:
Three items changed code, two are declines with the reasoning below, and the gate at this head is included.

**Blocking — `testInstallingApps` unmarked by a refused duplicate test install.** Fixed in `04214c4fb`.

The teardown and the unmark are bound to the hold that set them: `if (test && acquired)` at `appInstaller.js:704`. `acquired` is set with nothing awaited between it and the mark going in, so holding it is the same fact as having marked — a separate flag would be a synonym for it on this path. A call that did not take the hold now leaves both the mark and the app alone. The forced `removeAppLocally` sits inside the same guard, so a refused duplicate no longer tears down an app the node holds and tells the network nothing.

`tests/unit/appInstaller.test.js` — *"leaves a running test install its mark and its app when it is refused"*. It asserts `InstallOutcome.REFUSED` first, so it cannot pass against an install that ran to the end and cleared its own mark. Dropping `&& acquired` fails that test and only that test.

**The gate at this head.** Full gate at `731e5e591`, which contains `b73e5bfee`: **104 of 105 suites pass** — `RESULT suites_pass=104 suites_fail=1 FAILED:[ 84 ]`, 105 suites, `MAXN=6`. Suite **89**, which the cgroup commit fixes, passes. So do **83**, **85**, **221** and **222**.

The one red is suite `84`. Its invariant requires what a reader loses to be a whole number of cut lines; docker delivers the log in 16 KB frames, and here the notice reported 486,532 characters against a cut line of 251,458 — a shortfall from two lines of exactly 16,384 bytes, one frame. The frame-based assertion (`shortfall % DOCKER_FRAME_BYTES`) is on #1808 and not on this branch.

**Dead harness code from dropping suite 223.** Kept, with the reason recorded at each piece.

`test-infra/runner/framework/daemon-control.js`, above `holdRpc`: a hold stops every caller of the method, and `getbenchmarks` is shared across most of the app lifecycle, so it is not a scalpel. `ZelBack/src/services/appMessaging/peerNotification.js:140` now states what the `app:announcing` event is: nothing waits on it, `cycleLock` is what orders a removal against an announcement. The primitive is general and the comments say what it does and does not serialize.

**The ordering fix's coverage.** It is covered at both ends, each end with its own mutation. Dropping the departing filter from the announcement at `peerNotification.js:193` fails three tests: an app whose removal this node has broadcast is not announced, it is announced again once that removal ends, and a node whose every app is departing announces nothing. Dropping the mark from the removal at `appUninstaller.js:853` fails two: the mark stands for the length of a broadcast removal, and a removal that finishes does not unmark an overlapping one still running.

**`delay: 30` in `tests/unit/globalconfig/default.js`.** Deleted in `731e5e591`. It had no reader in `ZelBack/src`, `tests/` or `test-infra/` — only `composedDelay` and `probability` are read — and the fixture's `redeploy` block now diffs identically against `ZelBack/config/default.js`.

**The cgroup write, one shot rather than per controller.** Keeping it as one write.

A node that cannot delegate `memory` refusing to boot is the behaviour that commit is for. Writing the controllers one per line swaps a loud total failure for a quiet partial one: `memory` is enabled, the node boots, and some other controller is silently missing. Only `memory` is checked, so nothing catches it. If the missing one is `cpu` or `pids`, docker cannot apply limits and the node looks healthy while failing every app it is handed — which is the class the commit exists to kill. The write source is `/sys/fs/cgroup/cgroup.controllers`, which lists only what was already delegated to the container, so a controller present there that cannot be enabled is close to hypothetical.

One correction on that item: the `|| :` is still in place at `entrypoint.sh:129`. The `exit 1` sits behind it, after the retries.

**PR body.** Updated: the unit count, the gate line at this head, and the ordering coverage above.

Cabecinha84 (commented):
What I would post

  Request the rebase, then approve. Concretely, three conditions, none of which need design discussion:

  1. Rebase onto 1370ac54d; resolve the two conflicts (jsdoc — keep dev's @returns {Promise<string>} One of
     InstallOutcome.; test file — keep both tests).
  2. tests/unit/appInstaller.test.js:1315 → InstallOutcome.BUSY.
  3. Re-run the full gate at the rebased head. Not ceremony this time: the merged tree runs the spawner through #1804's
     policy/blocklist gating, which suites 221 and 222 have never met. policyReachable is derived automatically by
     createTestEnv, so I expect them to pass — but "expect" is not "ran".

  If the team is comfortable treating §2 as a merge-time fix, a plain ACK with those conditions written down is
  defensible. I would not merge the branch as it stands, because it does not merge cleanly and the first thing CI would
  report is a red unit suite.

MorningLightMountain713:
All three conditions are done, and the two older items are answered below.

**1. Rebase.** The branch is on `8520cacf3` rather than `1370ac54d` — development took #1806 about two hours after the review was written. Tip is `163aa9439`, 22 commits, mergeable.

Both conflicts were where you said. In `appInstaller.js` the jsdoc keeps development's `@returns {Promise<string>} One of InstallOutcome.`, and keeps this branch's `@param sendRemovalMessage` text with it — that paragraph says when the flag is true and when it is false, which the one-line version it replaced does not. In `tests/unit/appSpawner.test.js` both tests are kept, written out as two separate blocks: they interleave across three hunks and share their `buildModule` call and their `await`, so taking both sides as-is fuses them into one test that asserts neither thing.

**2. `tests/unit/appInstaller.test.js:1315`.** Now `InstallOutcome.BUSY`. The test sets `installationInProgress`, which is the guard at `appInstaller.js:417`. It is folded into the commit that owns the assertion rather than added on top, so no commit on the branch asserts against a key that does not exist at that point in its own history. `InstallOutcome.REFUSED` now appears nowhere in the tree.

**3. Full gate at the rebased head.** `163aa9439` on chud, 108 suites, **107 pass**. Suites **221** and **222** both pass, so the spawner running through #1804's policy and blocklist gating is now run rather than expected — `policyReachable` derives true for both fleets and they take the default bundle. Suite **84** passes as well.

The one red is suite **89** `restore-propagates-real-syncthing`, and it is the promotion standoff #1808 fixes. Promotion out of receive-only defers to a running peer unless the node holds nothing, and the emptiness test is an item count. A volume's own scaffolding — `lost+found` from mkfs, `appdata` from the primary mount, `.stfolder` — turns that count non-zero on every node at once, so both instances defer to each other and neither seeds. In this run one node reached `streak=1/2, elected=true, folderEmpty=true` and lost the election on the next tick when its own scaffolding was indexed. #1808 keys the deferral on whether a source exists and nothing else, and reads holdings from syncthing's `db/localchanged` with directories excluded by type.

**The crash-loop relocation question from the first round is signed off explicitly.** It is stated in the body under Deliberate choices: an app that crash-loops on one node is recovered in place and never relocated, the instance costs the owner until they act, and the load balancer takes the dead one out of rotation.

**On the retraction limitations list**, the `'Flux App not found'` path is deliberately not added. The announcement is built from the app's local row, and that throw fires only when there is no local row — so a node reaching it has never claimed the app, and its peers hold nothing of its to retract. The missing broadcast has nothing to take back. The `getLocalSocketAddress` case is different and is listed, because there the claim exists and the retraction is what fails.

Unit is **7097 passing, 0 failing**, lint clean. One commit on the branch is unrelated to the defect and worth naming so it is not a surprise in the diff: the dockerService unit tests now assert against docker responses recorded by `tests/unit/fixtures/docker/record.js` instead of driving a live container, which also takes the container start/stop out of the unit script.

Cabecinha84 (approved):
ack

PR #1808, opened 2026-09-17
A holder can seed, what it holds is read not counted, and a component can keep a directory off the network

MorningLightMountain713:
## The problems

**A node holding the only copy of an app's data could never publish it.** The promotion guard made a holder defer to a running peer, so that it would sync from a source rather than publish data nothing had verified. But a receive-only folder's local files are published with a zeroed version vector — syncthing's `prepareFileInfoForIndex`: *"we do not want it to ever become the globally best version"* — so no peer can take them, and no peer can become the source being waited for. The wait had no end. Seen in production: an app with 5.8 GB of a customer's world on one node, an empty peer on the other, both instances down, neither able to promote.

**What a node "holds" was counted, and a count cannot tell data from scaffolding.** `db/status` counts local-changed entries, and FluxOS leaves entries on every volume it creates — `mkfs` leaves `lost+found`, the primary mount leaves `appdata`, an `f:` mount leaves a zero-length file because docker would otherwise create a directory in its place. Counting those says "I am holding something" on a volume holding nothing, on every node at once.

**The check that stops a stale index gutting the cluster could not fire.** Before a folder may send, the index's claim is tested against the disk: an index claiming bytes over a wiped volume must not go sendreceive, because sendreceive broadcasts every missing file as a deletion. The disk side walks the volume root and counts a directory as content, and every app volume is formatted ext4 — so `lost+found` is at that root before the app is, the volume reads as occupied whatever else is on it, and the test returns the same answer on every node in every state.

**A component had no way to keep a directory off the network.** Everything under the app's volume is synced, so a component whose working files must stay local had nowhere to put them but the container's writable layer — which is a flat 10 GiB whatever the app bought.

**An operation's staging was three rules that had to agree.** Staging directories sat at the volume root as `.flux-op-<id>`, and their name had to be recognised independently by the ignore policy, the reserved-name check and the reclaimer.

## What this changes

**`ml:<subdir>:<path>`** — a mount form declaring a directory the component keeps off the network. The `.stignore` policy lines, including one per declared `ml:` directory, are written before the folder is ever handed to syncthing, because the first scan indexes whatever it finds.

**The specification decides the ignores, and the volume does not.** What may leave a node is derived from the spec afresh each monitor pass and set exactly, rather than merged into what is already in `.stignore`. Every node must compute the same set from the same specification, so the answer comes from the spec and never from what happens to be on one node's disk. Deriving the whole set also means a spec that drops an `ml:` mount has its exclusion removed with it.

**An `ml:` name stands for itself, as one pattern and as one line.** An ignore line is a pattern, not a name, so a name carrying pattern syntax describes some other set of entries: `/[a]ppdata` excludes the component's synced storage and leaves the directory literally named `[a]ppdata` replicating — the reverse of what the spec asked for, on every node at once and with nothing logged anywhere. The file is also line-oriented and the derived lines are joined into one document, so a name carrying a line terminator writes a second line the specification never named, and the converge then reads back more lines than it derived and rewrites the file and rescans the folder every pass for as long as the app exists. The name is held to the class `pathSecurity` refuses in a path component, for the reason that file states.

**`POST /syncthing/db/ignores` and `POST /syncthing/config/defaults/ignores` are removed.** They set ignore patterns from outside the specification, and what they wrote to an app folder is replaced by the next monitor pass. FluxOS's own converge never passed through either handler — it calls `getFolderIgnores` and `setFolderIgnores` directly — so it is unaffected.

**Deferral is decided by whether a source exists, and by nothing else.** Where several holders remain, the claim decides rather than the address: the largest copy first, most recently written only to separate equal byte counts, address only between equals. A size and a timestamp cannot express containment, so the order is chosen for the direction it fails in — a near-empty volume can never publish over a full one, which is the loss that cannot be recovered from. There is no ranking at all unless every candidate answered, because ranking a field with a silent holder in it puts that holder last, and a node that cannot read its own holdings stands down rather than fall through to the address order: unknown is not empty, and it is not a claim.

**A claim that cannot be compared is not an answer.** Both fields cross the network in plaintext from a peer that asserts them, and subtracting one that is not a number gives NaN, which `sort` reads as equal — the comparison returns before the address tiebreak and the winner becomes whichever candidate the peer list carried first, which is a database read order and differs per node. A claim is required to carry finite numbers and counts as no answer otherwise, which is the case the address order exists for. Comparability is the whole requirement: a figure no local measurement could produce still orders against the rest, and refusing it would collapse the field to the address order over one peer's reply. The ranking is a better answer among honest candidates and it is not a defence — a candidate that overstates what it holds wins, and no ceiling a node can derive changes that, because whatever it can derive the candidate can claim.

**Which of the two rules applies is decided from one node's own answers, and this change does not close that.** A candidate silent to one node and answering another puts the two on different rules over one field, and two rules can elect two winners of the same folder. The address order does not have that property, which is why it remains the fallback rather than a legacy path: every node computes it from the same input. Narrowing it costs more than it saves — a node that stands down whenever a candidate is silent to it will not seed while a peer is merely dead, and a dead holder must never strand an app with no writable copy anywhere. What bounds it today is the connectivity floor applied before a win counts, and what is left is two nodes that each saw a whole field and were handed different numbers by the same peer. An agreed view of the field is the only thing that closes it.

**What a node holds is answered to a node, or to the flux team.** `holding` is a size and a last-write time for every app whose data a node carries, which is the tenant's, and the endpoint it rode on took no authentication. So the open GET answers `ready` and `folders` and nothing else — folder ids are app names, and the location table already publishes those. A POST beside it answers the holdings to a caller that signs as a node on the deterministic list, the way every node-to-node call on this API proves itself, or that holds flux team privilege. The signature names the node being asked and the moment it was made, so one captured in flight cannot be turned on the rest of the fleet or kept, and the moment is read against the bound every signed Flux broadcast is held to. A caller that shows neither is answered without the holdings rather than refused, and a peer whose release has no POST route is asked again over the GET — both read as claiming nothing, which is what the address-order election already expects. Nothing is probed at all unless a folder is awaiting promotion, so the signature is a cold start's cost.

**The caller proves itself before the node looks anything up, and the moment is bounded both ways.** The route is open, so every check ahead of the signature runs for whoever sends bytes at it: shape and timestamp answer from the body alone, and the node's own address — the one thing here that reaches outside it — is read only once a signature has verified. `target` is checked last and loses nothing by it, since it refuses a request captured here and replayed elsewhere wherever it sits. Asking only whether a request is too old leaves the one a sender can mint deliberately: stamped far enough ahead, a body signed once against a chosen target authorises whoever holds it for as long as they hold it, so what is served only to a node becomes public the moment that body is. The forward bound is the tolerance the network already keeps for clocks that do not agree, named now rather than restated, so both places holding a signed message to it hold it to one value.

**A node that cannot say who is asking refuses, rather than waiting to find out.** Establishing the caller reads the deterministic node list, and those accessors wait for it to arrive — which is right for callers that cannot tell an unknown list from an empty one, and wrong for this one, because every peer asks this on every monitor pass. Parked there, a request holds a handler open per peer per pass and the peer learns nothing until its own probe expires, which it reads as this node being gone. Gone and starting are opposite instructions: one promotes, the other waits, and the node still starting may hold the only writable copy. It answers 503 instead, and the prober records that as alive-and-not-ready rather than covering it with the open endpoint — which reports syncthing's readiness, set by a different pass, and would read as this node holding nothing.

**Holdings are read from syncthing's own `db/localchanged`** — the name, size and modification time of every entry the cluster's index does not have — rather than from a count. Directories are excluded by type, not by size, and zero-byte entries contribute no timestamp, so an `f:` mount's touched file cannot make a node holding nothing look more recently written than one holding the world. Peers already probe `/apps/promotedfolders` once per monitor pass, so the claim rides on a round trip already paid for; a node too old to publish it reads as claiming nothing, which is what its behaviour has always amounted to.

**A claim is read to the end of its pages, folded as it arrives, and bounded.** `db/localchanged` is paged, and a total taken from one page is a prefix of any larger folder — two nodes each totalling their own prefix produce figures that no longer order by how much each holds. Pages are handed to the caller and dropped, so the cost is flat whatever the folder holds, and the walk stops at sixteen pages, because how many entries exist is the app's to choose and these requests are paid again every pass. A folder that reaches the bound is reported truncated and its total read as a floor, which is all the ranking needs from it.

**A claim dies with the data it describes.** It is dropped at promotion, which answers the receive-only question the claim asks, at hard removal, and at the volume recreate's point of no return. A claim that outlives its volume offers data the node no longer has, and peers rank a seed on it.

**A claim is withdrawn by the condition the node stands down on.** A read this node cannot make is not a smaller claim, it is no claim, and the election refuses to lead on exactly that. Published on success only, the last good figure stayed up: peers rank the node first and defer to it while it defers to its own unknown, and nobody seeds for as long as the read keeps failing — the standoff this change set exists to remove, reached by a syncthing restart. Removed instead, a peer reads it as not having answered, which is the case the address order exists for.

**The index says which claim it is making, and the disk is read on those terms.** What is the owner's data and what is the volume's scaffolding is one rule now, applied by the index read and the disk walk alike: not syncthing's control files, not `lost+found`, not the executor's staging directory, not `/backup`, and not a directory the specification declared `ml:`. It is anchored to the volume root, because that is where each of those names is an ignore line. On top of that, `globalFiles` separates the two claims an index can make. Files claimed are answered by files on disk — the primary mount and every `m:` directory survive a wipe, so counting directories answers every volume the same way and the check decides nothing. Bytes claimed with no files is a folder whose payload is empty directories, where the directory count is the only honest reading and a files-only walk is the 2026-07-04 false positive that stopped healthy apps. The harness test that recorded this gap as pending is now asserting it.

**The disk is read by a process that can see all of it.** Syncthing indexes the volume as root on every node; FluxOS is root on Arcane and the account the operator installed it under elsewhere. A volume holds what the app's container wrote, under the ownership and mode that container chose — postgres keeps its data directory at 0700 and refuses to start otherwise, on every start — so the index and a walk this process makes itself can be reading different filesystems, and the side that sees less is the one deciding whether to stop the app. A directory this process was not shown is now reported rather than counted as empty, and root is asked one bounded question about it: is there content here. A volume no process on the node can read is recorded as the fault it is and the folder is left as it stands, because stopping a running app over the node's own blindness is not what the guard is for.

**An empty file at the volume root is not evidence of bytes.** An `f:` mount binds a single file into the container, and docker creates a directory where a bind source is missing — so FluxOS touches the file itself when it builds the volume, at the root, zero-length, and a wipe leaves it there. Counted, it made a component declaring an `f:` mount read as occupied whatever its volume held. The walk answers whether the disk holds the bytes the index claims, and a file carrying none of them answers no. Anchored at the root like every other exclusion, because deeper down an empty file is the owner's and a payload of them is a payload.

**The synced scope is one set, and the ignore lines are it.** A component's volume root is its syncthing folder and every volume is formatted ext4, so `lost+found` sits inside the replicated tree; syncthing's own internal names are `.stfolder`, `.stignore` and `.stversions` and nothing else, so what one node's fsck recovers replicated to every other holder. It is now an ignore line, and the disk side reads the scope off those lines rather than restating it — a second list of the same names is a second rule that has to agree with the first.

**What the browser hides, it also refuses to be asked for.** The volume root carries entries that are not the owner's, and the listing filtered them out of what it offered; what an owner may ask for by name is a different question, and the read endpoints never asked it. They refuse it now, checked against what the path resolves to as well as what was asked for, because a name that is reserved nowhere can still arrive in the same place — the reason the write side resolves too.

**A node asks the benchmark daemon for its own address once a minute, not once a caller.** The value is invariant except on a reassignment, and it was re-derived on every one of thirty-one call sites. It is served from the cached value while that is fresh, on a monotonic clock so a step cannot serve it stale or expire it early, and a null is never cached.

**Staging is one `.flux-op/<id>` directory.** One reserved name at the volume root instead of a family of them, so the ignore policy, the reserved-name check and the reclaimer share a single rule.

**The permission widen names one path, and FluxOS owns it.** `chmod` resolves a symbolic link given as an argument and ignores one met inside a traversal, so a path it is handed that another writer can replace with a link has that link's target widened, as root, at a path of that writer's choosing. Every name inside the volume is such a path — the owner's file API copies and moves a link as a link and accepts any unreserved name in the volume root as a destination, and syncthing replicates a link into the folder it replicates, which is that same root — so the volume root is the only argument. The executor's staging directory is widened with the rest: root owns it and the executor runs as root, and reaching it needs a path to the volume root that no app container is given.

## Before release

`ml:` is parsed by this release and not by earlier ones, and app registration length-checks `containerData` rather than validating its syntax. The FluxOS version floor must be at the release carrying this before the form is used, so that every node placing the app can install it.

## Verification

Unit suite: **7214 passing, 0 failing, 17 pending.** `npx eslint ZelBack/src tests test-infra` clean. A cross-module resolution sweep of every `alias.method()` against its target's exports resolves identically to `development`.

**The full gate is queued and has not run at this tip.** The last full run was 108 of 109 suites at an earlier head, and its one failure was a harness fixture reaching the public registry for a digest the repository already carries — fixed since, and proven both ways with the registry made unreachable. The suites below have been run individually at this tip; the gate is what will say whether the rest still hold, and it will be run before this is merged.

Four harness suites carry the behaviour no unit test can reach, and three of them were run against a tree with the fix reverted to show which assertion catches it.

**FluxOS runs under the account its fleet gives it.** Every node in the harness ran as root, so no suite could reach what a node is refused — and that is half the field, since FluxOS is root on Arcane and the operator's account elsewhere while syncthing is spawned with sudo on both. A fleet now names the nodes that run it unprivileged; the container stays root, because making loop devices, delegating cgroups and supervising a dockerd is the host's work on a real node, and only the supervised process drops. Suite 1802 puts a running sendreceive app on such a node, locks its data directory to the container's own uid, and holds the folder to the rule that only a definite answer moves it. With the disk read reverted, its first assertion fails on the app being stopped: `mount safety block: phantom_index_empty_disk`.

**The mount-safety suite's phantom app declares an `f:` mount**, because a bare `r:/appdata` app puts no regular file at the volume root at all and cannot tell a walk that counts any file from one that counts the owner's data. It reads the file's size off the volume before asserting, so the case cannot quietly revert to the one it replaced. With the rule reverted, the demotion never comes.

**The cold-start election is asked what happens when the field is not the same from every node**, by refusing one candidate's answers to one other node. Both directions are read from inside the fleet, so the split is asserted rather than assumed.

**A node whose network state genuinely never starts is asked for its holdings**, and held to the bound that makes the answer worth having: sooner than the probe gives up. The ready node answering the same call at the same moment is what stops the refusal reading as a broken route.

Unit tests cover `localHoldings` and its paging, the mount parser, the folder state machine's ranking and claim lifecycle, the ignore policy and the scope derived from it, the staging rules, the permission widen's argument list, the disk walk's refused-versus-absent split and its privileged read, the holdings authorisation and its ordering, and the probe's transport and its fallbacks.

The sixteen-page bound is set against what the network carries: across **456 syncthing folders on 214 nodes**, the largest holds 63,999 entries and the 99th percentile 40,771, against a page size of 65,536. Every folder on the network is read in a single request, and no app resembling one on it reaches the bound.


Cabecinha84 (commented):
1. fixAppdataPermissions now chmods symlink targets, as root

  syncthingFolderStateMachine.js — the widen was changed from chmod -R 777 <appPath> to:

  serviceHelper.runCommand('find', { runAsRoot: true,
    params: [appPath, '-path', stagingPath, '-prune', '-o', '-exec', 'chmod', '777', '{}', '+'] });

  GNU chmod -R ignores symlinks encountered during traversal. find -exec chmod passes them as command-line arguments,
  where chmod dereferences them. I verified this empirically rather than arguing it:

  before                        600 target/secret.txt
  after  chmod -R 777 vol       600 target/secret.txt
  after  find ... -exec chmod   777 target/secret.txt

  The app container writes freely into its own bind-mounted appdata/ (and any m: dir). Planting appdata/x -> /etc/shadow
  and waiting for the next promotion gets that host file chmod'ed 777 by root — on every node running the app.
  fixAppdataPermissions runs on both the leader-seed path and the synced-promote path, and sudo is unrestricted
  (NOPASSWD:ALL).

  This is a regression: the code being replaced was not vulnerable.

  Fix: exclude symlinks from the expression, e.g.

  find "$appPath" -path "$stagingPath" -prune -o \( -type d -o -type f \) -exec chmod 777 {} +

  (There is still a TOCTOU window between find's stat and chmod's resolve; an alternative that closes it is chmod -R 777
  over each non-symlink top-level entry except .flux-op, since chmod -R never follows links it meets inside.) Note
  tests/unit/syncthingFolderStateMachine.test.js:683 pins the exact -exec tail and must change with the fix.

  ---

  Should be settled before merge

  2. bestHolder ranks mtime above size — the one ordering that can still lose the world

  if (right.newestModified !== left.newestModified) return right.newestModified - left.newestModified;
  if (right.bytes !== left.bytes) return right.bytes - left.bytes;

  A node holding a single 200-byte file written a minute ago outranks a node holding 5.8 GB written three days ago. The
  near-empty node seeds; the holder then syncs the tiny global, reads isSynced && receiveOnlyChangedFiles > 0, and the
  existing db/revert branch deletes its 5.8 GB. That is exactly the failure the PR's own comment on bestHolder describes
  ("the full one's files then become local changes that a later revert deletes") — the ranking order just doesn't close
  it.

  The scaffolding exclusions make this much less likely than it was, but not impossible: any >0-byte non-reserved file
  on a nearly-empty volume (a partial sync from a previous life, a half-run container, a restored fragment) is enough.
  I'd rank bytes first, or use mtime only to separate claims of comparable size. Worth an explicit decision from the
  author rather than a silent tiebreak order.

  3. globalState.folderHoldings is write-only

  It is created lazily and only ever .set(). Compare promotedFolderIds, which is rebuilt as a fresh Set every pass
  (syncthingMonitor.js:529). Consequences:

  - An entry survives app uninstall forever — unbounded growth across app churn on a node.
  - Once a folder is promoted, handleReceiveOnlyTransition stops running for it, so the node keeps publishing its last
    receive-only holdings indefinitely on the unauthenticated /apps/promotedfolders.
  - A reinstall of the same app can be ranked on the previous install's numbers before the first pass overwrites them.

  Prune on uninstall and drop/refresh on promotion.

  ---

  Minor / non-blocking

  - claims key inconsistency. Peers are keyed by peer.ip, self by localSocketAddr, and bestHolder reads them with exact
    string lookup — while the two lines immediately above use socketAddressesMatch for the same comparison. It works
    today only because the running-message ip is literally localSocketAddr (peerNotification.js:239). If those ever
    diverge (bare IP vs ip:port), everyoneAnswered silently goes false and the entire ranking degrades to lowest-IP with
    no log line saying so. Normalize the keys.
  - Stale .stignore line on spec change. Dropping an ml: mount leaves the exclusion behind — documented and warned
    saying so. Normalize the keys.
  - Stale .stignore line on spec change. Dropping an ml: mount leaves the exclusion behind — documented and warned about.
    But an app changing ml:cache → m:cache keeps cache unreplicated forever with only a log.warn. A //-comment sentinel
    delimiting the FluxOS-derived block would let it be rewritten wholesale instead.
  - db/localchanged paging. The JSDoc says "the default page size is 65536 entries, so one call covers any app volume." That
    is an assumption, not a guarantee — a folder with more locally-changed entries under-reports bytes. Only affects
    ranking (bytes stays >0, so folderIsEmpty is unaffected), but passing an explicit perpage would make it true.
  - ml: rollout. Spec validation only length-checks containerData, so an ml: app propagates network-wide, but
    parseContainerData throws Unknown mount syntax on any node that hasn't upgraded — install fails there. Not a consensus
    break, but the form shouldn't be advertised until the release is fleet-wide. The PR body doesn't mention this.
  - Stale comments. removeStagingPath still says "the mount plus one minted component with nothing to traverse" — it's now
    two components. unsyncedSubdirsOf carries getUnsyncedSubdirs's JSDoc (@param {string} containerData) while taking
    parsedMounts.
  - Scope. Suites 84 (log streaming) and 96 (surplus trim) are harness-flake fixes unrelated to mounts or cold start.
    Test-only and well-argued, but they'd review more easily on their own.

  ---

MorningLightMountain713:
All nine answered. Head is `3942436a7`, rebased onto current `development`.

## 1. `fixAppdataPermissions` chmods symlink targets, as root — fixed, `d274e489c`

Confirmed, and reproduced on a node rather than from the man page. GNU findutils 4.10.0 / coreutils 9.5: `chmod -R 777 vol` leaves a 600 target at 600, `find … -exec chmod 777 {} +` sets it to 777. Reachable end to end — the container is bind-mounted `appdata/` and each `<subdir>/`, `runAsRoot` is `sudo`, and both call sites sit in `handleReceiveOnlyTransition`, so it re-fires on every promotion.

The fix is your parenthetical rather than your primary suggestion. `\( -type d -o -type f \)` leaves the TOCTOU window you named, and with `-exec … +` the chmod runs batched well after `find` stat'd each path, against a process that knows when promotion happens. Instead each top-level entry of the volume is widened with its own `chmod -R`, `.flux-op` excluded, and the root itself without `-R` — recursing from there would walk into the staging directory the exclusion exists for. That is TOCTOU-free here specifically because the volume root is never mounted into a container, only the directories beneath it, so every path named as an argument is one FluxOS created and the app cannot replace.

The assertion you flagged at `syncthingFolderStateMachine.test.js:683` is replaced rather than adjusted, since there is no `-exec` tail left to pin, and a second test asserts that no path below the volume root ever reaches chmod's argument list. Three mutations break separately: the path walk, the staging exclusion, and `-R` on the root.

## 2. `bestHolder` ranks mtime above size — reordered, `239b821e7`

Bytes decide; recency now separates only equal byte counts, which is the one case a timestamp reads as the same content written at different moments.

Worth stating the limit rather than implying it is solved: a size and a timestamp cannot express containment, so no ordering of them is right in every case. Ranking bytes first still loses a smaller copy's unique recent writes. It is chosen for the direction it fails in — a near-empty volume can no longer publish over a full one, which is the loss that cannot be recovered from.

## 3. `globalState.folderHoldings` is write-only — fixed, `b273a02ca`

Dropped at the three points the data stops being what the claim says: promotion (which answers the receive-only question the claim asks, and is where `promotedFolderIds` is already reconciled), hard removal, and the volume recreate's point of no return, beside the synced-mark invalidated there for the same reason.

One consequence beyond the three you listed, and it is the sharp one. `bestHolder` gates on `everyoneAnswered`, so a node that has not looked at an app since reinstall *should* be absent from `claims` and force the lowest-IP fallback. A stale entry makes it present. So it does not merely misrank — it unlocks the claim-based ranking on figures describing a volume that no longer exists.

The removal cleanup also now clears node-local state before stopping syncthing rather than after. The volume is deleted whether or not that call succeeds, and it sat inside a catch that swallowed the failure.

## 4. `claims` key inconsistency — fixed, `e086defeb`

Fixed by removing the second source rather than normalising it. One pass over the election list keys every claim by `peer.ip`, the string the lookups read; being this node now decides only where the answer comes from — local holdings rather than a probe — and no longer how it is filed. The separate write above the loop and the `.filter` are both gone.

A missing claim is not a tie: it drops the whole field back to the address order with nothing logged, which is why this was worth closing rather than documenting.

## 5. Stale `.stignore` line on spec change — fixed, `554188f21`, and further than proposed

The root issue was that FluxOS wrote lines into that file and kept no record of which were its own, though it knew exactly at write time. A sentinel block records it, but only for volumes created afterwards; the first converge on an existing volume has to decide which lines to absorb, and the one ambiguous line is the stale `ml:` leftover the sentinel was meant to remove.

So the set is derived from the spec afresh each pass and set exactly, instead. That removes a dropped `ml:` mount's exclusion with the mount.

It is a wider behaviour change than you asked for and should be read as one: patterns FluxOS did not derive no longer survive a pass, where the merge preserved whatever it found on the volume. Deriving the set is also what makes it the same on every node — every node computes it from the same specification, and a file on one node's disk is not an input the others have.

## 6. `db/localchanged` paging — fixed, `4029badb4` and `8db885974`

Correct, and it matters more than when you wrote it: bytes is now the primary ranking key, so a truncated read corrupts the decision rather than a tiebreak. Two nodes both truncated are summing different prefixes, so a node holding more can report less and lose to one holding less.

Passing an explicit `perpage` is not sufficient on its own. The walk is now read to the end of its pages in the service, folded per page and never accumulated, and bounded at sixteen pages. How many entries exist is the app's to choose — it writes the files — and these requests are paid again every monitor pass, so an unbounded read over that list is its own problem. A folder that reaches the bound is reported truncated and its total is a floor, which is all the ranking needs.

The bound is set against what the network carries rather than picked: across 456 syncthing folders on 214 nodes, the largest holds 63,999 entries and the 99th percentile 40,771, against a page size of 65,536. Every folder on the network is read in a single request today, and the largest is at 98% of one page — so this was about to start truncating rather than being theoretical.

One page-boundary detail: a folder whose entry count is an exact multiple of the page size answers the next request with no list at all rather than an empty one, so "no list" means unreadable only on the first page. Read the other way, a full folder reads as unreadable, and unreadable is treated as holding data.

## 7. `ml:` rollout — not gated in code

The mechanism for this is the FluxOS version floor, raised to the release carrying the form; a height-gated rejection in new code would make new nodes refuse what old nodes accept, since registration only length-checks `containerData`. The requirement is stated in the PR body under **Before release**.

The failure mode meanwhile is contained: `appReconciler.js:258-267` catches an unparseable spec and refuses with a reason instead of looping, and that guard is already on `development`.

## 8. Stale comments — fixed, `78574b8a6`

Both. `unsyncedSubdirsOf` documents the parsed mounts it takes, and the empty answer for an unparseable spec is stated at `getUnsyncedSubdirs`, which is what parses.

`removeStagingPath` was the one worth catching: its sentence was a safety argument, not a description. It now names what actually makes the path safe to remove — the staging root and the minted component are both FluxOS's, and the app cannot interpose a link on either because the volume root is not mounted into any container.

## 9. Scope — staying as one PR

The three strands are not disjoint: `volumeReservedNames.js` is changed by the mounts and the staging work, and `syncthingFolderStateMachine.js` by the election and the staging refactor, so a split is a three-deep stack rather than independent PRs.

On suites 84 and 96 specifically: 89's fix is in the product code here, and `development` now needs it, since removing `fluxapps.redeploy.delay` was what exposed the cold-start standoff it was hiding.

## Also on the branch

`3942436a7` stops `portManager.test.js` binding a real port. The block asked the OS for a free port and released it before use, so the rest of the suite could be handed it first; the bind then failed and all eight tests in the block failed together on `reason: 'error'`. The listener is stubbed like every other collaborator there — the subject is the witness rules, and `fluxHttpTestServer` has its own suite against real sockets. Dropping the corroboration threshold from 2 to 1 still fails five of those tests.

Unit suite **6899 passing, 0 failing, 17 pending**; `npx eslint ZelBack/src tests` clean.

MorningLightMountain713:
Two routes removed since the reply above, at `12c330df6`: `POST /syncthing/db/ignores` and `POST /syncthing/config/defaults/ignores`.

What may leave a node for an app its operator does not own is decided by the specification, and these set ignore patterns from outside it — so what either wrote to an app folder was replaced by the next monitor pass. Neither had a caller: nothing in FluxOS routed through them, and the converge calls `getFolderIgnores` and `setFolderIgnores` directly, which is unchanged.

Noted in the PR body under **What this changes**, since it removes two endpoints.

MorningLightMountain713:
One thing to be explicit about before this is read as ready: **the end-to-end harness has not run at this tip.** A gate is queued.

What has run at `12c330df6` is the unit suite (6897 passing, 0 failing, 17 pending), `npx eslint ZelBack/src tests`, and a cross-module resolution sweep of every `alias.method()` against its target module's exports — which resolves identically to `development`, so the rebase and the removed exports introduced no unresolved call.

That is deliberate rather than an oversight: this branch was rebased across 163 commits today and has taken eleven commits since, so the earlier gate result does not describe this tip and is not cited anywhere in the PR body.

Cabecinha84 (commented):
Blocker: the new permission widen still makes root follow a link, reachable by the app owner

  The new fixAppdataPermissions code:
  const entries = await fs.promises.readdir(appPath);
  const widen = entries.filter((e) => e !== STAGING_ROOT).map((e) => path.join(appPath, e));
  chmod 777 appPath
  chmod -R 777 ...widen        // runAsRoot

  It is only safe if no top-level entry can ever be a symlink. The author's reasoning is that the volume root is never
  mounted into a container. That's true, but the container isn't the only thing that writes to the root:

  1. The owner's file API. Copy and move both keep links. Harness suite 72 asserts it: cp -a copies photos/evil ->
     /etc/shadow as a link. The path check allows a link as the source ("Operations act on a link rather than through
     it"), and nothing stops a plain name at the root as the destination. So the attack is:
     - the app creates appdata/evil -> / (or /etc, /root/.ssh) from inside its container;
     - the owner calls copyobject or moveobject with destination: "evil".
  2. Syncthing. The synced folder is the volume root (appsFolder + appId), and syncthing replicates symlinks. A link
     placed at the root on one node reaches every peer. A hostile node operator running an instance of the app can also
     publish one directly.

  chmod follows a symlink it is given as an argument. I checked with GNU coreutils 9.4, running the exact commands this
  code builds (chmod 777 vol && chmod -R 777 vol/appdata vol/evil vol/evildir). Both the linked file and the linked
  directory went 600/700 → 777, and so did the file inside the linked directory. So a link to / makes root run chmod -R
  777 over the host's whole filesystem. That breaks sudo and ssh at the very least, and opens local privilege
  escalation.

  When it fires: at the next promotion of any node running the app. That's either path in handleReceiveOnlyTransition
  (the leader seed or the synced promote). In particular, every new instance hits it after it has synced the link from
  its peers, and the owner can force new instances by changing the instance count or redeploying.

  It's a regression: the old chmod -R 777 <appPath> doesn't follow links it meets during the walk, so it wasn't
  vulnerable. The new unit test "never names anything the app could have made a link" builds its fixture with a
  top-level entry called escape and checks that it is passed to chmod. The test encodes the wrong assumption.

  Suggested fixes, strongest first:
  - Do the widen inside the executor container, with only the volume mounted. A link then resolves inside the
    container's read-only root filesystem, never on the host.
  - Or go back to chmod -R 777 <appPath> (immune to links) and accept that .flux-op gets widened too. Only root and the
    executor write inside it, and the owner API already refuses it.
  - If readdir output must stay: lstat each entry and skip symlinks. This still leaves a race window (syncthing or the
    executor can swap an entry for a link after the check), so it's the weakest option.

  In every case, a test should plant a real top-level link and check the target's mode.

  Non-blocking (fine as follow-ups)

  - Forged holdings claims. /apps/promotedfolders has no authentication, and the ranking trusts whatever holding a peer
    sends. A hostile peer can claim 2^53 bytes and always win the seed. That doesn't add much to what a hostile peer can
    already do once it's sendreceive, but a plain number check (Number.isFinite, >= 0) is cheap. The endpoint also now
    exposes each receive-only app's size and last-write time.
  - Glob characters in ml: names. validateSubdirOrFilename allows *?[]{}. So ml:*:/x becomes the ignore line /*, and the
    whole volume, appdata included, stops replicating. Only the owner is hurt, but escaping or refusing these
    characters is cheap.
  - .stignore is now an exact set. Patterns the Flux team adds through /syncthing/db/ignores are wiped within one pass.
    This is intended, and the author says so; it just needs to be known operationally.
  - On-disk emptiness check vs. new directories. checkDirectoryHasSyncScopedContent (the check against an index that
    claims bytes over an empty disk) doesn't exclude ml: directories or the permanent .flux-op. In practice lost+found
    already makes that check fire rarely, so the PR adds little here.
  - Rollout of ml:: raising the minimum FluxOS version before any app uses ml: has to be on the release checklist.

MorningLightMountain713:
Answered. Head is `4e52e847e`.

## The blocker

You were right, and the reasoning I gave for the previous version was the wrong rule. It checked the container and stopped there. The container is not the only writer to the volume root: the file API accepts a link as a source deliberately and any unreserved name at the root as a destination, and syncthing replicates a link into the folder it replicates, which is that same root.

`fixAppdataPermissions` is one command again — `chmod -R 777 <appPath>`, the volume root as its only argument. That is the directory FluxOS created, and nothing can substitute it. The executor's staging directory is widened with the rest; the mode buys nothing there, since reaching it needs a path to the volume root that no app container is given.

I reproduced the behaviour on GNU coreutils 9.5 rather than arguing it: a link given as an argument has its target widened and the walk recurses into it, while the same link met inside a traversal is left alone.

The unit test that went with the old version asserted how many path separators were in the argument list, which encoded exactly the assumption that turned out to be wrong. It is replaced by one asserting the argument list itself — one chmod, one target, the volume root — and that nothing runs a path walk to build it. Mutation-tested three ways: the per-entry form, the `find … -exec` form, and dropping the widen, each killed separately.

## Forged holdings claims

Taken further than a range check, because a range check does not answer it — a fabricated figure is finite and positive, and any ceiling a node can derive a caller can claim.

The open `GET /apps/promotedfolders` no longer carries `holding` at all. It answers `ready` and `folders`, which is what it answered before this branch. A `POST` beside it answers the holdings to a caller that signs as a node on the deterministic list, the way every node-to-node call on this API proves itself, or that holds flux team privilege. The signature names the node being asked and the moment it was made, so one captured in flight cannot be turned on another node or kept, and the moment is read against the bound every signed Flux broadcast is held to.

A caller that shows neither is answered without the holdings rather than refused, and a peer whose release has no `POST` route is asked again over the `GET`. Both read as claiming nothing, which is the case the address order already exists for, so a mixed-version fleet degrades to the behaviour it has today rather than breaking.

Separately, a claim now has to be comparable before it counts as an answer. Subtracting a field that is not a number gives `NaN`, which `sort` reads as equal — so the comparison returns before the address tiebreak and the winner becomes whichever candidate the peer list carried first. That list is a database read ordered by when each node received the broadcasts, so it differs per node: two nodes rank the same field differently and both elect themselves. An empty holding reaches this, and an older peer sends one.

## Glob characters in `ml:` names

Refused, and the rule is wider than the characters you named, because the file is line-oriented as well as pattern-oriented. A name carrying a line terminator writes a second ignore line the specification never named, and the converge then derives four lines, reads back five, and rewrites the file and rescans the folder every pass for as long as the app exists. The name is held to the class `pathSecurity` refuses in a path component, for the reason that file already states.

`*` and `?` were in fact already refused by the existing character check; `[`, `]`, `{`, `}` and `\` were not.

## The on-disk emptiness check

This one was worth more than it looked. The check could not fire at all — not rarely, never. Every app volume is formatted ext4 so `lost+found` is at the folder root before the app is, the primary mount and every `m:` directory survive the data wipe because the wipe preserves the mount structure, and a directory counted as content. So the disk always read as occupied and a stale index over emptied data was never caught.

Both sides now apply one rule for what is the owner's and what is the volume's scaffolding, and the index says which kind of claim it is making: files claimed are answered by files on disk, while bytes claimed with no files is a folder whose payload is empty directories, where counting directories is the only honest reading. That second half is the 2026-07-04 false positive, which is why a files-only walk cannot simply replace the old one.

The harness recorded this gap as a pending test rather than deleting it, with a comment naming this exact fix. That test is un-skipped and passing.

## `.stignore` as an exact set

The endpoint this concerns, `POST /syncthing/db/ignores`, was removed on this branch at `12c330df6` — about five minutes before your review, so the note reads as current. Nothing in the repository calls it, and the only remaining callers of the underlying service functions are FluxOS's own convergence pass.

## `ml:` rollout

Already in the description under **Before release**, added in the previous round.

## Verification

Full gate on chud: **108 passing, 1 failing** over 109 suites, 1h35m. The failure is suite 82, whose fixture mirrors the published executor image and reached the public registry to resolve a pinned tag — an unretried request for a digest the repository already carries and the box already had cached. Fixed at `4e52e847e` and proven both ways with that registry made unreachable: without the change the suite fails at that call, with it both affected suites pass.

Unit **6934 passing, 0 failing, 17 pending**; `npx eslint ZelBack/src tests test-infra` clean; the cross-module resolution sweep resolves identically to `development`.

Cabecinha84 (commented):
Blocker: files-only phantom check hides data FluxOS can't read (1f1fef634)

  verifySendReceiveFolderSafety now decides how to walk the disk from what the index claims. If globalFiles > 0, it
  counts only regular files (countDirs: !claimsFiles). The walk (countFilesUpTo) silently skips any directory it can't
  list, "like find does".

  In production FluxOS runs as a normal user. It uses sudo for everything privileged (62 runAsRoot call sites, including
  the chmod itself). App data is written by containers with their own user ids, and databases lock their data directory
  down on every start. The Postgres entrypoint, for example, sets PGDATA to 0700 owned by uid 999. With the primary
  mount on appdata, the FluxOS user can't list that directory, so:

  - The walk finds 0 files, and the check reports phantom_index_empty_disk.
  - syncthingAppsCore raises a SAFETY BLOCK: the folder is demoted to receive-only and the container is stopped.
  - Every later seed or promote attempt runs the same check and fails the same way, so the app stays down.
  - The first monitor pass after a restart checks every installed app. Every node restarts on upgrade, so every node
    running such an app stops it on the upgrade to this release.

  Before this PR, the check counted directories. The appdata directory itself counted as content, so the check could
  never reach this state. The PR body is right that the old check never fired. This change makes it fire on the wrong
  condition, which is the same kind of false positive as the 2026-07-04 incident the code comments cite.

  Reproduced. As uid 1000, on a volume holding .stfolder, .stignore, lost+found and an unreadable appdata/:
  - Files-only walk (this PR): {"hasContent":false,"fileCount":0}
  - Old directory-counting walk: {"hasContent":true,"fileCount":1}

  Why the gate can't catch it. test-infra/Dockerfile.fluxos has no USER line, so FluxOS runs as root in the harness and
  can always list the directory. Suites 61 and 1801 pass for that reason alone.

  Suggested fix, any one of these:
  - Treat a directory it can't list as content (unknown is not empty — the rule this PR already applies to holdings).
  - Do the count through the root executor.
  - Use files-only counting only when the whole tree was readable.

  Add a unit test with an unreadable subtree, and ideally a harness suite that runs FluxOS as non-root.

  Previous blocker: fixed

  d7e60be6b goes back to a single chmod -R 777 <root>. chmod does not follow links it meets inside a recursive walk, and
  the volume root is created by FluxOS. So an app can no longer plant a root-level link to get a path of its choosing
  widened. The staging directory now gets widened too, but it is owned by root and no app container can reach it. That's
  acceptable.

  New code since the last review that checks out

  - Signed POST /apps/promotedfolders. Holdings (size and last-write time per app) are no longer on the open GET. The
    POST serves them only to:
    - a signer on the deterministic node list, whose target is this node's own address and whose timestamp is within the
      5-minute broadcast window, or
    - a caller with Flux team privilege.

    JSON bodies are parsed for every route (express.json() in fluxServer.js). An older peer that answers 404 is asked
    again over GET. Holdings are new in this PR, so there's no mixed-version regression.
  - isRankableClaim. A non-numeric or NaN claim now counts as no answer, so every node computes the same ranking.
  - Withdrawal. A holdings figure that can no longer be read is withdrawn instead of left published. This removes the
    stand-off after a syncthing restart.
  - ml: names. Names are now held to literal ignore-pattern characters and control characters, so each name becomes
    exactly one .stignore line. appdata, backup and the reserved names were already refused.
  - POST /syncthing/db/ignores and /config/defaults/ignores removed. Nothing called them.

  Minor, non-blocking

  - verifyTimestampInFluxBroadcast has no lower bound, so a request dated in the future stays valid longer than 5
    minutes. Only a listed node can sign one, and listed nodes are entitled to the answer anyway.
  - isSyncedPayloadName excludes lost+found, but .stignore doesn't. Files fsck recovers into it are indexed but not
    counted on disk. This is rare, and it is another way the phantom check can misfire.

MorningLightMountain713:
All three are addressed, and the blocker was right in a way the fix first proposed for it would not have covered.

**The blocker — a files-only walk reading a refused directory as an empty one.** Correct, and reproduced here before changing anything: with two files under an unreadable `appdata`, the walk returned `hasContent: false, fileCount: 0`, where counting directories returned true. One correction to the premise, which narrows it rather than removing it — FluxOS is root on Arcane and the operator's account only on legacy nodes, so the exposure is legacy-only. It is also not hypothetical there: syncthing is spawned with sudo on every node, so it indexes as root what an unprivileged FluxOS is refused, and postgres holds its data directory at 0700 and chmods it on every start, not only at init.

Neither of the first two suggestions is taken, and for the same reason: treating a refused directory as content disables the guard wherever the node cannot see, which is exactly the nodes that need it. The walk now reports a directory it was not shown rather than counting it as empty, and root is asked one bounded question per refused subtree — is there content here — with the payload rule applied at the volume root before anything is handed over, so the privileged side never decides what the owner's data is. A volume no process on the node can read is recorded as the fault it is and the folder left as it stands. The third suggestion, counting only when the whole tree was readable, is what remains when root cannot answer either.

**The harness could not have caught it**, as noted. That is now fixed rather than worked around: a fleet can name the nodes that run FluxOS unprivileged, the container staying root because making loop devices and supervising a dockerd is the host's work on a real node. The new suite fails on the app being stopped when the fix is reverted.

**`verifyTimestampInFluxBroadcast` having no lower bound.** Taken, and it matters more than non-blocking: a node can mint a request dated far ahead against a chosen target and publish the body, at which point anyone holding it reads that node's holdings with no key. The request is bounded both ways now, forward by the tolerance the network already keeps for clocks that do not agree — named and exported rather than restated, so both places holding a signed message to it hold it to one value.

**`lost+found` indexed but not counted.** Taken, and the larger half of it is that nothing kept it off the network at all: syncthing's internal names are `.stfolder`, `.stignore` and `.stversions` and nothing else, so what one node's fsck recovers replicated to every other holder of the app. It is an ignore line now, and the disk side reads the scope off those lines rather than keeping a second list that has to agree with them.

Reviewing the change again after those, a few more went in: an `f:` mount's zero-length file at the volume root made the phantom guard read every such volume as occupied, so the guard was off for that whole class of app; the holdings caller now proves itself before the node looks anything up about itself; a node whose network state has not started answers rather than parking the request, since a peer reads silence as that node being gone when it means the opposite; the staging directory is refused to a reader that asks for it by name, not only hidden from the listing; and the own-address lookup is served from a cache rather than asking the benchmark daemon on each of thirty-one call sites.

One thing this does not close, and the code now says so where the election is decided: which of the two election rules applies is read from one node's own answers, so a candidate silent to one node and answering another can put two nodes on two rules and elect two winners. Narrowing it costs more than it saves, because a node that stands down whenever a candidate is silent will not seed while a peer is merely dead. An agreed view of the field is what closes it.

Unit suite 7214 passing, lint clean, cross-module sweep identical to `development`. The affected suites have been run individually at this tip and the full gate is queued — it will be run before this is merged.

Cabecinha84 (commented):
Blocker (new)

  The boot sweep deletes as root through a .flux-op symlink (volumeExecutor.js ~2102 for the sweep, ~1682 for mkdir -p).
  - What the code does: sweepStagingDirectories lists <mount>/.flux-op, then runs sudo rm -rf <mount>/.flux-op/<uuid>.
    Nothing checks with lstat that .flux-op is a real directory.
  - How the link gets there: on current development, .flux-op is not a reserved name. The owner can have their app
    create a link in appdata and move it to the volume root as .flux-op with the file API; the third pass verified that
    the file API accepts a link moved to an unreserved root name. The old .stignore only excludes /.flux-op-*, so
    syncthing also copies the link to every node holding the app.
  - What happens after the upgrade: at boot, root deletes every UUID-named entry in whatever host directory the owner
    chose. That can include another tenant's volume at a predictable path. Separately, mkdir -p creates directories as
    root through the same link.
  - Reproduced: I pointed .flux-op at a victim directory and ran the real sweep with sudo swapped for a plain local
    command. It removed victim/<uuid> and reported it as .flux-op/<uuid>.
  - Fix: before the sweep and before any mkdir, lstat .flux-op. If it is not a real directory, unlink it and recreate
    it, or refuse to continue.

  Should-fix (all verified)

  1. The "not too far in the future" timestamp check can be bypassed with a string (appQueryService.js ~621).
     verifyTimestampInFluxBroadcast does timestamp + maxOld, and a string makes that a concatenation. A timestamp of
     "<now − 1 year>" or [ts] reads as fresh, which is exactly what commit 9a6225b0d was meant to stop. Fix: require
     Number.isFinite(body.timestamp) and a string target.
  2. Whitespace in an ml: name. ml:cache :/b is accepted, with subdir "cache ". Syncthing trims the ignore line, so it
     becomes /cache, and the audit showed this against a real syncthing 1.27.2. The result: the m:cache directory stops
     replicating, the "local" directory replicates instead, and the .stignore is rewritten and rescanned on every pass.
     Fix: reject leading and trailing whitespace.
  3. Nodes can see different election fields. The new signed endpoint answers "no holdings" to a caller it refuses, for
     example because of clock skew over 120 s or a node-list mismatch. That caller then falls back to the address order
     while the others rank by bytes. Using the real isDesignatedLeader, this produced two leaders in one case and no
     leader in another, which is the same stall this PR exists to remove. Fix: when the caller is refused, say so
     explicitly instead of returning an empty answer, and don't fall back to the address order on it.
  4. False "phantom index" on syncthing v1 (7a308b8d9). Take a volume holding only the zero-byte f: file plus empty
     directories. v1 reports globalFiles=1, globalBytes=256, the walk counts files only and skips the empty file at the
     root, so the check reports phantom_index_empty_disk and the app is stopped. Syncthing v2 reports 0 bytes for the
     same tree, so it is not affected.
  5. An ml: directory added by an app update replicates once. The ignore line is only written before the first scan on a
     fresh install. On a soft redeploy it arrives at the next monitor pass, after the container has started writing.

MorningLightMountain713:
All six are dealt with. Four were real and are fixed, one is a ruling this PR already records, and one cannot occur on the fleet.

**The blocker — the boot sweep deleting through the staging directory.** Right, and reproduced here before anything changed: pointing that name at another directory and running the real read-and-remove leaves that directory's UUID-named entries gone and reports them under a staging path they were never in. The link survives, so it fires again every boot. The correction to the reach is that it is not the sweep alone — an operation's `mkdir -p` creates through the same name, so a link there is also a root-privileged write at a path of the owner's choosing. The name is now established to be a directory before either is done, and what is not one is removed with `unlink`, which never follows a link and refuses a directory. That is defence where the privilege is rather than at the door: the name is refused to the file API on this branch, but a volume carries what earlier releases allowed onto it and what syncthing copies from a holder still running one, so the name being refused today says nothing about the entry sitting there now.

**The timestamp.** Taken, and the fix is one step further out than suggested. The bound is arithmetic, so the value is required to be a number — added to a string or an array it concatenates, and a message of any age then reads as fresh. Held at the helper every signed message passes through rather than at the one call site, because that helper has nine callers and the same reading is wrong at each of them. One correction to the framing: the forward bound is not what breaks. A numeric string compares numerically, so a far-future stamp is still refused; what a string defeats is the check that a request is not too old.

**Whitespace in an `ml:` name.** Taken, and settled from syncthing's source rather than from the symptom: each line is trimmed as the file is read and again as it is reported, and nothing else in the parse touches what is reported — the dedup, the comment skipping, the escape handling and the unicode normalisation all build the patterns. So the consequence is larger than the rewrite loop. The name stands for a different name as a pattern, which means the directory the specification declared local replicates while one named the trimmed way stops, and the disk walk and the index then disagree about which names are the owner's — the two sides of the phantom check on different scopes. Leading and trailing whitespace is refused at the name, on Unicode's set rather than ASCII's, and interior whitespace is left alone because it survives the round trip and is the owner's.

**Nodes seeing different election fields.** This is the residual limitation the description already records, and the trigger named here sits inside it: which of the two rules applies is read from one node's own answers, so any candidate that does not answer *this* node puts it on a different rule from the nodes that were answered. Closing it needs an agreed view of the field, and half of what is suggested here is the narrowing that section explains the cost of — a node that stands down whenever a candidate is silent to it will not seed while a peer is merely dead, and a dead holder must never strand an app with no writable copy anywhere. It stays recorded rather than half-closed.

**The phantom index on syncthing v1.** This one cannot happen. The fleet is held to a syncthing version floor — a daily check against the published minimum, with the apt sources moved to the v2 series and the package upgraded below it — and a node reports `v2.0.15` today. On v2 that tree reports zero bytes, which takes the early return and never reaches the check, as noted.

**An `ml:` directory added by an update.** Real, and the path is narrower and the consequence longer-lived than stated. An update that leaves `hdd` alone takes the soft redeploy, which installs onto the existing volume and so never reaches the seed that volume creation writes; syncthing's watcher is enabled by default with a ten second delay, against a converge up to one monitor pass away. The policy is now asserted before the directory is created. Running that down turned up why it would not have held on a legacy app: a version 3 or lower redeploy reaches the installers directly rather than through the path that writes the local row, and neither uninstall on that branch touches it, so the row kept the old specification for as long as the app stayed installed and every reader of it — this node's own report, the reconciler, the ignore policy — worked from that. Fixed with it.

Rebased onto `development` at `074adae73`; every file both sides touch was checked by content rather than by the rebase running quietly. Unit suite 7325 passing, 0 failing, 17 pending, lint clean.

**The full gate is running now** at this tip, 112 suites. I will report the result here. Nothing merges before it returns.

Cabecinha84 (commented):
Two things must
  happen before this merges:
  1. The full harness gate has to finish green at this head. The author says the 112-suite gate is running now, and the
     PR body agrees. No result has been posted yet.
  2. The FluxOS version floor has to be raised first. It must reach the release carrying this change before anyone uses
     ml:. The PR body states this: registration only checks the length of containerData, so it does not reject ml:. An
     old node would fail the soft redeploy, fall back to a forced removeAppLocally, and delete the app's volume.

Non-blocking polish (can go in a follow-up)

  - volume_unreadable is scored as tampering. It has severity 1 in EVENT_SEVERITY, but #1809 set the rule that a fault
    of the host's own weighs 0 (volume_host_fault: 0). It is recorded once per hour per app and adds to the tamper
    score, and a score above 10 plus a blocklist entry puts a non-ArcaneOS node on DOS. In practice this only matters
    for a node that is already blocklisted, but it is inconsistent with #1809's rule. Suggest 0.
  - Duplicated doc comment: in syncthingIgnorePolicy.js, the docblock of ensureStignoreCovers repeats the paragraph that
    starts ".stignore is syncthing's own control file…".
  - Silent path on install: if ensureStignoreCovers fails during the soft install (syncthing is down, or the folder is
    unknown), it only logs, and the install goes on to create the ml: directory without the ignore. It is low risk,
    because soft uninstall does not remove the syncthing folder.

MorningLightMountain713:
**The full gate is green: 112 of 112 suites, no failures.** Run on a dedicated box at `MAXN=6`, from images the digest guard verifies against the tree.

`volume_unreadable` weighs 0. It is recorded when no process on the node, root included, could read a volume, which describes what this node can see rather than what was done to it — the class `volume_host_fault` and `recreation_failed` are already in. The reason sits beside theirs in `EVENT_SEVERITY`.

The ignore policy's docblock says once what it was saying twice.

The install path's handling of a failed converge is deliberate. Refusing an app's redeploy because syncthing was briefly unreachable costs more than the window it closes, and the monitor converges on its next pass; the directory is created either way, so the alternative is an app that does not come back.

The version floor is the one condition this still needs, as the description says.

Rebased onto `development` at `b8e66367f`. Every file both sides touch was checked by content — each commit's contribution is unchanged by the replay and the merged work's symbols resolve identically to `development`. Unit suite 7362 passing, 0 failing, 17 pending; `npx eslint ZelBack/src tests test-infra` clean.

Cabecinha84 (approved):
ack

PR #1809, opened 2026-09-17
An app's volume goes on a disk the node owns, and FluxShare keeps only a way to collect its files

MorningLightMountain713:
Two changes: where an app's volume image is placed, and FluxShare reduced to a way of collecting the files it already holds.

## Placement

An app's `hdd` allocation is enforced by giving the app a filesystem of exactly that size: a file is allocated, an ext4 image is made inside it, and that image is mounted at the app's directory. The image is a file, so where it goes is a question about a directory. It was being answered from device names.

### What was wrong

- **Nothing asked whether the mount point was a directory.** Where docker binds its own files into a containerised FluxOS — `/etc/hostname`, `/etc/hosts`, `/etc/resolv.conf`, each reporting the host disk's free space — the first of them won and `fallocate` answered `Not a directory`. Every install on such a node failed, and `testAppMount` failed the same way at boot, which puts the node in DOS with *Unavailability to mount applications volumes*.
- **One filesystem was counted once per bind mount.** Those binds are one device reported four times and their free space was summed, so a node with 1.8 TiB free believed it had 7.3 TiB.
- **A multi-disk node only ever used one disk.** First-with-room filled whichever disk the mount table happened to name first, reaching the others only for an app too big for what was left. The apps contended for one disk's I/O while the rest of the operator's hardware sat idle.
- **A ZFS- or btrfs-rooted node could not install an app at all.** The candidate filter required the source to contain `/dev/`, and those filesystems are named by dataset, so the candidate list came back empty.

### What it is now

There is one notion of a disk. `isHostFilesystem` is a filesystem the node may use: real, surviving a reboot, on this machine, not the boot disk, not FAT, and not an app volume it has already placed — a loop mount is such a volume, except at the root, where a loop is the host disk itself.

`findmnt --real` excludes pseudo filesystems and nothing else, so an NFS export, a CIFS share, a virtiofs passthrough and a fuse-mounted object store all arrive in the mount table looking like local disks. They are named and refused. `fallocate` is unsupported on CIFS and on NFSv3, so an image cannot be created on one at all, and where it can the app's data is hostage to a share that can go away while the node keeps running.

Where an image may be PUT and where one may be FOUND are asked separately, because they are different questions. **Placement** adds the filesystem types that can carry a new image, that the mount point is a writable directory, and that the path resolves through this mount rather than one stacked over it; it counts each disk once and takes the emptiest. **Discovery** asks only location and ownership, minus `/`, whose images live in `appvolumes`, and does not deduplicate, because two directories on one disk are two places an image can sit.

A search is not narrowed by filesystem type. An image already written to a disk is readable whatever that disk is formatted as, and an earlier release placed by source alone — so refusing to look at a type reports images missing that are sitting on nodes today, which is a tampering event against the operator and an image orphaned by its own removal.

Free space is never totalled across mounts. A total means something only when no two rows share any, and they can: `findmnt` reports a ZFS pool once per dataset with every dataset naming the pool's free space. An app is placed on a volume whose own free space clears its `hdd` plus the system reserve, which is a question about one disk. What each row has *used* is its own and does add up, and that is what the reserve is computed from.

## Finding an image

FluxOS chooses where an image goes, so it is the authority on where one is. Deriving that again on every boot by scanning the disks for a filename is what makes any written file a candidate to be mounted as root over an app's data, and it is what every exclusion rule above exists to compensate for.

The path, and a filesystem UUID the node stamps the image with when it creates it, are recorded against the component on its runtime-state document — node-local, already keyed per component, and not the owner's signed specification. A later boot looks the image up, and checks the stamp before mounting: a file substituted where this node put its image is refused. An image found somewhere else is a stale record rather than a foreign file — a volume re-created, a release rolled back — so it is adopted and the record replaced, not refused.

Nothing has to be migrated by hand. A node upgrading with its apps running never searches for them: a mounted volume's loop device names its own backing file, so the path comes from the kernel. One whose app was down is searched for once, under the rules above, and recorded when the mount proves it real.

## FluxShare

FluxShare stops being a file store. FluxOS runs other people's applications; a personal one for the node operator is not part of that job. What survives is a way to collect what is already there — a listing and a download, and nothing else.

Nothing in FluxOS creates `<appsFolder>/ZelShare`. The service's only `mkdir` is non-recursive and takes a folder name, so on a node without that directory the listing, the upload and every named folder creation end in `ENOENT`. Confirmed on a live node.

That is the Arcane story, and it is not the whole fleet. `ZelApps/ZelShare/.gitkeep` is tracked, and a legacy install resolves the apps folder to the repo root — so on those nodes the directory arrives with the checkout and this particular failure does not apply. Whether the feature works there beyond that is untested. A fleet sweep found it provably unused on 5,129 Arcane nodes and holding data on two, both belonging to one operator who created the directory from a shell. The 393 legacy nodes cannot be checked at all: no team SSH, and every route is node-operator-only.

`ZelApps/ZelShare/.gitkeep` goes, since nothing needs that directory to pre-exist now. `ZelApps` itself stays — that is the apps folder, and every app volume mounts under it. A node holding files in `ZelShare` keeps them: git does not remove a directory that still has untracked files in it, so what disappears is an empty directory on the nodes that never used the feature.

### What is left

Two routes, `NODE_OPERATOR` as every FluxShare route has always been, in a service of 160 lines where there were 920:

- `GET /apps/fluxshare/getfolder/:folder?`
- `GET /apps/fluxshare/getfile/:file?`

No upload, no rename, no delete, no folder creation, no stats, no database — and **no token**. `getfile` used to fall through to a `{name, token}` lookup when the privilege check failed, which served the file to whoever held the link. There is no such path, and a test asserts a refusal ends the request rather than branching in it.

A node with no share directory answers an **empty list**. Absent and empty are the same thing to a caller collecting files, and the `ENOENT` that came back instead is what made the feature look broken from its own interface.

Containment is resolved rather than spelled — `sanitizePath`, then `verifyRealPathOfExistingPath` against the share root, so a link out is refused even where it names something that does not exist yet.

**Proved on a real node** (Arcane, branch deployed, files seeded by hand): the listing returns the directory and the file, descending into a subfolder works, both files download, `../../../etc/passwd` is refused with `Path validation failed..`, unauthenticated listing and download are 401, `getfile/:file/:token` does not route, and `stats`, `sharedfiles`, `createfolder` and `removefile` all 404. The frontend page is RunOnFlux/fluxos-frontend#172.

Removed with it: twelve routes and the share token on a thirteenth — `getfile` keeps its path and loses `/:token?` — plus the FluxShare database configuration, and `capacityVolumesInGib`, whose only caller totalled disk size for the FluxShare quota.

### Release order

The frontend page is reduced to matching read-only shape in RunOnFlux/fluxos-frontend#172, and **that should ship first**. The two routes kept here already exist on `development` at the same paths, so the new page works against an old node. In the other order the old full page meets a node that no longer has `stats`, `uploadfile` and the rest, and 404s on them.

## A host that cannot mount volumes

A mount can fail for reasons that describe the host rather than the volume: no loop device, a disk the kernel remounted read-only, a mount table that will not read, a mountpoint that cannot be created. None of that is the operator's doing, so none of it scores against them — but it was then recorded nowhere at all, and nothing else on a node reports it, so a host that could not mount volumes left no trace.

It is now recorded as `volume_host_fault`, in the same zero-weighted class as `recreation_failed`: visible on the tampering events a node already publishes, countable across the fleet, and weighing nothing. Given any weight it would gather in every hourly bucket for as long as a disk stayed broken, which is the outcome keeping it out of the score exists to prevent.

**Nothing acts on it.** That is deliberate, and it replaces something that would have.

FluxOS carried an unfinished feature meaning "if this node cannot mount app volumes, take it out of service". The half that decides never reached the half that acts — the mount test wrote a module-local, and the state object the availability checker reads is a different one nothing ever wrote to — so it has never run on any node. Completing it was not the fix available: the value it reaches for is the one at which a node removes every app it holds and repeats every ten minutes, reached from a single attempt of a test that a busy loop device or an unreadable mount table can trip. A transient would have destroyed the data of every app on the node. It is removed here rather than left for someone to finish.

The judgement it wanted is the one the incident model already makes, and makes with the guards that one had none of: a per-type severity, hourly buckets so a persistent fault accumulates rather than firing once, a score threshold, and a check that the node is listed on chain before anything is enforced.

So the path forward, if the fleet data ever says a node like this should be acted on, is to give `volume_host_fault` a weight and let that machinery apply — measuring the population first, since nothing has ever counted it. Not a wire back to a value that means delete everything.

## Evidence

- **Unit: 7195 passing / 0 failing / 17 pending.**
- **31 placement tests over mount tables read off the fleet** rather than composed: two Arcane nodes, a legacy node and a containerised FluxOS, plus a loop-mounted root, a dataset-named root, network storage and a FAT ESP for the shapes the fleet may hold and those four did not.
- **Each of the twenty rules a candidate passes fails a test of its own when removed**, measured across both suites — neither file's green is evidence by itself.
- **What the ZFS change does and does not claim.** A ZFS-rooted node is no longer *refused* a candidate list, which it was. Whether the install then completes is untested: `fallocate` has historically been unsupported on older OpenZFS, and there is no ZFS node in the fleet or the lab to run one on — Arcane is XFS, legacy is ext4, the harness is ext4. Every ZFS case here is a mount-table fixture.
- **An image is verified before it is mounted**, on a real node: `2001-app-volume-is-the-one-this-node-made` installs an app, substitutes a complete ext4 filesystem at the image's own path, and asserts the reconciler reports the volume unavailable and mounts nothing — then puts the genuine image back and waits for it to mount, so the refusal is a refusal and not a node that mounts nothing.
- **A call between services is checked against what the callee exports**, statically, over the real sources. A member that is not exported is `undefined`, so the call throws where it runs and a proxyquired caller cannot see it; the check found three, one of them introduced here.
- **A fault is never read as a fact about the disk.** A mount table that did not arrive is not "nothing is read-only"; a database that would not answer is not "this node recorded no image"; a probe that could not run - including one asked about a file that is no longer there - is not "the image holds no filesystem"; a removal that failed is not a volume cleaned. Each of those answered in the direction that does damage, and each now defers or reports what actually happened.
- **An image found somewhere other than where this node recorded it** is reported as `volume_image_moved`, at no weight. Refusing it would refuse a volume legitimately re-created, so the record is replaced and the fact recorded rather than swallowed — deleting an image is recorded, and deleting it and leaving another was the quieter of the two. Whichever pass mounts the volume records it, because the mount replaces the record and no later pass has anything left to compare against.
- **An image left behind the one the search answers with is named.** The candidate order decides between two files under the same name - the node's own directories outrank a mount, because nothing else can write into them - and the one that answers is then recorded, so every later lookup reads the record without searching. A second image is reported at the one moment anything sees it.
- **A container runtime's storage is refused** — Docker, containerd, LXD and kubelet data directories — for placement and for finding an existing image. On a storage driver backed by real filesystems each container's root is a mount like any other, so an image could otherwise be written into another application's container, or a planted one answered from it and loop-mounted as root at the victim's directory.
- **The full harness returns 109 of 110 across the stack**, measured at `991c18751`; the three commits after it carry unit coverage only. The single red was a pre-existing restart-timing race in the location-table suite — outside this PR's code, and fixed on the branch stacked on it; that suite re-runs **10 of 10**.


Cabecinha84 (commented):
Finding 1 — BLOCKING: remote filesystems are now valid placement targets, and they sort first

  isHostFilesystem replaced the old source.includes('/dev/') requirement with an fstype denylist (EPHEMERAL_FSTYPES +
  vfat). The assumption behind that is that findmnt --real has already removed everything that is not a local disk. It
  has not. --real filters against libmount's hardcoded pseudofs list — network and passthrough filesystems are real
  filesystems and are listed.

  Reproduced by running the PR's own volumeService against this machine's live findmnt --real output:

  --- placementVolumesInGib (emptiest first) ---
  │ 0 │ '/dev/sdd' │ '/'      │ 1007 │ 203 │ 753 │
  │ 1 │ 'C:\\'     │ '/mnt/c' │  952 │ 649 │ 303 │   ← 9p passthrough, accepted

  --- capacityVolumesInGib ---
  │ 0 │ 'drivers'  │ '/usr/lib/wsl/drivers' │ 952 │ ← same remote volume,
  │ 1 │ '/dev/sdd' │ '/'                    │1007 │    counted twice
  │ 2 │ 'C:\\'     │ '/mnt/c'               │ 952 │
  summed size GiB = 2911     (local disk is 1007)

  nfs, nfs4, cifs, 9p, virtiofs, fuse.* (rclone, sshfs) all pass the predicate. On development today every one of them
  is excluded by the /dev/ test.

  Failure scenario. A node operator with a 4 TB NAS at /mnt/nas (NFS or CIFS), or a Proxmox guest with a virtiofs share.
  Because placementVolumesInGib sorts emptiest first, that share is now the first candidate offered, ahead of every
  local disk:

  1. testAppMount() runs at boot, picks the share, and runs fallocate → mke2fs → mount -o loop on it. fallocate is
     EOPNOTSUPP on CIFS and on NFSv3; mount -o loop over a network share is fragile. Any throw lands in the catch at
     advancedWorkflows.js:3268 and sets
     dosMountMessage = 'Unavailability to mount applications volumes. Impossible to run applications.'
     — the node goes to DOS. That is the exact failure mode this PR exists to eliminate, re-introduced for a different
     node shape.
  2. If it succeeds, app data silently lives on remote storage; a NAS outage puts the mount in D-state and hangs FluxOS.
  3. canHoldAppVolume calls fs.stat(mount.target) with no timeout. On a hung NFS/CIFS mount that call is
     uninterruptible, so placementVolumesInGib — and therefore both createAppVolume and the boot-time mount test — block
     indefinitely. development never stats a mount point, so this is new too.

  Fix (~3 lines + one fixture): a REMOTE_FSTYPES denylist beside EPHEMERAL_FSTYPES (nfs, nfs4, cifs, smb3, 9p, virtiofs,
  ceph, glusterfs, afs, sshfs, and fstype.startsWith('fuse.')), or invert to an allowlist of local fstypes (ext2/3/4,
  xfs, btrfs, zfs, f2fs, fuseblk, ntfs3). A NAS is not "a filesystem the node may use for its own storage" under the
  PR's own definition, and the rule deserves to be stated rather than inherited from a /dev/ substring test that no
  longer exists.

  ---

  Finding 2 — non-blocking: the "one filesystem counted once" invariant does not hold for ZFS

  oneRowPerFilesystem keys on source.split('[')[0]. That is a device identity for block-backed mounts (and correctly
  collapses btrfs subvolumes, which findmnt names /dev/sda2[/@home]), but ZFS names each dataset separately while every
  dataset reports the pool's free space. Run through the real code with a stock ZFS-on-root layout:

  rows kept: 5   | real pool free: 900 GiB
  summed available: 4500 GiB
  summed size     : 4510 GiB

  Allowing non-/dev/ sources is precisely what this PR adds (claim #4 in the body), so the PR opens this door and the
  dedup does not cover it. Impact is bounded — placement itself still lands somewhere sane, and the authoritative
  app-space gate is nodeSpecs.ssdStorage from fluxbench, not this sum — but it inflates getSpaceAvailableForFluxShare
  and the secondary availableSpace guard in createAppVolume. The PR's own ZFS fixture is a single-dataset LXC root
  (rpool/lxc/ct-101), which is the realistic Flux-on-Proxmox shape, so this is a "know about it", not a blocker.

  If you want it closed: for a zfs fstype, dedup on the pool name (source.split('/')[0]). Note that stats.dev from the
  fs.stat already being performed would not help here — each ZFS dataset gets its own superblock and its own st_dev.

  ---

  Finding 3 — non-blocking: PR body overstates test independence by one rule

  ▎ Every one of the nine rules fails independently when broken.

  Eight do. Mutating each rule in turn:

  ┌─────────────────────┬─────────────────┬──────────────────────────────────────────────┐
  │    rule removed     │ placement suite │                 both suites                  │
  ├─────────────────────┼─────────────────┼──────────────────────────────────────────────┤
  │ read-only check     │ 1 failing       │ 1 failing                                    │
  ├─────────────────────┼─────────────────┼──────────────────────────────────────────────┤
  │ apps-folder guard   │ 1 failing       │ 1 failing                                    │
  ├─────────────────────┼─────────────────┼──────────────────────────────────────────────┤
  │ emptiest-first sort │ 3 failing       │ 3 failing                                    │
  ├─────────────────────┼─────────────────┼──────────────────────────────────────────────┤
  │ deduplication       │ 2 failing       │ 2 failing                                    │
  ├─────────────────────┼─────────────────┼──────────────────────────────────────────────┤
  │ loop-not-at-root    │ 1 failing       │ 1 failing                                    │
  ├─────────────────────┼─────────────────┼──────────────────────────────────────────────┤
  │ isDirectory()       │ 2 failing       │ 2 failing                                    │
  ├─────────────────────┼─────────────────┼──────────────────────────────────────────────┤
  │ ephemeral fstypes   │ 0 failing       │ 1 failing (covered in volumeService.test.js) │
  ├─────────────────────┼─────────────────┼──────────────────────────────────────────────┤
  │ /boot prefix        │ 0 failing       │ 1 failing                                    │
  ├─────────────────────┼─────────────────┼──────────────────────────────────────────────┤
  │ vfat                │ 0 failing       │ 0 failing                                    │
  └─────────────────────┴─────────────────┴──────────────────────────────────────────────┘

  vfat is redundant with the /boot rule in every fixture, because the only vfat row in the suite is /boot/efi. The rule
  is worth keeping — a systemd-boot ESP at /efi, or a FAT32 stick at /media/x, is caught by nothing else — but nothing
  pins it. One extra row (row('/dev/sda1', '/efi', 'vfat', 1)) closes it. Worth also adding an ext4 /boot row, since a
  separate ext4 /boot partition is the common case and is currently only caught jointly with vfat.

  ---

  Finding 4 — cosmetic

  capacityVolumesInGib does not apply the directory check, and dedup keeps the shortest target. On the containerised
  node that means the returned row is mount: '/etc/hosts' (10 chars) rather than /mnt/appdata (12). The byte counts are
  identical and no caller reads mount off the capacity set — but it will read oddly in logs.

  Also: the capacityVolumesInGib JSDoc ("A wider set than eligibleHostMounts: a loop-mounted ROOT is included") is now
  only part of the story — the set also differs by deduplication and by the absence of the writable-directory check.

MorningLightMountain713:
All four answered. Two are fixed, one is dissolved rather than patched, and one closed itself when FluxShare went. Thanks — finding 1 was a real regression and I would not have found it from the diff.

## 1 — remote filesystems are placement candidates. Fixed, `1a4275ad1`.

Confirmed the mechanism in util-linux source rather than by inspection: `--real` tests `mnt_fs_is_pseudofs()` and only that (`findmnt.c`), and libmount keeps a *separate* `mnt_fstype_is_netfs()` list — `cifs`, `smb3`, `smbfs`, `nfs*`, `afs`, `ncpfs`, `glusterfs`, `fuse.sshfs`, `9p*`. `virtiofs` and `ceph` are on neither list, so they are real too. The old `source.includes('/dev/')` test had been excluding all of them as a side effect.

A `REMOTE_FSTYPES` denylist now sits beside `EPHEMERAL_FSTYPES`, plus the `fuse.` family — gluster and sshfs arrive as `fuse.glusterfs` and `fuse.sshfs`, and the object-store drivers are open-ended. A local fuse pool loses nothing by being refused there, because the disks it pools are mounted in their own right.

Deliberately **not** the alternative of inverting to an allowlist of local fstypes. An allowlist that forgets `bcachefs` or `f2fs` returns an empty candidate list, which is exactly the "no app can install at all" failure this PR exists to remove. I also checked `findmnt --local`, which would delegate to libmount's own netfs list: it exists only on util-linux **master**, not in 2.34 through 2.41, so it would break every node we have.

Two corrections. The reproduction is a WSL 9p mount, which is not a fleet shape, and we have no way to count how many nodes mount network storage — so the severity rests on an unmeasured population. And *"development never stats a mount point, so this is new too"* is not right: `findmnt` cannot produce the `SIZE`/`USED`/`AVAIL` columns without a `statfs` per mount target, and `development` already requests those columns, so the hung-mount exposure pre-dates this change.

## 2 — deduplication does not hold for ZFS. Dissolved, `57a36d0b3`.

No dedup key is correct here, which is why this is not a key change. `used` and `available` have different sharing relations: bind mounts and btrfs subvolumes share both, ZFS datasets share only the free space. Keying on the pool would fix the free-space over-count by introducing a `used` under-count.

So the sum of free space across mount rows is gone instead. Its only consumer was the *"space is already assigned to system files"* guard, which was provably redundant: the per-volume check below it needs some row with `available > hdd + minSystemReserve`, and since every row's available is at most the total, any case where that loop succeeds is a case where the guard did not fire. It could only ever pre-empt a failure the loop produces anyway, with a less accurate message.

`capacityVolumesInGib` is gone with it — its only caller totalled disk size for the FluxShare quota, and FluxShare is removed in this PR. `Σ used` stays, and is correct under device keying on all three filesystem shapes. No total of free space across mount rows survives anywhere.

Removing that guard turned no test red, which is how I learned the refusal path had no coverage at all. It now has two tests pinning both sides of the bound with every term named — reserve floored at `extrahdd` 20, a 1 GiB app refused 21 GiB and placed on 22 — and three mutations each break exactly one of them.

## 3 — the FAT rule is unpinned. Correct. Fixed, `36750393e`.

Re-measured on the current tree by removing each rule in turn, rather than relying on the table, since the rule set has grown. FAT was the only one that could be deleted with nothing going red anywhere: every fixture named its FAT filesystem `/boot/efi`, which the boot rule rejects on its own. It is now pinned by a systemd-boot ESP at `/efi`, the emptiest disk in its fixture so ranking alone would take it, and the rule says why it exists — FAT caps a file at 4 GiB and carries no ownership or permissions.

All eleven rules a candidate passes now fail a test of their own when removed.

One correction: the `/boot` rule is not only caught jointly with FAT. `volumeService.test.js` already carries an ext4 `/boot` row, so it is pinned in its own right and needed nothing.

## 4 — cosmetic. Closed.

Both halves were about `capacityVolumesInGib`, which no longer exists. Underneath it, the deduplication JSDoc justified keeping the shortest target as *"the mount the others are subpaths of"*, which is not true of the containerised case it cites — those four are independent binds of one device. It now says what it is: an arbitrary tie-break between equal-capacity views of one disk, which settles only the directory an image is written into.

## Since this was written

FluxShare stops being a file store on the node. It keeps one thing: a listing and a download, so an operator can collect files a previous release let them put there. No upload, no rename, no delete, no folder creation, and no share token — `getfile` used to fall through to a `{name, token}` lookup when the privilege check failed, and that path is gone.

On Arcane the feature could not be used from its own interface, which a fleet sweep confirmed: unused on 5,129 nodes, data on two, both one operator's. On the 393 legacy nodes the share directory arrives with the checkout, and none of them can be checked — which is why the read path stays rather than going with the rest.

The frontend is RunOnFlux/fluxos-frontend#172, and the order is now **FluxOS first**: the page calls two endpoints that only exist here.

**Unit: see the PR body.** **A gate covering this tip has not returned yet** — the last full run was 108 of 109 at an earlier tip, its one red being suite 89, whose fix is on #1808. A run at the current tip is queued.

Cabecinha84 (commented):
Commit 0638db772 quietly lets the flux team read operator files. Both routes now check
  Privilege.NODE_OPERATOR_OR_FLUX_TEAM (ZelBack/src/services/fluxshareService.js). On development, every FluxShare route
  was NODE_OPERATOR only. The flux team has never been able to list or download what an operator stored in ZelShare.
  With this PR, a flux team login can do that remotely on any node.

  - It contradicts the PR itself. The comment on the routes in routes.js says "Read only, and the node's operator only".
    The commit before it (cf3f8bc25) and the PR description both say "the node operator's".
  - The reason given doesn't hold. The commit says "the flux team is admitted everywhere else in administration." Dozens
    of routes are still NODE_OPERATOR only (benchmark, fluxService, idService, all the daemon RPC routes).
  - The actual problem is in the frontend. The menu entry in fluxos-frontend#172 is privilege: ['admin', 'fluxteam'], so
    it shows the page to the team. The fix is to remove 'fluxteam' from that menu entry, not to give the team access to
    the files.
  - These are operator files on operator hardware. Given your GDPR work, giving the company read access to them is a
    data-access decision. It needs explicit product or legal sign-off, and it shouldn't arrive as a "fix" commit on the
    last day.

  What to ask for: revert the privilege to NODE_OPERATOR on both routes and update the test at
  tests/unit/fluxshareService.test.js:39. Alternatively, get written sign-off and fix the route comment and the PR
  description to match.

  Should be fixed before release (not blocking)

  - The two PRs give opposite release orders.
    - This PR says fluxos-frontend#172 must ship first.
    - #172 (head 05f7a629, still OPEN) says FluxOS must ship first or together, because the two endpoints "only exist on
      #1809". That's wrong: getfolder/:folder? and getfile/:file?/:token? already exist on development with the same
      paths.
    - The new page works against an old node for the operator. It only fails for the team, which old nodes refuse. The
      old full page against a new node gets 404s on stats, upload and the other removed routes.
    - Frontend first is the safe order. Ask the author to fix the description of #172.
  - The CI gate at this head hasn't finished. The build check is pending. The last full run was 108/109, with the
    failing suite fixed by #1808.
  - The product decision on FluxShare is still open. It's now reduced to collecting existing files rather than removed
    completely, but it's still bundled into a placement PR.

MorningLightMountain713:
All correct, including the two that are mine rather than judgement calls. Reverted and fixed.

## The privilege — reverted, `2fb4ef948`

A straight revert of `0638db772`. Both routes are `NODE_OPERATOR` again, with the test and the JSDoc back with them.

You were right on all three sub-points, and two of them catch a real inconsistency I introduced:

- The comment above the routes still said *"the node's operator only"* while the code no longer did. I widened the check and left the comment behind.
- The justification in that commit message — "the flux team is admitted everywhere else in administration" — is false. There are **84 `NODE_OPERATOR` call sites** across the services; I checked after reading this.
- The menu entry was the actual fault, and it is narrowed to `['admin']` in RunOnFlux/fluxos-frontend#172. Widening the node to match the menu was the wrong direction to resolve the mismatch.

On the data-access point: agreed, and it is why this is a revert rather than a discussion. Whether the flux team should be able to read an operator's files is a decision to take deliberately and on its own, not as a fix commit.

## Release order — you are right, and #172's description was wrong

`getfolder/:folder?` and `getfile/:file?/:token?` do already exist on `development` at the same paths. So the new page works against an old node for the operator, and the old full page against a new node is the failing direction — it 404s on `stats`, `uploadfile` and the rest.

**Frontend first.** Both descriptions say so again; #172's claim that the endpoints "only exist on #1809" was mine and it was wrong.

## The gate

A full run is going now at the tip of the stack (#1810), so one run covers both branches. Its head is `6b4d19ab9`, which is this branch's tree plus #1810 — it predates the privilege revert above, so what it does not cover is two `Privilege.` constants, asserted by unit tests on both sides of the change.

**This should not hold up #1809.** If the run turns up anything, it gets fixed on #1810 rather than reopening this. The previous full run was 108 of 109, its one red being suite 89 — a cold-start deadlock in the receive-only election, fixed on #1808 and red on any branch based on `development` today.

## Scope

Fair, and it is a product decision rather than a defence: FluxShare is reduced to collecting existing files instead of being removed, because the 393 legacy nodes cannot be checked and may be holding an operator's data. That it rides a placement PR is deliberate — the two touch the same service and the same apps folder — and the description leads with it rather than burying it.

Cabecinha84 (commented):
Blocking finding: ZFS container datasets become places to put and find app images

  development only accepted mounts whose source contains /dev/, which excluded every ZFS dataset. This PR drops that
  rule so ZFS-rooted nodes can place apps. The side effect is that every mounted ZFS dataset now passes isHostFilesystem
  and canHoldAppVolume.

  That includes each running container's root filesystem on a host where Docker uses the ZFS storage driver: those are
  mounted at /var/lib/docker/zfs/graph/<id>. I reproduced both problems below against the PR's code, using the same
  proxyquire setup the new tests use.

  1. Placement (placementVolumesInGib). Every dataset in a pool reports the same pool-wide free space. The order is
     decided by the exact byte count, so a few MB of change between reads is enough. In my test, a co-tenant's container
     rootfs ranked first: [ '/var/lib/docker/zfs/graph/3f9c1e', '/', '/var/lib/docker' ]. The app's raw ext4 image then
     gets created inside another tenant's container, where that tenant can read and change it. When that container is
     removed on its next redeploy, Docker destroys the dataset and the image with it.
  2. Discovery (getVolumeFilePath). It checks every mount before appvolumes, and appvolumes is where images live on a
     root-hosted node. So a file named /<appId>FLUXFSVOL, even one baked into a co-tenant's Docker image, is found ahead
     of the real volume. ensureAppVolumeMounted then mounts it as root at the victim's app directory. That swaps the
     victim's data, and it also means mounting an untrusted ext4 image as root. Reproduced: it returns
     /var/lib/docker/zfs/graph/3f9c1e/fluxweb_victimFLUXFSVOL.

  Scope: only ZFS hosts where Docker uses the ZFS driver, which is Docker's usual choice on older OpenZFS when
  /var/lib/docker is on ZFS. Arcane (XFS with overlay2) and ext4 legacy nodes are not affected, since overlay is
  excluded. The group is probably small. But the PR advertises ZFS support, and this is a hole this PR introduces.

  Fix: in isHostFilesystem, exclude mount targets under the container runtimes' data directories: Docker's root
  (/var/lib/docker/, /dat/var/lib/docker/, or read DockerRootDir from docker info), containerd, LXD and kubelet. Add a
  ZFS fixture with a graph-driver mount that must fail both placement and discovery.

  Non-blocking

  - Discovery now requires a writable mount. It refuses read-only and FAT mounts. An image that older code placed on a
    FAT data mount, or on a disk that comes up read-only after a reboot, is reported volume_file_missing. The app stays
    inert and the node gets a volume_missing tamper event (weight 1) it doesn't deserve. Finding an image shouldn't
    depend on being able to write there.
  - The remote-filesystem blocklist is incomplete. It misses lustre, gpfs, beegfs and plain fuse. These are rare.
  - ZFS placement is only tested on mount-table fixtures. Whether fallocate works on the target OpenZFS version hasn't
    been shown on a real node.
  - Codecov reports 1 uncovered line.

MorningLightMountain713:
Three fixed, one recorded, one stale. Thanks — the blocking finding is a real hole and the one filed non-blocking was the better catch of the two.

## Container runtime storage — fixed, `a14681c73`

Docker, containerd, LXD and kubelet data directories are refused, for placement and for finding an existing image.

I took the rule wider than the finding: it is not about ZFS. A runtime owns what is under its data directory and reaps it, so our files have no business there on any filesystem. That way it does not depend on the precondition, which is the part I could not confirm — Arcane is XFS with overlay2 and legacy is ext4, so nothing in the fleet is known to run the ZFS storage driver.

Both directions are covered and mutation-proven: a fixture with `/var/lib/docker` and two container datasets must leave only the node's own root, and a planted `<appId>FLUXFSVOL` inside a container dataset must not be returned by discovery. Removing the rule fails both and nothing else.

## Discovery required a writable mount — fixed, `6ba248fe4`, and it became a different change

You were right, and the reason I got it wrong is in the commit message I wrote: I wanted placement and discovery to share one predicate. They should share containment, not writability. Where an image may be **put** is a stricter question than where one may be **found**.

But finding it does not rescue the app — a read-only host filesystem means the loop mount fails anyway. What the wrong answer actually cost was the diagnosis: `volume_file_missing` is what the reconciler turns into a `volume_missing` **tampering event**, so a disk that remounted read-only after an I/O error was recorded as operator interference, and the uninstaller could not find the image to remove.

So discovery no longer asks the write question, and the state is named instead: `host_filesystem_readonly`. True, no tampering event, and it tells the operator which of the two problems they have.

One thing that came out of writing it: the filesystem holding a path is the **deepest** mount the path sits under. Keyed the other way, a read-only `/` would condemn every volume on every other disk on the node. That has its own test.

## The blocklist — three added, `6a4db8726`

`lustre`, `gpfs` and `beegfs` are in, with fixture rows, so the additions are pinned rather than a list nothing tries.

**Plain `fuse` is already excluded.** It is in libmount's pseudofs list, so `findmnt --real` never returns it — which is also where the `fuse.` prefix rule came from.

## ZFS `fallocate` — recorded, not fixed

Agreed and unanswerable here. There is no ZFS node in the fleet or the lab, so every ZFS case is a mount-table fixture. The PR body now says what the change claims — that a ZFS-rooted node is no longer refused a candidate list — and what it does not, which is that the install then completes.

## Codecov

That comment was measured at 09:02 today, before the two cases covering `createAppVolume`'s refusal path landed. It is against an old head.

---

All twelve rules a candidate passes still fail a test of their own when removed, re-swept after these changes. The gate in progress is at `6b4d19ab9` and predates all four commits above, so it does not cover this tip.

Cabecinha84 (commented):
Blocking: the new rule also hides a disk mounted exactly at /var/lib/docker

  isHostFilesystem now refuses any mount where mount.target === dir || mount.target.startsWith(dir + '/'). The problem
  from last time was only the paths under the runtime directory (each container's root, e.g.
  /var/lib/docker/zfs/graph/<id>). The === dir half also drops the runtime's own data disk.

  A separate disk mounted at /var/lib/docker is a common setup on legacy nodes. On development that disk counts as a
  normal place for app images (capacityVolumesInGib accepts any /dev/ source that isn't boot or loop). Once the root
  disk is full, images go there as /var/lib/docker/<appId>FLUXFSVOL.

  I reproduced this with a proxyquire harness: root at /dev/sda1 with 5 GiB free, /dev/sdb1 mounted at /var/lib/docker,
  and an existing image on it.

  ┌───────────────────────────────────┬──────────────────────────────────────┬─────────────────────┐
  │                                   │             development              │       PR head       │
  ├───────────────────────────────────┼──────────────────────────────────────┼─────────────────────┤
  │ getVolumeFilePath                 │ /var/lib/docker/fluxweb_appFLUXFSVOL │ null                │
  ├───────────────────────────────────┼──────────────────────────────────────┼─────────────────────┤
  │ ensureAppVolumeMounted            │ tries to mount                       │ volume_file_missing │
  ├───────────────────────────────────┼──────────────────────────────────────┼─────────────────────┤
  │ getComponentAppIdsFromVolumeFiles │ ['fluxweb_app']                      │ []                  │
  ├───────────────────────────────────┼──────────────────────────────────────┼─────────────────────┤
  │ disks available for placement     │ / + the 838 GiB docker disk          │ / only (5 GiB)      │
  └───────────────────────────────────┴──────────────────────────────────────┴─────────────────────┘

  What this means after the upgrade:
  - The app stops working after the next reboot. While the volume stays mounted, isPathMounted short-circuits, so
    nothing happens until then. After a reboot the image can't be found, so the reconciler holds back every action on
    the app and records a volume_missing tampering event against the operator. Startup also records mount_vanished.
  - The image is left behind forever. The uninstaller's getVolumeFilePath can't see it, so removal leaves the app's data
    on disk.
  - The node loses its big disk for new installs. The docker disk is no longer a placement target, so new installs fail
    with "No useable volume found" even though the node advertises that storage.

  A container's owner can't control anything at the data-directory root itself; only the paths under it are unsafe. Fix:
  exclude strict descendants only, i.e. drop mount.target === dir. Add a fixture test: a disk at /var/lib/docker with
  an existing image must still be found. The ZFS_DOCKER_GRAPH fixture then expects ['/var/lib/docker', '/'] or similar,
  and the container datasets stay excluded.

  Non-blocking

  1. The runtime directories are hardcoded. Docker resolves a symlinked /var/lib/docker and a custom data-root, so on
     those nodes the container mounts show up under the real path and the rule misses them. This only matters on ZFS
     with the zfs storage driver, which no fleet node is known to run. dockerService.dockerInfo().DockerRootDir would be
     a sturdier source. The existing code in dockerService.js:1238 makes the same Arcane/legacy assumption, so this
     matches current practice.
  2. The "no tampering event" goal is only half met. host_filesystem_readonly avoids volume_missing in the reconciler.
     But crontabAndMountsCleanup.js:122 still records mount_vanished for any failure reason at startup, and startup is
     exactly when a disk tends to come up read-only.
  3. A code comment is inaccurate. It says "The mount would fail anyway", which isn't strictly true. The mount options
     are loop,nosuid,nodev with no explicit rw, so util-linux falls back to a read-only mount for a clean ext4 image.
     Refusing to mount is a defensible choice, but the comment should say it's a choice.
  4. Small edge case in isOnReadOnlyFilesystem. When two mounts sit on the same path, the strict > in the reduce picks
     the first row (the covered one) rather than the visible one. This only affects which reason gets reported.
  5. Carried over: whether fallocate works on ZFS is still untested on a real node. The PR body now says this honestly.

MorningLightMountain713:
Thanks — the blocking one was right, and so were four of the five below it.

**The disk mounted exactly at `/var/lib/docker`.** Fixed: strict descendants only, `mount.target === dir` is gone. Both directions are pinned — restoring the equality reddens the placement test and the discovery test, and there is a fixture with a disk at `/var/lib/docker` holding an image that must still be found.

**1 — the runtime directories are hardcoded.** Recorded rather than chased, deliberately. `dockerInfo().DockerRootDir` is the sturdier source and it is a call to a daemon that may be down at exactly the moment placement is asked, so it trades one failure mode for another; the list stays, and the precondition it depends on is not one anything in the fleet is known to meet.

**2 — the "no tampering event" goal was only half met.** Fixed, and it went further than the finding. A mount failure that describes the host rather than the volume is now its own zero-weighted event class, `volume_host_fault`, so the startup sweep no longer records `mount_vanished` for a disk that came up read-only, a missing loop device, an unreadable mount table, or a mountpoint it could not create. Separately, an image that mounts but is not the one this node made now records under its own name from both the startup sweep and the reconciler, instead of being folded into `mount_vanished` at boot and recorded nowhere at all mid-run.

**3 — the comment was inaccurate.** Fixed. It now says the refusal is a choice and why: the mount options carry no explicit `rw`, so util-linux falls back to a read-only loop mount for a clean image, and an app whose volume cannot be written to is down regardless. The reason it is refused under names the disk rather than the image, because calling the image missing blames an operator for a hardware fault.

**4 — the tie in `isOnReadOnlyFilesystem`.** Fixed. Equal-length targets are mounts stacked on one path and `findmnt` lists them in mountinfo order, so the reduce takes the later row, which is what the kernel resolves through. Both directions are pinned.

**5 — `fallocate` on ZFS.** Still untested, and the body says so. There is no ZFS node in the fleet or the lab; every ZFS case here is a mount-table fixture.

## What has changed since you last looked

Enough that it is worth saying plainly rather than leaving you to diff it.

**A node records the image it made, and knows it by more than its name.** `createAppVolume` writes down the absolute path and stamps the filesystem with a UUID the node generates, and later boots look the record up instead of scanning every mounted filesystem for a file called `<appId>FLUXFSVOL`. A filename proves nothing — anything on the node can write one — so the search is the weakness, not the safeguard. A file substituted at the recorded path is refused; an image found elsewhere means the record is stale rather than the volume hostile, and the record is replaced. Migration needs no separate pass: a node upgrading with its apps running reads the path from the kernel's `backing_file`.

**An unfinished node-level DOS path is removed.** `testAppMount` set a value the availability checker never read, because the two halves referred to different objects. Completing the wire was the obvious tidy-up and would have been the wrong one: the value is the threshold at which a node removes every app it holds, repeating every ten minutes, reached from a single failure of a one-shot test that a busy loop device or an unreadable mount table can trip. It is removed rather than left for someone to finish, and the intent it served has a home in `volume_host_fault`. If the fleet data ever says a node like this should be acted on, that is a weight on that event type and the incident machinery's existing guards — measuring the population first, since nothing has ever counted it. The boot rehearsal itself is gone too: its caller discarded the result, so a node behaved identically whether it passed or failed.

**A command that did not answer no longer establishes anything.** Several places read a failure as a fact about the disk — an unreadable mount table as "nothing is read-only", a probe that could not run as "there was nothing to wipe", a probe of the image that could not run as "the image holds no filesystem", a failed `rm -rf` as "cleaned", a successful `umount` as nothing at all. Each answered in the direction that does damage, and the one that asked a command what the kernel could answer now asks the kernel.

**Where an image is, and whether this node still knows where it put one, are two questions.** The state read answered `null` both for "no record" and for "could not read", and a caller acts on `null` by searching the disks and trusting what it finds. The search still runs — a removal needs the image found whatever the database is doing, and not searching leaves a disk nothing will ever account for again — but an image found somewhere other than the record, while the record is unsettled, is not adopted: mounting it skips the stamp for want of anything to compare against, and then writes it down as this node's own.

**A call between services is checked against what the callee exports**, statically, over the real sources. A member that is not exported arrives `undefined`, the call throws where it runs, and a proxyquired caller cannot see it. The check found three, one of them introduced here.

**Suite `2001`** installs an app on a real fleet, substitutes a complete ext4 filesystem at the image's own path, asserts the reconciler reports the volume unavailable and mounts nothing, then puts the genuine image back and waits for it to mount — so the refusal is a refusal and not a node that mounts nothing either way.

Cabecinha84 (commented):
Non-blocking notes (worth a follow-up)

  1. Search order was flipped. appvolumes is now searched before the disks; development searched the disks first. If an
     app has no settled record and two copies of its image exist, which one gets mounted depends only on that order, and
     the choice is then recorded permanently. Neither order is correct in that case. It's rare: it only happens when
     the app wasn't mounted at the first start after the upgrade. Better: when more than one candidate exists, log or
     record the ambiguity instead of picking silently.
  2. blkid exit code 2 also means "file not found". I confirmed this locally. If the image disappears between fs.access
     and the failed mount (for example, an uninstall racing the reconciler), the failure is classified as mount_failed →
     volume_image_unrecognised, weight 1, against the operator. Re-checking that the file exists before blaming the
     image would close this.
  3. ensureMountPathsExist ignores imageMoved. If that path is the one that mounts the volume, the weight-0 event is
     never recorded.
  4. Soft removal can now stop partway. removeControllerState throws where remove() used to swallow errors. A database
     error during a soft removal now aborts it after the containers are removed but before the local app entry is
     deleted. This is deliberate (so the operator's stop flag isn't silently kept), but it is a behavior change.
  5. The symlink error message lost a hint. The pathSecurity error for symbolic links no longer suggests "compress the
     folder" to users of the app volume browser, which does support that.
  6. Carried over from before, and both acknowledged in the PR: the runtime-directory list is hardcoded, and fallocate
     on ZFS is untested.

MorningLightMountain713:
Thanks — three of these are fixed, two are deliberate and stay, and the last pair is unchanged.

**1 — the search order, and choosing silently between two images.** The order stays and the silence goes. The node's own two directories outrank the mounts because nothing else can write into them, which is what stops a planted file outranking the genuine image on a node that has no record yet. Where more than one file under the name exists, every candidate is checked and the ones behind the answer are named in the log — the one that mounts is recorded, and every later lookup reads the record without searching, so that line is the only moment anything sees the second. What the search meets *after* the image turns up cannot rule anything in or out, so it does not mark the result as having left a question open; old and new agree on `path`, `conclusive`, `blocked` and `recordSettled` across all 2048 combinations of per-candidate outcome, record state and mount-table readability, and the comparison was checked against a deliberately wrong variant that it flags in 110 of them.

**2 — `blkid` exit 2 also means the file is not there.** Fixed, and confirmed against a real `blkid` rather than taken on trust: exit 2, empty stdout, empty stderr, for an empty file and a missing one alike. The gap between the search finding the image and the mount failing is five awaits wide, so an uninstall crossing a reconcile could put a weight-1 `volume_image_unrecognised` on the operator for a file the node itself had removed. `imageFsType` asks whether the file is still there before giving that answer, and a file that is gone was not probed — which the caller already treats as saying nothing about the image. A volume whose image really is deleted is reported by the next pass, from the search rather than from a probe.

**3 — `ensureMountPathsExist` ignored `imageMoved`.** Fixed, at both of its mounts: the component's own volume and the referenced component's. It is the pass that mounts on an app restarted through the API while its volume is down, and the fact is unrecoverable afterwards because the mount replaces the record — so it records, as the reconciler, the boot sweep and the syncthing mount repair do. The two sites are pinned separately: removing either reddens only its own test.

**4 — a soft removal can stop partway.** Deliberate, and it stays. The alternative is what it replaced: swallowing the failure leaves the operator's stop flag in place and reports the redeploy done, so the component the operator asked to run stays down and nothing says why. Failing loudly leaves the app installed with its containers gone, which the reconciler recovers. Between a recoverable state that is visible and an unrecoverable one that is silent, the visible one is the right failure.

**5 — the symlink message lost the "compress the folder" hint.** Deliberate. That error is raised in `pathSecurity`, which serves callers on both the app volume browser and the FluxShare routes, and FluxShare has no compress route — this PR leaves it with `getfolder` and `getfile` only. The advice names only what is true for every caller of it.

**6 — the hardcoded runtime directories, and `fallocate` on ZFS.** Unchanged, as the body says.

Unit is **7195 / 0 / 17** at the tip. The harness figure in the body is measured at `991c18751` and these three commits are unit-covered only; the body now says so rather than claiming the tip.

Cabecinha84 (approved):
ack

PR #1810, opened 2026-09-17
Five surfaces narrowed to what they are actually for

MorningLightMountain713:
Five hardening changes, each one narrowing a surface to what it is actually for, plus a harness suite for the one of them a fleet can demonstrate.

## The changes

**Backups are addressed by the app whose volume they are.** Three backup endpoints authorised against an app name and then operated on a file path supplied separately, so the path was never checked against the name. The path is now rebuilt from the authorised name and compared whole, reusing the existing app-identifier helpers rather than spelling the volume naming rule a third time. Two plain defects went with it: `appname` was read from `req.query.number`, so every `?appname=` caller authorised against the empty string and was refused, and `if (!path)` tested the imported `path` module. The flux team still reaches any app's archives, but must name the app it is addressing.

**A fluxnode has no wallet, so FluxOS stops publishing one.** 63 wallet and Zcash RPC routes removed, verified uncalled across FluxOS, fluxos-frontend, fluxmarketplace, fiat-orchestration-backend, cloud_admin, arcane-gui-backend and deployer. `fluxRpc.js`'s method map is enforced, so dropping 67 names from it makes those calls uncallable rather than merely unrouted. `sendrawtransaction` stays public — broadcasting is a service a node can offer — and the five `getaddress*` GETs move to the public block, where their POST twins already were. −11,440 lines.

**A storage link is not an address the specification author chooses.** `F_S_ENV`/`F_S_CMD` URLs come from the app owner's own specification, and the node fetches them and signs a message to whatever host they name. All 21,357 storage links in the chain's entire message history point at one host, so the rule is an allowlist: host compared whole, `https` only, and `maxRedirects: 0`, because an allowed host answering a 302 would otherwise reopen it. Enforced at the fetch, which covers every app at any height, and at submission, held to `liveSubmission` alongside the existing owner-identity rule so that historical specifications are unaffected.

**A node's word is not a payment, and the rendezvous it was attached to is.** `/payment/verifypayment` checked only that a payment id existed and was under an hour old, then stored caller-supplied txid, amount, coin and addresses verbatim and answered success — while `/payment/paymentrequest` issues ids to anyone. Flux already verifies payments the way that cannot be forged, every node reading the chain and matching a transaction against the app message hash and the minimum price for that height, so nothing downstream ever read what this recorded.

What the three routes carry is not that verification. A wallet opened through the `zel:` handler is a separate application with no route back to the page that opened it, so the page asks a node for an id, listens on `/ws/payment/<id>`, and hands the wallet a callback carrying the same id; the transaction id the wallet posts is delivered to that listener and to nothing else. `palworld-server-website`, `minecraft-server-website` and `games-website` all depend on it, from their deployment, hardware and management dashboards.

So the rendezvous stays and the record goes. The id and the transaction id sit in a small bounded cache; the amount, coin and addresses are recorded nowhere; `activepaymentrequests` and `completedpayments` are dropped at boot with their indexes. Issuance is rate-limited per address and each address may hold only a few ids at once, so no single caller fills the cache and evicts another's, and a delivered id is dropped rather than left to time out. The id is unguessable, because it is the only thing the callback is held to, and its own lifetime is the wait — there is no second clock that could give up on a wallet the id would still have accepted.

**A signing authority that signs nothing.** `certs/v1rootcert.privkey.pem` and its certificate, CA:TRUE and valid to 2043, have been tracked since 2023 — force-added past a `.gitignore` rule that already covered `certs/`. Nothing has ever trusted them: they are in no trust store, no code references them, and the certificate a node serves is a different, per-node self-signed one. Removed. No rotation and no revocation, because publishing either would assert the authority mattered.

## Four more of the same kind

**A backup path is resolved against the app's own volume.** Three endpoints authorise over one app and then operate on a path: the local backup list, a file removal and a file download. Only the download resolved symlinks at all, and it resolved against the apps folder — which holds every app's volume, so a link from one into another stayed inside the boundary it was checked against. All three now resolve against the volume the tenancy check has already identified. A container is given `<id>/appdata` and cannot write into its own backup directory, so no app can plant such a link today; the guard no longer rests on that staying true.

**A containment check is held to the part of the path that exists.** `verifyRealPath` returns a path it cannot resolve unchanged, so a name that has not been created yet passes even when a directory above it is a link out. The folder browser and both file downloads asked it that question and now ask `verifyRealPathOfExistingPath`, which walks to the deepest part that does exist. `volumeSession` keeps the original deliberately — it verifies the parent with the stronger call and uses this one to compare where two paths land — and the contract now says which is which. The proxyquire stubs in one suite named the old export: with `noCallThru` the renamed one arrived `undefined`, the call threw, the handler's catch swallowed it, and every test still passed on an assertion made earlier in the handler.

**The payment collections are dropped at boot, with their indexes.** `activepaymentrequests` and `completedpayments` are on every node, empty — boot created indexes on them whether or not anything ever wrote a document, and nothing ever did. Measured on a live node: both present, zero documents, each carrying a `createdAt` TTL index, so mongo's TTL monitor scans them every sixty seconds for documents that cannot arrive.

**The 413 in the refusal list names the layer that answers it** — express's default 100kb body limit, and a handler that reads its own body for the content types no mounted parser claims.

## The volume-identity work underneath

Four changes to the placement and volume-identity code this stacks on:

- **An unformatted volume with no record is an unfinished install, not tampering.** An image that holds no filesystem and has no record — what a crash between allocating a volume file and formatting it leaves — is `volume_incomplete_install`, recorded at no weight. An image that holds no filesystem where a record says one belongs is that image overwritten, and scores.
- **One classifier maps a mount fault to its event, for the boot sweep and the reconciler both.** `volume_host_fault` is recorded whether a mount fails at boot or while the node runs, so a fleet query for it sees every node rather than only those that met the fault at boot.
- **A folder download names itself through the response and streams once.** The download name is set through the response's own `attachment`, so a folder name the app chose reaches the header encoded rather than raw; the archiver carries an error listener; the archive streams once with no length announced; and a root download hides the same reserved entries the listing does.
- **App identifiers derive one way.** `KadenaChainWebNode` and `FoldingAtHomeB`, the two names once forced to a `zel` prefix, are registered nowhere, so every name takes `flux` and every derivation of an identifier agrees.

Each is mutation-covered.

## The harness suite

`1901-storage-links-are-flux-storage-only` — 7/7 on a real fleet, covering the storage-link rule end to end. It needed a second TLS certificate under the CA the fleet already trusts, a stub serving that host name over TLS on its own listener, and a Docker network alias so the fleet's resolver answers for it. Because the routes exist only on that listener, a recorded request is one that satisfied protocol *and* host, and an empty log is a node that refused before asking.

**All four of its rules are independently mutation-proven**: each broken alone, rebuilt and run, each turning exactly one test red and leaving the other six green. `isFluxStorageUrl` is two rules in one expression, so host and protocol were broken separately — breaking both at once would only have shown that something was covered.

## Evidence

- **Unit: 6912 passing / 0 failing / 17 pending**, against a base branch (#1809) measuring 7195 in the same environment, run sequentially and never overlapping.
- **The full harness runs green with these fixes.** The last full run returned **109 of 110**; the single red was a restart-timing race in the location-table suite — a node restarted before it had adopted a policy bundle that names no artifact, and refreshed against the previous one for a stray fetch. Fixed here: the policy store persists an adopted bundle before it notifies subscribers, and the suite waits for the node to adopt the tableless bundle before restarting it. That suite re-runs **10 of 10**.

## Note for whoever lands #1767

It moves `obtainPayloadFromStorage` into a shared resolver and makes `F_S_CONTACTS` a fetch path. **`F_S_CONTACTS` must be added to `STORAGE_MARKERS` when it does**, or that marker reopens what this closes.


Cabecinha84 (commented):
🔴 Blocking: the FluxShare filename check kills the node process

  ZelBack/src/services/fluxshareService.js:818

  filename: (name, ext, part, form) => validateFilename(part.originalFilename),

  validateFilename throws. Formidable calls it from _getNewName() inside the async _handlePart(), and
  formidable/src/plugins/multipart.js:163 does await this.onPart(part) inside an un-awaited async listener. The
  rejection is never handled.

  I reproduced this against the PR head with a real multipart request through the handler:

  ┌───────────────────┬─────────────────────────────────────────────────────────────┐
  │   filename sent   │                           result                            │
  ├───────────────────┼─────────────────────────────────────────────────────────────┤
  │ bad<0x01>name.txt │ unhandled rejection, request hangs forever, nothing written │
  ├───────────────────┼─────────────────────────────────────────────────────────────┤
  │ "" (empty)        │ unhandled rejection, request hangs                          │
  ├───────────────────┼─────────────────────────────────────────────────────────────┤
  │ ..                │ unhandled rejection, request hangs                          │
  └───────────────────┴─────────────────────────────────────────────────────────────┘

  Standalone (no mocha rejection handler installed), Node exits:

  Error: Invalid filename: path separators and control characters are not allowed
      at validateFilename (ZelBack/src/services/utils/pathSecurity.js:430:11)
      at IncomingForm.filename (ZelBack/src/services/fluxshareService.js:818:44)
      at IncomingForm._getNewName (formidable/dist/index.cjs:1605:38)
      at IncomingForm._handlePart (formidable/dist/index.cjs:1348:30)
  EXIT CODE: 1

  In production this reaches apiServer.js:232's uncaughtException handler, which is logErrorAndExit(err, { exitCode: 1
  }) — FluxOS exits.

  Three things make this matter:

  1. It is a regression this PR introduces. On the base branch the callback returned originalFilename unconditionally
     and could not throw.
  2. It contradicts the commit's own stated intent — "a bad name is refused rather than quietly stored under
     formidable's fallback." The observed behaviour is not refusal; it is a node restart and a client that never gets an
     answer.
  3. The new test does not cover it. tests/unit/fluxshareUpload.test.js only exercises traversal names — and those pass
     for a different reason than the author's: formidable strips directories from originalFilename before the callback
     sees it, so ../../outside/pwned arrives as pwned and validateFilename never fires. The three cases that do reach
     the throw are untested.

  Mitigating: the endpoint is Privilege.NODE_OPERATOR, so this is operator-triggered self-DoS, not remote. But a browser
  is not the only client, and the fix is small — the check has to report through formidable's error channel (which
  form.on('error') already handles) or run somewhere that can answer the request, rather than throwing synchronously
  inside _getNewName. Whatever shape it takes, please add the control-char / empty / .. cases to the test.

  The underlying fix in this commit — deleting the fileBegin override that replaced formidable's built path with the
  unvalidated field name — is correct and is the real vulnerability closure. That part is good.

  ---

  🟡 Confirm before merge

  1. The storage-link check at the fetch is not grandfathered, and will break existing apps.
  The submission check is correctly gated to liveSubmission (consensus-safe, and the harness proves a replayed message
  still validates). But obtainPayloadFromStorage enforces isFluxStorageUrl for every app at any height. The commit says
  "166 links on chain would fail this rule" while the PR body says all 21,357 point at one host — so those 166
  presumably fail on http: or on URL parse. Every one of them now throws at container creation. Please confirm none
  belong to a currently-deployed, unexpired app; if any do, they fail on the next install/redeploy on every node
  simultaneously. Everything else about this change is sound: single chokepoint, whole-host comparison, both spec field
  spellings handled, compose array validated upstream, storageHost present in the config app.js actually pins.

  2. maxRedirects: 0 against the real storage. Correct reasoning (an allowed host answering 302 re-opens the hole), and
  the harness proves the refusal against a stub. It does not prove storage.runonflux.io answers 200 directly today. One
  curl before merge settles it.

  3. Cross-repo verification has two gaps. I cloned and grepped fluxos-frontend: getWalletInfo and rescanDaemon exist in
  src/services/DaemonService.js but are never called — dead wrappers, so the claim holds. However
  RunOnFlux/fluxos_flutter (lib/ui/routes/daemon/daemon.dart) and RunOnFlux/titan_ui_old both reference
  /daemon/getwalletinfo and /daemon/rescanblockchain, and neither repo is in the PR's verified list. fluxos_flutter was
  last pushed 2024-02-15, so it is almost certainly dormant — but say so explicitly rather than leaving it unstated.
  (GitHub code search rate-limited me partway through the full 58-route sweep; the publicly-reachable removals —
  gettransaction, both fundrawtransaction — came back clean.)

Minor, non-blocking
  - Two route-wiring floors were lowered (>200→>150, >400→>300). Justified by 63 deletions, but it does weaken the guard;
    the new exact-list assertions compensate.
  - downloadLocalFile still calls verifyRealPath(filepath, appsFolder), not the app's own volume dir — so a symlink planted
    inside one app's backup/ still resolves cross-tenant. Not reachable today (backup is a reserved mount subdir in
    mountParser, and the container only gets <id>/appdata), so it is defence-in-depth, not a hole.
  - activepaymentrequests / completedpayments are dropped from config but left behind on existing nodes as orphan
    collections. Harmless.
  - messageHelper.js's edited comment now reads "a request too large to accept (a body too large to accept 413)" — circular
    after the paymentService reference was removed.

MorningLightMountain713:
Every one of these checked out. Three needed code, and one of the three was answered by the feature above it rather than by a fix.

## Blocking — the filename check reaching formidable's callback

Right, and it no longer exists on this branch. FluxShare is reduced to download-only in the base branch: `fluxshareService` exports `fluxShareGetFolder` and `fluxShareDownloadFile` and nothing else, so there is no upload path and no `filename` callback for a throw to escape from.

The other caller of `validateFilename` is sound, and worth saying so rather than leaving it implied. `fileSystemManager.js` calls it inside `receiveOne`, whose `catch` records the failure and drains the incoming stream, so a bad name settles the operation that is waiting on it instead of rejecting into nothing. The three cases named — a control character, an empty name, `..` — reach a handler that answers the request.

## Confirm 1 — the storage-link check at the fetch

Measured against the current global list rather than the chain's history, which is the population the question is actually about: **1,740 apps, 162 `F_S_ENV`/`F_S_CMD` links, every one of them `https://storage.runonflux.io`, none failing the rule.** So no deployed, unexpired app fails at container creation.

Built twice so the answer is not a single reading: a structured walk of `compose[].environmentParameters` and `commands` plus the v1–v3 top-level spellings, and a field-agnostic scan of the raw document. Both give 162. Breaking the rule two ways — requiring `http:`, requiring a different host — flips all 162 to failing, so it discriminates.

Two things that belong with the number rather than under it. **691 of the 1,740 carry an encrypted specification and expose an empty `compose`**, so their links are not in that count and cannot be read from any public listing; the fetch is what decides for those, on each node that runs them. And the 768 `F_S_CONTACTS` markers live in the app's `contacts` field rather than the environment — nothing dereferences them, so they are not in scope.

The address check has since been tightened: a port and userinfo are part of an address a specification author does not choose, and neither passes now. None of the 162 carries either.

## Confirm 2 — `maxRedirects: 0` against the real storage

Covered twice. Harness suite `1901` has *a storage that answers a redirect is not followed, wherever it points* — 7/7 on a real fleet, and one of four rules each broken alone and each turning exactly one test red. Separately, six distinct links taken off the chain answer with **zero redirects** when requested directly; unsigned they answer 403, since the node fetches these carrying its own signed identity, so what that shows is that the host does not redirect on that path rather than that the authorised path returns 200.

## Confirm 3 — the cross-repo sweep

Neither `RunOnFlux/fluxos_flutter` nor `RunOnFlux/titan_ui_old` resolves for us, in the organisation listing or in code search, so we could not confirm or contradict those two.

We swept the organisation for all 22 removed route stems instead, and it found a consumer neither of us had named: **`RunOnFlux/fluxcloud-analytics`** matches `/daemon/rescanblockchain` in `src/config/eventMappings.js`. It classifies events rather than calling the route, so removal leaves that mapping dead rather than broken.

The same sweep found something that did matter, and it is the reason the payment routes are back in a different shape. **`palworld-server-website`, `minecraft-server-website` and `games-website` all use `/payment/paymentrequest`, `/ws/payment/{id}` and `/payment/verifypayment`** against `api.runonflux.io`, from their deployment, hardware and management dashboards. So "its only consumer was removed from the frontend in 2025" was wrong, and the PR body has been corrected.

What those sites need is the rendezvous, not the verification. A wallet opened through the `zel:` handler is a separate application with no route back to the page that opened it, so the page asks a node for an id, listens on `/ws/payment/<id>`, and hands the wallet a callback carrying that id. The verification was never what they relied on: the page stores the transaction id and the chain decides whether the app is paid.

So the three routes are restored as exactly that and no more. The id and the transaction id sit in a small bounded in-memory cache; the amount, coin and addresses a wallet sends are recorded nowhere; the two collections stay dropped. Issuance is rate-limited per address and each address may hold only a few ids at once, so no single caller fills the cache and evicts another's; a delivered id is dropped on delivery rather than left to time out. The id is unguessable, because it is the only thing the callback is held to, and its own lifetime is the wait — there is no second clock that could give up on a wallet the id would still have accepted.

One thing the callback now does that the old one did not: it answers whatever content type the wallet chose. `express.json` is the only parser mounted, so a JSON body arrives already parsed with its stream spent while every other type arrives unread — and the old handler waited on an `end` that, for the first kind, had already fired.

## The four minor ones

**Route-wiring floors lowered.** Acknowledged — the exact-list assertions are what carry it now, and the pinned list matches the route table exactly.

**`downloadLocalFile` resolving against `appsFolder`.** Fixed: a backup path is resolved against the app's own volume directory, so a link planted inside one app's `backup/` no longer has a base wide enough to land in another's.

**The orphan payment collections.** Fixed: both are dropped at boot with their indexes, so the TTL monitor stops scanning them on every node that never used them.

**The circular 413 comment in `messageHelper`.** Fixed: it names the layer that answers.

## Since then

The storage-link rule is enforced on the whole address, not the host alone, and the fetch reads the marker the way the door does — splitting on the marker truncated a link carrying the same text in its query, so the value approved was not the value requested.

`sanitizeAndVerifyPath` and `verifyRealPathSync` are removed. Neither had a caller, and the first offered itself as the most secure option while calling the check that answers "does this resolve outside" rather than "is this contained".

Cabecinha84 (commented):
🔴 Blocking: a peer can restart FluxOS with one gossip message

  Commit ea96a818a ("persist an adopted bundle before announcing") makes adopt, consider and offerBundle in
  ZelBack/src/services/policyStore.js async.

  The gossip dispatcher in ZelBack/src/services/fluxCommunication.js (~line 720) runs handlers like this:

  setImmediate(() => { try { handler(msgObj, peerSocket); } catch (e) { log.error(e); } });

  It is a synchronous try/catch around a call it does not await. Anything offerBundle now throws becomes an unhandled
  rejection. That reaches apiServer's uncaughtException handler, which calls logErrorAndExit(exitCode: 1).

  How it's triggered: verifyBundle calls Buffer.byteLength(raw) before it checks the signature. A fluxpolicy message
  whose bundle is not a string (for example { length: 1 }) makes that call throw ERR_INVALID_ARG_TYPE.

  I reproduced it through the same dispatch pattern:

  ┌───────────────────┬─────────────────────────────────────────────┐
  │       Head        │                   Result                    │
  ├───────────────────┼─────────────────────────────────────────────┤
  │ #1809 (bdc7de260) │ The try/catch catches it, the node survives │
  ├───────────────────┼─────────────────────────────────────────────┤
  │ #1810 (b34c47aaa) │ UNCAUGHT → the process exits                │
  └───────────────────┴─────────────────────────────────────────────┘

  Who can do it: fluxpolicy is declared ANSWER, which isn't deduplicated and is delivered even when this node didn't
  ask. So any registered fluxnode connected as a peer can send it with its own valid signature. It can restart every
  node it connects to, and again on each reconnect. This is the same class of bug as the previous blocker.

  Fix (one line either way): add a .catch(log.error) to the register('fluxpolicy', ...) handler, or have verifyBundle
  reject a raw that isn't a string or Buffer before it measures it. A unit test that sends a non-string bundle through
  the dispatcher would pin it.

  ---

  🟠 Should fix: the payment relay's per-address cap works against real users but not attackers

  The payment routes are back as a small in-memory relay in ZelBack/src/services/paymentRelayService.js. The PR body
  claims that "no single caller fills the cache and evicts another's". Two problems:

  1. An attacker can bypass the cap with their own callbacks. answerCallback overwrites the entry with { txid } and
     drops its ip, so answered ids no longer count toward the cap of 10. Writing the entry again also resets its
     one-hour lifetime.
     - I reproduced this from one address: request 5 ids per second, answer each one yourself, repeat. That reached 550
       ids minted with 0 refused, filled the 500-entry cache, and evicted a victim's pending id.
     - Fix: keep ip when writing the txid ({ ...held, txid }).
  2. The cap lands on real users, because the sites go through the domain proxy. games-website's walletService.js
     reaches nodes via api.runonflux.io or a sticky *.node.api.runonflux.io address. Both go through FDM, the Flux
     domain proxy, so req.socket.remoteAddress is the proxy's address, not the user's.
     - Every user coming through one proxy shares 5 requests per second and 10 pending ids per node.
     - A closed WebSocket doesn't free its id. So 10 dialogs abandoned in an hour, for example users who switch to SSP,
       make that node answer 429 to the next ZelCore payment through that proxy.
     - The old implementation had no cap, so this is a user-facing regression on exactly the flow this PR says it
       preserves.
     - Fixes to consider: drop the pending id when its WebSocket closes, and trust x-forwarded-for only from known FDM
       addresses, or raise the cap for them.

  Dropping amount, coin and addresses is safe: every consumer (DeploymentDialog, HardwareTab, VersionTab,
  ServerManagementPanel) reads only txid.

MorningLightMountain713:
Thanks — both were right, and both are fixed. Each one is now proven against a real node in both directions: with the fix, and rebuilt from the tree without it.

**The blocking one.** Correct in full, including that it is this PR's doing: `adopt`, `consider` and `offerBundle` were synchronous before `ea96a818a`, so the dispatcher's `try` caught exactly this. Reproduced on a fleet from the stub peer, framed and signed as any peer's broadcast — without the fix the node's pid changes and its log carries `ERR_INVALID_ARG_TYPE` from `Buffer.byteLength` followed by `exiting with code: 1`; with it, the pid is unchanged and the bundle is refused as what it is.

Fixed at both ends, but not in the dispatcher. `verifyBundle` refuses a bundle that is not text before it measures one, and `offerBundle` holds its own failure and settles the peer's ask on it — which is what its eight sibling handlers on that route already do, each wrapping its whole body. A `.catch` on the dispatcher looked equivalent and is not: `processSyncChunk` **awaits** its handler, and the drain loop's `catch` ends the peer's stream rather than continuing into the next chunk. Catching at that level would have taken that away silently, for every ordered type, to fix one gossip one. Either end alone stops the reported message; both are there because the first is the honest answer for that input and the second is what stops the next one.

**The cap.** All three parts, and the reproduction of the first was exactly right.

1. **The address now stays with an answered id** (`{ ...held, txid }`), so an id its asker answered themselves goes on costing them their own allowance instead of becoming a free slot. The remaining lifetime is carried across the answer too — a meeting is an hour old from when it was made, and answering it again does not buy another.
2. **An id the browser stopped waiting on is dropped when its socket closes.** Nothing else collects one, so it would otherwise hold a slot, and its asker's allowance, for the rest of the hour. An answered id is left for the delivery path, which spends it once a browser has it.
3. **The address counted is the caller's, not the connection's.** You are right that `req.socket.remoteAddress` is the proxy for everyone arriving that way — `api.runonflux.io` is a CNAME to `fdm-lb-main`, and a response comes back naming the backend it was relayed to. Rather than inventing a rule for this, the node now uses the one already written for it: `resolveClientIp` believes `x-forwarded-for` only from an address in `fdmAddresses`, and reads the **last** entry, which is the only one a caller cannot write past. A header from anything else is ignored, so the limit cannot be stepped around by setting one.

`1902-payment-ids-are-held-against-the-caller` covers all three on a real node. The forwarding case needs the node to see the suite as a balancer, which the fleet's own config override does — the shipped list is addresses no suite can send from. Against the unfixed tree the three fail one apiece: the id an abandoned listener held is never given back, an id its asker answered stops counting against them, and a second caller behind one balancer is refused for what the first one holds. The round-trip test passes either way, which is what it is for.

**On the last note** — agreed, and checked against all three sites: `walletService.js` reads only `txid`, and the close code.

Unit is **6934 / 0 / 17**. The harness figure in the body is five commits back and the body now says so rather than claiming the tip.

**A full harness run is queued and has not run at this tip.** The box is occupied, and the run itself takes several hours; the result will be posted here when it returns. What stands behind these changes today is the unit suite, the targeted suites above, and their unfixed-tree counterparts — not a gate.

Cabecinha84 (commented):
Conditions

  1. Harness gate at the tip. The PR body says the last full run (109/110) was five commits back, and the run at this
     tip is queued but hasn't run. The two new commits have unit tests and targeted e2e runs only.
  2. Ops check of fdmAddresses (22 hard-coded IPs in ZelBack/config/default.js, which can only change with a FluxOS
     release):
     - api.runonflux.io resolves to 5.39.57.40, which is not on the list. The list names 128.199.246.121 and
       5.161.44.226 for "main" instead.
     - For the nodes balancer, the address its DNS returns (5.39.57.41) is listed.
     - Someone should confirm the actual outgoing addresses for main. This can't cause harm either way: a missing entry
       just falls back to the socket address, which is the old behaviour. But the websites default to api.runonflux.io,
       so if it's wrong, their users still share one allowance per node.

  Non-blocking polish

  - paymentRelayService.js ~line 86 has a stale comment ("the callback overwrites the entry with just the txid"). This
    commit made the opposite true.
  - The PR body and the module's doc comment name 3 dependent sites; about 15 site repos actually use this relay. That
    matters for the plan to remove it once the sites host the callback themselves.
  - Residual: the per-IP cap doesn't stop an attacker who controls many addresses (e.g. an IPv6 block) from filling the
    500-entry cache. It is out of scope and is not a regression.

MorningLightMountain713:
**Full gate at `9fc715536`: 113 of 114.** Archived as `9fc715536-fullgate-RED-13`.

The single red is `13-orchestrator-state-machine`. **It passed the previous 40 gate runs in a row** — this is its first failure in 43 recorded runs across both harness boxes.

It is not this PR. The cause is a pre-existing crossing-dial race in the peer manager, already fixed on the v9 lineage, and we may backport it separately. This branch touches no part of the peer manager; its only peer-path changes are a comment block and the `paymentService` → `paymentRelayService` rename.

**On the `fdmAddresses` condition — real, and fixed.** `fdm-lb-main` is geo-steered across three regional balancers, and the EU one, `5.39.57.40`, was not on the list, so European callers were keyed to the balancer rather than to themselves. Confirmed on the balancer itself: it answers on `5.39.57.40`, reaches nodes from that same address, and has no source rewrite — which also settles the question about outgoing addresses. Every other entry is correct and nothing listed is stale.

**On the polish points.** The stale comment at `paymentRelayService.js:86` is corrected, as is the module header. The relay stays on the node, so the count of dependent sites gates nothing; for the record, an org-wide search finds the three the module names. The cache-fill residual is out of scope and unchanged.

This branch is now rebased onto `development` at `0bea2c166`. Unit suite: **7101 passing, 17 pending, 0 failing**.

Cabecinha84 (approved):
ack

PR #1812, opened 2026-09-21
The compliance sweep follows the policy, not a clock

MorningLightMountain713:
## What this changes

The blocklist reaches a node in seconds now that it travels in the signed bundle, so the hourly `setInterval` the compliance sweep was armed on became the whole of the wait between the network banning something and a node acting on it — up to an hour, and a full hour after boot before anything looked at all.

It is driven by the bundle instead: a pass when the policy gate opens, and a pass on every bundle change while it is open. That is what every other consumer of the bundle already does (`ipLocationSync`, `enterpriseNetwork`).

Both triggers are staggered. `policyStore.adopt()` opens the gate *before* it announces, and `AsyncGate.open()` drains its waiters synchronously, so an unstaggered gate trigger runs first and puts the whole fleet's removals — and the broadcast of each one — into the same instant.

## A refusal is not a removal

`removeAppLocally` returned `void`. Its guards return having touched nothing when the node is already installing or removing, and its catch does not rethrow — so a caller could not tell "removed" from "refused, node busy", and a blocked application could be marked done while still running, with nothing coming back for it.

It now answers a `RemovalOutcome`: `REMOVED`, `NOT_INSTALLED`, `BUSY`, `FAILED`. This mirrors the `InstallOutcome` the install side has carried since it had the same problem, and its doc states the same principle. **37 callers ignore the return value and are unaffected.**

What tells `NOT_INSTALLED` from `FAILED` is an `AppNotFoundError`, not the wording of a message. The message and the error's `name` are what these endpoints answer with and are unchanged, so reword one and the outcome a caller acts on does not move.

## The rest

- **The list is asked again before each removal, and so is the record.** A pass holds itself open for minutes: an entry lifted while it ran was still acted on and the uninstall broadcast, and an application redeployed onto a clean image was judged on the one it no longer runs. A record that cannot be read is not an answer that the node has let the application go — that is owed a later pass, not settled.
- **What has no event is owed by name**, on a self-cancelling backoff: a specification that did not decrypt, a removal the node was too busy to take, one that failed, a record that could not be read, and the applications a stopped pass never reached — a pass that throws included. A node owing nothing runs no timer, and what the node no longer holds is owed nothing.
- **The wait is decided from what is owed**, at the moment of arming: it doubles for as long as the same thing is owed and starts again the moment that changes, for the whole node exactly as for one application.
- **A record rewritten without a redeploy is judged.** `reinstallOldApplications` writes the local record and rebuilds nothing when a specification differs only in the fields its comparison deletes — `owner` among them — so an application transferred to an owner the network refuses was never checked by any path. Those names are collected and swept once the pass is over, because both of them uninstall and a removal the sweep attempts while this pass holds the node's install and removal flags is refused.
- **The sweeper is an object that owns its state and takes its dependencies**, including the clock and its knobs. As file-scope state it was reachable from two entry points that disagreed about it.
- `imageComplianceIntervalMs` is removed; four knobs replace it, with harness-compressed values and their `HARNESS_TUNABLES.md` rows.

## Verified

| | |
|---|---|
| unit suite | **7129 passing, 0 failing** |
| `imageManager` tests | **87 passing** |
| `appUninstaller` tests | **35 passing** |
| full gate on chud, at `e5dfd1d7f` | **108/108, `suites_fail=0`** |
| suite 1601 green/red pair, at `e5dfd1d7f` | green 3/3, red 0/3 on the right assertion |
| lint | clean on every changed file |

### Full gate

```
15:35:20 START   MAXN=6 MIN_FREE_MB=15000 MAX_LOAD=12 suites=108 head=e5dfd1d7f
17:10:52 RESULT  suites_pass=108 suites_fail=0 FAILED:[ ]
17:10:52 ###PAR-DONE
```

No non-zero `rc` on any suite and no `###ABORT`. This is the run that matters for blast radius: the change adds `return` statements inside `removeAppLocally`'s `try`/`catch`/`finally`, which manages `removalInProgress` and `departingApps`. A leaked lock there fails the *next* removal rather than this one, which no unit test or single suite would catch. App removal, image-update redeploy, container lifecycle, node-status removal and the reconciler restart set are all green.

### Suite 1601

It no longer restarts every node to make them notice a new document, and the bound is 20s. Same processes, same mesh, same installed application — the only thing that changes is the document.

| arm | code | interval | result |
|---|---|---|---|
| green | this branch | none — event-driven | `suites_pass=1 suites_fail=0` |
| red | `development` | production 3600000 | `suites_pass=0 suites_fail=1` |

The red arm fails on the assertion the suite exists for, not a boot error — `###ABORT`, `before all` hook failures and `Pool overlaps` all checked and zero:

```
not ok 1 … removes an installed application once the document names it
  Timeout after 20000ms waiting for event: app:removed
```

Test 3 reproduces the production behaviour this change is about, in three containers:

```
node 1 installed a blocked application more often than it removed it: expected 1 to be at most +0
```

The red arm runs at production's `imageComplianceIntervalMs` deliberately: the harness compresses it 3600000 → 60000, shorter than any bound the suite could reasonably use, so at the harness value a generous bound passes against code with none of the behaviour it names.

Archives on chud: `e5dfd1d7f-fullgate-green`, `e5dfd1d7f-1601-green`, `dev-1601-red-beforestate`.


MorningLightMountain713:
## Verified on the harness

Run on chud against `head=e5dfd1d7f`, images built and checked with `verify-images.sh` against their own trees.

### Full gate — GREEN

```
15:35:20 START   MAXN=6 MIN_FREE_MB=15000 MAX_LOAD=12 suites=108 head=e5dfd1d7f
17:10:52 RESULT  suites_pass=108 suites_fail=0 FAILED:[ ]
17:10:52 ###PAR-DONE
```

**108/108, no non-zero `rc` on any suite, no `###ABORT`.** 95 minutes, `inflight=6` sustained throughout.

This was the run that mattered for blast radius: `removeAppLocally` has 37 callers and this change adds `return` statements inside its `try`/`catch`/`finally`, which manages `removalInProgress` and `departingApps`. A leaked lock there would not fail a unit test or suite 1601 — it would fail the *next* removal. The gate covers app removal, image-update redeploy, container lifecycle, node-status removal and the reconciler restart set, and all of them are green.

### Suite 1601 — green/red pair

| arm | tree | code | interval | result |
|---|---|---|---|---|
| green | `07da8ef5…` | this branch | none — event-driven | `suites_pass=1 suites_fail=0`, 3/3 |
| red | `2452c1b0…` | `development` | production 3600000 | `suites_pass=0 suites_fail=1`, 0/3 |

The red arm fails on the assertion the suite exists for, not on a boot error — checked explicitly for `###ABORT`, `before all` hook failures and `Pool overlaps`, all zero:

```
not ok 1 … removes an installed application once the document names it
  Timeout after 20000ms waiting for event: app:removed
```

Test 3 is the more telling one, and it reproduces the production behaviour this change is about:

```
node 1 installed a blocked application more often than it removed it: expected 1 to be at most +0
```

A node taking on a banned application because nothing had told it yet. Green's identical assertion passes.

The red arm runs at production's `imageComplianceIntervalMs` deliberately: the harness compresses it 3600000 → 60000, which is shorter than any bound the suite could reasonably use, so at the harness value a generous bound passes against code with none of the behaviour it names.

Archives on chud: `e5dfd1d7f-fullgate-green`, `e5dfd1d7f-1601-green`, `dev-1601-red-beforestate`.

Cabecinha84 (commented):
Findings (non-blocking, both reproduced)

  I called createComplianceSweeper with injected timers, DB and blocklist to reproduce these.

  1. A DB error during the per-app re-read silently drops a blocked app (new finding).
  subjectNow() returns null both when the app is gone and when installedApps(name) returns status: 'error'. The caller
  then treats it as "no longer holds it": it calls owed.delete(appName) and continues.
  - Repro: a blocked app, and the re-read returns an error. Result: 0 removals and no retry scheduled.
  - Impact: the app stays running until the next bundle change, owner transfer or restart. On development, the hourly
    timer would have caught it within an hour.
  - This goes against the PR's own rule that anything it couldn't check is held for later, never assumed either way.
  - Fix, a few lines: tell a read error apart from "not found", and hold(appName, 'unexamined') on the error.

  2. The whole-node retry never backs off.
  The armRetry callback clears owedWholeNode before the pass runs. holdEverything() then sees it as false and resets
  retryDelayMs to the base value.
  - Measured: 60 s, 60 s, 60 s, 60 s, 60 s. For comparison, named holds do back off correctly: 60 s → 120 s → … → 3600 s
    cap.
  - When it happens: only if a signed bundle contains a malformed blocklist, or the installed-apps DB read keeps
    failing.
  - Cost: one DB query and one warning log line per minute, in a state that should be rare.

  3. Minor polish:
  - A full pass never removes held names for apps that are already gone, so there is one redundant scoped pass.
  - NOT_INSTALLED is detected by matching the text 'Flux App not found', which is fragile.
  - No unit test covers the removeAppLocally results or the owner-transfer → requestComplianceSweep path.
  - Two code comments claim more than the code does:
    - The finally in reinstallOldApplications says the sweep's removals "cannot meet this pass's". The next reinstall
      pass can still overlap with the sweep, but this was already true before the PR.
    - The same block says "the sweep takes the policy gate". In fact a request doesn't wait on the gate; it holds the
      apps and retries.

MorningLightMountain713:
Both reproduced here independently before touching anything, driving `createComplianceSweeper` with injected timers, table and blocklist. All of it is fixed on the branch.

## 1 — a read that failed is not a record saying the application is gone

`subjectNow` collapsed three answers into `null`: the node holds it, the node does not hold it, and the record could not be read. It now answers `{ answered, subject }`, and the removal loop takes three branches by name. A read that failed holds the application under its own reason, `unread`, so the log of a node that keeps asking says which question went unanswered.

Reproduced before: one blocked application, the per-app re-read errors, `0 removals, 0 timers armed`. After: the application is held and the retry is armed. `b4cd3e2b9`.

## 2 — the wait is decided from what is owed

The reset was written by the act that recorded a hold, and the retry path clears what it is about to re-ask before the pass runs, so `holdEverything()` read a debt the node was still carrying as one it had never carried.

The decision now lives at the moment of arming, and nowhere else: it doubles for as long as the same thing is owed and starts again when that changes. `hold` and `holdEverything` only record. Measured on both paths, production values: 60s → 120 → 240 → … → 3600s cap, where the whole-node path was flat at 60s. `c53bb2396`.

## 3 — the rest

**The redundant scoped pass.** The prune was asked of the scope; it is now asked of the owed list against the whole table, which every pass already reads. One branch and one wakeup removed rather than added. `6e28800f7`.

**`NOT_INSTALLED` by string match.** It is an `AppNotFoundError` now, thrown at all three sites in `appUninstaller.js`, with the message and the error's `name` untouched so the API error body is identical. Six tests pin what `removeAppLocally` actually answers — removed, busy for each guard, not installed for the unforced and the forced absence, failed for anything else — and each branch of the mapping was mutation-checked separately. `06efc9b2b`.

Writing those turned up something in the existing harness: `createWarningMessage` was absent from the `messageHelper` stub, and under `proxyquire.noCallThru()` that makes it `undefined`, so both BUSY guards were throwing into the catch instead of returning. The two existing "refused duplicate" tests passed either way — what they assert is true of a crash as much as of the guard. The stub key is added and they now exercise the guard.

**The owner-transfer path.** Two tests, with the canary: a pass whose application differs only in the fields the comparison deletes asks the sweep about it by name, and a pass that genuinely redeploys does not, having been proved to reach the redeployment first. `e6339d4d0`.

**The two comments.** Both rewritten to what the code does. What holds the sweep and the reinstall pass apart is the node's install and removal flags, not that function's lock: a removal the sweep attempts while the pass holds them is refused and goes onto its backoff. The request is not awaited because a pass spaces its removals over minutes; a node without confirmed policy holds the applications and asks again rather than waiting on the gate. The set is also renamed for what it collects — `owner` is one of six fields deleted from that comparison, so an expiry, a description or an instance count reaches the same branch. `53d8c9ab4`.

## One more of the same kind

The pass's outer `catch` logged and held nothing, so a throw was the one way out of a pass that forgot it had work to do. It holds the node now, which makes the rule true by construction rather than by inspecting every callee. `94108e0bd`.

## Verified

Unit suite **7129 passing, 0 failing** (7116 before, plus the 13 tests added). Lint clean on every changed file. Nothing here touches the sweep's triggers, `removeAppLocally`'s control flow or its lock handling.

The harness run is queued behind other runs on the box. The 108/108 gate and the 1601 green/red pair in the description are at `e5dfd1d7f`, and the description says so.

Cabecinha84 (commented):
New findings (non-blocking)

  I reproduced both. Neither can be worse than development, which sweeps every hour.

  1. The backoff remembers old debt. The last owed set (owedWhenLastArmed) is never cleared once nothing is owed. If the
     same thing becomes owed again later, the first retry waits the full 3600s instead of 60s.
     - Repro: a named app went 60→…→3600, was removed, and when the same name was held again its first wait was 3600s.
     - The same happens for the whole-node path: it capped once, recovered, then a later failure also started at 3600s.
     - Fix is one line: reset owedWhenLastArmed = null in armRetry when nothing is owed.
  2. A new debt waits behind a timer that is already armed. If a bundle-triggered pass holds a new app while a 3600s
     retry is pending, armRetry returns early. The new app waits out the remaining time, up to 1h. The comment saying
     the wait "starts again the moment that changes" is therefore only true at arming time.

MorningLightMountain713:
Both reproduced here before touching anything, driving `createComplianceSweeper` with injected timers at production values. Both fixed in `da0e9a059`.

## They are one defect

`armRetry` read "a timer is armed" as "the debt is unchanged". Those are different questions, and each of your two findings is that conflation failing in one direction: finding 1 keeps a memory the debt has outlived, finding 2 keeps a wait the debt has outgrown.

It has a third face the review did not name. The function's own `SELF-CANCELLING` docblock claims a node owing nothing runs no timer. Discharge the debt with a bundle-triggered pass while a retry is pending and the wait stays armed, on a node with nothing left to ask about, for the remainder of whatever it was armed for.

## What it does now

The wait is decided against the debt the pass ends with, at the end of every pass rather than only at the one that armed it. A wait already running for exactly that debt is left alone, so an unchanging debt still backs off. One armed for anything else is replaced at the base rate. Nothing owed leaves no timer and no memory of what was.

## Why not the one-liner

Resetting `owedWhenLastArmed` only where nothing is owed does not reach the case where the debt was cleared underneath a running timer: `armRetry` returns at the `retryTimer` guard before the reset, and that timer then fires, finds an empty scope and returns without calling `request`, so nothing reaches `armRetry` at all and the stale value survives to the next hold. The rule has to be asked of the debt on every pass, not of the branch that happens to be taken.

## Measured

Production values, `60000` base and `3600000` ceiling, on the injected clock.

| | before | after |
|---|---|---|
| a name owed, discharged, owed again | 3600000 | 60000 |
| the whole node owed, discharged, owed again | 3600000 | 60000 |
| an application held while another sits at the ceiling | waits out 3600000 | 60000 |
| a debt settled underneath a running wait | timer left armed | no timer |
| an unchanging named debt | 60000 → … → 3600000 | unchanged |
| an unchanging whole-node debt | 60000 → … → 3600000 | unchanged |
| the ceiling | held | held |
| `stop()` | drops both | unchanged |

## Verified

Three tests, red before and green after, each breaking under exactly one targeted mutation and no other: one that withholds the reset when nothing is owed, one that never replaces a running wait while something is owed, and one that never drops a running wait when the debt empties. Each turns exactly its own test red and leaves the other two green.

Unit suite **7132 passing, 0 failing** (7129 before), across nine full runs. `imageManager` **90**. Lint clean on both changed files.

Cabecinha84 (approved):
ack

